PR #20, opened 2018-04-05
Version.TryParse returns null on conversion failure.

GiriB:
The current documentation states that Version.TryParse returns a Version number - 0.0 - on conversion failure. But it actually returns null.


GiriB:
Reference: https://www.ideone.com/jsh62w

PR #35, opened 2018-04-10
Update Validator.xml

agrtga:
The overload for TryValidateObject that provides a Boolean value to recursively check each property has remarks that are the same as the other overload that doesn't. The remarks indicate it won't recursively check properties which is contrary to the corresponding parameter. Clarity has been added to the remarks.


agrtga:
The suggested changes have been committed.

agrtga:
You're welcome.

PR #40, opened 2018-04-13
Fixed 'unable to parse type System.T&' build warnings

rpetrusha:
## Fixed 'unable to parse type System.T&' build warnings

The OPS build is emitting warnings and not displaying a property type because the property type in these two cases is identified as System.T&, which is incorrect; the property type is simply T.

This issue will eventually have to be fixed in the reflection routine that outputs type and member information. In the meantime, I'm fixing it here to eliminate the build warnings.

//cc @dend @mairaw

### DIFF
--- a/xml/System/ReadOnlySpan`1.xml
+++ b/xml/System/ReadOnlySpan`1.xml
@@ -364,11 +364,11 @@
       </Docs>
     </Member>
     <Member MemberName="Item">
-      <MemberSignature Language="C#" Value="public T&amp; this[int index] { get; }" />
-      <MemberSignature Language="ILAsm" Value=".property instance class System.T&amp; modreq(System.Runtime.InteropServices.InAttribute) Item(int32)" />
+      <MemberSignature Language="C#" Value="public T this[int index] { get; }" />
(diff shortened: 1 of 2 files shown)


xml/System/ReadOnlySpan`1.xml
```diff
@@ -364,11 +364,11 @@
       </Docs>
     </Member>
     <Member MemberName="Item">
-      <MemberSignature Language="C#" Value="public T&amp; this[int index] { get; }" />
-      <MemberSignature Language="ILAsm" Value=".property instance class System.T&amp; modreq(System.Runtime.InteropServices.InAttribute) Item(int32)" />
+      <MemberSignature Language="C#" Value="public T this[int index] { get; }" />
```
svick: I believe [this should be `public ref readonly T this[int index] { get; }`](https://github.com/dotnet/corefx/blob/dc2eefdfcbc17c229f62ad3f01f0524de0880d1f/src/System.Memory/ref/System.Memory.cs#L153). Similar change should be done for other languages and the `Enumerator` too.

I'm not sure how `ref readonly` should be represented in `<ReturnType>`.

Also, there's a similar issue with [`Span<T>` indexer](https://docs.microsoft.com/en-us/dotnet/api/system.span-1.item?view=netcore-2.1), which should be `ref T`, but isn't.

PR #41, opened 2018-04-13
Add missing ArgumentException to CroppedBitmap ctor

ChrisMaddock:
## Summary

Encountered this constructor throwing an undocumented `ArgumentException` today. 

Not sure if this is the correct fix for these sort of API issues, let me know if I should file an issue instead. 🙂


ChrisMaddock:
Thanks @rpetrusha - I'd just copy-pasted the one above! Change made. 😄

PR #45, opened 2018-04-14
Update SecureString.xml

yevster:
## Summary
SecureString does exist on .Net core. In fact, it appears to be a part of .Net standard 2.0.

However, the implementation does not behave the same way absent Windows' ProtectedData, as evidenced in https://github.com/PowerShell/PowerShell/issues/1654.

### DIFF
--- a/xml/System.Security/SecureString.xml
+++ b/xml/System.Security/SecureString.xml
@@ -130,7 +130,7 @@
  Although data stored in a <xref:System.Security.SecureString> instance is more secure than data stored in a <xref:System.String> instance, there are significant limitations on how secure a <xref:System.Security.SecureString> instance is. These include:  
   
  Platform  
- On the Windows operating system, the contents of a <xref:System.Security.SecureString> instance's internal character array are encrypted. However, whether because of missing APIs or key management issues,  encryption is not available on all platforms. Because of this limitation, <xref:System.Security.SecureString> may not work correctly on .Net core. 
+ On the Windows operating system, the contents of a <xref:System.Security.SecureString> instance's internal character array are encrypted. However, whether because of missing APIs or key management issues,  encryption is not available on all platforms. Because of this platform dependency, <xref:System.Security.SecureString> does not encrypt the internal storage on non-Windows platform. Other techniques are used on those platforms to provide additional protection.
   
  Duration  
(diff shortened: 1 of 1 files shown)


BillWagner:
closing and re-opening to force a new build.

GrabYourPitchforks:
It'll work correctly on all platforms in that the APIs will behave as expected, but it won't encrypt its contents on certain platforms, correct?

rpetrusha:
@terrajobst, could you take a look? Should we provide any additional guidance or detail?

xml/System.Security/SecureString.xml
```diff
@@ -130,7 +130,7 @@
  Although data stored in a <xref:System.Security.SecureString> instance is more secure than data stored in a <xref:System.String> instance, there are significant limitations on how secure a <xref:System.Security.SecureString> instance is. These include:  
   
  Platform  
- On the Windows operating system, the contents of a <xref:System.Security.SecureString> instance's internal character array is encrypted. However, whether because of missing APIs or key management issues,  encryption is not available on all platforms. Because of this, <xref:System.Security.SecureString> is available on Desktop (Windows only), and not on .NET Core.  
+ On the Windows operating system, the contents of a <xref:System.Security.SecureString> instance's internal character array are encrypted. However, whether because of missing APIs or key management issues,  encryption is not available on all platforms. Because of this limitation, <xref:System.Security.SecureString> may not work correctly on .Net core.
```
bartonjs: > Because of this limitation, <xref:System.Security.SecureString> may not work correctly on .Net core.

That depends on what "correctly" means.  The contents are stored outside of the managed heap in memory marked to not page out (absent hibernation), so it still protects against heap compaction copies.

The contents _are_ only encrypted on Windows (during times that the contents are not being utilized), but the API functions on all platforms.  Note the name of the type is not "EncryptedString".

BillWagner: @yevster Can you update this PR with the following text (based on product team feedback):

> Because of this platform dependency, xref:System.Security.SecureString does not encrypt the internal storage on non-Windows platform. Other techniques are used on those platforms to provide additional protection.

yevster:
Changed. Thanks for the help. I may make another PR if/when I get to the bottom of why https://github.com/PowerShell/PowerShell/issues/1654 occurs.

BillWagner:
Thanks for working with us to update this @yevster  I'll :shipit: now.  You should see the updates in the next day or two on the live site.

Good luck with the powershell issue. I'm sure that team is looking at that one.

PR #76, opened 2018-04-25
HttpHeaders.GetValues throws InvalidOperationException

dimaaan:
System.Net.Http.Headers.HttpHeaders.GetValues method throws InvalidOperationException when header is missing instead of returning null.
This should be mentioned in the docs!


dimaaan:
@rpetrusha thank you too.

PR #93, opened 2018-04-27
Replaced crdefault token with token text, deleted token

rpetrusha:
## Replaced crdefault token with token text, deleted token

This PR replaces the crdefault-md token with its token text and removes the token. In two cases, it corrects the use of the wrong token.

The token text was replaced and the token removed from the dotnet/docs repo by [dotnet/docs#5054](https://github.com/dotnet/docs/pull/5054)


rpetrusha:
It's hard to name the most hated token -- there are too many to choose from.

PR #97, opened 2018-04-28
The only Introduction sentence omits half the description.

MightyPen:
Method Enum.IsDefined is useful when the program has a integer value to test, and the lone sentence that comprises the entire Introduction mentions this.  But .IsDefined is equally useful when instead the program has a string that might match the Name that is assigned to the integer (integral value) inside the enum body; yet the Introduction improperly omits this important fact and functionality.
I realize this text change really needs to be made in the .NET source code ///summary for the System.Enum class, for the text change to live on indefinitely.
I work at Microsoft.  GitHub MightyPen = GeneMi at MS.  2018/04/28

## Summary

Describe your changes here.
Added other half of functionality to the summary Introduction.

n/a

### DIFF
--- a/xml/System/Enum.xml
+++ b/xml/System/Enum.xml
@@ -965,7 +965,7 @@ thisInstance And flag = flag
       <Docs>
         <param name="enumType">An enumeration type.</param>
         <param name="value">The value or name of a constant in <c>enumType</c>.</param>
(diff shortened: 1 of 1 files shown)


MightyPen:
I do not see any Errors.  But I am unfamiliar with this particular type of Build feedback.
What am I supposed to do next?  Am I now supposed to enter the keyword string HASH SIGN OFF?

PR #137, opened 2018-05-03
Replaced FwLink and changed link format

ChrisMaddock:
Relates to https://github.com/dotnet/docs/issues/3391

This single fwlink was in this repo nearly 2000 times. 🙂 It was repeated in the same phrase, so I created an include file instead.

I wasn't totally sure how to do this though, so some questions:
- Is my include file ok with a relative link in, when it's used in places like the `<summary>` section? I see the current syntax doesn't render correctly in the API lookup

image

- I wasn't sure what the text in the square brackets of the `[!INCLUDE[asp mobile homepage](../includes/asp-mobile-md.md)]` was meant to be? Is "asp mobile homepage" ok?

### DIFF
--- a/xml/System.Web.Mobile/CookielessData.xml
+++ b/xml/System.Web.Mobile/CookielessData.xml
@@ -18,7 +18,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>Encapsulates data into an encrypted <see cref="T:System.Web.Security.FormsAuthenticationTicket" /> for a page adapter to include in links in mobile controls. For information about how to develop ASP.NET mobile applications, see <see href="https://docs.microsoft.com/aspnet/mobile/overview">Mobile Apps & Sites with ASP.NET</see>.</summary>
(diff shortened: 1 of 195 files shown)


ChrisMaddock:
Again, this is a big PR sorry - but I didn't see much sense in splitting it up. Let me know if I should do!

mairaw:
You can see [here](https://docs.microsoft.com/en-us/dotnet/api/?view=netcore-2.0) what happens when you use an include:

image

ChrisMaddock:
Oh! 😞 I thought I was being clever!

That lookup doesn't seem to support markdown links at all. Would writing it as plaintext be better? Or maybe we should just cut the text altogether - I'm not sure how much value it's adding?

image

mairaw:
No, it doesn't. 😢 

You can see a PR that solves this issue [here](https://github.com/dotnet/dotnet-api-docs/pull/14/files). 

So try changing them like this:
`<see href="https://docs.microsoft.com/aspnet/mobile/overview">www.asp.net/mobile</see>`.

Without https://docs.microsoft.com/ might work but because this content also goes into IntelliSense I kept the domain in there until I had time to test the whole workflow.

ChrisMaddock:
Great - thanks @mairaw - updated!

PR #161, opened 2018-05-09
Correct TreeViewItem.IsSelectionActive description

idg10:
The docs incorrectly stated that TreeViewItem.IsSelectionActive indicated whether the TreeViewItem itself has the keyboard focus. By inspection this is not true. In fact, if the containing TreeView has the keyboard focus then *every* TreeViewItem in it will return true from IsSelectionActive. So the docs need to make it clear that although this property is defined as a member of TreeViewItem, it actually reports information about the containing TreeView.

This can be demonstrated by experiment. Moreover, it appears to be by design. The TreeViewItem.IsSelectionActive property is in fact an alias for Selector.IsSelectionActive, an inherited property which is already correctly documented as reporting whether the containing Selector has the focus.

Furthermore, inspecting the reference source reveals not only that this property is an alias, but also that WPF makes no attempt to make TreeViewItem.IsSelectionActive specific to any single TreeViewItem. The handling for this property exists entirely at the TreeView level.

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Windows.Controls/TreeViewItem.xml
+++ b/xml/System.Windows.Controls/TreeViewItem.xml
@@ -517,14 +517,14 @@
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets a value that indicates whether the <see cref="T:System.Windows.Controls.TreeViewItem" /> has keyboard focus.</summary>
+        <summary>Gets a value that indicates whether the containing <see cref="T:System.Windows.Controls.TreeView" /> has keyboard focus.</summary>
         <value>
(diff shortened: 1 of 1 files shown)


mairaw:
@Tanya-Solyanik can you take a look at this PR?

Tanya-Solyanik:
This is a WPF treeview, adding the team to look at this PR

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

mairaw:
@SamBent would you be the right person to review this then?

mairaw:
Internal review URL is the following:
https://review.docs.microsoft.com/en-us/dotnet/api/system.windows.controls.treeviewitem.isselectionactive?view=netframework-4.7.2&branch=pr-en-us-161

mairaw:
Thanks for your PR @idg10 and apologies on the delay on getting it reviewed. Your fix should go live within a day.

PR #162, opened 2018-05-09
Remove bad Interlocked.Exchange example

svick:
Contributes to https://github.com/dotnet/docs/issues/5227.

The example is not thread-safe and I don't know of an easy way to fix it. I also think that having no example is better than having a bad example, so I think it should be removed for now.

A good example would still be useful, which is why I'm not saying this PR fixes the linked issue.

### DIFF
--- a/xml/System.Threading/Interlocked.xml
+++ b/xml/System.Threading/Interlocked.xml
@@ -952,8 +952,10 @@
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
-## Remarks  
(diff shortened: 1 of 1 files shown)


pkulikov:
@svick what do you think of the example for [Exchange(Object, Object)](https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.exchange?view=netframework-4.7.2#System_Threading_Interlocked_Exchange_System_Object__System_Object_)? Given that this method is from the pre-generic era, should we remove such an example and convert **Remarks** to the note?

svick:
@pkulikov Yeah, that's a good point. Since the only thing that example really shows is how to deal with the lack of generics for `Exchange`, I think it's not useful.

svick:
@rpetrusha Done.

PR #163, opened 2018-05-09
Improve ConcurrentDictionary.TryUpdate

BrunoJuchli:
## Summary
Improve the code-doc summary of ConcurrentyDictionary.TryUpdate. The summary mentions parameters in a different order and with different names than they are represented in the method signature. This is confusing:
image

To clear things up, I've rephrased the summary to mention parameters in the same order as they occur in the method signature and used `paramref` to reference them (although I can see that the generated seems to treat it just the same as `c`, and it seems some of the apidoc has been converted to use `c` while some still uses `paramref`. I don't know what the preference is).

### Note
I'm "cross-PR-ing" this from https://github.com/dotnet/corefx/pull/29599
Though I'm wondering a bit about the duplication of the api-doc.
Will we really have to maintain the code-doc twice? Will one become obsolete? Will they be merged (one-way or two-way) automagically?
I haven't found anything on that topic in the contribution guide. But then I admit I didn't read every word of it, either...


rpetrusha:
@BrunoJuchli, code comments are often (but not always) used for the documentation. The two, however, are independent of one another; changes to documentation in the .NET codebase aren't migrated to the .NET docs, and vice versa.

PR #165, opened 2018-05-10
Modify description to correspond to code changes

rpetrusha:
## Modify description to correspond to code changes

This change is related to dotnet/docs#5044, which replaces calls to the AppDomain.CurrentDomain.DefineDynamicAssembly method with calls to the static AssemblyBuilder.DefineDynamicAssembly method. The reasons for the change are:

- The example contains too many comments, which are not localized. I've moved some (though very few) into the text.
- The example was modified to work with .NET Core. This means that it is no longer is hard-coded to look for mscorlib.dll and System.dll.
- There is no need to use snippet tags, since the entire example is used.
- Since the intent is to make this work with .NET Core, I've removed the C++ example.

### DIFF
--- a/xml/System.Reflection/EventInfo.xml
+++ b/xml/System.Reflection/EventInfo.xml
@@ -165,9 +165,9 @@
 You might use the `AddEventHander` method when you load a type after the program is already compiled, when it is not possible to use the C# += syntax or the Visual Basic `WithEvents` and `Handles` keywords to hook up the event.   
   
 ## Examples  
- The following example loads an assembly, creates an instance of a type from the assembly, creates an event handler using a dynamic assembly, and hooks up the dynamic event handler. All actions are performed using late binding.  
+ The following example creates an instance of the <xref:System.Timers.Timer?displayProperty=nameWithType> class, creates an event handler using a dynamic assembly, and hooks up the dynamic event handler. All actions are performed using late binding.  
   
- The example creates an instance of the <xref:System.Timers.Timer?displayProperty=nameWithType> class. The <xref:System.Timers.Timer> instance is stored in a variable of type <xref:System.Object>, and all code that accesses the Timer does so late-bound. The example uses the <xref:System.Type.GetEvent%2A?displayProperty=nameWithType> method to get the <xref:System.Timers.Timer.Elapsed> event, and the <xref:System.Reflection.EventInfo.EventHandlerType%2A> property to get the delegate type for the event.  
+ The <xref:System.Timers.Timer> instance is stored in a variable of type <xref:System.Object>, and all code that accesses the <xref:System.Timers.Timer> does so late-bound. The example uses the <xref:System.Type.GetEvent%2A?displayProperty=nameWithType> method to get the <xref:System.Timers.Timer.Elapsed> event, and the <xref:System.Reflection.EventInfo.EventHandlerType%2A> property to get the delegate type for the event.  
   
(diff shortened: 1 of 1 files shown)


xml/System.Reflection/EventInfo.xml
```diff
@@ -160,22 +160,21 @@
...
+ The following example loads an assembly, creates an instance of a type from the assembly, creates an event handler using a dynamic assembly, and hooks up the dynamic event handler. All actions are performed using late binding.  
   
- The code example loads System.dll and creates an instance of the <xref:System.Timers.Timer?displayProperty=nameWithType> class. The example uses the <xref:System.Type.GetEvent%2A?displayProperty=nameWithType> method to get the <xref:System.Timers.Timer.Elapsed> event, and the <xref:System.Reflection.EventInfo.EventHandlerType%2A> property to get the delegate type for the event.  
+ The example creates an instance of the <xref:System.Timers.Timer?displayProperty=nameWithType> class. The <xref:System.Timers.Timer> instance is stored in a variable of type <xref:System.Object>, and all code that accesses the Timer does so late-bound. The example uses the <xref:System.Type.GetEvent%2A?displayProperty=nameWithType> method to get the <xref:System.Timers.Timer.Elapsed> event, and the <xref:System.Reflection.EventInfo.EventHandlerType%2A> property to get the delegate type for the event.
```
BillWagner: The one instance of "Timer" that is capitalized, and not an xref: Should it be code-fenced so it is not localized?

xml/System.Reflection/EventInfo.xml
```diff
@@ -160,22 +160,21 @@
 ## Remarks  
  This method attempts to add a delegate to synchronize the event on the target object.  
   
- Each time the event is raised by the target parameter, the method or methods encapsulated by the handler will be invoked.  
+ Each time the event is raised by the target parameter, the method or methods encapsulated by the handler are invoked.  
   
-   
+You might use the `AddEventHander` method when you load a type after the program is already compiled, when it is not possible to use the C# += syntax or the Visual Basic `WithEvents` and `Handles` keywords to hook up the event.
```
svick: This technically doesn't apply anymore to the modified sample, since `typeof()` requires that the type is known during compilation.

Does that matter?

xml/System.Reflection/EventInfo.xml
```diff
@@ -160,22 +160,21 @@
...
   
- Each time the event is raised by the target parameter, the method or methods encapsulated by the handler will be invoked.  
+ Each time the event is raised by the target parameter, the method or methods encapsulated by the handler are invoked.  
   
-   
+You might use the `AddEventHander` method when you load a type after the program is already compiled, when it is not possible to use the C# += syntax or the Visual Basic `WithEvents` and `Handles` keywords to hook up the event.   
   
 ## Examples  
- The following code example shows how to load an assembly, create an instance of a type from the assembly, create an event handler using a dynamic assembly, and hook up the dynamic event handler. All actions are performed using late binding.  
+ The following example loads an assembly, creates an instance of a type from the assembly, creates an event handler using a dynamic assembly, and hooks up the dynamic event handler. All actions are performed using late binding.
```
svick: The example doesn't load the assembly anymore, because of the change in https://github.com/dotnet/samples/pull/63.

rpetrusha: Thanks for catching this, @svick!

PR #180, opened 2018-05-14
Changed list format to ordered

pkulikov:
Also, in the third step changed the last word from *checked* to *created* to be consistent with the method name and wording of the last step.

Fixes dotnet/docs#5275


pkulikov:
@rpetrusha should we open/close that PR then, as it's already waiting for 13 hours?

PR #206, opened 2018-05-18
ArgumentException.InnerException -> Exception.InnerException

ChrisMaddock:
This xref from a change I made the other day was being flagged in CI as invalid - I guess because ArgumentException doesn't override the InnerException property.


ChrisMaddock:
No worries @rpetrusha. Just spotted it in the CI for another PR, and wanted to fix it! 😄

PR #212, opened 2018-05-21
Removed ArgumentNullException

rpetrusha:
## Removed ArgumentNullException

The MemoryMappedFile creation methods only throw an ArgumentNullException when calling the CreateOrOpen and OpenExisting methods. This PR corrects exception information for MemoryMappedFile methods. 

Fixes dotnet/docs#5404


rpetrusha:
I think I've fixed the xrefs, @mairaw, so this should build.

PR #231, opened 2018-05-30
Span<T>.Enumerator: revised text about undefined behavior of Current

pkulikov:
Mention that an exception is thrown if `MoveNext` returns `false`. Also synced phrasing with the one used in the [Current property description](https://docs.microsoft.com/en-us/dotnet/api/system.span-1.enumerator.current?view=netcore-2.1)

### DIFF
--- a/xml/System/Span`1+Enumerator.xml
+++ b/xml/System/Span`1+Enumerator.xml
@@ -83,8 +83,6 @@ Unlike some other enumerator structures in .NET, the <xref:System.Span%601.Enume
 
       ]]></format>
         </remarks>
-        <exception cref="T:System.IndexOutOfRangeException">
-          The enumerator is positioned before the first item in the <see cref="System.Span`1" />.</exception> 
       </Docs>
     </Member>
     <Member MemberName="MoveNext">


xml/System/Span`1+Enumerator.xml
```diff
@@ -29,11 +29,11 @@ The C# [foreach](~/docs/csharp/language-reference/keywords/foreach-in.md) of the
...
+If <xref:System.Span%601.Enumerator.MoveNext%2A> passes the end of the <xref:System.Span%601>, the enumerator is positioned after the last item in the <xref:System.Span%601>, and <xref:System.Span%601.Enumerator.MoveNext%2A> returns `false`. When the enumerator is at this position, subsequent calls to <xref:System.Span%601.Enumerator.MoveNext%2A> also return `false`. If the call to <xref:System.Span%601.Enumerator.MoveNext%2A> returns `false`, <xref:System.Span%601.Enumerator.Current> throws an <xref:System.IndexOutOfRangeException>. You cannot set <xref:System.Span%601.Enumerator.Current> to the first item in the <xref:System.Span%601> again; you must create a new enumerator instance instead.
```
ahsonkhan: This isn't true. Current doesn't throw IndexOutOfRangeException if MoveNext returns false.

```C#
Span<byte> span = new byte[10];

var enumerator = span.GetEnumerator();
while (enumerator.MoveNext())
    Console.WriteLine(enumerator.Current);

Console.WriteLine(enumerator.MoveNext());

Console.WriteLine(enumerator.Current);
Console.WriteLine(enumerator.MoveNext());
Console.WriteLine(enumerator.MoveNext());

Console.WriteLine(enumerator.Current);
Console.WriteLine(enumerator.MoveNext());
Console.WriteLine(enumerator.MoveNext());
Console.WriteLine(enumerator.MoveNext());
Console.WriteLine(enumerator.Current);
```

**Output:**
```text
0
0
0
0
0
0
0
0
0
0
False
0
False
False
0
False
False
False
0
```

We should also update [the Current property description](https://docs.microsoft.com/en-us/dotnet/api/system.span-1.enumerator.current?view=netcore-2.1)

cc @rpetrusha, @KrzysztofCwalina, @stephentoub

pkulikov: Indeed, [source code](https://github.com/dotnet/coreclr/blob/f31097f14560b193e76a7b2e1e61af9870b5356b/src/System.Private.CoreLib/shared/System/Span.cs#L122) shows that enumerator never passes the last element and `Current` returns the last element when `MoveNext` returns `false` on non-empty spans.

@ahsonkhan thanks for the catch! 

So, `Current` throws if enumerator is just created. `Current` is undefined (I'm not sure docs should say *returns the last element of the span*) if the last call to `MoveNext` returns `false` for non-empty span.

That implementation of `Current` with `MoveNext` returning `false` differs from other enumerator implementations in .NET: throw `InvalidOperationException` for arrays or return the default value of `T` for `List<T>`. What is the reason for that?

pkulikov: I'll also rephrase this sentence then:
>If `MoveNext` passes the end of the `Span<T>`, the enumerator is positioned after the last item in the `Span<T>`, and `MoveNext` returns `false`. When the enumerator is at this position, subsequent calls to `MoveNext` also return `false`.

stephentoub: > What is the reason for that?

This is undefined behavior in general, as highlighted by the docs at https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerator-1.current?view=netframework-4.7.2:
```
Current is undefined under any of the following conditions:
...
- The last call to MoveNext returned false, which indicates the end of the collection.
...
```
There isn't one consistent approach that all enumerators in .NET take, but this behavior of just returning whatever was last current is used in other places, for example:
- https://source.dot.net/#System.Collections/System/Collections/Generic/LinkedList.cs,568
- https://source.dot.net/#System.Collections.Concurrent/System/Collections/Concurrent/ConcurrentBag.cs,1039

pkulikov: I was aware of LinkedList implementation, but thought it was an exception (generic dictionaries, SortedSet, List follow the return-default approach, while queue and stack throw exceptions). @stephentoub thank you for more examples, I didn't explore concurrent collections.

BillWagner: @pkulikov This illustrates why I get really nervous anytime we change the docs from saying that "behavior is undefined" to explaining what an implementation does. We don't want to unnecessarily constrain other implementations.

stephentoub: Yes, we should leave undefined behavior undefined.

pkulikov: So, we don't want readers to pick up the edge behavior of one implementation of `Current` and think that that same behavior is valid for other implementations, while those implementations might realize different behavior?

What should be done with the cases when exceptions are thrown, for example, [IEnumerable.Current for List<T>](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.enumerator.system-collections-ienumerator-current?view=netcore-2.1)?

Exceptions are documented in the dedicated **Exceptions** section. The topic might not have the **Remarks** section, but almost always has the **Exceptions** section. The above referenced topic has the **Remarks** section, which starts with 
>`Current` is undefined...

However, the **Exceptions** sections clearly defines what happens. Looking at the **Exceptions** section and the remarks text, one might think that "undefined behavior" always means throwing exceptions. It's not logically correct, but it's possible train of thoughts.

Two ways to deal with such a situation:
- remove **Exceptions** section and exception info: to not explain what an implementation does. Doesn't look ideal.
- sync the **Remarks** section by specifying that `Current` throws. In all other cases keep saying that "behavior is undefined" (but at least doesn't throw). Also might not be ideal. However, I'm for this variant.

What do you think?

BillWagner: @pkulikov I'm not sure yet. The point is that "behavior is undefined" **could** mean "this throws an exception". The phrase "The behavior is undefined" means that an implementation is free to do anything: it might throw an exception (any exception), or not. It might return a stale value. It might return a default value. Any implementation specific behavior satisfies the spec for "the behavior is undefined".

For developers using the API, we (likely) need to do a better job of explaining that this phrase means "we can't tell you exactly what happens, but it's almost certainly not what you want. And, even if it is, it might change across implementations or time".

pkulikov: Oh, docs are hard. Especially, if the developers don't read them and go directly to the GitHub repo to learn from the source what implementation does. In that case, "undefined behavior" in docs might mean "whatever you see in the source code might change at any moment because of the internal reasons, since that behavior is really not relevant to the API and you don't want to use it in certain cases". It's more like, behavior is not relevant.

xml/System/Span`1+Enumerator.xml
```diff
@@ -29,17 +29,17 @@ The C# [foreach](~/docs/csharp/language-reference/keywords/foreach-in.md) of the
 
 Enumerators can be used to read the data in the <xref:System.Span%601>, but they cannot be used to modify it. 
 
-Initially, the enumerator is positioned before the first element in the <xref:System.Span%601>. At this position, attempting to retrieve an element from the <xref:System.Span%601.Enumerator.Current> property throws an <xref:System.IndexOutOfRangeException>. To prevent an exception, you must call <xref:System.Span%601.Enumerator.MoveNext%2A> to advance the enumerator to the first item in the <xref:System.Span%601> before reading the value of <xref:System.Span%601.Enumerator.Current>.
+Initially, the enumerator is positioned before the first element in the <xref:System.Span%601>. At this position, <xref:System.Span%601.Enumerator.Current> is undefined. You must call <xref:System.Span%601.Enumerator.MoveNext%2A> to advance the enumerator to the first item in the <xref:System.Span%601> before reading the value of <xref:System.Span%601.Enumerator.Current>.
```
ahsonkhan: Just to confirm, did we decide to change this to "undefined" even though we throw an exception here?

pkulikov: My reasoning is that *is undefined* here and in other places communicates to the user of the API: *it's not relevant, you better not to use this API in this state, implementation might change because of some internal reasons, various implementations of that member for different classes differ*. So, yes, to avoid leaking implementation details. With that said, the `Span<T>.Enumerator.Current` page still contains the **Exceptions** section. In that case, the rule to document exceptions overrules intention to hide implementation details.  Any solution looks to be not perfect.

/cc @BillWagner @stephentoub

stephentoub: > In that case, the rule to document exceptions overrules intention to hide implementation details.

FWIW, I disagree.  The exceptions document exceptions that can occur for defined behavior.  Undefined behavior is undefined behavior and isn't relevant to any portion of the documentation, including Exceptions.

pkulikov: @stephentoub thank you for the reply. That is convincing. @BillWagner thanks for the thumbs-up.  I've removed the exceptions section from the docs of the `Current` property.

Now this PR looks to be ready to merge.

PR #172, opened 2018-05-11
Clarified undefined behaviors for SortedDictionary

pkulikov:
Related to dotnet/docs#3575

I've fixed `GetEnumerator` time complexity to O(log n) based on the fact that SortedDictionary is a [SortedSet](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.sortedset-1?view=netcore-2.0) of key/value pairs and uses SortedSet's enumerator. [`SortedSet.GetEnumerator`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.sortedset-1.getenumerator?view=netcore-2.0) is O(log n) operation.

PR #232, opened 2018-05-30
SortedList: clarified behavior of the invalidated enumerator

pkulikov:
Related to dotnet/docs#3575


BillWagner:
@pkulikov That's very close:

> - [IDictionaryEnumerator.Entry property](https://docs.microsoft.com/en-us/dotnet/api/system.collections.idictionaryenumerator.entry?view=netcore-2.0) has the **Exceptions** section. Is it correct for the interface member to have such an section? I think it should be removed.

Some exceptions are part of the contract and defined: In the case of enumerators, the fact that `Current` or `Entry` throws an exception before the first call to `MoveNext` is part of the contract. (Historical note: That's so the first call to `MoveNext` can determine an empty sequence. The first `MoveNext` returns `false` when the sequence is empty.)

> - The **Remarks** section of the aforementioned property states both things: that `Entry` is undefined and that `Entry` throws. I guess, only "undefined behavior" should be mentioned.

That's a bit mixed. The fact that `Entry` throws an `InvalidOperationException` before the first call to `MoveNext` is part of the contract. That fact that `Entry` throws after `MoveNext` returns false is optional.

> - At last, the [Remarks on IDictionaryEntry interface](https://docs.microsoft.com/en-us/dotnet/api/system.collections.idictionaryenumerator?view=netcore-2.0#remarks) talk about the `Current` property that throws an exception. I'm not sure if the `Entry` property should be mentioned here, but "undefined behavior" should be used instead of exceptions.

The same comment applies here: The fact that `Entry` or `Current` throws before the first call to `MoveNext` is part of the contract. The behavior of `Entry` or `Current` after `MoveNext` returns `false` is undefined.

I hope that makes sense.

BillWagner:
@pkulikov wrote:

> Now, I would like to revert all updates to the Current property description I've made so far in this and earlier PRs. My last commit in this PR brings back "undefined behavior" to SortedList.

For review notes, the changes from these PRs are affected:

#59 
#152 
#160 
#172 
#220 

#231 (still open)

pkulikov:
@BillWagner I'm confused by the following sentence:
>In the case of enumerators, the fact that `Current` or `Entry` throws an exception before the first call to `MoveNext` is part of the contract. 

Does it also apply to [non-generic](https://docs.microsoft.com/en-us/dotnet/api/system.collections.ienumerator.current?view=netframework-4.7.2) and [generic](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerator-1.current?view=netframework-4.7.2) enumerators?

The current documentation doesn't document any exceptions for the `Current` property of either enumerator. Moreover, in both places docs say that 
>`Current` is undefined under any of the following conditions:
>- The enumerator is positioned before the first element in the collection, immediately after the enumerator is created. `MoveNext` must be called to advance the enumerator to the first element of the collection before reading the value of `Current`.
>- The last call to `MoveNext` returned `false`, which indicates the end of the collection.
>- The enumerator is invalidated due to changes made in the collection, such as adding, modifying, or deleting elements.

The first item states that even before the first call to the `MoveNext` method, `Current` is undefined and, thus, an exception is optional.

Can it be that the contract mentioned by you was relaxed when generic collections were added? The `List<T>.Enumerator.Current` never throws exception, thus breaks the contract.

BillWagner:
@pkulikov Well, that's what I get for looking through old (possibly draft) specs for answers. 😞 That said, thank you for your diligence in following through and helping to investigate all these issues. 👏 

Let's leave any descriptions of "undefined behavior" as is. As for the remarks and exception sections: where extra information is provided, I think we're safe leaving that in place, especially in those locations where the text also explains that the behavior is undefined.

Yes, it's a bit ugly, but I think it is the best course of action. Do you agree?

pkulikov:
@BillWagner let's sync our thoughts based on some concrete examples as I might think of several interpretations of your suggestion.

First, I suggest to make descriptions of interfaces consistent and not contradictory. PR #241 does that for the non-generic `IEnumerator` interface; the generic `IEnumerator<T>` interface's docs look fine. As for the `IDictionaryEnumerator`, I suggest to sync it with the `IEnumerator` interface's docs (it's ok to assume that `Entry` behaves the same way as `Current`, so if `Current` is undefined, then `Entry` is also undefined): (1) state that behavior of `Entry`, `Key`, and `Value` is undefined (instead of throwing exception) in those cases when `Current` is undefined, (2) remove exception information related to that undefined behavior.

Second, documentation of a certain implementation. Let's take `SortedSet<T>` for the example. There are two `Current` properties:
- [generic](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.sortedset-1.enumerator.current?view=netcore-2.0)
- [non-generic](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.sortedset-1.enumerator.system-collections-ienumerator-current?view=netcore-2.0)

When I edited those descriptions, I've removed all remarks about undefined behavior and clearly described what happens in the cases that the interface's docs claim as undefined behavior. After updating more and more different collections' docs, I admit that specifying what happens in edge cases might be not a good idea. I suggest to replace phrases *...throws an `InvalidOperationException`...* and *...returns the default value of type `T`...* with *...is undefined...*. Along with that, also remove the exception info from the non-generic `Current` properties. However, as I understand you are not sure about that last suggestion about the exceptions section?

BillWagner:
@pkulikov 

> Along with that, also remove the exception info from the non-generic Current properties. However, as I understand you are not sure about that last suggestion about the exceptions section?

The more I thought about it, the more I agree with @stephentoub 's comment: undefined is undefined. Removing the exceptions section from those APIs where the behavior is undefined is the best course of action.

Thanks again for all the work on this.

BillWagner:
@pkulikov Is this still WIP, or is it ready for final review?  (If it's ready for final review and merge, just remove "wip" from the title).

pkulikov:
@BillWagner it's still WIP as I want to double-check that the work is going on the right track. Please, check the discussion in dotnet/docs#3575

PR #247, opened 2018-06-05
IEnumerator interfaces: changing collection makes enumerator's behavior undefined

pkulikov:
In this PR, I've clarified the undefined behavior of the invalidated enumerator based on the following:
- [Remarks of the `IEnumerator` interface docs](https://docs.microsoft.com/en-us/dotnet/api/system.collections.ienumerator?view=netcore-2.0#remarks) clearly state that `MoveNext` and `Reset` throw when the collection is modified.
- [Docs of the `IEnumerator.MoveNext` method](https://docs.microsoft.com/en-us/dotnet/api/system.collections.ienumerator.movenext?view=netcore-2.0) mention `InvalidOperationException` both in the exceptions and the remarks sections.
- [Docs of the `IEnumerator.Reset` method](https://docs.microsoft.com/en-us/dotnet/api/system.collections.ienumerator.reset?view=netcore-2.0) also clearly say what happens when the collection is modified.
- [Remarks of the generic `IEnumerator<T>` interface docs](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerator-1?view=netcore-2.0#remarks) also don't forget to mention `InvalidOperationException` as a response to a collection change.
- At last, the discussion in dotnet/docs#3575 makes me think that the behavior is really defined.

@BillWagner please review.
@safern @ianhays @stephentoub please confirm that this PR updates docs in the correct direction.

### DIFF
--- a/xml/System.Collections.Generic/IEnumerator`1.xml
+++ b/xml/System.Collections.Generic/IEnumerator`1.xml
@@ -58,7 +58,7 @@
   
  The <xref:System.Collections.IEnumerator.Reset%2A> method is provided for COM interoperability. It does not necessarily need to be implemented; instead, the implementer can simply throw a <xref:System.NotSupportedException>. However, if you choose to do this, you should make sure no callers are relying on the <xref:System.Collections.IEnumerator.Reset%2A> functionality.  
   
- An enumerator remains valid as long as the collection remains unchanged. If changes are made to the collection, such as adding, modifying, or deleting elements, the enumerator is irrecoverably invalidated and the next call to <xref:System.Collections.IEnumerator.MoveNext%2A> or <xref:System.Collections.IEnumerator.Reset%2A> throws an <xref:System.InvalidOperationException>.  
+ If changes are made to the collection, such as adding, modifying, or deleting elements, the behavior of the enumerator is undefined.  
   
  The enumerator does not have exclusive access to the collection; therefore, enumerating through a collection is intrinsically not a thread-safe procedure. To guarantee thread safety during enumeration, you can lock the collection during the entire enumeration. To allow the collection to be accessed by multiple threads for reading and writing, you must implement your own synchronization.  
   


xml/System.Collections.Generic/IEnumerable`1.xml
```diff
@@ -110,6 +110,8 @@
...
   
  If <xref:System.Collections.IEnumerator.MoveNext%2A> passes the end of the collection, the enumerator is positioned after the last element in the collection and <xref:System.Collections.IEnumerator.MoveNext%2A> returns `false`. When the enumerator is at this position, subsequent calls to <xref:System.Collections.IEnumerator.MoveNext%2A> also return `false`. If the last call to <xref:System.Collections.IEnumerator.MoveNext%2A> returned `false`, <xref:System.Collections.Generic.IEnumerator%601.Current%2A> is undefined. You cannot set <xref:System.Collections.Generic.IEnumerator%601.Current%2A> to the first element of the collection again; you must create a new enumerator instance instead.  
+
+ An enumerator remains valid as long as the collection remains unchanged. If changes are made to the collection, such as adding, modifying, or deleting elements, the enumerator is irrecoverably invalidated and the next call to <xref:System.Collections.IEnumerator.MoveNext%2A> or <xref:System.Collections.IEnumerator.Reset%2A> throws an <xref:System.InvalidOperationException>.
```
stephentoub: This is not true for all enumerators.

xml/System.Collections/IEnumerable.xml
```diff
@@ -108,7 +108,7 @@
...
   
- An enumerator remains valid as long as the collection remains unchanged. If changes are made to the collection, such as adding, modifying, or deleting elements, the enumerator is irrecoverably invalidated and its behavior is undefined.  
+ An enumerator remains valid as long as the collection remains unchanged. If changes are made to the collection, such as adding, modifying, or deleting elements, the enumerator is irrecoverably invalidated and the next call to <xref:System.Collections.IEnumerator.MoveNext%2A> or <xref:System.Collections.IEnumerator.Reset%2A> throws an <xref:System.InvalidOperationException>.
```
stephentoub: Ditto.

xml/System.Collections.Generic/IEnumerable`1.xml
BillWagner: So, in cases, this should change to "the behavior is undefined".

stephentoub: In this case, I think it's undefined behavior by default, but a particular implementation might document it otherwise.  For example, `ConcurrentQueue<T>.GetEnumerator()` has this in the remarks:
https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentqueue-1.getenumerator?view=netframework-4.7.2#System_Collections_Concurrent_ConcurrentQueue_1_GetEnumerator
"The enumeration represents a moment-in-time snapshot of the contents of the queue. It does not reflect any updates to the collection after GetEnumerator was called. The enumerator is safe to use concurrently with reads from and writes to the queue."

pkulikov: @stephentoub thank you for the reply, I would like to clarify two points.

> it's undefined behavior by default

If default is undefined behavior, then the documentation of `IEnumerator` and `IEnumerator<T>` interfaces should be updated to reflect that? If yes, I'll make that in the scope of this PR.

> ...but a particular implementation might document it otherwise.

So, that's fine if the documentation of a particular implementation mentions `InvalidOperationException`? For example, see [remarks of `Queue<T>.Enumerator.MoveNext` method](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.queue-1.enumerator.movenext?view=netcore-2.0#remarks). Also that means the decision made in dotnet/docs#3575 to specify behavior of the invalidated enumerator for particular implementations is fine.

Is my understanding correct?

pkulikov: By the way, thank you for the example with the `ConcurrentQueue`: it shows that the following wording
> ...the enumerator is irrecoverably invalidated...

is also not true for all enumerators.

BillWagner: > So, that's fine if the documentation of a particular implementation mentions InvalidOperationException?

It depends.  This is where this entire rabbit hole started.  When the spec, or tests declare the implementation behavior, then yes. However, if I'm writing docs based on looking at what *the current code does*, I check with the product team to make sure it's by design. I don't want to authoritatively document behavior when it might change because of performance improvements, xplat updates or other future concerns. 

This exercise got really complicated because of the interfaces involved: The interface docs need to be very general, and each implementation may or may not be more specific. And IEnumerable / IEnumerable<T> are implemented by many different types.

pkulikov: I suspect the story about this side of `IEnumerator` has evolved the following way. At the beginning (.NET Framework 1.0 and 2.0 (generics)), the specs demanded the enumerator to be invalidated (and, thus, throw exceptions at `MoveNext` and `Reset` calls) whenever the collection was modified. I think that explains how this information got into the documentation of the `IEnumerator` and `IEnumerator<T>` interfaces; also might be the reason why collection implementations from the `System.Collections.Generic` namespace maintain the `version` field that allows invalidating an enumerator when the collection is modified; this part of implementation is quite consistent for different collections types, not like `Current` property, implementations of which have different flavors even within the `System.Collections.Generic` namespace . So, I assume that throwing `InvalidOperationException` by the collections from the `System.Collections.Generic` namespace is by design and might be documented. However, I would be happy to get confirmation of that from the product team.

Then, the concurrent collections were added to the framework (.NET Framework 4?). Their implementations handled the collection modifications in a different way that immediately made the documentation of the `IEnumerator` and `IEnumerator<T>` interfaces not correct. Thus, the interface docs must be made more general than they are now.

pkulikov:
@BillWagner @stephentoub I've updated the interfaces docs: the enumerator's behavior is undefined if changes are made to a collection. Exception and invalidation remarks are removed.

Please review made changes.
If review is fine, this PR is ready for merge.

As for the following question for the collections from the `System.Collections.Generic` namespace:
>So, that's fine if the documentation of a particular implementation mentions `InvalidOperationException`?

If there is no definite answer on it yet, I'll continue the discussion in dotnet/docs#3575

PR #264, opened 2018-06-13
Update InputEventArgs.xml

MichaelRumpler:
Add an explanation what int Timestamp actually represents.

## Summary

Integer is a bad type for a timestamp. If you still use it for performance reasons, it should be explained, what that int really means. It could be ticks, seconds since 1970 or anything else.

### DIFF
--- a/xml/System.Windows.Input/InputEventArgs.xml
+++ b/xml/System.Windows.Input/InputEventArgs.xml
@@ -126,7 +126,7 @@
       </ReturnValue>
       <Docs>
         <summary>Gets the time when this event occurred.</summary>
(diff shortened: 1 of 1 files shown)


xml/System.Windows.Input/InputEventArgs.xml
```diff
@@ -126,9 +126,9 @@
       </ReturnValue>
       <Docs>
         <summary>Gets the time when this event occurred.</summary>
-        <value>The timestamp.</value>
+        <value>The milliseconds since last reboot. After about 24.9 days this value reaches Int32.MaxValue and restarts at (the negative) Int32.MinValue.</value>
```
rpetrusha: - Better: The milliseconds since last --> The number of milliseconds that have elapsed since the last 
- Int32.MaxValue can be a link: \<xref:System.Int32.MaxValue?displayProperty=nameWithType>
- Int32.MinValue can be a link: \<xref:System.Int32.MinValue?displayProperty=nameWithType>

PR #267, opened 2018-06-14
Removed example of obsolete member

rpetrusha:
## Removed example of obsolete member

A customer commented that the example didn't use the IPAddress.Address property, which is true. The example description also didn't correspond to the example code. Since the property is deprecated, I've chosen to remove the example rather than create one.


rpetrusha:
Yes, I definitely agree, @BillWagner.

PR #284, opened 2018-06-18
Fix xrefs in Window.WindowStartupLocation

ChrisMaddock:
## Summary

We've had some issues in production with trying to have a WPF window startup in the centre of a Winforms owner, using the `WindowStartupLocation` property.

Eventually, we stumbled across this page:

https://blogs.msdn.microsoft.com/wpfsdk/2007/04/03/centering-wpf-windows-with-wpf-and-non-wpf-owner-windows/

Which states:

> When a WPF window has a non-WPF owner, WPF honors the owner window/owned window relationship eg when a non-WPF owner window is minimized, the owned WPF window follows suit. However, if the owned WPF window has its WindowStartupLocation property set to WindowStartupLocation.CenterOwner, WPF does not center the owned WPF window over the non-WPF owner window. Instead, you’ll need to set the WindowStartupLocation of the owned window to WindowStartupLocation.Manual and manually calculate its Top and Left properties with values that make it appear centered:

The blogs page matches our experience, and makes me believe the docs.microsoft page is incorrect (as per edit) - however it would be good to have that verified from someone in the know, as it may be us 'using it wrong'!

### DIFF
--- a/xml/System.Windows/Window.xml
+++ b/xml/System.Windows/Window.xml
@@ -2363,13 +2363,11 @@
  Setting the `WindowStartupLocation` property to <xref:System.Windows.WindowStartupLocation.CenterScreen> causes a window to be positioned in the center of the screen that contains the mouse cursor.  
   
  Setting the `WindowStartupLocation` property to <xref:System.Windows.WindowStartupLocation.CenterOwner> causes a window to be positioned in the center of its owner window (see <xref:System.Windows.Window.Owner%2A?displayProperty=nameWithType>), if specified. The owner window can be either another WPF window or a non-WPF window.
-  
- If an owner window is not specified, the position of the window is determined in the same way as if the `WindowStartupLocation` property is set to <xref:System.Windows.WindowStartupLocation.Manual>.  
-  
+
 > [!NOTE]   
 >  For more information about WPF windows with non-WPF windows, see [WPF and Win32 Interoperation](~/docs/framework/wpf/advanced/wpf-and-win32-interoperation.md) and <xref:System.Windows.Interop.WindowInteropHelper>. 
 
-If an owner window is not specified, the position of the window is determined in the same way as if the `WindowStartupLocation` property is set to <xref:System.Windows.WindowStartupLocation.Manual>.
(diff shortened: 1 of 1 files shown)


BillWagner:
@SamBent Can you comment on this PR? Is this correct?

xml/System.Windows/Window.xml
```diff
@@ -2362,10 +2362,7 @@
   
  Setting <xref:System.Windows.WindowStartupLocation.CenterScreen> causes a window to be positioned in the center of the screen that contains the mouse cursor.  
   
- Setting <xref:System.Windows.WindowStartupLocation> to <xref:System.Windows.WindowStartupLocation.CenterOwner> causes a window to be positioned in the center of its owner window (see <xref:System.Windows.Window.Owner%2A>), if specified. The owner window can be either another WPF window or a non-WPF window.  
-  
-> [!NOTE]
->  For more information about WPF windows with non-WPF windows, see [WPF and Win32 Interoperation](~/docs/framework/wpf/advanced/wpf-and-win32-interoperation.md) and <xref:System.Windows.Interop.WindowInteropHelper>.  
+ Setting <xref:System.Windows.WindowStartupLocation> to <xref:System.Windows.WindowStartupLocation.CenterOwner> causes a window to be positioned in the center of its owner window (see <xref:System.Windows.Window.Owner%2A>), if specified. The owner window must be a WPF window.
```
rpetrusha: Two nits unrelated to your PR, but important to fix:
- The first xref should be \<xref:System.Windows.Window.WindowStartupLocation>. Or better, just, "The \`WindowStartupLocation\` property".
- The second xref should end with "?displayProperty=nameWithType"

ChrisMaddock: Thanks! I’m away at the moment, but will get back to this.

ChrisMaddock: Done - good spot @rpetrusha! I also change a couple of other cases of the same enum/property xref mix-up which were out of view of the diff.

SamBent:
https://developercommunity.visualstudio.com/spaces/61/index.html - click “Report a problem”.   Repro will be essential, as it looks like your scenario is different somehow from what others are doing.

ChrisMaddock:
Thanks @SamBent

PR #312, opened 2018-06-26
Corrected links to downloadable samples

rpetrusha:
## Corrected links to downloadable samples

Our downloadable samples were moved from dotnet/docs to the dotnet/samples repo, but we forgot to update the links.


rpetrusha:
I've added issue dotnet/docs#6147

PR #316, opened 2018-06-27
Clarify what 'namespace-qualified assembly' is

ddizh:
fix for dotnet/docs#6125

but personally I feel like there is no need for these remarks at all. I can see at least three places where the doc describes what `namespace-qualified name` is and what `assembly-qualified name` is and the difference. Also it clearly describes which one should be used in order to get type from an assembly other then current and/or mscorlib.

/cc @BillWagner @StevenTCramer


StevenTCramer:
@d-dizhevsky  Thanks for the update.  I agree that the section stating.

"You can use the GetType method to obtain a Type object for a type in another assembly, if you know its namespace-qualified name."

could just be removed.  As it was contradicting the other parts of the document.  Now  at least it will be accurate although superfluous.

PR #328, opened 2018-07-04
Add macOS and Linux information to GetEnvironmentVariable/SetEnvironmentVariable

rpetrusha:
## Add macOS and Linux information to GetEnvironmentVariable/SetEnvironmentVariable

Fixes dotnet/docs#6277

//cc @jkotas @JeremyKuhne @stephentoub

### DIFF
--- a/xml/System/Environment.xml
+++ b/xml/System/Environment.xml
@@ -675,13 +675,11 @@ Environment variable names are case-sensitive on Linux and macOS but are not cas
  
 ### On Windows systems
 
- On Windows systems, the environment block of the current process includes the following environment variables:  
-  
-- All per-machine environment variables that are defined at the time the process is created, along with their values.  
-  
-- All per-user environment variables that are defined at the time the process is created, along with their values.  
+ On Windows systems, the environment block of the current process includes:
+ 
+- All environment variables that are provided to it by the parent process that created it. For example, a .NET application launched from a console window inherits all of the console window's environment variables. 
 
-- All environment variables that are provided to it by the parent process that created it. For example, if a .NET application is launched from a Console window, it inherits all of the Console window's environment variables.
+  If there is no parent process, per-machine and per-user environment variables are used instead. For example, a new console window has all per-machine and per-user environment variables defined at the time it was launched.
 
 - Any variables added to the process block while the process is running by calling either the <xref:System.Environment.SetEnvironmentVariable%28System.String%2CSystem.String%29> method or the <xref:System.Environment.SetEnvironmentVariable%28System.String%2CSystem.String%2CSystem.EnvironmentVariableTarget%29> method with a `target` value of <xref:System.EnvironmentVariableTarget.Process?displayProperty=nameWithType>. These environment variables persist until the .NET application terminates.  
   


xml/System/Environment.xml
```diff
@@ -676,11 +685,9 @@ goto :EOF
...
  If environment variables are created after the process has started, you can use this method to retrieve only those variables that were created by calling the <xref:System.Environment.SetEnvironmentVariable%28System.String%2CSystem.String%29> method or the <xref:System.Environment.SetEnvironmentVariable%28System.String%2CSystem.String%2CSystem.EnvironmentVariableTarget%29> method with a `target` value of .<xref:System.EnvironmentVariableTarget.Process?displayProperty=nameWithType>.  
   
- To retrieve all environment variables along with their values, call the <xref:System.Environment.GetEnvironmentVariables%2A> method.  
-  
- Environment variable names are not case-sensitive.  
-  
-   
+### On macOS and Linux systems
+
+On macOS and Linux, the `GetEnvironmentVariable(String)` method retrieves the value of any environment variable that is defined within the scope of the dotnet process. Once the dotnet process ends, that environment variable ceases to exist.  .NET Core does not support per-machine or per-user environment variables.
```
stephentoub: This description might be misleading.  A process inherits the environment variables provided to it by the parent process that created it, e.g. if it's launched from the shell, it's going to have whatever environment variables were configured for that user.  I want to make sure we don't imply that the only environment variables available are those explicitly set by a call to Environment.SetEnvironmentVariable.

xml/System/Environment.xml
```diff
@@ -666,7 +666,16 @@ goto :EOF
...
+
+The <xref:System.Environment.GetEnvironmentVariable%28System.String%29> method retrieves an environment variable from the environment block of the current process only. It is equivalent to calling the <xref:System.Environment.GetEnvironmentVariable%28System.String%2CSystem.EnvironmentVariableTarget%29> method with a `target` value of <xref:System.EnvironmentVariableTarget.Process?displayProperty=nameWithType>. 
+
+To retrieve all environment variables along with their values, call the <xref:System.Environment.GetEnvironmentVariables%2A> method.  
+  
+Environment variable names are not case-sensitive.  
+ 
+### On Windows systems
+
+ On Windows systems, the environment block of the current process includes the following environment variables:
```
jkotas: This description might be misleading as well. On both Windows and Unix, a process inherits the environment variables provided to it by the parent process that created it. 

The per-machine and per-user environment variables are used to initialize the environment block of processes with no parent (e.g. processes started from Windows Start menu - the operating system itself is the parent). The per-machine or per-user environment variable can be modified or cleared by the current process or one of its parent. The value of per-machine or per-user variable in the current process may be different from its defined value, or the current process might not have that variable set at all.

xml/System/Environment.xml
```diff
@@ -666,7 +666,16 @@ goto :EOF
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Environment.GetEnvironmentVariable%28System.String%29> method retrieves an environment variable from the environment block of the current process only. It is equivalent to calling the <xref:System.Environment.GetEnvironmentVariable%28System.String%2CSystem.EnvironmentVariableTarget%29> method with a `target` value of <xref:System.EnvironmentVariableTarget.Process?displayProperty=nameWithType>. The environment block of the current process includes the following environment variables:  
+
+The <xref:System.Environment.GetEnvironmentVariable%28System.String%29> method retrieves an environment variable from the environment block of the current process only. It is equivalent to calling the <xref:System.Environment.GetEnvironmentVariable%28System.String%2CSystem.EnvironmentVariableTarget%29> method with a `target` value of <xref:System.EnvironmentVariableTarget.Process?displayProperty=nameWithType>. 
+
+To retrieve all environment variables along with their values, call the <xref:System.Environment.GetEnvironmentVariables%2A> method.  
+  
+Environment variable names are not case-sensitive.
```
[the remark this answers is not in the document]
BillWagner: They would also be case-sensitive on Mac.

xml/System/Environment.xml
```diff
@@ -743,14 +750,21 @@ goto :EOF
...
-## Remarks  
- The `target` parameter specifies whether the environment variable is retrieved from the current process or from the Windows operating system registry key for the current user or local machine. All per-user and per-machine environment variables are automatically copied into the environment block of the current process. However, environment variables added only to the environment block of the current process  persist only for the duration of the process.  
-  
- To retrieve all environment variables along with their values, call the <xref:System.Environment.GetEnvironmentVariables%2A> method.  
+## Remarks 
+
+To retrieve all environment variables along with their values, call the <xref:System.Environment.GetEnvironmentVariables%2A> method.  
   
- Environment variable names are not case-sensitive.  
+Environment variable names are not case-sensitive.
```
BillWagner: same note on case-sensitive on unix & Mac

xml/System/Environment.xml
```diff
@@ -826,7 +840,19 @@ goto :EOF
...
+  
+- All per-machine environment variables that are defined at the time the process is created, along with their values.  
+  
+- All per-user environment variables that are defined at the time the process is created, along with their values.  
+
+- Any variables added to the process block while the process is running by calling either the <xref:System.Environment.SetEnvironmentVariable%28System.String%2CSystem.String%29> method or the <xref:System.Environment.SetEnvironmentVariable%28System.String%2CSystem.String%2CSystem.EnvironmentVariableTarget%29> method with a `target` value of <xref:System.EnvironmentVariableTarget.Process?displayProperty=nameWithType>.  
+
+### On macOS and Linux systems
+
+On MacOS and Linux, calling the `GetenvironmentVariables` method retrieves the name and value of all environment variables that are defined within the scope of the dotnet process itself. Once the dotnet process ends, that environment variables cease to exist.  .NET Core does not support per-machine or per-user environment variables.
```
BillWagner: typo:  "that environment variables" => "those environment variables"

xml/System/Environment.xml
```diff
@@ -880,11 +906,18 @@ goto :EOF
...
+
+The names and values of the environment variables are stored as key/value pairs in the returned <xref:System.Collections.IDictionary> object.  
+
+### On Windows systems
+
+On Windows systems, the `target` parameter specifies whether the source is the current process, the registry key for the current user, or the registry key for the local machine.  
+
+### On macOS and Linux systems
+
+On macOS and Linux, only a `target` value of <xref:System.EnvironmentVariableTarget.Process?displayProperty=nameWithType> is supported. Per-process environment variables are defined within the scope of the dotnet process. Once the dotnet process ends, that environment variables cease to exist.
```
BillWagner: same typo "that environment variables"

xml/System/Environment.xml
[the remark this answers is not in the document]
stephentoub: macOS is a Unix :)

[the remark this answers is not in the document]
BillWagner: Yes, but we often call it out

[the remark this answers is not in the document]
stephentoub: That's strange to me.  I can understand that if it's "Linux and macOS".  But "Unix and macOS" implies that the latter isn't part of the former, which is itself misleading IMHO.

xml/System/Environment.xml
```diff
@@ -666,21 +666,36 @@ goto :EOF
...
- Environment variable names are not case-sensitive.  
-  
-   
+Environment variable names are case-sensitive on Linux and macOS but are not case-sensitive on Windows.  
+ 
+### On Windows systems
+
+ On Windows systems, the environment block of the current process includes the following environment variables:  
+  
+- All per-machine environment variables that are defined at the time the process is created, along with their values.
```
jkotas: These two points should be deleted even for Windows. They are not accurate statements.

> All per-machine environment variables that are defined at the time the process is created, along with their values.  
> All per-user environment variables that are defined at the time the process is created, along with their values.

Instead, I would rephrase the third point to be something like this:

All environment variables that are provided to it by the parent process that created it. For example, .NET application launched from a Console window inherits all of the Console window's environment variables.

If there is no parent process, per-machine and per-user environment variables are used instead. For example, a new Console window has all per-machine and per-user environment variables defined at the time it was launched.

rpetrusha: Thanks for your review, @jkotas. I've made the modification as you suggested.

mairaw:
@BillWagner do you wanna double check that the changes look good now?

BillWagner:
I just checked the updates, and this is ready to merge @rpetrusha

PR #331, opened 2018-07-05
Expand ITaskItem.GetMetadata return value docs

richardszalay:
Makes explicit the return value when no metadata can be found.

(Migrated from [dotnet/docs/#4738](https://github.com/dotnet/docs/pull/4738))

## Summary

The behavior of ItaskItem.GetMetadata was previously ambiguous when called with no metadata is found (ie. Exception vs null vs empty string). I double checked the source before adding the note.


BillWagner:
Also, thanks for following through our migration and rewriting history to split the repos. All the :bacon: goes to you!

PR #336, opened 2018-07-10
Document that a DirectoryNotFoundException is thrown

jpreese:
## Summary

Documentation pages for _System.IO.Directory.SetLastWriteTime_ and _System.IO.Directory.SetLastWriteTimeUtc_ state that a FileNotFoundException is thrown when the directory cannot be found. After some local testing, what is actually thrown is a DirectoryNotFoundException.

A FileNotFoundException is thrown if you attempt call System.IO.Directory.SetLastWriteTime or System.IO.Directory.SetLastWriteTimeUtc on a file.

Fixes dotnet/docs#6394

### DIFF
--- a/xml/System.IO/Directory.xml
+++ b/xml/System.IO/Directory.xml
@@ -3705,9 +3705,10 @@ Directory::CreateDirectory("Public\\Html");
   
  ]]></format>
         </remarks>
-        <exception cref="T:System.IO.FileNotFoundException">The specified path was not found.</exception>
+        <exception cref="T:System.IO.FileNotFoundException">
+          <paramref name="path" /> was not found (for example, the directory doesn’t exist or it is on an unmapped drive).</exception>
         <exception cref="T:System.IO.DirectoryNotFoundException">
-          <paramref name="path" /> is invalid (for example, the directory doesn’t exist or it is on an unmapped drive).</exception>
+          <paramref name="path" /> was not found (for example, the directory doesn’t exist or it is on an unmapped drive).</exception>
         <exception cref="T:System.ArgumentException">
           <paramref name="path" /> is a zero-length string, contains only white space, or contains one or more invalid characters. You can query for invalid characters with the  <see cref="M:System.IO.Path.GetInvalidPathChars" /> method.</exception>
         <exception cref="T:System.ArgumentNullException">
@@ -3778,9 +3779,10 @@ Directory::CreateDirectory("Public\\Html");
   
  ]]></format>
         </remarks>
-        <exception cref="T:System.IO.FileNotFoundException">The specified path was not found.</exception>
+        <exception cref="T:System.IO.FileNotFoundException">
+          <paramref name="path" /> was not found (for example, the directory doesn’t exist or it is on an unmapped drive).</exception>
         <exception cref="T:System.IO.DirectoryNotFoundException">
-          <paramref name="path" /> is invalid (for example, the directory doesn’t exist or it is on an unmapped drive).</exception>
+          <paramref name="path" /> was not found (for example, the directory doesn’t exist or it is on an unmapped drive).</exception>
         <exception cref="T:System.ArgumentException">
           <paramref name="path" /> is a zero-length string, contains only white space, or contains one or more invalid characters. You can query for invalid characters with the  <see cref="M:System.IO.Path.GetInvalidPathChars" /> method.</exception>
         <exception cref="T:System.ArgumentNullException">


rpetrusha:
Thanks for contributing to the dotnet/docs repo and for submitting this pull request, @jpreese. However, I can't reproduce the DirectoryNotFoundException that you've reported. The following code, when run using both .NET Framework and .NET Core, throws FileNotFoundException rather than DirectoryNotFoundException:
``` 
using System;
using System.IO;

public class Example
{
   public static void Main()
   {
      try {
         Directory.SetLastWriteTime(@"C:\temp2\", DateTime.Now.AddDays(-3));
      }
      catch (Exception e) {
          Console.WriteLine($"{e.GetType()}: {e.Message}");
      }
      try {
         Directory.SetLastWriteTime(@"C:\temp\file1.txt", DateTime.Now.AddDays(-3));
      }
      catch (Exception e) {
          Console.WriteLine($"{e.GetType()}: {e.Message}");
      }

      try {
         Directory.SetLastWriteTimeUtc(@"C:\temp2", DateTime.UtcNow.AddDays(-3));
      }
      catch (Exception e) {
          Console.WriteLine($"{e.GetType()}: {e.Message}");
      }
      try {
         Directory.SetLastWriteTimeUtc(@"C:\temp\file1.txt", DateTime.UtcNow.AddDays(-3));
      }
      catch (Exception e) {
          Console.WriteLine($"{e.GetType()}: {e.Message}");
      }
   }
}
```
Do you have any code that repos the exception?
@JeremyKuhne, do you know whether DirectoryNotFoundException is thrown under a particular set of conditions?

JeremyKuhne:
>  do you know whether DirectoryNotFoundException is thrown under a particular set of conditions?

It will throw either DirectoryNotFound or FileNotFound depending on what the OS gives us as an error code. I just looked at NetFX and CoreFX to validate that we don't do any special mapping of File to Directory or vice-versa in this case.

Windows doesn't document when they return which error so the best we can do is say you can get either of the two. Nesting might change it (e.g. under `"C:\notthere\nothereeither\nope.txt"`), but I'd be very hesitant to attempt to "document" undocumented behavior in Windows. :)

```
        ERROR_FILE_NOT_FOUND = 0x2;
        ERROR_PATH_NOT_FOUND = 0x3;
```

In general you can get either of these two exceptions from _any_ IO activity in .NET. There are a few places where we change error codes, but mostly we always go through a helper that just funnels back what the OS gives us. On NetFX that would be this method:

https://referencesource.microsoft.com/#mscorlib/system/io/__error.cs,a2e22bc07d9bab0a

In CoreFX, it is this (for Windows):

https://source.dot.net/#System.IO.FileSystem/Common/CoreLib/System/IO/Win32Marshal.cs,8cf456a56c1e487e

The Unix code does the equivalent.

I think I suggested writing an article about catching IO errors in .NET, I'll try to see if I can find where that might be...

jpreese:
Thanks for the response, @rpetrusha. I looked into this a bit more and I think I know whats going on (at least on my machine). For your code above, I get the following result:

image

I wrote a suite of tests, and these all pass. Note that some include a DirectoryNotFoundException.

```csharp
    public class DirectoryTests
    {
        [Fact]
        public void SetLastWriteTime_FileThatDoesNotExist__ThrowsFileNotFoundException()
        {
            Assert.Throws<FileNotFoundException>(() => System.IO.Directory.SetLastWriteTime(@"C:\doesnotexist", new DateTime()));
        }

        [Fact]
        public void SetLastWriteTime_DirectoryThatDoesNotExist_TrailingSlash_ThrowsFileNotFoundException()
        {
            Assert.Throws<FileNotFoundException>(() => System.IO.Directory.SetLastWriteTime(@"C:\doesnotexist\", new DateTime()));
        }

        [Fact]
        public void SetLastWriteTime_FileInDirectoryThatDoesNotExist_ThrowsDirectoryNotFoundException()
        {
            Assert.Throws<DirectoryNotFoundException>(() => System.IO.Directory.SetLastWriteTime(@"C:\doesnot\exist", new DateTime()));
        }

        [Fact]
        public void SetLastWriteTime_FileThatDoesNotExist_WithFileExtension_ThrowsFileNotFoundException()
        {
            Assert.Throws<FileNotFoundException>(() => System.IO.Directory.SetLastWriteTime(@"C:\doesnotexist.txt", new DateTime()));
        }

        [Fact]
        public void SetLastWriteTime_FileThatDoesNotExist_NestedInDirectoryThatDoesNotExist_ThrowsDirectoryNotFoundExceptionn()
        {
            Assert.Throws<DirectoryNotFoundException>(() => System.IO.Directory.SetLastWriteTime(@"C:\doesnot\exist.txt", new DateTime()));
        }
    }
```

In summary, assuming none of these paths exist..

C:\doesnotexist - FileNotFoundException (probably assumes a file with no extension)
C:\doesnotexist\ - FileNotFoundException (this is a little strange to me)
C:\doesnot\exist - DirectoryNotFoundException
C:\doesnotexist.txt - FileNotFoundException
C:\doesnot\exist.txt - DirectoryNotFoundException

jpreese:
@rpetrusha I have pushed the requested changes.

PR #358, opened 2018-07-17
Added docs for ReadOnlySpan

Thraka:
## Summary

Documented ReadOnlySpan and ReadOnlySpan+Enumerator

Fixes dotnet/docs#4946

### DIFF
--- a/xml/System/ReadOnlySpan`1.xml
+++ b/xml/System/ReadOnlySpan`1.xml
@@ -533,11 +533,11 @@ This reference is used for pinning and must never be dereferenced.]]></format>
       <Docs>
         <param name="left">The first span to compare.</param>
         <param name="right">The second span to compare.</param>
-        <summary>Returns a value that indicates whether two <see cref="T:System.ReadOnlySpan`1" /> value types are equal.</summary>
+        <summary>Returns a value that indicates whether two <see cref="T:System.ReadOnlySpan`1" /> instances are equal.</summary>
         <returns>
-          <see langword="true" /> if the two <see cref="T:System.ReadOnlySpan`1" /> value types are equal; otherwise, <see langword="false" /></returns>
+          <see langword="true" /> if the two <see cref="T:System.ReadOnlySpan`1" /> instances are equal; otherwise, <see langword="false" /></returns>
         <remarks>
-          <format type="text/markdown"><![CDATA[This tests if two <xref:System.ReadOnlySpan`1> value types point to the same memory location, and have the same <xref:System.ReadOnlySpan`1.Length> values. This does not compare the contents of two `ReadOnlySpan` value types.]]></format>
+          <format type="text/markdown"><![CDATA[This tests if two <xref:System.ReadOnlySpan`1> instances point to the same starting memory location, and have the same <xref:System.ReadOnlySpan`1.Length> values. This does not compare the contents of two <xref:System.ReadOnlySpan`1> instances.]]></format>
         </remarks>
       </Docs>
     </Member>


mairaw:
Internal review URLs:
[ReadOnlySpan<T>](https://review.docs.microsoft.com/en-us/dotnet/api/system.readonlyspan-1?view=netcore-2.1&branch=pr-en-us-358)
[ReadOnlySpan<T>.Enumerator](https://review.docs.microsoft.com/en-us/dotnet/api/system.readonlyspan-1.enumerator?view=netcore-2.1&branch=pr-en-us-358)

Adding some folks here that reviewed the Span<T> work earlier.

/cc @stephentoub @ahsonkhan @jkotas

xml/System/ReadOnlySpan`1.xml
```diff
@@ -29,9 +29,13 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <typeparam name="T">To be added.</typeparam>
-    <summary>To be added.</summary>
-    <remarks>To be added.</remarks>
+    <typeparam name="T">The type of items in the <see cref="System.ReadOnlySpan`1" />.</typeparam>
+    <summary>Provides a type- and <see cref="System.ReadOnlySpan`1" /> memory-safe read-only representation of a contiguous region of arbitrary memory.</summary>
```
ahsonkhan: typo: `Provides a type- and <see cref="System.ReadOnlySpan1" /> memory-safe`

xml/System/ReadOnlySpan`1.xml
```diff
@@ -688,9 +733,11 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Returns the string representation of this <see cref="T:System.ReadOnlySpan`1" /> object.</summary>
+        <returns>The string representation of this object.</returns>
```
ahsonkhan: I don't think this is entirely accurate. The remark is more accurate.

The only time it returns the string representation is if T=char. For all other Ts, it returns the type and number of elements. We should say that instead since it is more generally true, and the remark should specify that for T=char, we return the string representation.

Thoughts?

xml/System/ReadOnlySpan`1.xml
```diff
@@ -716,10 +763,15 @@
         <Parameter Name="destination" Type="System.Span&lt;T&gt;" />
       </Parameters>
       <Docs>
-        <param name="destination">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <param name="destination">The target of the copy operation.</param>
+        <summary>Attempts to copy the contents of this <see cref="System.ReadOnlySpan`1" /> into a <see cref="System.Span`1" /> and returns a value to indicate whether or not operation succeeded.</summary>
```
ahsonkhan: whether or not **the** operation succeeded.

xml/System/ReadOnlySpan`1.xml
```diff
@@ -395,8 +424,10 @@
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Returns a value that indicates the current read-only span is empty.</summary>
+        <value>
+          <format type="text/markdown"><![CDATA[`true` when the current span is empty; otherwise `false`.]]></format>
```
ahsonkhan: The true and false keywords aren't formatted correctly.
image

xml/System/ReadOnlySpan`1.xml
```diff
@@ -422,16 +453,18 @@
...
       <Docs>
-        <param name="index">To be added.</param>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <param name="index">The zero-based index of the element to get.</param>
+        <summary>Gets an item from the read-only span at the specified zero-based index.</summary>
+        <value>The element from the read-only span.</value>
         <remarks>To be added.</remarks>
+        <exception cref="T:System.IndexOutOfRangeException">
+          <paramref name="index" /> is less then zero or greater than or equal to <see cref="P:System.SpanReadOnly`1.Length" />.</exception>
```
ahsonkhan: System.SpanReadOnly`1.Length is not formatted correctly in the final document.

xml/System/ReadOnlySpan`1.xml
```diff
@@ -666,9 +709,11 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Copies the contents of this read-only span into a new array.</summary>
+        <returns>An array containing the data in the current span.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[This method performs a heap allocation and therefore should be avoided if possible. It it sometimes necessary when APIs that support <xref:System.ReadOnlySpan%601> objects are not available but APIs that work with arrays are.]]></format>
```
ahsonkhan: ... performs a heap allocation **and data copy** and ...

xml/System/ReadOnlySpan`1+Enumerator.xml
```diff
@@ -44,12 +60,20 @@
...
       </Attributes>
       <ReturnValue>
-        <ReturnType>System.T&amp;</ReturnType>
+        <ReturnType>T@</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <summary>Gets a reference to the item at the current position of the enumerator.</summary>
```
ahsonkhan: `<MemberSignature Language="C#" Value="public T&amp; Current { get; }" />`

[Remove the ampersand](https://github.com/dotnet/dotnet-api-docs/pull/358/files#diff-eb540cf08a14e0c666aab66049b91c77R46), like span:https://github.com/dotnet/dotnet-api-docs/blob/88663122a2ba4a93aa7d133c3423bd477688b928/xml/System/Span%601%2BEnumerator.xml#L56

xml/System/ReadOnlySpan`1.xml
ahsonkhan: The same is true for Span.

xml/System/ReadOnlySpan`1+Enumerator.xml
Thraka: Is that normal to remove? I think all that header stuff was auto generated (I'm new to working on ref files)

ahsonkhan: Good question, I am not sure.

The signature for Span is `public ref T Current` and for ReadOnlySpan is `public ref readonly T Current`

@weshaggard, any ideas?

mairaw: If the signature is wrong, we'd need to open an issue at https://github.com/mono/api-doc-tools/ 

/cc @joelmartinez

weshaggard: I'm not sure what tool was used here to generate these but I would not be surprised at all if the tool didn't understand `ref` and `ref readonly` as they are pretty new concepts.

xml/System/ReadOnlySpan`1.xml
mairaw: don't use markdown here otherwise IntelliSense will also be broken, since they're generated from the docs. you need to use the language keyword tag.

xml/System/ReadOnlySpan`1.xml
mairaw: Probably due to the typo.
SpanReadOnly -> ReadOnlySpan

xml/System/ReadOnlySpan`1+Enumerator.xml
BillWagner: @weshaggard is correct: the tool loses the `ref` and `readonly` modifiers. See   MicrosoftDocs/feedback#468

xml/System/ReadOnlySpan`1.xml
Thraka: I think this is pretty standard wording for ToString. Remarks delves into more details about what that representation can be.

xml/System/ReadOnlySpan`1+Enumerator.xml
Thraka: what do we do about this issue?

mairaw: Sometimes, we've done workarounds to fix the docs manually here. The problem is that the next time we run CI, the workaround is gone. We can follow up with @dend and @joelmartinez to see when this issue will be picked up.

Thraka: Verified with @mairaw that this will be fixed by mdoc and not by hand editing.

xml/System/ReadOnlySpan`1.xml
```diff
@@ -29,9 +29,13 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <typeparam name="T">To be added.</typeparam>
-    <summary>To be added.</summary>
-    <remarks>To be added.</remarks>
+    <typeparam name="T">The type of items in the <see cref="System.ReadOnlySpan`1" />.</typeparam>
+    <summary>Provides a type-safe and memory-safe read-only representation of a contiguous region of arbitrary memory.</summary>
+    <remarks>
+      <format type="text/markdown"><![CDATA[`ReadOnlySpan<T>` is a [ref struct](~/docs/csharp/reference-semantics-with-value-types.md#ref-struct-type) that is allocated on the stack rather than on the managed heap. Ref struct types have a number of restrictions to ensure that they cannot be promoted to the managed heap, including that they can't be boxed and they can't be assigned to variables of type <xref:System.Object>, `dynamic`, and to any interface type.
```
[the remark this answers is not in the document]
BillWagner: Instead of "...that is allocated on the stack rather than the managed heap", I'd prefer "must be allocated on the stack and can never escape to the managed heap."  

My reason is that we should start promoting the language in the [draft C# spec](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md#draft-language-specification) for ref like structs.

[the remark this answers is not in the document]
Thraka: OK, I updated those sections. Please review to make sure I correctly described it 😄

xml/System/ReadOnlySpan`1+Enumerator.xml
```diff
@@ -22,8 +22,24 @@
...
+
+<xref:System.ReadOnlySpan`1.Enumerator.Current> returns the same value until <xref:System.ReadOnlySpan`1.Enumerator.MoveNext%2A> is called. <xref:System.ReadOnlySpan`1.Enumerator.MoveNext%2A> sets <xref:System.ReadOnlySpan`1.Enumerator.Current> to the next item in the <xref:System.ReadOnlySpan`1>.
+
+If <xref:System.ReadOnlySpan`1.Enumerator.MoveNext%2A> passes the end of the <xref:System.ReadOnlySpan`1>, <xref:System.ReadOnlySpan`1.Enumerator.MoveNext%2A> returns `false`. When the enumerator is at this state, subsequent calls to <xref:System.ReadOnlySpan`1.Enumerator.MoveNext%2A> also return `false` and <xref:System.ReadOnlySpan`1.Enumerator.Current> is undefined. You cannot set <xref:System.ReadOnlySpan`1.Enumerator.Current> to the first item in the <xref:System.ReadOnlySpan`1> again; you must create a new enumerator instance instead. 
+
+The enumerator does not have exclusive access to the <xref:System.ReadOnlySpan`1>. In addition, the underlying data on which the span is based can also be modified. Therefore, enumerating through a span is intrinsically not a thread-safe procedure. To guarantee thread safety during enumeration, you must implement your own synchronization.
```
stephentoub: I don't understand the first sentence here.  The second sentence is true, but the first sentence suggests that someone else could be accessing that same span... but span is a stack-only data structure, so the only other code that can be accessing it is the current thread.

xml/System/ReadOnlySpan`1.xml
```diff
@@ -29,9 +29,13 @@
...
   <Docs>
-    <typeparam name="T">To be added.</typeparam>
-    <summary>To be added.</summary>
-    <remarks>To be added.</remarks>
+    <typeparam name="T">The type of items in the <see cref="System.ReadOnlySpan`1" />.</typeparam>
+    <summary>Provides a type-safe and memory-safe read-only representation of a contiguous region of arbitrary memory.</summary>
+    <remarks>
+      <format type="text/markdown"><![CDATA[`ReadOnlySpan<T>` is a [ref struct](~/docs/csharp/reference-semantics-with-value-types.md#ref-struct-type) that is allocated on the stack and can never escape to the managed heap. Ref struct types have a number of restrictions to ensure that they cannot be promoted to the managed heap, including that they can't be boxed, captured in lambda expressions, assigned to variables of type <xref:System.Object>, assigned to `dynamic` variables, and they cannot implement any interface type.
+
+A `ReadOnlySpan<T>` instance is often used to hold the elements of an array or a portion of an array. Unlike an array, however, a `ReadOnlySpan<T>` instance can point to managed memory, native memory, or memory managed on the stack.]]></format>
```
stephentoub: Nit: "to hold" suggests to me that it has its own storage... rather, then span is just referencing other storage.

xml/System/ReadOnlySpan`1.xml
```diff
@@ -84,10 +90,16 @@
         <Parameter Name="length" Type="System.Int32" />
       </Parameters>
       <Docs>
-        <param name="pointer">To be added.</param>
-        <param name="length">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="pointer">A pointer to the starting address of a specified number of bytes in memory.</param>
+        <param name="length">The number of bytes to be included.</param>
```
stephentoub: This is wrong.  It's not the number of bytes, it's the number of Ts:
https://github.com/dotnet/coreclr/blob/af0b442c17fe3a9a95370e506a36ce72ee99ea80/src/System.Private.CoreLib/shared/System/ReadOnlySpan.Fast.cs#L92

xml/System/ReadOnlySpan`1.xml
```diff
@@ -84,10 +90,16 @@
...
+        <summary>Creates a new <see cref="T:System.ReadOnlySpan`1" /> object from a specified number of bytes starting at a specified memory address.</summary>
```
stephentoub: See above.

xml/System/ReadOnlySpan`1.xml
```diff
@@ -165,9 +182,13 @@
         <Parameter Name="destination" Type="System.Span&lt;T&gt;" />
       </Parameters>
       <Docs>
-        <param name="destination">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="destination">The read-only span to copy items into.</param>
+        <summary>Copies the contents of this <see cref="System.ReadOnlySpan`1" /> into a destination <see cref="System.ReadOnlySpan`1" />.</summary>
+        <remarks>
```
stephentoub: Are we using the XML comments in the source to seed all of this? e.g.
https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/ReadOnlySpan.Fast.cs#L92
There's useful material there and it seems like it's not being considered, but maybe I'm just missing it.

xml/System/ReadOnlySpan`1.xml
```diff
@@ -245,8 +266,8 @@
         <ReturnType>System.ReadOnlySpan&lt;T&gt;</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Returns an empty <see cref="T:System.ReadOnlySpan`1" /> object.</summary>
+        <value>Returns an empty <see cref="T:System.ReadOnlySpan`1" /> object</value>
```
stephentoub: "object"... is that the nomenclature we use even for value types?

xml/System/ReadOnlySpan`1.xml
```diff
@@ -107,9 +119,9 @@
         <Parameter Name="start" Type="System.Int32" />
       </Parameters>
       <Docs>
-        <param name="array">To be added.</param>
-        <param name="start">To be added.</param>
-        <summary>To be added.</summary>
+        <param name="array">The array from which to create the <see cref="T:System.ReadOnlySpan`1" /> object.</param>
+        <param name="start">The starting index from which to copy.</param>
+        <summary>Internal only. Creates a new <see cref="T:System.ReadOnlySpan`1" /> object from an array starting at the specified index.</summary>
```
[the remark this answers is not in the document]
jkotas: `ReadOnlySpan(T[] array, int start)` constructor does not even exist in the final shipped implementation, neither as public nor internal. The documentation for this method should be deleted.

> I believe for legal reasons, it must be documented.

There is no rule like does. You must be confusing it with something else.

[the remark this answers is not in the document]
Thraka: @jkotas I'll investigate. I wonder why the xml file generated with that constructor? I did find it here in the source though: https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/ReadOnlySpan.Fast.cs#L114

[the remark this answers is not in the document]
Thraka: Whoops! OK Yes you're correct. My bad! I'll have the XML file regenerated.

BillWagner:
I like the latest updates a lot @Thraka You can :shipit: when ready.

xml/System/ReadOnlySpan`1.xml
```diff
@@ -483,11 +531,14 @@
...
       </Parameters>
       <Docs>
-        <param name="left">To be added.</param>
-        <param name="right">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <param name="left">The first span to compare.</param>
+        <param name="right">The second span to compare.</param>
+        <summary>Returns a value that indicates whether two <see cref="T:System.ReadOnlySpan`1" /> objects are equal.</summary>
```
ahsonkhan: > "object"... is that the nomenclature we use even for value types?

I think @stephentoub's previous comment applies here (and elsewhere) within the doc as well.

xml/System/ReadOnlySpan`1.xml
```diff
@@ -688,9 +748,11 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Returns the string representation of this <see cref="T:System.ReadOnlySpan`1" />.</summary>
+        <returns>The string representation of this object.</returns>
```
ahsonkhan: Using "object" for value type.

xml/System/ReadOnlySpan`1.xml
Thraka: What do I use?

Thraka: @BillWagner

BillWagner: If you know type specific type, I'd use the specific type (e.g. `int`, `ReadOnlySpan<T>`, etc. If it must be a value type (but not a specific type), use the phrase "value type".

xml/System/ReadOnlySpan`1.xml
```diff
@@ -727,7 +727,7 @@ This reference is used for pinning and must never be dereferenced.]]></format>
         <summary>Copies the contents of this read-only span into a new array.</summary>
         <returns>An array containing the data in the current span.</returns>
         <remarks>
-          <format type="text/markdown"><![CDATA[This method performs a heap allocation and data copy, and therefore, should be avoided if possible. It it sometimes necessary when APIs that support <xref:System.ReadOnlySpan%601> objects are not available, but APIs that work with arrays are.]]></format>
+          <format type="text/markdown"><![CDATA[This method performs a heap allocation and data copy, and therefore, should be avoided if possible. It it sometimes necessary when APIs that support <xref:System.ReadOnlySpan%601> value types are not available, but APIs that work with arrays are.]]></format>
```
ahsonkhan: nit: this is awkward to read, imo, and could cause confusion (are we talking about the T being the value type or referring to Span itself as the value type).

How about removing `value types` all together:
"It is sometimes necessary when APIs that support `<xref:System.ReadOnlySpan%601>` are not available, ..."

Thraka: Yeah, I was trying to make it plural. I'm not sure how to word that.

Thraka: @ahsonkhan I appreciate you looking at this 😄 I changed the "value types" to "instances" which is used in other struct equals docs.

xml/System/ReadOnlySpan`1.xml
```diff
@@ -483,11 +531,14 @@
...
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <param name="left">The first span to compare.</param>
+        <param name="right">The second span to compare.</param>
+        <summary>Returns a value that indicates whether two <see cref="T:System.ReadOnlySpan`1" /> instances are equal.</summary>
+        <returns>
+          <see langword="true" /> if the two <see cref="T:System.ReadOnlySpan`1" /> instances are equal; otherwise, <see langword="false" /></returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[This tests if two <xref:System.ReadOnlySpan`1> instances point to the same starting memory location, and have the same <xref:System.ReadOnlySpan`1.Length> values. This does not compare the contents of two <xref:System.ReadOnlySpan`1> instances.]]></format>
```
ahsonkhan: I think the change from `value types` to `instances` here is great. Can we apply the same change to the other areas, like op_InEquality:
https://github.com/dotnet/dotnet-api-docs/pull/358/files#diff-6f643f9c5e5dcd33509a5c8e52ef06aeR634

And here (ToString):
https://github.com/dotnet/dotnet-api-docs/pull/358/files#diff-6f643f9c5e5dcd33509a5c8e52ef06aeR752

PR #376, opened 2018-07-23
Update with the current library configuration

dend:
Part of the MSDN migration effort.


rpetrusha:
@dend, this has enormously increased the number of build warnings from 106 to 292. Almost all the new ones are  "<member>'s name and signature are not unique." Is that to be expected?

PR #377, opened 2018-07-24
Remove obsolete information for RSAParameters

kappa7194:
## Summary

The RSAParameters fields are no longer marked as not serializable, the bit saying so should be removed as besides being wrong can lead to security vulnerabilities if one trusts the documentation and serializes the object without checking.

**N.b.** This change applies only to .NET Core ≥ 2.1.0, however I have no idea how to apply it only on that version's documentation.

Fixes dotnet/docs#6637


rpetrusha:
Thanks for contributing to the dotnet/docs repo and for submitting this PR, @kappa7194. And thanks for your review, @bartonjs. We should explicitly call out the version difference. In terms of how to do it, although the API Browser recognizes which API are present in which implementations/versions, a single topic serves all versions. So the best way to identify .NET Core 2.1+ content is to explicitly identify it as: 
\*\*.NET Core 2.1 and later only\*\*: \<text that applies to .NET Core 2.1+>

kappa7194:
I've updated the text as requested.

PR #383, opened 2018-07-25
Nullable<T>: clarified GetValueOrDefault

pkulikov:
> object's default value.

That phrasing is confusing and even might be misleading. Object is of type `Nullable<T>`; its default value is an instance with `HasValue` returning `false`. That's why, I've rephrased "object's default value" to "default value of the underlying type".


pkulikov:
@rpetrusha I've hesitated the same way as you. So, I've removed mention of the binary zeros and added clarification in the remarks with the link to the table of default value of value types.

pkulikov:
Closed by accident.

PR #387, opened 2018-07-26
Cleaned up the WaitCallback article

pkulikov:
I've removed examples because:
- those examples are primarily of the `QueueUserWorkItem` method (by the way, identical snippets are used at the `QueueUserWorkItem` API topic)
- remarks contain links to the `QueueUserWorkItem` API page and I assume if one reads the `WaitCallback` page, she/he most probably ends up at `QueueUserWorkItem` page. Even more probable that one comes to the `WaitCallback` page from the `QueueUserWorkItem` page. Thus, example duplication, which is not necessary.

Without examples, a reader can focus on the remarks section that seems to be more useful in the case of this page.

### DIFF
--- a/xml/System.Threading/WaitCallback.xml
+++ b/xml/System.Threading/WaitCallback.xml
@@ -51,6 +51,8 @@
 >  Visual Basic users can omit the <xref:System.Threading.WaitCallback> constructor, and simply use the `AddressOf` operator when passing the callback method to <xref:System.Threading.ThreadPool.QueueUserWorkItem%2A>. Visual Basic automatically calls the correct delegate constructor.  
   
  If you want to pass information to the callback method, create an object that contains the necessary information and pass it to the <xref:System.Threading.ThreadPool.QueueUserWorkItem(System.Threading.WaitCallback,System.Object)> method as the second argument. Each time the callback method executes, the `state` parameter contains this object.  
+
+ For examples that use the `WaitCallback` delegate, see the <xref:System.Threading.ThreadPool.QueueUserWorkItem%2A?displayProperty=nameWithType> method.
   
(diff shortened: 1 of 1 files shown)


xml/System.Threading/WaitCallback.xml
```diff
@@ -45,40 +45,17 @@
...
 >  Visual Basic users can omit the <xref:System.Threading.WaitCallback> constructor, and simply use the `AddressOf` operator when passing the callback method to <xref:System.Threading.ThreadPool.QueueUserWorkItem%2A>. Visual Basic automatically calls the correct delegate constructor.  
   
- If you want to pass information to your callback method, create an object that contains the necessary information and pass it to <xref:System.Threading.ThreadPool.QueueUserWorkItem%2A> when you queue your task for execution. Each time your callback method executes, the `state` parameter contains this object.  
+ If you want to pass information to the callback method, create an object that contains the necessary information and pass it to the <xref:System.Threading.ThreadPool.QueueUserWorkItem(System.Threading.WaitCallback,System.Object)> method as the second argument. Each time the callback method executes, the `state` parameter contains this object.  
   
- For more information about using callbacks to synchronize thread pool threads, see [The Managed Thread Pool](~/docs/standard/threading/the-managed-thread-pool.md).  
-  
-   
-  
-## Examples
```
rpetrusha: Although removing the example is fine, it would be good to mention and link to the examples: "For examples that use the \`WaitCallback\` delegate, see the \<xref:System.Threading.ThreadPool.QueueUserWorkItem%2A?displayProperty=nameWithType> method."

pkulikov: @rpetrusha that's a good remark. I've added that sentence to the docs. Thank you for review.

PR #392, opened 2018-07-27
Add ISO 8601 parsing examples

rpetrusha:
## Add ISO 8601 parsing examples

This PR:

- Adds an example to two DateTimeOffset.ParseExact overloads.
- Adds an example to one DateTimeOffset.TryParseExact overloads.
- Documents new DateTimeOffset.ParseExact overloads that include ReadOnlySpan<Char> parameters.
- Documents new DateTimeOffset.TryParseExact overloads that include ReadOnlySpan<Char> parameters.

A successful build depends on the examples in dotnet/samples#208.

Fixes dotnet/docs#6331

//cc @iant-travelopia

### DIFF
--- a/xml/System/DateTimeOffset.xml
+++ b/xml/System/DateTimeOffset.xml
@@ -5458,7 +5458,7 @@ Me.ToUtcDateTime().ToFileTime()
         <param name="input">A character span that contains the representation of a date and time to convert.</param>
         <param name="format">A format specifier that defines the required format of <paramref name="input" />.</param>
         <param name="formatProvider">An object that supplies culture-specific formatting information about <paramref name="input" />.</param>
-        <param name="styles">A bitwise combination of enumeration values that indicates the permitted format of <paramref name="input" />. A typical value to specify is <see cref="F:System.Globalization.DateTimeStyles.None>" /> </param>
+        <param name="styles">A bitwise combination of enumeration values that indicates the permitted format of <paramref name="input" />. A typical value to specify is <see cref="F:System.Globalization.DateTimeStyles.None" /> </param>
         <param name="result">When the method returns, contains the <see cref="T:System.DateTimeOffset" /> equivalent to the date and time of <paramref name="input" />, if the conversion succedded, or <see cref="F:System.DateTimeOffset.MinValue" /> if the conversion failed. The conversion fails if the </param>
         <summary>Converts the representation of a date and time in a character span to its <see cref="System.DateTimeOffset" /> equivalent using the specified format, culture-specific format information, and style. The format of the date and time representation must match the specied format exactly.</summary>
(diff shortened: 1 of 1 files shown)


xml/System/DateTimeOffset.xml
```diff
@@ -5472,13 +5662,23 @@ Me.ToUtcDateTime().ToFileTime()
...
  [!code-vb[System.DateTimeOffset.TryParseExact#1](~/samples/snippets/visualbasic/VS_Snippets_CLR_System/system.DateTimeOffset.TryParseExact/vb/TryParseExact.vb#1)]  
+
+The following example uses a variety of <xref:System.Globalization.DateTimeStyles> values to parse an array of strings that are expected to conform to [ISO 8601](~/docs/standard/base-types/standard-date-and-time-format-strings.md#the-round-trip-o-o-format-specifier). As the output from the example shows, strings that are in the proper format fail to parse if:
+
+- they contain white space, and an appropriate <xref:System.Globalization.DateTimeStyles> flag (such as <xref:System.Globalization.DateTimeStyles.AllowWhiteSpace?displayProperty=nameWithType> has not been supplied in the method call.
+
+- they contain date and time elements that are out of range.
+
+Strings that do not specify a UTC offset are assumed to have the offset of the local time zone (in this case, -07:00) unless the <xref:System.Globalization.DateTimeStyles.AssumeUniversal?displayProperty=nameWithType> flag is supplied in the method call. In that case, they are assumed to be Universal Coordinated Time.
```
BillWagner: Hmm. The phrase "(in this case, -7:00)" is only accurate during daylight savings time. Otherwise, it would be -8:00 UTC. These samples aren't interactive, so the behavior won't change from the comments, but it might be worth mentioning.

xml/System/DateTimeOffset.xml
```diff
@@ -3659,13 +3665,47 @@
...
+  
+-   A <xref:System.Globalization.CultureInfo> object that represents the culture based on which `input` is interpreted. The <xref:System.Globalization.DateTimeFormatInfo> object returned by its <xref:System.Globalization.CultureInfo.DateTimeFormat%2A> property defines the symbols and formatting in `input`.  
+  
+-   A <xref:System.Globalization.DateTimeFormatInfo> object that defines the format of date and time data.  
+  
+ If `formatprovider` is `null`, the <xref:System.Globalization.CultureInfo> object that corresponds to the current culture is used.  
+  
+ ]]></format>
+        </remarks>
+        <exception cref="T:System.ArgumentException">The offset is greater than 14 hours or less than -14 hours.
```
BillWagner: The limits of **14** and **-14** surprised me. I want to make sure those aren't typos.

rpetrusha:
Closing and reopening to begin new build after examples merged.

xml/System/DateTimeOffset.xml
rpetrusha: I double checked, @BillWagner. Those are still the correct ranges.

PR #394, opened 2018-07-30
CI Update - mdoc 5.7.2.2 + 1.1 assemblies removal

dend:
This PR addresses inconsistencies caused by an older version of mdoc, and also removes redundant APIs.


dend:
@mairaw it does seem like the latest build is ✅ so we should be good to go. There are some warnings about signatures not being unique but I assume it's related to the fact that we changed the entity types between frameworks (e.g. struct to class).

mairaw:
I'll check this one now.

xml/Accessibility/IAccIdentity.xml
```diff
@@ -12,15 +12,18 @@
...
-    <Attribute>
+    <Attribute FrameworkAlternate="netframework-2.0;netframework-3.0;netframework-3.5;netframework-4.0;netframework-4.5;netframework-4.5.1;netframework-4.5.2;netframework-4.6;netframework-4.6.1;netframework-4.6.2;netframework-4.7;netframework-4.7.1;netframework-4.7.2">
       <AttributeName>System.Runtime.InteropServices.Guid("7852B78D-1CFD-41C1-A615-9C0C85960B5F")</AttributeName>
     </Attribute>
-    <Attribute>
+    <Attribute FrameworkAlternate="netframework-4.0;netframework-4.5;netframework-4.5.1;netframework-4.5.2;netframework-4.6;netframework-4.6.1;netframework-4.6.2;netframework-4.7;netframework-4.7.1;netframework-4.7.2">
       <AttributeName>System.Runtime.InteropServices.InterfaceType</AttributeName>
     </Attribute>
+    <Attribute FrameworkAlternate="netframework-2.0;netframework-3.0;netframework-3.5">
+      <AttributeName>System.Runtime.InteropServices.InterfaceType(1)</AttributeName>
```
mairaw: this new entry looks a bit odd @dend @joelmartinez 

From what I see on MSDN, .net 3.5 was System.Runtime.InteropServices.InterfaceType and later versions InterfaceTypeAttribute(ComInterfaceType.InterfaceIsIUnknown). Shouldn't that be showing here?

mairaw:
@dend also please make sure this API is included in the test cases for the front-end work of frameworkalternates to avoid dupe attributes being listed:
https://review.docs.microsoft.com/en-us/dotnet/api/accessibility.iaccpropserver?view=netframework-4.7.2&branch=pr-en-us-394

xml/Accessibility/IAccIdentity.xml
joelmartinez: @mairaw the `(1)` is likely the value of `ComInterfaceType.InterfaceIsIUnknown`, and the formatter always drops the `Attribute` suffix on attributes, since you can omit that when you apply an attribute in the source code. 

As to why it didn't expand the enum name, that might be a different issue, but "strictly speaking" this seems correct.

mairaw: @joelmartinez but the later versions are the ones that had the parameter and earlier versions were parameterless. So something looks not quite right here.

joelmartinez: @mairaw I just loaded up and inspected the `Accessibility.dll` from the `netframework-2.0` folder.
image
This is in the VSMac assembly browser, and you can see that it lists `[InterfaceType (1)]` :)

mairaw: You're right. MSDN has that for 2.0 as well. Can you check the 3.5 version since you have that assembly browser handy? That's the one I was looking at this morning on MSDN.

mairaw: Thank you for looking into this @joelmartinez.

joelmartinez: Of course! 👏

mairaw:
@dend I don't really understand why the change here would make the members to be listed twice at https://review.docs.microsoft.com/en-us/dotnet/api/system.servicemodel.clientbase-1.channelbase-1?view=netframework-4.7.2&branch=.pr-en-us-394#explicit-interface-implementations.

I can only see one entry for IRequestChannel.Via, for example, in the xml file.

mairaw:
Also, how do I fix the moniker2assembly mapping issue? Not sure where that is configured.

dend:
@mairaw I am actually not sure why we have two of the same member. @joelmartinez - do you know what is the significance of the `ToDelete` node attribute?

joelmartinez:
@dend @mairaw This seems like perhaps it was a confluence of two separate changes that didn't interact well (the EII name change, and the dupe removal enhancement). I will work on a mitigation to recover from this state.

dend:
@joelmartinez @mairaw I kicked off [a new build](https://apidrop.visualstudio.com/Content%20CI/_build/results?buildId=26303&_a=summary&view=logs) and closed this PR to get a fresh view.

mairaw:
Perfect @dend. I was just going to ask you about that.

PR #396, opened 2018-07-30
Interlocked class: added missing words

pkulikov:
@BillWagner @rpetrusha related question: both [Interlocked.Increment](https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.increment?view=netcore-2.1) and [Interlocked.Decrement](https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.decrement?view=netcore-2.1) pages contain two examples: one using `Thread` API, another is based on the `Task` class. Should we remove the `Thread` API examples at all, as using TPL is recommended?


rpetrusha:
I would leave the threading example, but move it to follow the task-based example. Do you agree, @BillWagner?

PR #424, opened 2018-08-08
Update NetNamedPipeBinding.xml

chrysikos:
Important info about port concept.

### DIFF
--- a/xml/System.ServiceModel/NetNamedPipeBinding.xml
+++ b/xml/System.ServiceModel/NetNamedPipeBinding.xml
@@ -23,6 +23,8 @@
     <remarks>
       <format type="text/markdown"><![CDATA[  
   
+> [!IMPORTANT]
+>  For NetNamedPipeBinding there is no the notion of ports. Trying to add port number <xref:System.ArgumentException> will be thrown on process of parsing the address.
 ## Remarks  
  The <xref:System.ServiceModel.NetNamedPipeBinding> generates a run-time communication stack by default, which uses transport security, named pipes for message delivery, and a binary message encoding. This binding is an appropriate Windows Communication Foundation (WCF) system-provided choice for on-machine communication. It also supports transactions.  
   


rpetrusha:
@davidsh, could you take a look, please?

rpetrusha:
@mconnew and @Lxiamail, could you review this please?

mconnew:
Besides the grammar issues, I don't believe this change needs to be made for several reasons.  
1. NetNamedPipeBinding doesn't accept a URI on any api's so talking about the supported URI's here seems out of place.
2. NetNamedPipeBinding isn't a transport, it's a standard binding which combines binding elements together in any easy way. NamedPipeTransportBindingElement is the actual transport binding element which cares about the URI. Even that class doesn't have any api's which use a URI, that would be the ChannelFactory created by NamedPipeTransportBindingElement.  
3. Documenting this is setting a precedent. Would we need to document NetMsmqBinding, MsmqIntegrationBinding and NetPeerTcpBinding?
4. The port number of the URI is just one of many differences about the URI between different transports. For example, using a URL fragment isn't applicable to anything other than HTTP. Documenting core concepts such as the HTTP protocol or what is a named pipe is out of scope for WCF documentation. 
5. A port number is only applicable when using a TCP connection (which HTTP and Net.Tcp both do). It's clearly documented that a named pipe is used, which is not using TCP/IP so a port number is not applicable. The documentation on the same page says (emphasis mine):
> The NetNamedPipeBinding generates a run-time communication stack by default, which uses transport security, **named pipes** for message delivery

Basically it's clearly documented that this uses named pipes, and it's expected that if someone is choosing to use a named pipe for their transport, that they have a basic understanding of what a named pipe is before using it. Without that basic understanding, you can't make an educated decision about if the transport is the right one to use.

rpetrusha:
Thank you for contributing to the dotnet/docs repo, @chrysikos. However, based on the review from @mconnew, I'm afraid that we can't accept your PR, so I'm going to close it.

PR #425, opened 2018-08-08
Typo fix

pkulikov:
Or it's not a typo, but a way to mark the C++ related note?


pkulikov:
@rpetrusha indeed that's better; thank you for review. I've merged two notes.

PR #440, opened 2018-08-16
specify what the Multiply(Vector3, Vector3) operator overload does

Double-A-92:
The Multiply(Vector3, Vector3) operator overload seems to multiply the two given Vector3s element-wise. Added this information to the docs so we don't have to guess and try it out anymore. 👍

### DIFF
--- a/xml/System.Numerics/Vector`1.xml
+++ b/xml/System.Numerics/Vector`1.xml
@@ -1264,8 +1264,8 @@
       <Docs>
         <param name="left">The first vector.</param>
         <param name="right">The second vector.</param>
-        <summary>Multiplies two vectors together.</summary>
-        <returns>The product vector.</returns>
+        <summary>Returns a new vector whose values are the product of each pair of elements in two specified vectors.</summary>
+        <returns>The element-wise product vector.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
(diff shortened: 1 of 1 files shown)


BillWagner:
/cc @tannergooding @VicktorHofer

These changes LGTM. Can you check that this is how the `Vector3.Multiply` method should be described?

tannergooding:
@Double-A-92, would you also be willing to fix-up the documentation for `Vector2` and `Vector4`, as well as the `op_Multiply(Vector, Vector)` methods for each?

It would be nice to get them all fixed up at once, if possible.

Double-A-92:
@tannergooding Sure, I could do that. Do you think the text / description is good as it is in this PR?

BillWagner:
@Double-A-92 

> Do you think the text / description is good as it is in this PR?

I would say, yes, the description is fine. Thank you for adding the other related classes as well.

Double-A-92:
@tannergooding  @BillWagner Done.
I found an already existing description for that in the Vector (without number), and used that for all of the others.

Just out of curiosity... What happens to the in-code documentation comments? Will they get automatically updated based on this repository?

BillWagner:
@Double-A-92 

> What happens to the in-code documentation comments? Will they get automatically updated based on this repository?

It depends on the repository. In the case of the APIs that have multiple implementations (.NET Core, .NET Framework), the XML files were exported once, and that's now considered our source of truth. In many other repositories, the /// comments are the source of truth, and we don't edit the exported XML.  The tools we use are being updated to make this more clear and easier to work with.

Usually, if the docs repo is open for public contribution, the XML files in the docs repo are the source of truth. The README.MD or CONTRIBUTING.MD should also say. If you find any that aren't explicit, that's an issue we should fix.

PR #465, opened 2018-08-23
Corrected type self-reference

rpetrusha:
## Corrected type self-reference

Fixes dotnet/docs#7235


rpetrusha:
@krwq and @pjanotti, I've made the following changes:
- Removed the sentence that XmlMapping has no unique public or protected members, which is clearly not true.
- Changed the self-referencing statement to use XmlMapping instead of deriving from XmlMapping to recommend using XmlTypeMapping and XmlMemberMapping instead of deriving from XmlMapping.
 Is that OK?

PR #471, opened 2018-08-26
Update SpeechSynthesizer.xml

mikkelbu:
Corrected comment and made code format properly.

Fixes #7281


mikkelbu:
@rpetrusha Note that this should close the issue dotnet/docs#7281. I did not notice that the commit was in dotnet-api-docs and not docs as I was using the github interface.

PR #480, opened 2018-08-30
Removed reference to article without link

rpetrusha:
## Removed reference to article without link

Since I can't find the article anywhere in our doc sets, I've removed the unlinked mention to it.

Fixes dotnet/docs#7384


Thraka:
@kbdguy  @rpetrusha I found the article it was linking to: https://msdn.microsoft.com/en-us/library/bb384378

Should it be migrated? Is it slated for migration?

PR #486, opened 2018-08-31
Link to overview

breyed:
The docs page for the Dataflow namespace links to the dataflow overview, but none of the class pages do. You can get from a class page to the overview in two steps, but you'd have to be lucky enough to guess that you should click on the namespace first. Until you do, you're left wondering what makes a BufferBlock different from an array, a queue, or any other "buffer for storing data".

The PR shows one way of making the overview more discoverable. The PR should be updated as needed to become the best way to provide the link and then expanded to cover all dataflow classes.


rpetrusha:
Thanks, @breyed. I've created issue dotnet/docs#7544 to track this issue, and added it to the community contributors' project.

PR #491, opened 2018-09-03
Math.DivRem: remainder is calculated according to the remainder operator

pkulikov:
Not to confuse with the canonical modulus :).

.NET Framework code:
https://referencesource.microsoft.com/#mscorlib/system/math.cs,607

.NET Core code (according to the [spec of the remainder operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#remainder-operator)):
https://github.com/dotnet/corefx/blob/2dde65cc733a1b7f4560a7246b2b959710f5b784/src/Common/src/CoreLib/System/Math.cs#L113-L122

Because of the .NET Core implementation, I don't want to say *how* the remainder is calculated. However, it's important to mention to *what* value it calculates.

### DIFF
--- a/xml/System/Math.xml
+++ b/xml/System/Math.xml
@@ -1598,10 +1598,8 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The remainder is calculated by using the modulus operation.  
-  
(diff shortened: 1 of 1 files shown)


pkulikov:
Let's see if close/reopen would help.

PR #493, opened 2018-09-04
Escaped backslash in MaskedTextBox.Mask table

rpetrusha:
## Escaped backslash in MaskedTextBox.Mask table

Fixes https://github.com/dotnet/docs/issues/7473


mairaw:
The fix didn't seem to work:
image

Maybe you need to do something similar to https://github.com/dotnet/docs/pull/7478/files?

PR #499, opened 2018-09-05
Clarified versions affected in StringBuilder.MaxCapacity note

rpetrusha:
## Clarified versions affected in StringBuilder.MaxCapacity note

Under certain circumstances, a StringBuilder instance can grow beyond its predefined maximum capacity. This PR;

- Creates an include file with the boilerplate text, since it occurs in 33 places in the documentation.
- Notes that this issue applies to .NET Core and to .NET Framework 4.0 and later. Previously, the documentation indicated that this applied to .NET Framework 4.0 and 4.5.

Fixes dotnet/docs#6979

### DIFF
--- a/xml/System.Text/StringBuilder.xml
+++ b/xml/System.Text/StringBuilder.xml
@@ -273,7 +273,7 @@
  ]]></format>
     </remarks>
     <block subset="none" type="usage">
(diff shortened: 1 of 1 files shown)


rpetrusha:
This is ready for another review, @mairaw

PR #513, opened 2018-09-09
Updated InternalList property.

pcnova:
I have a collection that inherits from this class, and while debugging, I discovered this property actually returns the opposite of what its current documentation says (at least on .NET 4.6.1, but from looking at the reference source for 4.7.2, it looks like it's still the same). It seems to me like the documentation was just paraphrased from what the base's SourceCollection property says.

## Summary

Updated InternalList property to reflect its actual behavior.

### DIFF
--- a/xml/System.Windows.Data/ListCollectionView.xml
+++ b/xml/System.Windows.Data/ListCollectionView.xml
@@ -1143,8 +1143,8 @@
         <ReturnType>System.Collections.IList</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets the complete and unfiltered underlying collection.</summary>
-        <value>The underlying collection, which must implement <see cref="T:System.Collections.IList" />.</value>
+        <summary>Gets the filtered and sorted list of items.</summary>
+        <value>The <see cref="T:System.Collections.IList" /> on which filtering and sorting have been applied.</value>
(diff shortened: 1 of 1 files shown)


rpetrusha:
//cc @douglaslMS

douglaslMS:
@pcnova I apologize that I missed your PR while I was on vacation. Since it would take more than a few minutes for me to write a quick test in C#, I searched for other discussions of this property on the web. I found at least this one StackOverflow thread that seems to contradict your observations in the final comment, "I cannot use ... internal list, since it contains all items, not the filtered/sorted. ": [Iterating through ListCollectionView using foreach](https://stackoverflow.com/questions/10670483/iterating-through-listcollectionview-using-foreach)

douglaslMS:
@pcnova Do you have any comments in response to my previous comment? If the existing description of the property is correct, this does not surprise me - I would expect the class to keep a "hidden" copy of the unfiltered list.

douglaslMS:
@pcnova Thank you for the simple sample. I'll follow up by tomorrow.

douglaslMS:
@pcnova Thanks again for providing a sample to save me some time.

pcnova:
I just came back to this and saw my updates applied; you're welcome and thank you as well!

PR #523, opened 2018-09-13
 CI Update - mdoc 5.7.3.2

dend:

### DIFF
--- a/xml/System.Net/HttpStatusCode.xml
+++ b/xml/System.Net/HttpStatusCode.xml
@@ -24,10 +24,12 @@
     <AssemblyName>netstandard</AssemblyName>
     <AssemblyVersion>2.0.0.0</AssemblyVersion>
   </AssemblyInfo>
+  <!-- type forward info -->
   <AssemblyInfo>
     <AssemblyName>System.Net</AssemblyName>
     <AssemblyVersion>4.0.0.0</AssemblyVersion>
   </AssemblyInfo>
+  <!-- end type forward info -->
   <Base>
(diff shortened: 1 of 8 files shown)


mairaw:
Before merging, I need to check what happened to the APIs I've added the APIscan info to at PR #477

dend:
@mairaw I see that the build is green - is there anything that is missing?

mairaw:
I need to check the TypeForwardInfo stuff and I won't be able to get to this until next week

xml/FrameworksIndex/netframework-4.7.2.xml
```diff
@@ -15266,6 +15266,7 @@
     </Type>
     <Type Name="System.Lazy`1" Id="T:System.Lazy`1">
       <Member Id="M:System.Lazy`1.#ctor" />
+      <Member Id="M:System.Lazy`1.#ctor(`0)" />
```
mairaw: I don't understand where this new member is coming from @dend @joelmartinez

We only had 6 ctors on MSDN and this adds a 7th. I don't see this as well on ildasm?

image

joelmartinez: Ok ... so that particular ctor can be found in netcore 2 and 2.1. So it's a valid ctor, however, I'm not sure how it came to be referenced in that framework index for 4.7. I'm going to look into that now

mairaw: Thanks @joelmartinez. I'll merge this and then open a bug.

PR #563, opened 2018-09-27
Documented Path.Join

rpetrusha:
## Documented Path.Join

The PR:
- Documents the Path.Join overloads.
- Documents the Path.GetFullPath(String,String) method.
- Notes that Path.Combine is intended to be used with a base path and a relative path.
- Adds links to the Windows file path format topic to each method with a string path parameter.

Related to dotnet/samples#328

Fixes dotnet/dotnet-api-docs#970
Fixes dotnet/dotnet-api-docs#1213
Fixes dotnet/dotnet-api-docs#877

//cc @JeremyKuhne


xml/System.IO/Path.xml
```diff
@@ -1042,11 +1069,27 @@ For more information on file path formats on Windows, see [File path formats on
         <Parameter Name="basePath" Type="System.String" />
       </Parameters>
       <Docs>
-        <param name="path">To be added.</param>
-        <param name="basePath">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <param name="path">A relative path to concatenate to <see paramref="basePath" />.</param>
```
eerhardt: Can we add some `## Examples` for this overload as well? I think they really help illustrate how to use this method.

JeremyKuhne: The key thing this is supposed to replace is depending on the current working directory.

``` C#
Environment.CurrentDirectory = @"C:\temp";

// Current directory can change at any time by any thread in the process so this isn't deterministic.
string path = Path.GetFullPath(relativePath);

// Deterministic version (thread safe)
path = Path.GetFullPath(@"C:\temp", relativePath);
```

If you have any "relative" paths the new overload is the safe alternative. We also have a method for determining if the current directory will change paths when resolved/used: `Path.IsPathFullyQualified`

``` C#
// If this returns true the current directory will not change the location the path points to
bool currentDirectorySafe = Path.IsPathFullyQualified(path);
```

MSBuild reading project files would be one concrete example of where this API should be used for thread safety. Any given path in a project file should be resolved (evaluated in their terms) using `Path.GetFullPath(projectFilePath, itemPath)`.

eerhardt: A thing with this API that I really struggle with is that even you switched the parameters around in your write-up above....

`path = Path.GetFullPath(@"C:\temp", relativePath);`

But the signature is:

`Path.GetFullPath (string path, string basePath)`

rpetrusha:
@JeremyKuhne, @mairaw, and @BillWagner, this will be ready for review once dotnet/samples#328 merges.

rpetrusha:
Closing and reopening to begin new build after merge of dotnet/samples#328.

PR #568, opened 2018-10-01
Concurrent dictionary: AddOrUpdate and GetOrAdd methods are not atomic

pkulikov:
Fixes dotnet/docs#7704

Also fixed the description of the `TArg` type parameter.


pkulikov:
>I like the way you've described the behavior @pkulikov Well done.

@BillWagner thanks, though I cannot take it completely because I've taken descriptions from the other parts of the docs and modified it accordingly.

PR #583, opened 2018-10-04
Added a basic example

yashints:
## Summary

Added a basic example of the usage 

Fixes dotnet/docs#6272


xml/System.Threading.Tasks/Task.xml
```diff
@@ -5447,10 +5447,26 @@ Task t Status: RanToCompletion, Result: 42
   
 ## Remarks  
  You can use `await Task.Yield();` in an asynchronous method to force the method to complete asynchronously. If there is a current synchronization context (<xref:System.Threading.SynchronizationContext> object), this will post the remainder of the method’s execution back to that context. However, the context will decide how to prioritize this work relative to other work that may be pending. The synchronization context that is present on a UI thread in most UI environments will often prioritize work posted to the context higher than input and rendering work. For this reason, do not rely on `await Task.Yield();` to keep a UI responsive.  For more information, see the entry [Useful Abstractions Enabled with ContinueWith](http://blogs.msdn.com/b/pfxteam/archive/2008/07/23/8768673.aspx) in the Parallel Programming with .NET blog.  
+
+A good example of the usage would be a `Windows Forms` application where you would like to execute an action but at the same time release the thread so the UI would be accessible to user.
+```cs
+async void Form_Load(object s, object e) 
+{ 
+    await Task.Yield(); 
+    MessageBox.Show("Async message!");
```
BillWagner: This is a simple example, but I wonder if a contrived sample that points toward a more real use case might be better.

Suppose both were something like:

```csharp
async void Form_Load(object s, object e)
{
    await Task.Yield(); // unblock the UI before doing potentially lengthy operation
    LoadLocalConfiguration();
}
```

What do you think @yashints ?

yashints: agree

xml/System.Threading.Tasks/Task.xml
```diff
@@ -5447,10 +5447,26 @@ Task t Status: RanToCompletion, Result: 42
...
+
+A good example of the usage would be a `Windows Forms` application where you would like to execute an action but at the same time release the thread so the UI would be accessible to user.
+```cs
+async void Form_Load(object s, object e) 
+{ 
+    await Task.Yield();  // unblock the UI before doing potentially lengthy operation
+    LoadLocalConfiguration();
```
pkulikov: Am I correct that this method would be scheduled to run on the UI thread, thus, the example description should be updated?

yashints: Yes, I will update the description

xml/System.Threading.Tasks/Task.xml
```diff
@@ -5447,10 +5447,26 @@ Task t Status: RanToCompletion, Result: 42
   
 ## Remarks  
  You can use `await Task.Yield();` in an asynchronous method to force the method to complete asynchronously. If there is a current synchronization context (<xref:System.Threading.SynchronizationContext> object), this will post the remainder of the method’s execution back to that context. However, the context will decide how to prioritize this work relative to other work that may be pending. The synchronization context that is present on a UI thread in most UI environments will often prioritize work posted to the context higher than input and rendering work. For this reason, do not rely on `await Task.Yield();` to keep a UI responsive.  For more information, see the entry [Useful Abstractions Enabled with ContinueWith](http://blogs.msdn.com/b/pfxteam/archive/2008/07/23/8768673.aspx) in the Parallel Programming with .NET blog.  
+
+A good example of the usage would be a `Windows Forms` application where you would like to execute a lengthy operation, but at the same time release the thread so the UI would be accessible to user. This operation will get run at a later time (at which point, it still may run synchronously) on the current context.
```
pkulikov: @yashints thanks for the update. A question, though:
>...get run at a later time (at which point, it still may run synchronously)...

The statement is the parentheses contradicts to the first sentence in the **Remarks** section:
>...to force the method to complete asynchronously...

I still think that the comment in the code:
>// unblock the UI before doing potentially lengthy operation

is misleading. Yes, we return to the caller after `await Task.Yield();`, but then, eventually, UI thread will be occupied with `LoadLocalConfiguration` execution. So, we block UI anyway, but later.

Also, the question: why not to use
```csharp
await Task.Run(() => LoadLocalConfiguration());
``` 

instead of 
```csharp
await Task.Yield();  // unblock the UI before doing potentially lengthy operation
LoadLocalConfiguration();
```

yashints: Agree with the first point, in terms of using `Task.Yeild()` vs `Task.Run()` that was to let the user know what would happen more explicitly. Happy to change if needed tho

BillWagner: I agree with @pkulikov 's suggestions above.  

Also, I apologize for letting this get stale without a response.  I've resolved the conflicts. @yashints If you make the changes suggested above, I'll :shipit: today.

yashints: Resolved, did some minor changes to wording as well

xml/System.Threading.Tasks/Task.xml
```diff
@@ -5445,11 +5445,25 @@ Task t Status: RanToCompletion, Result: 42
...
+ You can use `await Task.Run();` in a method to force it to complete asynchronously. If there is a current synchronization context (<xref:System.Threading.SynchronizationContext> object), this will post the remainder of the method's execution back to that context. However, the context will decide how to prioritize this work relative to other work that may be pending. The synchronization context that is present on a UI thread in most UI environments will often prioritize work posted to the context higher than input and rendering work. For this reason, do not rely on `await Task.Yield();` or `await Task.Run();` to keep a UI responsive.  For more information, see the entry [Useful Abstractions Enabled with ContinueWith](http://blogs.msdn.com/b/pfxteam/archive/2008/07/23/8768673.aspx) in the Parallel Programming with .NET blog.
```
pkulikov: @BillWagner is there a need to mention `Task.Run` when talking about `Task.Yield`?

I'm not sure the existing remarks text should be altered (so suggesting a revert):
```suggestion
 You can use `await Task.Yield();` in a method to force it to complete asynchronously. If there is a current synchronization context (<xref:System.Threading.SynchronizationContext> object), this will post the remainder of the method's execution back to that context. However, the context will decide how to prioritize this work relative to other work that may be pending. The synchronization context that is present on a UI thread in most UI environments will often prioritize work posted to the context higher than input and rendering work. For this reason, do not rely on `await Task.Yield();` to keep a UI responsive.  For more information, see the entry [Useful Abstractions Enabled with ContinueWith](http://blogs.msdn.com/b/pfxteam/archive/2008/07/23/8768673.aspx) in the Parallel Programming with .NET blog.  
```

xml/System.Threading.Tasks/Task.xml
```diff
@@ -5445,11 +5445,25 @@ Task t Status: RanToCompletion, Result: 42
...
+
+A good example of the usage would be a `Windows Forms` application where you would like to execute a lengthy operation, but at the same time release the thread so the UI would be accessible to user. This operation will get run at a later time (at which point, it still may run synchronously) on the current context.
+```cs
+async void Form_Load(object s, object e) 
+{ 
+    await Task.Run(() => LoadLocalConfiguration());  // unblock the UI before doing potentially lengthy operation
```
pkulikov: Now, this example is not related to the containing article about [Task.Yield() method](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.yield?view=netframework-4.7.2). So, we've started with a `Task.Yield` example but ended up with a `Task.Run` alternative. Is there a short example of the `Task.Yield` use that *doesn't* have a better `Task.Run` alternative? If not, maybe the current **Remarks** section is good as it is now.

BillWagner: Unit testing is a good use case for `Task.Yield`:  I often use it when mocking services to ensure that the code under test is exercised when it receives a Task that's not complete.  Mock example:

```csharp
// In a fake service:
public async Task<int> CountOrderFromDb)
{
   // simulate async operation reading a database
   // but use yield, not Task.Delay so it doesn't slow down text execution.
   await Task.Yield();
   return 5;
}
```

xml/System.Threading.Tasks/Task.xml
yashints: agreed

yashints:
@BillWagner could you review the changes now? included a wrong use case and a unit test as you mentioned

xml/System.Threading.Tasks/Task.xml
```diff
@@ -5445,11 +5445,44 @@ Task t Status: RanToCompletion, Result: 42
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- You can use `await Task.Yield();` in an asynchronous method to force the method to complete asynchronously. If there is a current synchronization context (<xref:System.Threading.SynchronizationContext> object), this will post the remainder of the method's execution back to that context. However, the context will decide how to prioritize this work relative to other work that may be pending. The synchronization context that is present on a UI thread in most UI environments will often prioritize work posted to the context higher than input and rendering work. For this reason, do not rely on `await Task.Yield();` to keep a UI responsive.  For more information, see the entry [Useful Abstractions Enabled with ContinueWith](https://blogs.msdn.com/b/pfxteam/archive/2008/07/23/8768673.aspx) in the Parallel Programming with .NET blog.
```
pkulikov: The original paragraph is correct. I don't think it should be removed.

BillWagner:
I've been remiss in commenting on this PR; I apologize for that.

The delay is that I'm struggling with the right recommendation to update it. In general, we try to write positive guidance rather than negative guidance. There's two reasons:

1. A reader seeing "don't do this" asks the obvious question "what should I do instead?" 
1. APIs are created for a reason. Our role with docs is to identify those scenarios where an API should be used (and direct readers to alternatives when they land at the wrong solution).

I'm checking with the team to get recommendations.

BillWagner:
After a lot of debate, I'm closing this.

The API reference is the wrong location for conceptual guidance, which is what this PR does. That means developers find this only when they've already reach the point where they are searching for a specific method.

A better way to provide this guidance is to create an article in our conceptual section that provides the correct way to ensure that a GUI remains responsive during long-running tasks.

Consider this type of guidance as part fo dotnet/docs#6373 or dotnet/docs#951

PR #584, opened 2018-10-05
Added examples for each overload

yashints:
## Summary

Added examples for each of the overloads to show what header will be generated.

Fixes dotnet/docs#5864

### DIFF
--- a/xml/System.Net.Http.Headers/AuthenticationHeaderValue.xml
+++ b/xml/System.Net.Http.Headers/AuthenticationHeaderValue.xml
@@ -78,9 +78,14 @@
         <remarks>
 <format type="text/markdown"><![CDATA[ 
 Example:
-```cs
+```csharp
 HttpClient client = new HttpClient();
 client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue(ACCESS_TOKEN);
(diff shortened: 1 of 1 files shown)


xml/System.Net.Http.Headers/AuthenticationHeaderValue.xml
```diff
@@ -75,7 +75,21 @@
       <Docs>
         <param name="scheme">The scheme to use for authorization.</param>
         <summary>Initializes a new instance of the <see cref="T:System.Net.Http.Headers.AuthenticationHeaderValue" /> class.</summary>
-        <remarks>To be added.</remarks>
+        <remarks>
+<format type="text/markdown"><![CDATA[ 
+Example:
+```cs
```
Thraka: We should have at a minimum, C# and VB.NET examples in here

Thraka: nit: Additionally, I think we use **csharp** as the language identifier, not **cs**. But I think **cs** works.

yashints: @Thraka done

PR #586, opened 2018-10-05
Adding remark about changing the KeySize property #5307

kapiljaveri:
## Summary

Appended information about changing the KeySize property and its impact on the Key itself to the remarks section.

Let me know if this looks good or if you'd like any changes.

Fixes dotnet/docs#5307

### DIFF
--- a/xml/System.Security.Cryptography/AesCryptoServiceProvider.xml
+++ b/xml/System.Security.Cryptography/AesCryptoServiceProvider.xml
@@ -625,7 +625,7 @@
 ## Remarks  
  The minimum size of the key is 128 bits, and the maximum size is 256 bits.  
   
- Changing the KeySize value will reset the Key itself and cause a new random key to be generated. This will happen any time the setter of the KeySize property is invoked (including when invoked for the same value).  
+ Changing the `KeySize` value resets the key and generates a new random key. This happens whenever the `KeySize` property setter is invoked (including when it's assigned the same value).  
   
  ]]></format>
         </remarks>


rpetrusha:
@bartonjs, could you take a look, please?

xml/System.Security.Cryptography/AesCryptoServiceProvider.xml
```diff
@@ -625,6 +625,8 @@
 ## Remarks  
  The minimum size of the key is 128 bits, and the maximum size is 256 bits.  
   
+ Changing the KeySize value will reset the Key itself and cause a new random key to be generated. This will happen any time the setter of the KeySize property is invoked (including when invoked for the same value).
```
bartonjs: The technical aspects are accurate, invoking set_KeySize for AesCryptoServiceProvider (or, really, any SymmetricAlgorithm-derived type) will cause the Key property to get randomized to a new value of that size, even when doing `alg.KeySize = alg.KeySize`.

The sentence construction "will reset the Key itself" feels clunky to me, though.

kapiljaveri: I can see that. Any suggestions?

Could omit the word "itself"? Or possibly just simplify it to "Changing the KeySize value will cause a new random key to be generated."

kapiljaveri:
Thanks for all the feedback @rpetrusha and @bartonjs! Let me know if the changes look good.

rpetrusha:
Your changes look good, @kapiljaveri. We'll merge your PR now. I should be live on docs.microsoft.com in the next day or two.

kapiljaveri:
Awesome, thank you 😊

kapiljaveri:
Thanks again for all the feedback and patience @bartonjs & @rpetrusha

PR #590, opened 2018-10-05
CI Update - .NET Platform Extensions

dend:


xml/FrameworksIndex/netcore-1.0.xml
```diff
@@ -16423,17 +16423,24 @@
       <Member Id="F:System.Reflection.BindingFlags.CreateInstance" />
       <Member Id="F:System.Reflection.BindingFlags.DeclaredOnly" />
       <Member Id="F:System.Reflection.BindingFlags.Default" />
+      <Member Id="F:System.Reflection.BindingFlags.DoNotWrapExceptions" />
```
mairaw: not sure why we're getting this @dend @joelmartinez with this PR that doesn't touch the old assemblies

also, this is what I see for .NET Core 1.0. new mdoc bug?
image

mairaw: rolled back them manually with a new commit https://github.com/dotnet/dotnet-api-docs/pull/590/commits/65e78d0e5440931a939e7ff04332b9417ba15d94

joelmartinez: @mairaw I'm curious if you could send me a link to this build failure. Trying to understand what would fail from this being missing.

mairaw: @joelmartinez let me see if I still have that. Otherwise, I can create a copy of the branch and delete some data.

mairaw: but any idea why these are being added to these platforms now @joelmartinez?

joelmartinez: @mairaw I'm working to track it down (https://github.com/mono/api-doc-tools/issues/357) ... so far, I have been unable to reproduce this in all the tests that I've tried. It _almost_ seems like some sort of concurrency bug (because of how inconsistent it is), but there's no use of threads or async/await in mdoc 🙃 so I continue to try to track down the issue. I will of course let you know as soon as I have a solution

PR #605, opened 2018-10-10
Change ConcurrentStack to use two separate files

hymccord:
## Summary

The intent is that code examples in the dotnet/samples repo will compile. The original code doesn't meet that requirement since it combines two standalone examples in a single source code file

Related PR dotnet/samples#356


rpetrusha:

PR #610, opened 2018-10-11
Documented MemoryMarshal class

rpetrusha:
## Documented MemoryMarshal class

Fixes dotnet/docs#4912

@stephentoub @ahsonkhan @GrabYourPitchforks

### DIFF
--- a/xml/System.Runtime.InteropServices/MemoryMarshal.xml
+++ b/xml/System.Runtime.InteropServices/MemoryMarshal.xml
@@ -20,12 +20,12 @@
   </Docs>
   <Members>
     <Member MemberName="AsBytes&lt;T&gt;">
-      <MemberSignature Language="C#" Value="public static ReadOnlySpan&lt;Byte&gt; AsBytes&lt;T&gt; (ReadOnlySpan&lt;T&gt; span) where T : struct;" />
+      <MemberSignature Language="C#" Value="public static ReadOnlySpan&lt;byte&gt; AsBytes&lt;T&gt; (ReadOnlySpan&lt;T&gt; span) where T : struct;" />
       <MemberSignature Language="ILAsm" Value=".method public static hidebysig valuetype System.ReadOnlySpan`1&lt;unsigned int8&gt; AsBytes&lt;struct .ctor (class System.ValueType) T&gt;(valuetype System.ReadOnlySpan`1&lt;!!T&gt; span) cil managed" />
       <MemberSignature Language="DocId" Value="M:System.Runtime.InteropServices.MemoryMarshal.AsBytes``1(System.ReadOnlySpan{``0})" />
       <MemberSignature Language="VB.NET" Value="Public Shared Function AsBytes(Of T As Structure) (span As ReadOnlySpan(Of T)) As ReadOnlySpan(Of Byte)" />
       <MemberSignature Language="C++ CLI" Value="public:&#xA;generic &lt;typename T&gt;&#xA; where T : value class static ReadOnlySpan&lt;System::Byte&gt; AsBytes(ReadOnlySpan&lt;T&gt; span);" />
-      <MemberSignature Language="F#" Value="static member AsBytes : ReadOnlySpan&lt;'T (requires 'T : struct)&gt; -&gt; ReadOnlySpan&lt;Byte&gt; (requires 'T : struct)" Usage="System.Runtime.InteropServices.MemoryMarshal.AsBytes span" />
+      <MemberSignature Language="F#" Value="static member AsBytes : ReadOnlySpan&lt;'T (requires 'T : struct)&gt; -&gt; ReadOnlySpan&lt;byte&gt; (requires 'T : struct)" Usage="System.Runtime.InteropServices.MemoryMarshal.AsBytes span" />
       <MemberType>Method</MemberType>
       <AssemblyInfo>
         <AssemblyName>System.Memory</AssemblyName>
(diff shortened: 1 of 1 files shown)


BillWagner:
closing and re-opening to start a new build.

BillWagner:
This looks really good to me @rpetrusha 

One discussion item:  The C# spec using the term "unmanaged type" instead of "primitive type" See [the section on [pointer types](https://docs.microsoft.com/dotnet/csharp/language-reference/language-specification/unsafe-code#pointer-types).  I'd been following that example to describe the unmanaged constraint on generics, and the restrictions on `ref struct` types. 

Proposal:  We should standardize on the term "unmanaged type".  Vote 👍 or 👎 on this comment.  (If you vote it down, what would be better?

xml/System.Runtime.InteropServices/MemoryMarshal.xml
```diff
@@ -157,12 +185,23 @@
...
-        <typeparam name="TFrom">To be added.</typeparam>
-        <typeparam name="TTo">To be added.</typeparam>
-        <param name="span">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <typeparam name="TFrom">The type of the source span.</typeparam>
+        <typeparam name="TTo">The type of the target span.</typeparam>
+        <param name="span">The source slice to convert.</param>
+        <summary>Casts a read-only span of one primitive type to a read-only span of another primitive type.</summary>
```
BillWagner: I'm concerned about a possible misconception here.  Consider this cast of a scalar:

```csharp
long l = 5;
int i  = (int) l;
```

An int is smaller than a long, and only the lower 4 bytes of the long `l` are assigned to the int `i`.

What happens if we cast a `Span<long>` to a `Span<int>`? I'm guessing there are twice as many `int` elements than `long elements and the memory is re-interpreted. We should clarify that.

xml/System.Runtime.InteropServices/MemoryMarshal.xml
```diff
@@ -357,20 +447,28 @@
...
+
+If the span is empty, this method returns a reference to the location where the element at index 0 would have been stored. Such a reference may or may not be `null`. The returned reference can be used for pinning, but it must never be dereferenced.
+
+       ]]></format> 
+        </remarks>        
       </Docs>
     </Member>
     <Member MemberName="Read&lt;T&gt;">
-      <MemberSignature Language="C#" Value="public static T Read&lt;T&gt; (ReadOnlySpan&lt;byte&gt; source) where T : struct;" />
+      <MemberSignature Language="C#" Value="public static T Read&lt;T&gt; (ReadOnlySpan&lt;Byte&gt; source) where T : struct;" />
```
BillWagner: same comment as earlier on `Byte` vs. `byte`.

xml/System.Runtime.InteropServices/MemoryMarshal.xml
```diff
@@ -15,17 +15,17 @@
...
   <Interfaces />
   <Docs>
-    <summary>To be added.</summary>
+    <summary>Provides methods to interoperate with <see cref="T:System.Memory`1" />, <see cref="T:System.ReadOnlyMemory`1" />,  <see cref="T:System.Span`1" />, and  <see cref="T:System.ReadOnlySpan`1" />.</summary>
     <remarks>To be added.</remarks>
   </Docs>
   <Members>
     <Member MemberName="AsBytes&lt;T&gt;">
-      <MemberSignature Language="C#" Value="public static ReadOnlySpan&lt;byte&gt; AsBytes&lt;T&gt; (ReadOnlySpan&lt;T&gt; span) where T : struct;" />
+      <MemberSignature Language="C#" Value="public static ReadOnlySpan&lt;Byte&gt; AsBytes&lt;T&gt; (ReadOnlySpan&lt;T&gt; span) where T : struct;" />
```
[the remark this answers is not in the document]
rpetrusha: Thanks for noticing this, @BillWagner. No, there isn't a reason -- I must have changed it accidentally.,

xml/System.Runtime.InteropServices/MemoryMarshal.xml
```diff
@@ -84,11 +94,21 @@
...
+        <typeparam name="T">The type of items in the span.</typeparam>
+        <param name="span">The source slice to convert.</param>
+        <summary>Casts a <see cref="T:System.Span`1" /> of one primitive type, <paramref name="T" />, to a <see langword="Span&lt;Byte&gt;" />.</summary>
+        <returns>A span of type <see cref="T:System.Byte" />.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+  
+## Remarks  
+
+`T` cannot contain pointers or references. `T` is checked at runtime in order to preserve type safety.
```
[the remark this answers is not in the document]
rpetrusha: For the .NET docs, I think it's likely to cause a good deal of confusion, since "unmanaged type" is seen as the opposite of a managed type.

xml/System.Runtime.InteropServices/MemoryMarshal.xml
rpetrusha: Yes. The cast always appears to occur in an unchecked context.

xml/System.Runtime.InteropServices/MemoryMarshal.xml
rpetrusha: This seems to have been a copy and paste error -- I wanted to change byte to Byte in 2 places, and changed them also in C# and F# syntax. I've corrected them.

PR #656, opened 2018-10-23
Fix missing space

ChrisMaddock:
Fix a missing space.

I also code-fenced the enum values, as a xref would have linked back to the same page - is that the correct approach?


rpetrusha:
Yes, that is the correct approach, @ChrisMaddock.

PR #666, opened 2018-10-25
Documented VB optimization with CInt(Fix(value))

rpetrusha:
## Documented VB optimization with CInt(Fix(value))

Fixes dotnet/docs#8877

This PR:

- Documents the CInt(Fix(value)) optimization for the relevant overloads of Fix.
- Slightly modifies the description and return value sections of most (all?) methods.
- Replaces the `<c>` tags with `<paramref>` tags.
- Removes mention of a null return value for overloads whose parameter is a value type.
- Adds missing code identifiers.

//cc @KathleenDollard


xml/Microsoft.VisualBasic/Conversion.xml
```diff
@@ -335,8 +335,13 @@ End Class
   
  `Fix(number)` is equivalent to `Sign(number) * Int(Abs(number)).`  
   
+Starting with Visual Basic 15.8, the performance of decimal-to-integer conversion is optimized if you pass the value returned by the `Fix` method to the [CInt](~/docs/visual-basic/language-reference/functions/conversion-functions.md) function. For example:
```
BillWagner: Somewhere, we should clarify the impact of this optimization. Is it faster? By how much?

KathleenDollard: Faster. 

@DualBrain did some tests when he mentioned to me (I think over the phone). I don't recall the numbers. 

Do you think "significantly faster" is sufficient here?

DualBrain: Depending on the tests, it can be up to twice as fast.  Of course, this is *only* noticed if you are doing a ton of these operations as twice as fast as a little bit slow really isn't too noticeable except in the most specific of scenarios... in those scenarios... when it's noticed, it's noticed. ;-)

Although I didn't go into the performance differences... I did do a "thank you" (with some details) post over at http://addressof.com/blog/archive/2018/09/20/Performance-Improvement-_2D00_-CInt_2800_Fix_2800_value_29002900_.aspx

KathleenDollard: Ron, can we say something like "This optimization allows code to run faster - up to twice as fast for code that does a lot of conversion to integer types."

rpetrusha:
This is ready for a final review, @BillWagner @KathleenDollard @DualBrain @gafter

xml/Microsoft.VisualBasic/Conversion.xml
```diff
@@ -397,8 +395,13 @@ End Class
  The difference between `Int` and `Fix` functions is that if `Number` is negative, `Int` returns the first negative integer less than or equal to `Number`, whereas `Fix` returns the first negative integer greater than or equal to `Number`. For example, `Int` converts -8.4 to -9, and `Fix` converts -8.4 to -8.  
   
  `Fix(number)` is equivalent to `Sign(number) * Int(Abs(number)).`  
-  
-   
+
+Starting with Visual Basic 15.8, the performance of Double-to-integer conversion is optimized if you pass the value returned by the `Fix` method to the any of the [integral conversion functions](~/docs/visual-basic/language-reference/functions/conversion-functions.md), or if the Double value returned by `Fix` is implicitly converted to an integer with [Option Strict](~/docs/visual-basic/language-reference/statements/option-strict-statement.md) set to `Off`. This optimization allows code to run faster -- up to twice as fast for code that does a large number of conversions to integer types. The following example illustrates such an optimized conversion:
```
BillWagner: Do you want to code-fence `Double` in this paragraph?

xml/Microsoft.VisualBasic/Conversion.xml
```diff
@@ -713,8 +723,13 @@ End Class
   
  `Fix(number)` is equivalent to `Sign(number) * Int(Abs(number)).`  
   
+Starting with Visual Basic 15.8, the performance of Single-to-integer conversion is optimized if you pass the value returned by the `Fix` method to the any of the [integral conversion functions](~/docs/visual-basic/language-reference/functions/conversion-functions.md), or if the Single value returned by `Fix` is automatically converted to an integer with [Option Strict](~/docs/visual-basic/language-reference/statements/option-strict-statement.md) set to Off. This optimization allows code to run faster -- up to twice as fast for code that does a large number of conversions to integer types. The following example illustrates such an optimized conversion:
```
BillWagner: Same comment on code-fencing `Single`

xml/Microsoft.VisualBasic/Conversion.xml
```diff
@@ -1696,8 +1721,13 @@ End Class
   
  `Fix(number)` is equivalent to `Sign(number) * Int(Abs(number)).`  
   
-   
-  
+Starting with Visual Basic 15.8, the performance of Single-to-integer conversion is optimized if you pass the value returned by the `Int` method to the any of the [integral conversion functions](~/docs/visual-basic/language-reference/functions/conversion-functions.md), or if the Single value returned by `Int` is automatically converted to an integer with [Option Strict](~/docs/visual-basic/language-reference/statements/option-strict-statement.md) set to Off. This optimization allows code to run faster -- up to twice as fast for code that does a large number of conversions to integer types. The following example illustrates such an optimized conversion:
```
BillWagner: code-fence `Single`

PR #672, opened 2018-10-26
Noted APIs deprecated for new development

rpetrusha:
## Noted APIs deprecated for new development

Fixes dotnet/docs#917
Contributes to dotnet/docs#4121

//cc @terrajobst

### DIFF
--- a/xml/System/Environment.xml
+++ b/xml/System/Environment.xml
@@ -1485,7 +1485,7 @@ Per-machine and per-user environment variables are not supported. A `target` val
 
 The `Environment.OSVersion` property does not provide a reliable way to identify the exact operating system and its version. Therefore, we do not recommend that you use this method. Instead: 
 
-- To identify the operating system platform, use the <xref:System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform?displayProperty=nameWithType> method. 
+- To identify the operating system platform, use the <xref:System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform%2A?displayProperty=nameWithType> method. 
 
 - Avoid writing code that depends on a reported operating system version. Instead, check for the availability of the features that your application needs.
 


pkulikov:
@rpetrusha if not fixes (maybe even that), this PR contributes to dotnet/docs#4121

rpetrusha:
Thanks for pointing out dotnet/docs#4121, @pkulikov. I'll add it, and take a close look at ArrayList to see if it needs additional revision.

xml/System.Collections/ArrayList.xml
```diff
@@ -56,7 +56,11 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Collections.ArrayList> class is designed to hold heterogeneous collections of objects. However, it does not always offer the best performance. Instead, we recommend the following:  
+
+>  [!IMPORTANT]
+>  We don't recommend that you use the `ArrayList` class for new development. Instead, we recommend that you use the generic <xref:System.Collections.Generic.List%601> class. For more information, see [Non-generic collections shouldn't be used](https://github.com/dotnet/platform-compat/blob/master/docs/DE0006.md) on GitHub.
+
+The <xref:System.Collections.ArrayList> class is designed to hold heterogeneous collections of objects. However, it does not always offer the best performance. Instead, we recommend the following:
```
svick: With this change, there is now a note recommending to use `List<T>`, followed by another recommendation to use `List<T>`. Should these two be combined? Or at least connected somehow?

rpetrusha: Yes, it should be, @svick. Thanks for pointing that out. I should have read the documentation for ArrayList more carefully before adding the note.

imba-tjd:
About OSVersion.
I think the result of win8 or later should be kept a little bit, otherwise many people would have questions, although there are answers on StackOverflow. And actually editing manifest can get the right result on win10.
`System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform` accept `System.Runtime.InteropServices.OSPlatform`, which seems not accurate enough. It only has `Windows`, `OSX` and `Linux`.
`System.Runtime.InteropServices.RuntimeInformation.OSDescription` has more information, but I don't know OSVersion or OSDescription which one is more reliable on Linux and Mac OS.

PR #684, opened 2018-10-30
Clarify explanation of Regex.Unescape https://github.com/dotnet/docs/…

Coolerhino:
fixes dotnet/docs#8706

### DIFF
--- a/xml/System.Text.RegularExpressions/Regex.xml
+++ b/xml/System.Text.RegularExpressions/Regex.xml
@@ -5148,7 +5148,7 @@ On .NET Core, calls to the `Regex.CompileToAssembly` method throw a <xref:System
 > [!NOTE]
 >  <xref:System.Text.RegularExpressions.Regex.Unescape%2A> cannot reverse an escaped string perfectly because it cannot deduce precisely which characters were escaped,  
   
--   It replaces the hexadecimal values in verbatim string literals with the actual printable characters. For example, it replaces @"\0x07" with @"\a", or @"\x0A" with @"\n". It converts to supported escape characters such as \a, \b, \e, \n, \r, \f, \t, \v and alphanumeric characters.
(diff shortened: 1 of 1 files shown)


xml/System.Text.RegularExpressions/Regex.xml
```diff
@@ -5147,7 +5148,7 @@ On .NET Core, calls to the `Regex.CompileToAssembly` method throw a <xref:System
 > [!NOTE]
 >  <xref:System.Text.RegularExpressions.Regex.Unescape%2A> cannot reverse an escaped string perfectly because it cannot deduce precisely which characters were escaped,  
   
--   It replaces the representation of unprintable characters with the characters themselves. For example, it replaces \a with \x07. The character representations it replaces are \a, \b, \e, \n, \r, \f, \t, and \v.  
+-   It replaces the hexadecimal values in verbatim string literals with the actual printable characters. For example, it replaces @"\0x07" with @"\a", or @"\x0A" with @"\n". It converts to supported escape characters such as \a, \b, \e, \n, \r, \f, \t, \v and alphanumeric characters.
```
Thraka: Hi! Thank you so much for this pull request. Could you add a `,` after the "\v" and before "and" as we follow the oxford comma rule:

`\r, \f, \t, \v, and alphanumeric characters.`

elgonzo:
Note: This is a copy of the post i made in the (now closed) issue thread in response to this commit.


The approved commit of the updated documentation has two errors. First, it should be @"\x07" and not @"\0x07" (the latter would not represent a valid character code due to the 0 before the x).

Second, the results are not the verbatim strings @"\a" nor @"\n". It is wrong to use the verbatim string symbol for the replacement results; the results would be rather "\a", or "\n" respectively.

See this dotnetfiddle illustrating both errors: https://dotnetfiddle.net/EWLos0

Thraka:
I see what you mean,. I'll get that fixed.

PR #686, opened 2018-10-30
Remove regular expression as validation method

KyleMit:
Using regular expressions for Phone Validation was deprecated as of .Net Framework 4.7.2 and not in any version of .Net Core

Here's the compatability post on removing regex and the default validator in `[Phone]`:

https://github.com/Microsoft/dotnet/blob/master/Documentation/compatibility/aspnet-472-compat-doc.md

And .NET Core no longer users regular expressions to determine phone number format validity at all ([Presumably because of the best practices on Regex and injection/security concerns](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices))

Here's the source code for core without a Regex in sight:

https://github.com/dotnet/corefx/blob/v2.1.5/src/System.ComponentModel.Annotations/src/System/ComponentModel/DataAnnotations/PhoneAttribute.cs

Not sure if the description can be forked by API version / framework, but it looks like this single file is serving as the doc for both .NET Framework and .NET Core, and seems easier to just remove the regular expression bit as it's certainly incorrect in the case of Core, but if there's a way to version the comments, then let me know in the PR and I can update the commit to seperate by version.

### DIFF
--- a/xml/System.ComponentModel.DataAnnotations/PhoneAttribute.xml
+++ b/xml/System.ComponentModel.DataAnnotations/PhoneAttribute.xml
@@ -28,7 +28,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>Specifies that a data field value is a  well-formed phone number using a regular expression for phone numbers.</summary>
+    <summary>Specifies that a data field value is a well-formed phone number.</summary>
     <remarks>To be added.</remarks>
   </Docs>
   <Members>


chlowell:
@BillWagner I'm only on the compat docs because I wrote automation to generate them, I'm not involved with the content. @rpetrusha, what do you think?

BillWagner:
Sorry about that. I meant to ping @divega as owner for `System.ComponentModel.DataAnnotations`

PR #688, opened 2018-10-31
Standardizes Big O notation use of backticks

hop-along-polly:
There were some methods documented using Big O Notation in plain-text i.e. O(n log n) whileother used backticks around the internal variable of the Big O Notation i.e. O(`n` log `n`).
Since this notation is similar to a mathematical equation I switchd all the existing entries to use backticks around the entire Big O notation i.e. `O(n log n)` which also draws more
attention to it. I'm fine doing it any of the 3 way's shown as long as it's standardized.

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Collections.Generic/SortedSet`1.xml
+++ b/xml/System.Collections.Generic/SortedSet`1.xml
@@ -99,7 +99,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- This constructor is an O(1) operation.  
+ This constructor is an `O(1)` operation.  
   
  ]]></format>
         </remarks>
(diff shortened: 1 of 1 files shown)


xml/System.Collections.Generic/SortedSet`1.xml
```diff
@@ -1739,7 +1739,7 @@
 ## Remarks  
  Any duplicate elements in `other` are ignored.  
   
- If the `other` parameter is a <xref:System.Collections.Generic.SortedSet%601> collection with the same equality comparer as the current <xref:System.Collections.Generic.SortedSet%601> object, this method is an O(n log `m`) operation. Otherwise, this method is an O(`n` log `m`) + O(`n` log `n`) operation, where `n` is the number of elements in `other` and `m` is <xref:System.Collections.Generic.SortedSet%601.Count%2A>.  
+ If the `other` parameter is a <xref:System.Collections.Generic.SortedSet%601> collection with the same equality comparer as the current <xref:System.Collections.Generic.SortedSet%601> object, this method is an `O(n log m)` operation. Otherwise, this method is an `O(n log m) + O(n log n)` operation, where `n` is the number of elements in `other` and `m` is <xref:System.Collections.Generic.SortedSet%601.Count%2A>.
```
ezzy1337: I may need to brush up on my BIg O Notation but shouldn't this `O(n log m) + O(n log n)` be this instead `O((n log m) + (n log n))`.

BillWagner: I'm looking this up as well. I think you are right, but I want to verify.

PR #694, opened 2018-11-01
Update interfaces with mdoc 5.4.7.1

dend:
This PR includes more information on expanded interfaces. This is related to https://github.com/MicrosoftDocs/feedback/issues/223


dend:
@mairaw @BillWagner @rpetrusha [here is an example](https://review.docs.microsoft.com/en-us/dotnet/api/system.array?view=netframework-4.7&branch=CITest) of a page with "unfolded" interfaces.

cc: @joelmartinez

joelmartinez:
Just to pre-empt one possible question before it comes up ... you'll note that the signature:

```csharp
public abstract class Array : ICloneable, System.Collections.IList, System.Collections.IStructuralComparable, System.Collections.IStructuralEquatable
```

doesn't contain the whole list of interfaces:

> Implements
`ICollection`, `IEnumerable`, `IList`, `IStructuralComparable`, `IStructuralEquatable`, `ICloneable`

This is because the signature formatter still uses the explicitly listed list of interfaces ... ie. the ones that are literally in the source code, rather than the "expanded" list. This way, the formatted signature matches what is in the source code :)

rpetrusha:
@dend and @joelmartinez, I'm not sure that the list of assemblies per framework/core/starndard version is accurate. For example, core seems to include mscorlib, which is actually not included with core. And as far as I know, standard only includes a single netstandard assembly for each standard version, and not a series of assemblies, as the pre-2.0 standard versions have. @terrajobst and @weshaggard am I accurate here?

joelmartinez:
@rpetrusha Can you give me an example of what you're referring to?

mdoc processes the list of assemblies that are listed in the binaries repo. `netstandard-2.0` has only that single assembly as you mention:
https://apidrop.visualstudio.com/binaries/_git/binaries?path=%2Fdotnet%2Fnetstandard-2.0&version=GBmaster

And also lists only that single assembly: 
https://raw.githubusercontent.com/dotnet/dotnet-api-docs/bdad3dc3e8efc72d078164eee96565ec28ea1f92/xml/FrameworksIndex/netstandard-2.0.xml

The generated frameworks index file for `netcore-1.0` has no `mscorlib` listed:
https://raw.githubusercontent.com/dotnet/dotnet-api-docs/bdad3dc3e8efc72d078164eee96565ec28ea1f92/xml/FrameworksIndex/netcore-1.0.xml

joelmartinez:
@rpetrusha oh, wait … are you referring to the list of assemblies listed on the actual rendered page?

image

I _think_ there are still some engineering tasks TBD on the latter parts of the pipeline … so as of this recent release, mdoc now has all the information needed to filter the assemblies correctly per moniker, however there are still some changes downstream pending in order to actually do that. @dend, can you comment on that?

dend:
LGTM :shipit: @BillWagner - let me know if you can merge this, or should I?

PR #698, opened 2018-11-05
Add info on implicit file paths to System.Uri documentation.

rmkerr:
## Summary

This change updates the documentation to describe support for implicit file paths in System.Uri. It also adds a warning about depending on this behavior, as it is necessarily different on Windows and Unix based systems.

Fixes dotnet/corefx#33241

### DIFF
--- a/xml/System/Uri.xml
+++ b/xml/System/Uri.xml
@@ -164,12 +164,12 @@ http://myUrl/
  The <xref:System.GenericUriParserOptions.Idn?displayProperty=nameWithType> type indicates the parser supports Internationalized Domain Name (IDN) parsing (IDN) of host names. Whether IDN is used is dictated by the configuration values previously discussed.  
   
 ## Implicit File Path Support
- <xref:System.Uri> can also be used to represent local file system paths. These paths can be represented *explicitly* in URIs that begin with the file:// scheme, and *implicitly* in URIs that do not have a scheme. As a concrete example, the following two URIs are both valid, and represent the same file path:
-    ```csharp
-    Uri uri1 = new Uri("C:/test/path/file.txt") // Implicit file path.
-    Uri uri2 = new Uri("file:///C:/test/path/file.txt") // Explicit file path.
-    ```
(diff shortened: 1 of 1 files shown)


xml/System/Uri.xml
```diff
@@ -161,6 +163,14 @@ http://myUrl/
   
  The <xref:System.GenericUriParserOptions.Idn?displayProperty=nameWithType> type indicates the parser supports Internationalized Domain Name (IDN) parsing (IDN) of host names. Whether IDN is used is dictated by the configuration values previously discussed.  
   
+## Implicit File Path Support
+ <xref:System.Uri> can also be used to represent local file system paths. These paths can be represented *explicitly* in URIs that begin with the file:// scheme, and *implicitly* in URIs that do not have a scheme. As a concrete example, the following two URIs are both valid, and represent the same file path:
```
rpetrusha: nit:  a scheme --> the file:// scheme

xml/System/Uri.xml
```diff
@@ -161,6 +163,14 @@ http://myUrl/
...
+    ```csharp
```
rpetrusha: nit: the code block doesn't have to be indented.

xml/System/Uri.xml
```diff
@@ -161,6 +163,14 @@ http://myUrl/
...
+    Uri uri1 = new Uri("C:/test/path/file.txt") // Implicit file path.
+    Uri uri2 = new Uri("file:///C:/test/path/file.txt") // Explicit file path.
+    ```
+ These implicit file paths are not compliant with the URI specification, and so should be avoided when possible. When using .NET Core on Unix based systems implicit file paths can be especially problematic, as an absolute implicit file path is *indistinguishable* from a relative path. When such ambiguity is present, <xref:System.Uri> will default to interpreting the path as an absolute URI.
```
rpetrusha: - no comma after "specification"
- Unix based --> Unix-based
- comma after "Unix-based systems"
- as an absolute --> because an absolute
- will default --> default

rmkerr: I've addressed these issues and the others you mentioned in the latest commit. Thanks for the review!

rpetrusha:
Thanks, @rmkerr, for making the additional changes. I'll merge your PR now.

PR #707, opened 2018-11-07
Fix wrong remark on DistributedIdentifier

fredericDelaporte:
## Summary

Fix an erroneous remark for DistributedIdentifier: it was stating its value is `null` when the transaction is not escalated, although the property is not nullable. When the transaction is not escalated, its value is `Guid.Empty`.

Fix dotnet/docs#7942

Also update an old link in the same remark section.

### DIFF
--- a/xml/System.Transactions/TransactionInformation.xml
+++ b/xml/System.Transactions/TransactionInformation.xml
@@ -110,7 +110,7 @@
 ## Remarks  
  If the transaction is escalated to a two-phase commit transaction, this property returns its unique identifier. If the transaction is not escalated, the value is <xref:System.Guid.Empty>.  
   
- For more information on how a transaction's management is escalated, see [Transaction Management Escalation](http://msdn.microsoft.com/library/5b87ae58-8ccf-4b9b-9715-41e4aade89b7).  
+ For more information on how a transaction's management is escalated, see [Transaction Management Escalation](~/docs/framework/data/transactions/transaction-management-escalation.md).  
   
  ]]></format>
         </remarks>


xml/System.Transactions/TransactionInformation.xml
```diff
@@ -108,7 +108,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- If the transaction is escalated to a two-phase commit transaction, this property returns its unique identifier. If the transaction is not escalated, the value is `null`.  
+ If the transaction is escalated to a two-phase commit transaction, this property returns its unique identifier. If the transaction is not escalated, the value is <xref:System.Guid.Empty>.  
   
  For more information on how a transaction's management is escalated, see [Transaction Management Escalation](http://msdn.microsoft.com/library/5b87ae58-8ccf-4b9b-9715-41e4aade89b7).
```
fredericDelaporte: Side note: I wonder if the link for "Transaction Management Escalation", http://msdn.microsoft.com/library/5b87ae58-8ccf-4b9b-9715-41e4aade89b7 , is correct.

On a French desktop, it does redirect to an obsolete translated article: https://docs.microsoft.com/fr-fr/previous-versions/visualstudio/visual-studio-2008/ms229978(v=vs.90) . Trying to switch to English from there yields a 404. The correct version of the article seems to be https://docs.microsoft.com/en-us/dotnet/framework/data/transactions/transaction-management-escalation , but I do not know if this url is suitable for being used here.

Should this be opened as a new issue on dotnet/docs? (It is maybe the MSDN redirection which is to blame.)

rpetrusha: The best link, as you point out, @fredericDelaporte is <https://docs.microsoft.com/en-us/dotnet/framework/data/transactions/transaction-management-escalation>. If you'd like, you can make the change in this PR; it doesn't require a new issue or a separate PR. The correct format for the link is \(~/docs/framework/data/transactions/transaction-management-escalation.md)

fredericDelaporte: Ok, it is done, thanks for the review.

rpetrusha:
Thanks, @fredericDelaporte, for also updating the URL. I'll merge your PR now. Your changes should be live on docs.microsoft.com in the next day or two.

PR #736, opened 2018-11-15
Documented IPinnable, MemoryHandle

rpetrusha:
## Documented IPinnable, MemoryHandle

Fixes dotnet/docs#8255

### DIFF
--- a/xml/System.Buffers/IPinnable.xml
+++ b/xml/System.Buffers/IPinnable.xml
@@ -15,7 +15,15 @@
   <Interfaces />
   <Docs>
     <summary>Provides a mechanism for pinning and unpinning objects to prevent the garbage collector from moving them.</summary>
-    <remarks>To be added.</remarks>
+    <remarks>
+          <format type="text/markdown"><![CDATA[  
+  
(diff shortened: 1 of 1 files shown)


xml/System.Buffers/IPinnable.xml
```diff
@@ -40,10 +40,19 @@
...
-        <remarks>To be added.</remarks>
+        <param name="elementIndex">The offset to the element within the memory buffer to which the returned <see cref="T:System.Buffers.MemoryHandle" /> points.</param>
+        <summary>Pins a block of memory.</summary>
+        <returns>A handle to the block of memory.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+  
+## Remarks  
+
+Call this method to indicate that the <xref:System.Buffers.IPinnable> object cannot be moved by the garbage collector so that the address of the pinned object can be used.
```
BillWagner: I like this, but I'm wondering if it could be better. 

A developer can access the object without pinning it, but only through managed APIs. While pinned, unmanaged techniques can be used.

Issue #1153, opened 2018-11-21
Notes on C and C++ equivelance is misleading


jormundgand:
In C++ the closest equivalent concept to a delegate, would be a lambda. C++ lambdas have all of the qualities that is highlighted in the section as being the difference between a C-style function pointer and a delegate. A lambda may or may not contain copies or references to a selection or the entire scope as seen when the lambda is constructed. As such it is no less powerful than a delegate in terms of expressiveness or type safety.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5f87dc7a-d485-4667-8e99-d209aa9434c6
* Version Independent ID: c089f774-e9c1-a016-2ece-522bb2b8cb1d
* Content: [Delegate Class (System)](https://docs.microsoft.com/en-us/dotnet/api/system.delegate?view=netstandard-2.0)
* Content Source: [xml/System/Delegate.xml](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Delegate.xml)
* Product: **dotnet-api**
* GitHub Login: @rpetrusha
* Microsoft Alias: **ronpet**

svick:
@jormundgand I agree with you and so I have opened a Pull Request that fixes this: https://github.com/dotnet/dotnet-api-docs/pull/1169. Feel free to comment on that Pull Request if you think this should be fixed in a different way.

PR #1169, opened 2018-11-22
The closest equivalent of delegate in C++ is not function pointer

svick:
… that would be functor (function object).

I'm not sure that comparison with C++ functors would be useful here, so I have just removed the mention of C++.

Fixes https://github.com/dotnet/dotnet-api-docs/issues/1153.


jormundgand:
That certainly removes my primary objection. I don't think C is the primary language to compare C# with though. C++ has the high level concept of functors, which C# does not support fully; delegates are a sort of functor among many kinds of functors.

PR #1223, opened 2018-11-28
M32642: Truncated source

v-makoud:
Hello @mairaw,  @dend, @rpetrusha,
Translator has reported possible source content issue. 
Description: Would you please kindly advise how to translate these truncated sources - whose live pages are not found on the web?
Full Source String:
When the method returns, the manager of
When the method returns, the offset from the start of the
When the method returns, the length of the
When the method returns, an instance of
that the
represents.

Please review the comment tag added and reply with explanation if fix is needed or not. If you make related fix in another PR then share your PR number so we can confirm and close this PR.
Many thanks in advance.

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Runtime.InteropServices/MemoryMarshal.xml
+++ b/xml/System.Runtime.InteropServices/MemoryMarshal.xml
@@ -514,7 +514,7 @@ If the span is empty, this method returns a reference to the location where the 
         <exception cref="T:System.ArgumentException">
           <paramref name="T" /> contains references or pointers.</exception>
         <exception cref="T:System.ArgumentOutOfRangeException">
-          <paramref name="source" /> is smaller than <paramref name="T" /></exception>
+          <paramref name="source" /> is smaller than <paramref name="T" />.</exception>
       </Docs>
     </Member>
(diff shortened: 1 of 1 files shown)


xml/System.Runtime.InteropServices/MemoryMarshal.xml
```diff
@@ -661,7 +661,7 @@ This method allows a read-only memory buffer to be used in existing APIs that re
         <typeparam name="T">The type of the items in the read-only memory buffer.</typeparam>
         <typeparam name="TManager">The type of the <see cref="T:System.Buffers.MemoryManager`1" /> to retrieve.</typeparam>
         <param name="memory">The read-only memory buffer for which to get the memory manager.</param>
-        <param name="manager">When the method returns, the manager of <param name="memory" />. </param>
+        <param name="manager">When the method returns, the manager of <!-- incomplete sentence --><param name="memory" />. </param>
```
Thraka: Inside of these xml nodes, it should be `<paramref` not `<param` This whole XML file has a problem with this.

rpetrusha:
I've fixed the bad param elements, @v-makoud and @Thraka

PR #1339, opened 2018-12-11
C50190: Escaping @

v-makoud:
Hello @dend, @mairaw,
Description: The only way to fix this issue is to escape the @ character. ( or creating an include with plain @ character ) 
Please note that this character when not escaped in this format will cause everything after the @ character as a plain text = no format or code will render afterwards. ( Due to bilingual content )
Please review and merge the proposed file change to fix to target versions. If you make related fix in another PR  then share your PR number so we can confirm and close this PR.
Many thanks in advance.

## Summary

Describe your changes here.


v-maudel:
Thanks so much @rpetrusha

PR #1488, opened 2018-12-28
Different cross platform behavior of GetFileName

dvdme:
As it was discussed in this [coreclr issue 21597](https://github.com/dotnet/coreclr/issues/21597) the behavior of GetFileName is not the same and cannot be the same on Windows and Linux because `'\\'` is a legal file name on Linux. 

It was suggested on that issue discussion that this could be better clarified on the documentation, hence this proposed change.

Fixes [coreclr issue 21597](https://github.com/dotnet/coreclr/issues/21597)

### DIFF
--- a/xml/System.IO/Path.xml
+++ b/xml/System.IO/Path.xml
@@ -843,7 +843,7 @@ For more information on file path formats on Windows, see [File path formats on 
   
  The separator characters used to determine the start of the file name are <xref:System.IO.Path.DirectorySeparatorChar> and <xref:System.IO.Path.AltDirectorySeparatorChar>.  
  
- Because `'\\'` is a legal file name on Linux, `GetFileName` running under Linux-based platforms cannot correctly return the file name from a Windows-based path like `C:\mydir\myfile.ext` but `GetFileName` running under Windows-based platforms can correctly return the file name from a Linux-based path like `/tmp/myfile.ext`, so the behavior `GetFileName` will not be strictly the same on Linux-based and Windows-based platforms.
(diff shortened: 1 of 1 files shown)


xml/System.IO/Path.xml
```diff
@@ -842,7 +842,9 @@ For more information on file path formats on Windows, see [File path formats on
  The returned value is `null` if the file path is `null`.  
   
  The separator characters used to determine the start of the file name are <xref:System.IO.Path.DirectorySeparatorChar> and <xref:System.IO.Path.AltDirectorySeparatorChar>.  
-  
+ 
+ Because `'\\'` is a legal file name on Linux, `GetFileName` running under Linux-based platforms cannot correctly return the file name from a Windows-based path like `C:\mydir\myfile.ext` but `GetFileName` running under Windows-based platforms can correctly return the file name from a Linux-based path like `/tmp/myfile.ext`, so the behavior `GetFileName` will not be strictly the same on Linux-based and Windows-based platforms.
```
danmosemsft: I suggest you replace Linux with Unix in this text.

dvdme: I think it makes sense because Unix is indeed a better word to describe the platform as it is more general. 

I changed Linux to Unix on commit [f9d19fe7144dff3e514747929ebbb22d5b8cf71b](https://github.com/dotnet/dotnet-api-docs/pull/1488/commits/f9d19fe7144dff3e514747929ebbb22d5b8cf71b).

dvdme:
Thanks @BillWagner

PR #1611, opened 2019-01-15
fix <c> tags when it means <paramref>

TianqiZhang:
Replaced in Sublime Text with `<c>(@\w+)</c>` -> `<paramref name="$1" />`

### DIFF
--- a/xml/System.Data.Common/DbCommandBuilder.xml
+++ b/xml/System.Data.Common/DbCommandBuilder.xml
@@ -453,7 +453,7 @@
         <Parameter Name="useColumnsForParameterNames" Type="System.Boolean" />
       </Parameters>
       <Docs>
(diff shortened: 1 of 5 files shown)


mairaw:
Thanks @TianqiZhang. I have an IntelliSense build going on OPS to see the results of this change in the IntelliSense build.

mairaw:
@TianqiZhang this doesn't really solve the problem. See how that renders in the built IntelliSense for 

M:System.Data.Common.DbCommandBuilder.GetDeleteCommand(System.Boolean)
image

TianqiZhang:
@mairaw I see, you are right, we can't fix it this way, please ignore this PR. I've prepared a code fix in OPS build, let's discuss via original email thread.

PR #1698, opened 2019-01-26
Update Span`1.xml

genlu:
## Summary
Add a missing closing bracket.

n/a


genlu:
@mairaw Thanks for making it so easy to fix our documents :)

PR #1739, opened 2019-01-30
Revised ManualResetEvent remarks

pkulikov:
Accompanies dotnet/docs#10233
/cc: @rpetrusha

### DIFF
--- a/xml/System.Threading/ManualResetEventSlim.xml
+++ b/xml/System.Threading/ManualResetEventSlim.xml
@@ -307,7 +307,7 @@
       </ReturnValue>
       <Docs>
         <summary>Gets whether the event is set.</summary>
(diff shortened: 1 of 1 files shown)


xml/System.Threading/ManualResetEventSlim.xml
```diff
@@ -48,15 +48,15 @@
```
rpetrusha: Unrelated to your PR: The type description, "Provides a slimmed down version of ManualResetEvent", isn't helpful. Could you expand it by adding "to represent a thread synchronization event that, when signaled, must be reset manually."?

pkulikov:
@rpetrusha thank you for review and suggestions. I've addressed all the feedback (also replaced the "slimmed down" words with "lightweight", as it looks to be more common in the docs so far).

PR #1735, opened 2019-01-30
Remove unneeded sentences

nxtn:
In addition, `System.IO` is already included in these examples.

PR #1755, opened 2019-02-01
Remove unneeded sentences

nxtn:
Continued from #1735

### DIFF
--- a/xml/System.Drawing.Printing/PrintDocument.xml
+++ b/xml/System.Drawing.Printing/PrintDocument.xml
@@ -49,8 +49,6 @@
 > [!NOTE]
 >  The example requires that each line fits within the page width.  
   
- Use the <xref:System.ComponentModel>, <xref:System.Windows.Forms>, <xref:System.Drawing>, <xref:System.Drawing.Printing>, and <xref:System.IO> namespaces for this example.  
-  
  [!code-cpp[Classic PrintDocument Example#1](~/samples/snippets/cpp/VS_Snippets_Winforms/Classic PrintDocument Example/CPP/source.cpp#1)]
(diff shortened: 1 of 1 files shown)


NextTurn:
These sentences should appear before incomplete examples, but these are all complete ones with namespaces imported.

mairaw:
I agree but at least on the [first example](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.printing.querypagesettingseventargs?view=netframework-4.7.2) that I opened the example is not complete and it doesn't show the using statements:
image

They're helpful in this scenario, no?

NextTurn:
It was the only mistake. Sorry for that.

NextTurn:
All checked.

There must be a lot of incomplete examples without namespace hints. We could create a new issue to track these.

NextTurn:
I suggest to include the whole file for examples like [this](https://github.com/dotnet/samples/blob/master/snippets/csharp/VS_Snippets_CLR_System/system.Reflection.Emit.ILGenerator.EmitCalli%20Example/CS/source.cs), otherwise the readers will be confused about the parameter/field types.

mairaw:
I agree that we can either expand samples or include namespace hints. That's a massive task and not something we can take on right now. But feel free to open an issue to track this and will gladly accept community contributions with such changes. 

So closing this one based on what you said earlier that was a mistake in this particular PR.

NextTurn:
@mairaw I meant the removal in [QueryPageSettingsEventArgs Class](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.printing.querypagesettingseventargs?view=netframework-4.7.2) was a mistake. Others are OK.

PR #1805, opened 2019-02-06
Changed values of PI, E

### LINKED ISSUE #1094: fix value of Math.PI in syntax section
Please fix the first chapter on this page with the same value as in the remarks and source code:
https://referencesource.microsoft.com/#mscorlib/system/math.cs,fa5fd1f4e5f6f911

On the other hand, the value in remarks and source have 20 digits, which is higher than the significant digits of a double (17), explaining why reflection does NOT show the same value. It does not make much sense. 
Having PI as decimal internally would increase its accuracy, but I suppose it will apply a big change in the whole Math library. Even dedicated external library such as Math.NET is using double for PI constant.

---
#### Document Details

* ID: 8047bcfa-00dd-5106-d44d-9d8ac49e664f
* Version Independent ID: a73a472c-a3be-e54e-0354-7d3a224be282
* Content: [Math.PI Field (System)](https://docs.microsoft.com/en-us/dotnet/api/system.math.pi?f1url=https%3A%2F%2Fmsdn.microsoft.com%2Fquery%2Fdev15.query%3FappId%3DDev15IDEF1%26l%3DEN-US%26k%3Dk(System.Math.PI);k(DevLang-csharp)%26rd%3Dtrue&view=netframework-4.7.2)
* Content Source: [xml/System/Math.xml](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Math.xml)
* Product: **dotnet-api**
* GitHub Login: @rpetrusha
* Microsoft Alias: **ronpet**

rpetrusha:
## Changed values of PI, E

This PR modifies the documented value of Math.PI, Math.E, MathF.PI, and MathF.E.

It also changes their values in portions of the XML that are ordinarily generated by reflection only. This is because reflection formats floating point constants using the default "G" format string rather than the round-trippable "G17" format string (for Doubles) or the "G9" format string (for Singles). I've opened mono/api-doc-tools#412 to address this issue.

Fixes #1094 

//cc @tannergooding

### DIFF
--- a/xml/System/Math.xml
+++ b/xml/System/Math.xml
@@ -1778,11 +1778,11 @@ Dim i2 As Integer = CInt(Math.Ceiling(d2))        ' Result: 7969
       </Docs>
     </Member>
     <Member MemberName="E">
-      <MemberSignature Language="C#" Value="public const double E = 2.71828182845905;" />
-      <MemberSignature Language="ILAsm" Value=".field public static literal float64 E = (2.71828182845905)" />
+      <MemberSignature Language="C#" Value="public const double E = 2.7182818284590451;" />
+      <MemberSignature Language="ILAsm" Value=".field public static literal float64 E = (2.7182818284590451)" />
       <MemberSignature Language="DocId" Value="F:System.Math.E" />
-      <MemberSignature Language="VB.NET" Value="Public Const E As Double  = 2.71828182845905" />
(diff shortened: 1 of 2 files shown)


ladeak:
I think this change leads to some confusion around PI:
https://github.com/dotnet/dotnet-api-docs/issues/2043

PR #1872, opened 2019-02-16
Fix typos return value documentation for Add/AddRange in immutable list types.

PathogenDavid:
The documentation for `Add`/`AddRange` on both `IImutableList<T>` and `ImmutableList<T>` stated that the returned list would be unchanged if the list already contained the specified value(s).

This is inconsistent with both the typical .NET definition of a list and with the behavior of implementation. I suspect this was a copy+paste type from `IImutableSet<T>`. This PR removes the offending language.


PathogenDavid:
Thanks for the merge!

PR #1997, opened 2019-03-06
Update Equality Comparer to clarify statement.

BrianBu01:
It wasn't clear here that just having an equal HashCode would not result with two objects being equated as equal.  Although, it states "and" it isn't clear if the statement is a logical && or "It check both Equals and HashCode."

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Collections.Generic/EqualityComparer`1.xml
+++ b/xml/System.Collections.Generic/EqualityComparer`1.xml
@@ -143,7 +143,9 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Collections.Generic.EqualityComparer%601.Default%2A> property checks whether type `T` implements the <xref:System.IEquatable%601?displayProperty=nameWithType> interface and, if so, returns an <xref:System.Collections.Generic.EqualityComparer%601> that uses that implementation. Otherwise, it returns an <xref:System.Collections.Generic.EqualityComparer%601> that uses the overrides of <xref:System.Object.Equals%2A?displayProperty=nameWithType> and <xref:System.Object.GetHashCode%2A?displayProperty=nameWithType> provided by `T`.  
+ The <xref:System.Collections.Generic.EqualityComparer%601.Default%2A> property checks whether type `T` implements the <xref:System.IEquatable%601?displayProperty=nameWithType> interface and, if so, returns an <xref:System.Collections.Generic.EqualityComparer%601> that uses that implementation. 
+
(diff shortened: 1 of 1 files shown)


rpetrusha:
@safern, could you take a look at this suggested change? The equality predicate sentence seems to me to be incorrect.

safern:
> The equality predicate sentence seems to me to be incorrect.

Agreed. It is kind of confusing, @BrianBu01 from your Issue description I wasn't able to understand what are you trying to achieve with that predicate? Could you please elaborate?

BrianBu01:
Combining the original statement and the statement about IEquatable alone does not convey that both hashcode and equals are evaluated together.  It reads with the possibility of this code being (simplified ignoring nulls):

```
if((var equatable = left as IEquatable<T>) != null && equatable.Equals(right))
{
   return true;
}
if(left.GetHashCode() == right.GetHashCode())
{
   return true;
}
if(left.Equals(right))
{
    return true;
}
return false;
```

Being that is is (still simplified ignoring nulls):
`(var equatable = left as IEquatable<T>) != null && equatable.Equals(right)) || (left.GetHashCode() == right.GetHashCode() && left.Equals(right))`
The 'or' then 'and' makes it a bit hard to be clear in an English statement.  The effort here is attempting to make it 100% clear.  

I guess the English statement "Perform X then Perform Y and Perform Z" is really the confusion, as it is combining English logic with Binary Logic.  Cause it could be "Perform X then (Perform Y and Perform Z)" or "Perform X then Perform Y then Perform Z".   

I hope I have confused you by now lol.

rpetrusha:
Your code shows a test for equality that returns a Boolean, @BrianBu01. This is the Default property, which returns an `EqualityComparer<T>` -- an object that tests `T` for equality and supports customized equality comparisons. The documentation says that:
1. If `T`  implements the `IEquatable<T>` interface, the property returns an `EqualityComparer<T>` that implements `IEquatable<T>`.
2. If `T` does not implement `IEquatable<T>`, it returns an `EqualityComparer<T>` whose Equals method calls `T.Equals` and whose GetHashCode method calls `T.GetHashCode`. 

The implementation in the [.NET Framework](https://referencesource.microsoft.com/#mscorlib/system/collections/generic/equalitycomparer.cs,542680fa5b2d828d) and [.NET Core](https://source.dot.net/#System.Private.CoreLib/src/System/Collections/Generic/ComparerHelpers.cs,a25b39e86cd57fbc) is slightly more complicated, but I'm not sure that the precise detail adds a great deal of value to the documentation.
 
I'm going to close this PR, since the changes don't describe the Default property.

PR #2004, opened 2019-03-08
Added useful details to `PipeStream.ReadMode`

peteraritchie:
## Summary

Added some details to `ReadMode` with regard to implicit calls to `InitializeHandle` and exceptions before `NamedPipeClientStream.Connect` is called.

### DIFF
--- a/xml/System.IO.Pipes/PipeStream.xml
+++ b/xml/System.IO.Pipes/PipeStream.xml
@@ -1494,7 +1494,7 @@
 ## Remarks  
  Anonymous pipes do not support <xref:System.IO.Pipes.PipeTransmissionMode.Message> mode message transmission.  
 
- `ReadMode` will throw <xref:System.InvalidOperationException> if it set before <xref:System.IO.Pipes.PipeStream.InitalizeHandle> succeeds.  If using derived types like <xref:System.IO.Pipes.NamedPipeClientStream>, <xref:System.IO.Pipes.PipeStream.InitalizeHandle> is called by <xref:System.IO.Pipes.NamedPipeClientStream.Connect>.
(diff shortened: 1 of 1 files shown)


BillWagner:
Can you review this @JeremyKuhne 

Thank you.

rpetrusha:
Thanks, @peteraritchie, for submitting your PR. The exception, though, is documented in the Exceptions section, and of the four types derived from NamedPipeClientStream, only the NamedPipeClientStream type has a Connect method. A generalized alternative is to retrieve the value of the PipeStream.IsConnected property before accessing the ReadMode property. 

Thoughts, @jeremykuhne and @peteraritchie.

JeremyKuhne:
A number of properties throw if the handle hasn't been set up or if the class has been disposed. `NamedPipeClientStream` also throws when accessing properties if the connection isn't active. Not having a handle set up isn't something that happens outside of the NamedPipeClientStream as it gets set up in the constructors in the other classes.

As @rpetrusha points out, it is already documented that this throws in this way. I'm fine with calling out how to check if `NamedPipeClientStream` is connected.

peteraritchie:
Thanks for the feedback @rpetrusha.  As you say only NamedPipeClientStream has a Connect method--which makes IsConnected not as obvious.  I think the root of the issue is the opportunity to improve the inheritance hierarchy (if the opportunity for that level of change was there) and thus make PipeStream ReadMode more clear in the unique conditions of the inheritors.  Lacking that, it makes simply inheriting the documentation less than clear.  The message "handle not created" isn't clear when using a specialization of PipeStream that hides that implementation detail.  e.g. what you need to do to avoid the exception with ReadMode depends on the specialization and thus the PipeStream.ReadMode documentation becomes awkward because of it.
I suspect that details of that arguable Liskov violation isn't enough to do the same in the documentation (two wrongs don't make it right); so, I've updated the PR to focus on IsConnected to avoid the exception.

rpetrusha:
Closing and reopening to trigger a new CLA check.

Issue #1992, opened 2019-03-05
Archive bit


CeciAc:
@Snegovikufa commented on [Tue Mar 05 2019](https://github.com/dotnet/dotnet-api-docs.ru-ru/issues/14)

According to https://en.wikipedia.org/wiki/Archive_bit seems that documentation for "Archive" field is wrong.

---
#### Сведения о документе

⚠ *Не вносите правки в этот раздел. Это необходимо для связывания страницы сайта docs.microsoft.com с вопросом на GitHub.*

* ID: 7001712c-bcbb-d062-d857-cc94398fa32c
* Version Independent ID: bf49dcd7-ad7f-309a-53b0-c616a0f73951
* Content: [FileAttributes Enum (System.IO)](https://docs.microsoft.com/ru-ru/dotnet/api/system.io.fileattributes?view=netframework-4.7.2#feedback)
* Content Source: [xml/System.IO/FileAttributes.xml](https://github.com/dotnet/dotnet-api-docs.ru-ru/blob/live/xml/System.IO/FileAttributes.xml)
* Product: **dotnet-api**
* GitHub Login: @dotnet-bot
* Microsoft Alias: **dotnetcontent**

CeciAc:
Hi team,
I'm moving this here since the contributor is reporting an inconsistency
best

MSDN-WhiteKnight:
Dotnet docs are are actually consisent with WinAPI docs on this constant:

[dotnet docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.fileattributes?view=netframework-4.7.2#feedback)

>The file is a candidate for backup or removal.

[WinAPI docs](https://docs.microsoft.com/en-us/windows/desktop/fileio/file-attribute-constants)

>A file or directory that is an archive file or directory. Applications typically use this attribute to mark files for backup or removal .

Neither is contradicting Wikipedia, they just describe it differently. However, i do think that both can benefit from rewording, since current description is too brief in describing mechanics. Better wording should be something like that:

> This file is marked to be included in incremental backup operation. Windows sets this attribute whenever the file is modified, and backup software clears it when processing the file during incremental backup.

rpetrusha:
Your proposed wording sounds good, @MSDN-WhiteKnight, except that perhaps "backup software should clear it" may be better, since it's not done automatically. Would you like to submit a PR to make the correction?

PR #2030, opened 2019-03-13
Update FileAttributes.xml

MSDN-WhiteKnight:
## Summary

Reworded archive attribute description to better reflect mechanics behind it

Fixes https://github.com/dotnet/dotnet-api-docs/issues/1992

### DIFF
--- a/xml/System.IO/FileAttributes.xml
+++ b/xml/System.IO/FileAttributes.xml
@@ -98,7 +98,7 @@
       </ReturnValue>
       <MemberValue>32</MemberValue>
       <Docs>
-        <summary>The file is a candidate for backup or removal.</summary>
+        <summary>This file is marked to be included in incremental backup operation. Windows sets this attribute whenever the file is modified, and backup software should clear it when processing the file during incremental backup.</summary>
       </Docs>
     </Member>
     <Member MemberName="Compressed">


rpetrusha:
@JeremyKuhne, I assume that FileAttributes is only meaningful on a Windows system, since the attributes don't easily map to file attributes on Linux. Is that true? Should we note something about Linux/macOS?

MSDN-WhiteKnight:
Looking at GetAttributes method in [this file](https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/FileStatus.Unix.cs), a limited amount of attributes seems to be supported on non-Windows systems.

JeremyKuhne:
The update is fine.

The mapping for Unix is pretty limited. You can get `FileAttributes.ReadOnly`, `FileAttributes.ReparsePoint`, `FileAttributes.Directory`, and `FileAttributes.Hidden` OR `FileAttributes.Normal`. If the file doesn't exist (or isn't accessible) you get `(FileAttributes)(-1)` (Like Windows).

Hidden shows up if the file starts with a period on Unix. It recently was updated to also respect the hidden flag on MacOS, which is the only Unix OS where you can set the flag currently. `ReadOnly` is the only "flag" you can change universally on Unix (it is actually a permissions operation).

PR #2042, opened 2019-03-15
Updated StringBuilder.Clear

### LINKED ISSUE #1482: StringBuilder.Clear() can change Capacity now
Starting with Net Core 2.1 StringBuilder.Clear() has been optimized and in some way cannot keep the promise anymore to not change the Capacity so the documentation should be changed from:

> Clear is a convenience method that is equivalent to setting the Length property of the current instance to 0 (zero).
>
>Calling the Clear method does not modify the current instance's Capacity or MaxCapacity property. 

to:
> Clear is a convenience method that is equivalent to setting the Length property of the current instance to 0 (zero).

rpetrusha:
## Updated StringBuilder.Clear

Fixes #1482 

//cc @joshfree @maryamariyan @danmosemsft

### DIFF
--- a/xml/System.Text/StringBuilder.xml
+++ b/xml/System.Text/StringBuilder.xml
@@ -3553,17 +3553,15 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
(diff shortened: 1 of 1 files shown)


joshfree:
Thanks @rpetrusha !

PR #2057, opened 2019-03-21
Remove arbitrary time unit

breyed:
The property type is `TimeSpan`, which doesn't imply any particular unit. (If anything, it would be the 100 ns unit of `Ticks`.) Docs for other members of this class should be updated accordingly.

### DIFF
--- a/xml/System.Net/HttpListenerTimeoutManager.xml
+++ b/xml/System.Net/HttpListenerTimeoutManager.xml
@@ -138,8 +138,8 @@
         <ReturnType>System.TimeSpan</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets or sets the time, in seconds, allowed for the request entity body to arrive.</summary>
-        <value>The time, in seconds, allowed for the request entity body to arrive.</value>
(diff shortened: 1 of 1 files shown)


BillWagner:
@karelz Can you look at this PR?  The change LGTM, but I wonder if the original intent was to specify the resolution of the timeout, rather than the units.

Can you verify what the right fix is?

breyed:
I'm not tooled up with do widespread changes to the repro. It's easy for me to make a quick edit in GitHub of one to raise the issue and get a start on the solution, but I'd rather leave it to someone who works on docs day in and day out to do the systematic update.

breyed:
Regarding the unit, my guess is that way back when, some early version (pre-release?) of this API took an int in seconds. Initially, `TimeSpan`s weren't commonly used, but later most timing operations starting taking them as parameters to avoid bugs due to wrong unit assumptions. The code for call may have transitioned without the docs being updated.

rpetrusha:
Thanks, @breyed. I'll make the changes tomorrow, then merge your PR.

rpetrusha:
Thanks, @breyed. I've made the additional changes. We'll merge once our internal build completes successfully.

PR #2058, opened 2019-03-21
Remove exception from StreamWriter.Dispose docs

breyed:
This looks like a copy and paste typo. I'm assuming that `Dispose` isn't going to actually throw an exception.


breyed:
Never mind. I didn't realize that `Dispose` flush because the documentation was wrong. Fixed in #2059.

PR #2059, opened 2019-03-21
Document that StreamWriter.Dispose flushes

breyed:
This is an important detail. It tells you that `Dispose` is a blocking call. It also tells you that it can throw I/O exceptions. And of course that you don't have to call `Flush` or `FlushAsync` yourself first.


BillWagner:
Overall, this looks good to me.  

@JeremyKuhne Can you verify that this is how you'd like this behavior documented?

PR #2065, opened 2019-03-21
Make clear that the task constructor is for advanced use only and offer also TaskCompletionSource<T> as an alternative.

paulomorgado:
## Summary

Sometimes users use the constructor of `Task` and `Task<T>` in non advaced scenarios. Sometimes it's because they don't read the remarks until the end or they just need to control when the task completes.

This PR trys to make clear that the task constructor is for advanced use only by moving the last sentence of the paragraph to its own introdutory paragraph and offer also TaskCompletionSource<T> as an alternative.

/cc @stephentoub @stephencleary

### DIFF
--- a/xml/System.Threading.Tasks/Task.xml
+++ b/xml/System.Threading.Tasks/Task.xml
@@ -198,7 +198,11 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- Rather than calling this constructor, the most common way to instantiate a <xref:System.Threading.Tasks.Task> object and launch a task is by calling the static <xref:System.Threading.Tasks.Task.Run%28System.Action%29?displayProperty=nameWithType> or <xref:System.Threading.Tasks.TaskFactory.StartNew%28System.Action%29?displayProperty=nameWithType> method. The only advantage offered by this constructor is that it allows object instantiation to be separated from task invocation.  
+ This constructor should only be used in advanced scenarios where it is required that the creation and starting of the task is separated.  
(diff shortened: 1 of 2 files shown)


xml/System.Threading.Tasks/Task.xml
```diff
@@ -198,7 +198,11 @@
...
   
 ## Remarks  
- Rather than calling this constructor, the most common way to instantiate a <xref:System.Threading.Tasks.Task> object and launch a task is by calling the static <xref:System.Threading.Tasks.Task.Run%28System.Action%29?displayProperty=nameWithType> or <xref:System.Threading.Tasks.TaskFactory.StartNew%28System.Action%29?displayProperty=nameWithType> method. The only advantage offered by this constructor is that it allows object instantiation to be separated from task invocation.  
+ This constructor should only be used in advanced scenarios where it is required that the creation and starting of the task is separated.  
+  
+ Rather than calling this constructor, the most common way to instantiate a <xref:System.Threading.Tasks.Task> object and launch a task is by calling the static <xref:System.Threading.Tasks.Task.Run%28System.Action%29?displayProperty=nameWithType> or <xref:System.Threading.Tasks.TaskFactory.StartNew%28System.Action%29?displayProperty=nameWithType> method.  
+  
+ If a task with no action is need just for the consumer of an API to have something to await, a <xref:System.Threading.Tasks.TaskCompletionSource%601> should be used.
```
stephentoub: Nit: need => needed

PR #2073, opened 2019-03-22
Update ValueTask`1.xml

baj22:
There is a non-generic version of ValueTask. The documentation was misleading.

## Summary

Describe your changes here.


BillWagner:
Ping @stephentoub 

There is a non-generic `ValueTask`, as this PR mentions. Is that now preferred to the `CompletedTask` property? How should we phrase the guidance for these?

stephentoub:
> There is a non-generic ValueTask, as this PR mentions. Is that now preferred to the CompletedTask property?

No.

A full discussion is here:
https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/

But answering your question in short, for asynchronous methods that don't return a value, the non-generic `Task` is still the preferred choice. The only benefit in this situation of the non-generic `ValueTask` is if it's likely/possible that the implementation would be able to and would use a custom `IValueTaskSource`-based implementation in order to avoid an allocation in the asynchronous case.  That's super advanced, and only impacts the hottest of the hot paths.  And `ValueTask` incurs some overhead (both at runtime and concept count) that `Task` doesn't.

xml/System.Threading.Tasks/ValueTask`1.xml
```diff
@@ -47,7 +47,7 @@ There are tradeoffs to using a <xref:System.Threading.Tasks.ValueTask%601> inste
...
 
-As such, the default choice for any asynchronous method should be to return a <xref:System.Threading.Tasks.Task> or <xref:System.Threading.Tasks.Task%601>. Only if performance analysis proves it worthwhile should a <xref:System.Threading.Tasks.ValueTask%601> be used instead of a <xref:System.Threading.Tasks.Task%601>. There is no non-generic version of <xref:System.Threading.Tasks.ValueTask%601>, as the <xref:System.Threading.Tasks.Task.CompletedTask> property may be used to hand back a successfully completed singleton in the case where a method returning a <xref:System.Threading.Tasks.Task> completes synchronously and successfully.   
+As such, the default choice for any asynchronous method should be to return a <xref:System.Threading.Tasks.Task> or <xref:System.Threading.Tasks.Task%601>. Only if performance analysis proves it worthwhile should a <xref:System.Threading.Tasks.ValueTask%601> be used instead of a <xref:System.Threading.Tasks.Task%601>.
```
BillWagner: Per @stephentoub's comment, I suggest the following change instead: 

> The non generic version of <xref:System.Threading.Tasks.ValueTask>  is not recommended for most scenarios. The <xref:System.Threading.Tasks.Task.CompletedTask> property should be used to hand back a successfully completed singleton in the case where a method returning a <xref:System.Threading.Tasks.Task> completes synchronously and successfully.

baj22:
I made changes per @BillWagner's suggestion.

BillWagner:
Thanks for making the updates @baj22 

I've approved this PR, and I'll :shipit: once the build finishes.

PR #2208, opened 2019-03-28
Update EasingDoubleKeyFrame.xml

stmoy:
## Summary

Add a note that the default easing function for EasingDoubleKeyFrame is Linear.

Fixes this issue (from winrt-api repo): https://github.com/MicrosoftDocs/winrt-api/issues/200


stmoy:
Whoops - wrong doc.

PR #2213, opened 2019-03-29
fix: Udp TransportSettings message

nschonni:
- Tranport -> Transport
- Serivce -> Service


nschonni:
This might be reflected off the actual code and unfixable, so I submitted separately

PR #2221, opened 2019-04-01
Clarify that gcAllowVeryLargeObjects applies to .NET Framework only

jkotas:


jkotas:
Related to https://github.com/dotnet/coreclr/issues/8847

jkotas:
I do not think we want to documenting the internal environment variable in official docs for Core. I expect it to go away at some point. We left it around just in case somebody really needs it. We are removing settings like these over time unless we hear that they are really needed.

rpetrusha:
OK. I'll merge as soon as we're able to merge our current master to live.

PR #2233, opened 2019-04-02
Note that shutdown finalization is .NET Framework only

jkotas:


jkotas:
Related to https://github.com/dotnet/dotnet-api-docs/pull/363 , https://github.com/dotnet/coreclr/pull/23595#issuecomment-478505178

cc @janvorli

PR #2287, opened 2019-04-11
Specify how to identify culture

breyed:
The docs refer to [MS-LCID](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-lcid/a9eac961-e77d-41a6-90a5-ce1a8b0cdb9c) for a list of cultures, but don't say how you should refer to a culture. This change clarifies that the language tag (as specified in the 4th column of the table in MS-LCID) is used.

### DIFF
--- a/xml/System.Resources/NeutralResourcesLanguageAttribute.xml
+++ b/xml/System.Resources/NeutralResourcesLanguageAttribute.xml
@@ -193,7 +193,7 @@ csc Example.cs /resource:ExampleResources.resources
         <Parameter Name="cultureName" Type="System.String" />
       </Parameters>
       <Docs>
-        <param name="cultureName">The name of the culture that the current assembly's neutral resources were written in.</param>
+        <param name="cultureName">The language tag of the culture that the current assembly's neutral resources were written in.</param>
         <summary>Initializes a new instance of the <see cref="T:System.Resources.NeutralResourcesLanguageAttribute" /> class.</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[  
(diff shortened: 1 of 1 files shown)


rpetrusha:
Thanks, @breyed, for submitting this PR. It serves to identify a number of issues, including the phrase we use to describe language/region names throughout the documentation (as "culture name" or "name of culture") and the Go Global Developer Center, the original home of the linked table, has been retired.

There are a number of problems with changing the term we use to describe language/region:
- In changing the name so that it conforms to a column title in a table, we're making the documentation dependent on a table in the Windows documentation set that conceivably could change.
- **Language tag** is the term used by RFC 1766 and BCP 47. But we can't expect that developers are familiar with the standards terminology, especially when the string includes more than just a language identifier.  
- Page view data shows that only about 0.5% of users follow the link to the page that contains the table. We'd be standardizing terminology based on a column title that the vast majority of users don't see.
- The term introduces some confusion for developers who target Unix systems. The ICU documentation, for example, describes a language/region code as a **locale ID**. 

For these reasons, and after consulting with several colleagues, I've decided to close your PR. I will open an issue, however, that provides for renaming the link to the Go Global Developer Center this week, as well as adding a link to BCP 47.

breyed:
I wondered about that. I'd never heard of language tag before following the spec link while trying to understand a [VS designer bug](https://developercommunity.visualstudio.com/content/problem/527917/designer-shows-wrong-neutrallanguage-for-csproj.html).

More important than the particular terminology is that the docs make it clear what is expected. Can you use long names (like maybe "English (United States)"), short names (e.g. "en-US"), or either? The prior text was ambiguous. "Language tag" is precise, but as you noted, the term is obscure. Some clear and precise descriptive text along with a couple examples would probably be best.

rpetrusha:
I agree -- a clearer description would be useful, since long names don't work. Since culture name is used in some many place, the best solution is probably to create an include file that contains the appropriate tex. It can then be included in the numerous APIs that include a culture name. I'll open an issue for it.

PR #2300, opened 2019-04-13
Update AssemblyLoadContext.xml

jkotas:
Clarify return value of `AssemblyLoadContext.GetLoadContext` for assemblies not provided by the runtime.


jkotas:
cc @sdmaclea

PR #2344, opened 2019-04-20
Update CharSet.Ansi summary

nxtn:
Contributes to dotnet/docs#11396

/cc @jkoritzinsky

### DIFF
--- a/xml/System.Runtime.InteropServices/CharSet.xml
+++ b/xml/System.Runtime.InteropServices/CharSet.xml
@@ -94,7 +94,7 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>Marshal strings as multiple-byte character strings: ANSI on Windows, and UTF-8 on Unix.</summary>
+        <summary>Marshal strings as multiple-byte character strings: the system default [Windows (ANSI) code page](/windows/win32/intl/code-pages) on Windows, and UTF-8 on Unix.</summary>
       </Docs>
     </Member>
     <Member MemberName="Auto">


rpetrusha:
@AaronRobinsonMSFT @jkoritzinsky, could you review this PR, please?

xml/System.Runtime.InteropServices/CharSet.xml
```diff
@@ -94,7 +94,7 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>Marshal strings as multiple-byte character strings.</summary>
+        <summary>Marshal strings as multiple-byte character strings: ANSI on Windows, and UTF-8 on Unix.</summary>
```
AaronRobinsonMSFT: Broadly speaking, this is true but there is a little more nuance. Technically it is the system default ANSI code page - see the description of `CP_ACP` in https://docs.microsoft.com/windows/desktop/api/stringapiset/nf-stringapiset-widechartomultibyte.

AaronRobinsonMSFT: If interested, the specific code that does this marshaling is at:

Windows: https://github.com/dotnet/coreclr/blob/07b3afc27304800f00975c8fd4836b319aaa8820/src/System.Private.CoreLib/shared/System/Runtime/InteropServices/Marshal.Windows.cs#L58-L87

Non-Windows:
https://github.com/dotnet/coreclr/blob/07b3afc27304800f00975c8fd4836b319aaa8820/src/System.Private.CoreLib/shared/System/Runtime/InteropServices/Marshal.Unix.cs#L39-L53

NextTurn: The best solution may be to add more remarks in [Charsets and marshaling](https://github.com/dotnet/docs/blob/staging/docs/standard/native-interop/charset.md) and add a reference to it here (like #2342).

AaronRobinsonMSFT: @NextTurn I think that would be a reasonable solution.

xml/System.Runtime.InteropServices/CharSet.xml
```diff
@@ -94,7 +94,7 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>Marshal strings as multiple-byte character strings.</summary>
+        <summary>Marshal strings as multiple-byte character strings: the system default [Windows (ANSI) code page](/windows/win32/intl/code-pages) on Windows, and UTF-8 on Unix.</summary>
```
jkoritzinsky: On Windows it marshals to the MBCS of the current system culture code page, not the system default code page.

AaronRobinsonMSFT: Not according to the code I mentioned above. `CP_ACP` is specifically defined to be the system default. See the links to source above.

jkoritzinsky: Well I guess I'm wrong here then. When I was testing this a while back I must have figured out how to change the system default of my VM instead of the current culture. Sorry for the confusion.

AaronRobinsonMSFT: This is also a case where the documentation for these kind of APIs could be old/out of date. I am assuming the documentation is correct, I have _not_ verified the actual behavior of that argument.

NextTurn:
This is ready to merge, @rpetrusha.

PR #2349, opened 2019-04-23
Note that Path.GetExtension returns leading period

barnson:
## Summary

Add blurb (verbatim from `returns` text) to `summary` noting that the returned string contains a leading period. The `summary` text is what shows up in IntelliSense, so noting the leading period feels high-value.


IanYates:
100% yes!
I don't want to count the number of times I've cranked up linqpad to quickly run a Path.XYZ() method to see what's handled.
Thank you

PR #2364, opened 2019-04-25
Update ValueTask.Result remarks

StephenCleary:
## Summary

* Add note that `Result` is not intended for application code.

* Add section for semantics when wrapping `IValueTaskSource<T>`.

### DIFF
--- a/xml/System.Threading.Tasks/ValueTask`1.xml
+++ b/xml/System.Threading.Tasks/ValueTask`1.xml
@@ -47,6 +47,15 @@
 ## Remarks  
 A <xref:System.Threading.Tasks.ValueTask%601> instance may either be awaited or converted to a <xref:System.Threading.Tasks.Task%601> using <xref:System.Threading.Tasks.ValueTask%601.AsTask%2A>. A <xref:System.Threading.Tasks.ValueTask%601> instance may only be awaited once, and consumers may not call <xref:System.Threading.Tasks.ValueTask%601.GetAwaiter> until the instance has completed. If these limitations are unacceptable, convert the <xref:System.Threading.Tasks.ValueTask%601> to a <xref:System.Threading.Tasks.Task%601> by calling <xref:System.Threading.Tasks.ValueTask%601.AsTask%2A>.
 
+The following operations should never be performed on a <xref:System.Threading.Tasks.ValueTask%601> instance:
+
+  - Awaiting the instance multiple times.
+  - Calling <xref:System.Threading.Tasks.ValueTask%601.AsTask%2A> multiple times.
+  - Using `.Result` or `.GetAwaiter().GetResult()` when the operation hasn't yet completed, or using them multiple times.
+  - Using more than one of these techniques to consume the instance.
+
+If you do any of the above, the results are undefined.
+
 A method may return an instance of this value type when it's likely that the result of its operation will be available synchronously, and when it's expected to be invoked so frequently that the cost of allocating a new <xref:System.Threading.Tasks.Task%601> for each call will be prohibitive.   
 
 There are tradeoffs to using a <xref:System.Threading.Tasks.ValueTask%601> instead of a <xref:System.Threading.Tasks.Task%601>. For example, while a <xref:System.Threading.Tasks.ValueTask%601> can help avoid an allocation in the case where the successful result is available synchronously, it also contains two fields, whereas a <xref:System.Threading.Tasks.Task%601> as a reference type is a single field. This means that a method call returns two fields worth of data instead of one, which is more data to copy. It also means, that if a method that returns a <xref:System.Threading.Tasks.ValueTask%601> is awaited within an async method, the state machine for that async method will be larger, because it must store a struct containing two fields instead of a single reference.   


BillWagner:
Thanks for making this changes @StephenCleary This looks correct to me.

Ping @stephentoub for a review on the `IValueTaskSource<T>` remarks.

xml/System.Threading.Tasks/ValueTask`1.xml
```diff
@@ -661,9 +661,13 @@ As such, the default choice for any asynchronous method should be to return a <x
         <remarks>
           <format type="text/markdown"><![CDATA[  
 ## Remarks  
+ This method is intended for compiler use rather than for use in application code.
```
stephentoub: Why was this added?

xml/System.Threading.Tasks/ValueTask`1.xml
```diff
@@ -661,9 +661,13 @@ As such, the default choice for any asynchronous method should be to return a <x
...
+ If this <xref:System.Threading.Tasks.ValueTask%601> wraps a <xref:System.Threading.Tasks.Task%601>, the behavior of <xref:System.Threading.Tasks.ValueTask%601.Result%2A> is similar to the behavior of accessing <xref:System.Threading.Tasks.Task%601.Result%2A> on the wrapped task: if the task hasn't completed, accessing the property blocks the calling thread until it completes; if the task has completed successfully, the property returns the result; if the task has faulted or was cancelled, accessing the property throws an exception. The thrown exception is not wrapped in an <xref:System.AggregateException>, which is different from the behavior of <xref:System.Threading.Tasks.Task%601.Result%2A> in the same situation.
+ 
+ If this <xref:System.Threading.Tasks.ValueTask%601> wraps a <xref:System.Threading.Tasks.Sources.IValueTaskSource%601>, the behavior of <xref:System.Threading.Tasks.ValueTask%601.Result%2A> depends on the behavior of the wrapped <xref:System.Threading.Tasks.Sources.IValueTaskSource%601>. Accessing this property may throw an exception if the <xref:System.Threading.Tasks.Sources.IValueTaskSource%601> has not completed.
```
stephentoub: I'd like to be more strict than this and say that that the result should only be accessed when it's known that the ValueTask has already completed; the underlying IValueTaskSource may not expect a call in any other situation.  It also shouldn't be accessed more than once.

xml/System.Threading.Tasks/ValueTask`1.xml
StephenCleary: I inferred this from the "short rule" in [your blog post here](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/)

> with a ValueTask or a ValueTask<TResult>, you should either await it directly (optionally with .ConfigureAwait(false)) or call AsTask() on it directly, and then never use it again

So `Result` is off the beaten path. Especially with the newer `IValueTaskSource` scenario, where `Result` should only be called after you (somehow) know the `ValueTask` is completed. So the use case for `ValueTask<T>.Result` is much smaller than that of `Task<T>.GetAwaiter().GetResult()` (which is marked with this same warning).

That said, I have no problem removing this.

xml/System.Threading.Tasks/ValueTask`1.xml
StephenCleary: I'm totally fine with that.

Also, I would encourage massively simplifying all of this. To me, what the `ValueTask<T>` is wrapping is an implementation-specific detail. Thinking of a first version of a library that wraps `Task<T>`, and then v2 wraps a `IValueTaskSource<T>`. In this scenario, `Result` in particular has a significant semantic change with no API differences to act as a warning.

So, a more ideal documentation for this property would be something simpler (and stricter) saying pretty much what you said, and completely ignoring "if this wraps..." situations.

stephentoub: > So, a more ideal documentation for this property would be something simpler (and stricter) saying pretty much what you said, and completely ignoring "if this wraps..." situations.

I'm fine with that.

xml/System.Threading.Tasks/ValueTask`1.xml
stephentoub: I think we should remove it.  It's fine to warn people off from using it (i.e. "just use await"), but the C# compiler doesn't actually use this property :)

BillWagner:
@StephenCleary Can you ping me when you make the updates you and Stephen agreed to? After that, this should be ready.

BillWagner:
Ping @StephenCleary Do you want to update this PR? Or would you prefer we incorporate @stephentoub 's feedback on our side?

StephenCleary:
I added some usage comments based on [Stephen Toub's blog post](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/). The new Remarks paragraph is targeting consumers, which should be much more common than producers, so I put it at the top of the Remarks section.

I've written it from the perspective of my opinion: that consumers of `ValueTask<T>` should not assume anything about the style of that `ValueTask<T>`. Consumers are more restricted in what they can do, so that producers can change the style of their `ValueTask<T>` without causing compatibility problems.

This brought up an implementation question: for the `Task<T>`-wrapping style of `ValueTask<T>`, should `Result` be changed to throw if not completed? Currently, it still has the old blocking behavior.

BillWagner:
Thanks for updating this @StephenCleary. The new updates capture the limitations well.  I'm wondering if the new version begs some reader questions, namely "what happens if I do the wrong thing?"  

I think the key comes from this section of @stephentoub 's [blog post](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/) (edited for brevity below):

> In general, the following operations should never be performed on a `ValueTask` / `ValueTask<TResult>`:

> - Awaiting a `ValueTask` / `ValueTask<TResult>` multiple times. 
> - Awaiting a `ValueTask` / `ValueTask<TResult>` concurrently.
> - Using `.GetAwaiter().GetResult()` when the operation hasn’t yet completed. 

I'm thinking we should list those three operations in the **Remarks** section, and state that the result of any of those operations is "undefined". 

What are your thoughts @stephentoub ?

stephentoub:
> What are your thoughts

Makes sense to me to include those.  As @StephenCleary says, a developer needs to assume the least-common denominator for an arbitrary `ValueTask{TResult}` they're handed, which means in general they need to adhere to the listed constraints.  Only in the rare case where they have more explicit knowledge about how the `ValueTask{TResult}` was created could they choose to safely violate these.

StephenCleary:
Agreed.

@BillWagner As an old-school C++ "language lawyer", the term "undefined" has a very specific meaning to me, and it's the same way you're using it here. It's been a long time since I read the C# language specs, but I don't believe they use the term "undefined" like that - it's more like implicitly undocumented - just never mentioned at all. The C# language spec only has positive assertions. In the .NET world, should we use "undefined" or "undefined behavior" in a more C++-meaning way? Would that be understood?

BillWagner:
@StephenCleary 

> the term "undefined" has a very specific meaning to me, and it's the same way you're using it here.
That's the goal.

> but I don't believe they use the term "undefined" like that - it's more like implicitly undocumented - just never mentioned at all

There are a number of instances in both the language spec and APIs where we use it. (Hopefully small, because it is a bit of a crutch).  Examples include:

- The contents of the `Value` property on a nullable type when `HasValue` is false. 
- The state of an enumerator when a collection is modified.

I believe it that stating the behavior is undefined, and recommending against using it will be understood.

Thanks for raising the concern. It is important for us to make sure what we write is understood.

StephenCleary:
I've updated the text with a new paragraph @BillWagner. I did add a few more restrictions which I assume are correct but might actually be a bit too strict - would like @stephentoub to review:

```
The following operations should never be performed on a <xref:System.Threading.Tasks.ValueTask%601> instance:

  - Awaiting the instance multiple times.
  - Calling <xref:System.Threading.Tasks.ValueTask%601.AsTask%2A> multiple times.
  - Using `.Result` or `.GetAwaiter().GetResult()` when the operation hasn't yet completed, or using them multiple times.
  - Using more than one of these techniques to consume the instance.

If you do any of the above, the results are undefined.
```

I'm assuming that `await`, `AsTask`, `Result`, and `GetAwaiter().GetResult()` can all be used as "triggers" that the `ValueTask<T>` has been "consumed" and can be reused.

stephentoub:
@StephenCleary, everything you said is correct.

Issue #1770, opened 2019-02-03
Incorrect parameter names are used throughout the documents


Gnbrkm41:
the word `index` within the inline code block has been used throughout the document, despite the fact there's no parameter named `index`.

"`start` or `start` + `index` is less than zero or greater than Length.": it should be `length` instead of `index`.

"&gt;A span that consists of length elements from the current span starting at index.": Starting with a closing angle bracket, and it must be `start` instead of `index`.

And so on.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1e8a1983-0f84-abb2-fdb7-92b0c1fb0052
* Version Independent ID: 41e2a8f5-e794-629e-9720-d339f081fb8c
* Content: [Span&lt;T&gt;.Slice Method (System)](https://docs.microsoft.com/en-us/dotnet/api/system.span-1.slice?view=netcore-2.2#System_Span_1_Slice_System_Int32_)
* Content Source: [xml/System/Span`1.xml](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Span`1.xml)
* Product: **dotnet-api**
* GitHub Login: @dotnet-bot
* Microsoft Alias: **dotnetcontent**

Gnbrkm41:
This is also true for [documents for ReadOnlySpan<T>.Slice](https://docs.microsoft.com/en-gb/dotnet/api/system.readonlyspan-1.slice) method group.

PR #2466, opened 2019-05-17
correct span.slice docs

### LINKED ISSUE #1771: span.Slice(span.Length) do not throw, although the documentation says it should
See dotnet/corefx#35063 for more details.

---
#### Document Details

* ID: 1e8a1983-0f84-abb2-fdb7-92b0c1fb0052
* Version Independent ID: 41e2a8f5-e794-629e-9720-d339f081fb8c
* Content: [Span&lt;T&gt;.Slice Method (System)](https://docs.microsoft.com/en-us/dotnet/api/system.span-1.slice?view=netcore-2.2#System_Span_1_Slice_System_Int32_)
* Content Source: [xml/System/Span`1.xml](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Span`1.xml)
* Product: **dotnet-api**
* GitHub Login: @dotnet-bot

Anipik:
Fixes #1771

### DIFF
--- a/xml/System/Span`1.xml
+++ b/xml/System/Span`1.xml
@@ -922,7 +922,7 @@ Two <xref:System.Span%601> objects are equal if they have different lengths or i
         <returns>A span that consists of all elements of the current span from <paramref name="index" /> to the end of the span.</returns>
         <remarks>To be added.</remarks>
         <exception cref="T:System.ArgumentOutOfRangeException">
-          <paramref name="index" /> is less than zero or greater than or equal to <see cref="T:System.Span`1.Length" />.</exception>
+          <paramref name="start" /> is less than zero or greater than <see cref="T:System.Span`1.Length" />.</exception>
       </Docs>
     </Member>
     <Member MemberName="Slice">
--- a/xml/System/Memory`1.xml
+++ b/xml/System/Memory`1.xml
@@ -619,7 +619,7 @@ The garbage collector will not move the memory until the returned <xref:System.B
         <returns>An object that contains all elements of the current instance from <paramref name="start" /> to the end of the instance. </returns>
         <remarks>To be added.</remarks>
         <exception cref="T:System.ArgumentOutOfRangeExceptionException">
-          <paramref name="start" /> is less than zero or greater than or equal to <see cref="P:System.Memory`1.Length" />. </exception>
+          <paramref name="start" /> is less than zero or greater than <see cref="P:System.Memory`1.Length" />. </exception>
       </Docs>
     </Member>
     <Member MemberName="Slice">
@@ -686,7 +686,7 @@ The garbage collector will not move the memory until the returned <xref:System.B
         <returns>An object that contains <paramref name="length" /> elements from the current instance starting at <paramref name="start" />. </returns>
         <remarks>To be added.</remarks>
         <exception cref="T:System.ArgumentOutOfRangeExceptionException">
-          <paramref name="start" /> is less than zero or greater than or equal to <see cref="P:System.Memory`1.Length" />.
+          <paramref name="start" /> is less than zero or greater than <see cref="P:System.Memory`1.Length" />.
 
 -or-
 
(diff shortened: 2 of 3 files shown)


Anipik:
cc @ahsonkhan

xml/System/Span`1.xml
```diff
@@ -607,7 +607,7 @@ The `GetPinnableReference` method returns a [ref struct](~/docs/csharp/reference
         <value>The element at the specified index.</value>
         <remarks>To be added.</remarks>
         <exception cref="T:System.IndexOutOfRangeException">
-          <paramref name="index" /> is less than zero or greater than or equal to <see cref="P:System.Span`1.Length" />.</exception>
```
ahsonkhan: This is an incorrect change and should be reverted. The indexer will fail for equal to:

> Thrown when index less than 0 or index greater than or equal to Length
https://github.com/Anipik/coreclr/blob/83fc9359bbe09e10235545749b7794fa6dd1f114/src/System.Private.CoreLib/shared/System/Span.Fast.cs#L139-L155

xml/System/Span`1.xml
```diff
@@ -922,7 +922,7 @@ Two <xref:System.Span%601> objects are equal if they have different lengths or i
         <returns>A span that consists of all elements of the current span from <paramref name="index" /> to the end of the span.</returns>
         <remarks>To be added.</remarks>
         <exception cref="T:System.ArgumentOutOfRangeException">
-          <paramref name="index" /> is less than zero or greater than or equal to <see cref="T:System.Span`1.Length" />.</exception>
+          <paramref name="start" /> is less than zero or greater than <see cref="T:System.Span`1.Length" />.</exception>
```
ahsonkhan: What about all the other overloads? I would imagine this PR to update the documentation for all the APIs that were fixed in https://github.com/dotnet/coreclr/pull/24612

So, ReadOnlySpan, Span, ReadOnlyMemory, Memory, and MemoryExtensions

cc @mairaw

ahsonkhan:
@mairaw, @rpetrusha - there seems to be some confusing docs around span slicing (`start` vs `index`).

For example:
https://github.com/dotnet/dotnet-api-docs/blob/8b0ac1b1b4c78c681022eac86397e1de2d5a701a/xml/System/Span%601.xml#L990

> A span that consists of length elements from the current span starting at index.

https://github.com/dotnet/dotnet-api-docs/blob/8b0ac1b1b4c78c681022eac86397e1de2d5a701a/xml/System/Span%601.xml#L993

> start or start + index is less than zero or greater than Length.

I don't think either of these make much sense since we don't have an index parameter. The parameter is `start`.

Shouldn't it be something like:
> A span that consists of length elements from the current span beginning at start.

> start. length, or start + length is less than zero or greater than Length.

For reference, here's the source:
https://github.com/dotnet/coreclr/blob/0bd61335d749c4af972f83675b2e9e61149c5204/src/System.Private.CoreLib/shared/System/Span.Fast.cs#L339-L365

ahsonkhan:
I would leave the word-smithing to the docs team, but we should avoid using `index` and try to be precise on the semantics to avoid confusion.

cc @stephentoub, @jkotas, @GrabYourPitchforks

xml/System/Span`1.xml
Anipik: MemoryExtensions  and readonlyMemory doesnot contain this problem. I corrected the ReadOnlySpan in the next commit

ahsonkhan: > MemoryExtensions and readonlyMemory doesnot contain this problem. 

Oh I see. MemoryExtensions/ReadOnlyMemory don't contain the docs yet: https://docs.microsoft.com/en-us/dotnet/api/system.memoryextensions.asspan?view=netcore-3.0#System_MemoryExtensions_AsSpan_System_String_System_Int32_
https://docs.microsoft.com/en-us/dotnet/api/system.readonlymemory-1.slice?view=netcore-3.0#System_ReadOnlyMemory_1_Slice_System_Int32_System_Int32_

Gnbrkm41:
Please close #1770 as well once the start/index issue is resolved :^)

ahsonkhan:
@Gnbrkm41 - are you up for submitting a PR for the fix?

Gnbrkm41:
@ahsonkhan I could certainly look into it I guess?

PR #2471, opened 2019-05-20
Port System.Text.Json source comments to Docs (second attempt)

carlossanlop:
Automatically port triple slash comments found in System.Text.Json to dotnet-api-docs.

Second attempt - The first PR got in a bad state (unresolvable conflicts) because several APIs in this namespace got changed. This PR now contains the most up-to-date API versions.

### DIFF
--- a/xml/System.Text.Json/Utf8JsonWriter.xml
+++ b/xml/System.Text.Json/Utf8JsonWriter.xml
@@ -2030,7 +2030,7 @@ Validation is enabled, and this write operation would produce invalid JSON.</exc
       <Docs>
         <param name="utf8PropertyName">The UTF-8 encoded property name of the JSON object to be written.</param>
         <param name="value">The value to be written as a JSON string as part of the name/value pair.</param>
-        <summary>Writes a property name specified as a read-only span of bytes and a <see cref="T:System.DateTime" /> value (as a JSON string) as part of a name/value pair of a JSON object.</summary>
+        <summary>Writes a UTF-8 property name and a <see cref="T:System.DateTime" /> value (as a JSON string) as part of a name/value pair of a JSON object.</summary>
         <remarks>
           <format><![CDATA[
 
@@ -2064,7 +2064,7 @@ The property name is escaped before writing.
       <Docs>
         <param name="utf8PropertyName">The UTF-8 encoded property name of the JSON object to be written.</param>
         <param name="value">The value to be written as a JSON string as part of the name/value pair.</param>
-        <summary>Writes a property name specified as a read-only span of bytes and a <see cref="T:System.DateTimeOffset" /> value (as a JSON string) as part of a name/value pair of a JSON object.</summary>
+        <summary>Writes a UTF-8 property name and a <see cref="T:System.DateTimeOffset" /> value (as a JSON string) as part of a name/value pair of a JSON object.</summary>
         <remarks>
           <format><![CDATA[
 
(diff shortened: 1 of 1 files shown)


carlossanlop:
Adding area owners @ahsonkhan, @steveharter for review.

Please take a look whenever you get a chance.

I added some logic to the porting tool so it could automatically resolve the spacing issues we had last time. Let me know if you still see some inconsistencies.

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -198,9 +215,17 @@ The `CurrentDepth` property tracks the recursive depth of the nested objects / a
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Asynchronously commits any leftover JSON text that has not yet been flushed and releases all resources used by the current instance.</summary>
         <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
```
rpetrusha: Same comment as for Dispose.

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -174,8 +183,16 @@ The `CurrentDepth` property tracks the recursive depth of the nested objects / a
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <summary>Commits any leftover JSON text that has not yet been flushed and releases all resources used by the current instance.</summary>
+        <remarks>
```
rpetrusha: Are these remarks actually correct. Typically, Dispose closes the underlying stream, so advancing is not an issue, and reuse of the stream or buffer throws an ObjectDisposedException.

carlossanlop: @ahsonkhan can you confirm the remarks are describing the correct behavior?

@rpetrusha the remarks were correctly ported from the triple slash comments:

https://github.com/dotnet/corefx/blob/2fc59abd09c990121526360d90810e19723cb6cb/src/System.Text.Json/src/System/Text/Json/Writer/Utf8JsonWriter.cs#L313-L324

xml/System.Text.Json/Utf8JsonWriter.xml
carlossanlop: @ahsonkhan Can you also please confirm these remarks describe the correct behavior?

@rpetrusha Same case, the triple slash comments were properly ported:

https://github.com/dotnet/corefx/blob/2fc59abd09c990121526360d90810e19723cb6cb/src/System.Text.Json/src/System/Text/Json/Writer/Utf8JsonWriter.cs#L343-L354

xml/System.Text.Json/Utf8JsonWriter.xml
carlossanlop: Note: The tool detected a &lt;see cref="... inside the remark and automatically converted it to &lt;xref:

carlossanlop:
@rpetrusha I reviewed your two commits, thanks for doing those changes.

Can we incorporate the contents of your review #2422 into this one? I think my tool did not catch those modifications in this PR, but the easiest thing to do is to merge it all at once here. I'll let you decide, let me know and I'll submit the modification.

xml/System.Text.Json/Utf8JsonWriter.xml
ahsonkhan: > Are these remarks actually correct. 

Yes, at least given the implementation and intent of the type.

Since the `Utf8JsonWriter` has two modes and is re-settable (if you call the Reset without any parameters), it does not take ownership of the underlying stream or dispose of it when Dispose is called (in stream mode). In the case of the `IBufferWriter` mode, there is no concept of dispose at all, so the remark is just stating what dispose on the writer does to the underlying output sink (whether `Stream` or `IBufferWriter`).

> Typically, Dispose closes the underlying stream, so advancing is not an issue, and reuse of the stream or buffer throws an ObjectDisposedException.

We had discussed adding a `leaveOpen` (or `dispose`) bool to the ctor in the future (like [StreamReader](https://apisof.net/catalog/System.IO.StreamReader..ctor(Stream,Encoding,Boolean,Int32,Boolean))), which would dispose the stream if set to false.

xml/System.Text.Json/Utf8JsonWriter.xml
ahsonkhan: Yes, they are valid. If the question is, should the implementation/design change, that's a different story all together :)

carlossanlop:
@rpetrusha This one's ready. Ahson confirmed the remarks are correct.

rpetrusha:
Thanks, @carlossanlop. I'll merge now.

PR #2420, opened 2019-05-08
Update ns-System.Net.Mail.xml

stephenegriffin:
Lift the warning from SmtpClient to the namespace to make it clearer this is now obsolete.

## Summary

Describe your changes here.

Fixes #Issue_Number (if available)

PR #2483, opened 2019-05-22
Added obsolete note to SmtpClient type summary

rpetrusha:
## Added obsolete note to SmtpClient type summary

Related to #2420

### DIFF
--- a/xml/System.Net.Mail/SmtpClient.xml
+++ b/xml/System.Net.Mail/SmtpClient.xml
@@ -35,7 +35,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>Allows applications to send email by using the Simple Mail Transfer Protocol (SMTP).</summary>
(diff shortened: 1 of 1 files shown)


xml/System.Net.Mail/SmtpClient.xml
```diff
@@ -35,7 +35,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>Allows applications to send email by using the Simple Mail Transfer Protocol (SMTP).</summary>
+    <summary>Allows applications to send email by using the Simple Mail Transfer Protocol (SMTP). The <c>SmtpClient</c> type is now obsolete.</summary>
```
BillWagner: Is there are recommended alternative?  That's going to be an obvious question for readers when they see this.

rpetrusha: The recommended alternatives are listed in the obsolete message, which appears in the syntax section, @BillWagner. There is no .NET base class library alternative.

PR #2503, opened 2019-05-26
Improve Volatile class documentation

MSDN-WhiteKnight:
## Summary

- Replace incorrect statement about `volatile` keyword in remarks. `volatile` keyword does not techinically use Volatile.Read / Volatile.Write methods, but only causes all field access instructions to be volatile operations (which is effectively the same, but without a method call).

- Add See also link to the `volatile` keyword documentation.

Related to https://github.com/dotnet/dotnet-api-docs.ru-ru/issues/22

### DIFF
--- a/xml/System.Threading/Volatile.xml
+++ b/xml/System.Threading/Volatile.xml
@@ -50,7 +50,7 @@
   
  ]]></format>
     </remarks>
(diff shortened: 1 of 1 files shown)


BillWagner:
@kouvel Can you review this change?  It looks correct to me, but I'd like your review as well.

PR #2546, opened 2019-06-04
Emphasized that String.Copy is obsolete

### LINKED ISSUE #2382: Need to doc that String.Copy is obsolete as of .NET Core 3
Context: https://github.com/dotnet/corefx/issues/33602 and https://github.com/dotnet/corefx/blob/467d45e3950e8daeda9373396da8061f5b8b3787/src/System.Runtime/ref/System.Runtime.cs#L2330-L2331

We obsoleted (as warning) the `String.Copy` method in .NET Core 3 and created the fwlink https://go.microsoft.com/fwlink/?linkid=2084035 for inclusion in the obsolete message. Current the fwlink points back to the original GitHub issue, but we should update [the `String.Copy` docs](https://docs.microsoft.com/en-us/dotnet/api/system.string.copy?view=netcore-3.0) to mention alternatives and redirect the fwlink to the official documentation.

Ideally we'd want to say something like the below, with examples of proper usage.

* You probably didn't ever need to call this method at all, and you can just use the original `string` instance since `string` instances are immutable.

* If you need to create a mutable buffer with the same contents as the original string, use `string.ToCharArray()` or the `StringBuilder(string)` ctor.

* If you need to create a mutable copy of the string for interop purposes, use the method `Marshal.StringToHGlobalUni(string)`. Remember to free the unmanaged memory when you're done.

There may be other scenarios but I can't think of them offhand.

rpetrusha:
## Emphasized that String.Copy is obsolete

[Internal review URL](https://review.docs.microsoft.com/en-us/dotnet/api/system.string.copy?view=netframework-4.8&branch=pr-en-us-2546)

Fixes #2382 

//cc @GrabYourPitchforks

### DIFF
--- a/xml/System/String.xml
+++ b/xml/System/String.xml
@@ -4275,17 +4275,31 @@ You can download the [Sorting Weight Tables](https://www.microsoft.com/download/
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
-## Remarks  
- The <xref:System.String.Copy%2A> method returns a <xref:System.String> object that has the same value as the original string but represents a different object reference. It differs from an assignment operation, which assigns an existing string reference to an additional object variable. The example illustrates the difference.  
+## Remarks
+
+The `Copy` method returns a <xref:System.String> object that has the same value as the original string but represents a different object reference. It differs from an assignment operation, which assigns an existing string reference to an additional object variable.
+
+> [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


BillWagner:
As I read the alternatives to `String.Copy`, I found myself wondering why `Span<char>` wasn't mentioned. Would that be an alternative to unsafe code? I think we should recommend using it, or say why we don't in these scenarios.

Beyond that question, this LGTM.

cc @GrabYourPitchforks

rpetrusha:

PR #2619, opened 2019-06-19
Corrected SetText exception

Youssef1313:
The `Clipboard.SetText` function doesn't give an exception for me when I pass empty string. It does only with `null`.

image

-----------------
-----------------

image


Youssef1313:
This behavior was because I was referencing PresentationCore.dll instead of System.Windows.Forms.dll
The documentation for [SetText in PresentationCore is correct](https://docs.microsoft.com/en-us/dotnet/api/system.windows.clipboard.settext?view=netframework-4.8#System_Windows_Clipboard_SetText_System_String_)

rpetrusha:
Thanks for opening this PR, @Youssef1313. The documentation for System.Windows.Forms.Clipboard.SetText is correct, though; if String.IsNullOrEmpty returns true, the method throws an ArgumentNullException. 

Although similar exceptions for similar conditions is certainly desirable, it's also true that the WinForms code base is substantially older than WPF's, which does conform to the behavior that you'd expect. For compatibility reasons, there's really nothing that can be done about throwing an ArgumentNullException for an empty string. While it is possible to change an exception to a more derived exception (such as ArgumentException to ArgumentNullException), it's a breaking change to change to a less derived exception (such as ArgumentNullException to ArgumentException).

Since the documentation is correct, I'll close your PR.

PR #2642, opened 2019-06-26
Update Monitor.xml

ericmutta:
Move bracketed comment `(to be moved, the thread must be at the head of the waiting queue)` towards the end of the sentence so it doesn't interrupt the reading flow when the sentence refers to the `Pulse` and `PulseAll` methods. I have also clarified that the thread in question is the one that called the `Wait` method, which helps explain why it would be in the waiting queue and would later be moved to the ready queue when `Pulse` or `PulseAll` is called.


ericmutta:
Is anyone monitoring these PRs? This one has been idle for 13 days now!

ericmutta:
Pinging @mairaw

ericmutta:
@rpetrusha many thanks for following up!

PR #2657, opened 2019-06-28
Remove reference to InvalidOperationException in Uri.OriginalString

watfordgnf:
## Summary
Uri.OriginalString does not throw InvalidOperationException if a relative URI is used. This is true for .NET Core 3.0+ (at least) and the .NET Framework 4.8 (via ReferenceSource).

Sources: https://github.com/dotnet/corefx/blob/e2b4dc93c9f9482b67f962106937d4fdd46f5673/src/System.Private.Uri/src/System/Uri.cs#L1145-L1153
https://referencesource.microsoft.com/#System/net/System/URI.cs,1234

This caused some confusion developing Uri support in System.Text.Json https://github.com/dotnet/corefx/pull/39015

### DIFF
--- a/xml/System/Uri.xml
+++ b/xml/System/Uri.xml
@@ -3560,7 +3560,6 @@ If you used an escaped string to construct this instance (for example, "http://[
   
  ]]></format>
         </remarks>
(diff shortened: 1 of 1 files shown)


BillWagner:
Thanks for making this update @watfordgnf 

@karelz Can you verify this for accuracy? It looks good to me, but I want someone on the product team to verify.

PR #2658, opened 2019-06-28
Add missing exception to Uri.IdnHost

watfordgnf:
## Summary
`Uri.IdnHost` throws InvalidOperationException if the URI is not absolute. This is because it internally forwards to `DnsSafeHost` which throws. This is true for .NET Core 3+ (at least) and .NET Framework 4.8 (Reference Source).

Sources:
https://github.com/dotnet/corefx/blob/master/src/System.Private.Uri/src/System/Uri.cs#L1219-L1233
https://referencesource.microsoft.com/#System/net/System/URI.cs,988f6bd791dbb0dc,references

### DIFF
--- a/xml/System/Uri.xml
+++ b/xml/System/Uri.xml
@@ -2457,6 +2457,7 @@ If you used an escaped string to construct this instance (for example, "http://[
 
  ]]></format>
         </remarks>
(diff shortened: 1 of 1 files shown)


BillWagner:
This PR is related.  @karelz Can you take a look?

PR #2712, opened 2019-07-08
coreclr: Automatic port of triple slash from System.Runtime.InteropServices

carlossanlop:
Found some comments that could be automatically ported from dotnet/coreclr.
Adding area owners @AaronRobinsonMSFT, @jkoritzinsky to ensure the descriptions are correct. After this change, please proceed to manually update the rest of the APIs that do not have any triple slash comments that could be automatically ported.

### DIFF
--- a/xml/System.Runtime.InteropServices/NativeLibrary.xml
+++ b/xml/System.Runtime.InteropServices/NativeLibrary.xml
@@ -14,7 +14,7 @@
   </Base>
   <Interfaces />
   <Docs>
-    <summary>Provides APIs for managing Native Libraries.</summary>
+    <summary>Provides APIs for managing native libraries.</summary>
     <remarks>To be added.</remarks>
   </Docs>
   <Members>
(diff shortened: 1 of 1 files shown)


xml/System.Runtime.InteropServices/NativeLibrary.xml
```diff
@@ -37,8 +37,8 @@
         <Parameter Name="handle" Type="System.IntPtr" />
       </Parameters>
       <Docs>
-        <param name="handle">To be added.</param>
-        <summary>To be added.</summary>
+        <param name="handle">The native library handle to be freed.</param>
+        <summary>Frees a given a loaded library handle. No action if the input handle is <see langword="null" />.</summary>
```
AaronRobinsonMSFT: ~~No action if the input handle is null~~ -> No action if the input handle is `IntPtr.Zero`

This should not be `null` since `IntPtr` is a value type.

xml/System.Runtime.InteropServices/NativeLibrary.xml
```diff
@@ -168,12 +186,13 @@
...
       <Docs>
-        <param name="handle">To be added.</param>
-        <param name="name">To be added.</param>
-        <param name="address">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="handle">The  native library handle.</param>
+        <param name="name">The name of the exported symbol.</param>
+        <param name="address">The out-parameter for the symbol address, if it exists.</param>
+        <summary>Get the address of an exported Symbol, but do not throw.</summary>
```
AaronRobinsonMSFT: ~~Symbol~~ -> symbol

PR #2715, opened 2019-07-08
coreclr: Automatic port of triple slash from System.Runtime.CompilerServices.C*

carlossanlop:
Found some more triple slash comments that could be automatically ported from dotnet/coreclr
Adding area owner @Anipik to double check that the descriptions are correct. After this PR is checked-in, please proceed to manually document all the APIs that have missing documentation.


xml/System.Runtime.CompilerServices/ConfiguredValueTaskAwaitable`1.xml
```diff
@@ -33,7 +33,7 @@
   </Attributes>
   <Docs>
     <typeparam name="TResult">To be added.</typeparam>
```
Anipik: To be added -> The type of the result produced

carlossanlop: Thanks! If you press Control+G, you can add the code suggestion and I can merge it. Like this:
```suggestion
    <typeparam name="TResult">The type of the produced result.</typeparam>
```

PR #2718, opened 2019-07-08
Update Process.StartInfo to show exception thrown if Process.Start was not used

danmoseley:
## Summary
In .NET Core, https://github.com/dotnet/corefx/pull/1167/commits/16bda12d198864091029ff9ad5199e3de78d5768 changed Process.StartInfo to throw InvalidOperationException if Process.Start was not used to create the Process object. In .NET Framework, it simply returns a dummy ProcessStartInfo object. Clarify this in the text.


danmosemsft:
cc @stephentoub

danmosemsft:
cc @davster

xml/System.Diagnostics/Process.xml
```diff
@@ -5218,7 +5218,7 @@ There is a similar issue when you read all text from both the standard output an
...
+ If you did not use the <xref:System.Diagnostics.Process.Start%2A> method to start a process, the <xref:System.Diagnostics.Process.StartInfo%2A> property does not reflect the parameters used to start the process and should not be used. For example, if you use <xref:System.Diagnostics.Process.GetProcesses%2A> to get an array of processes running on the computer, the <xref:System.Diagnostics.Process.StartInfo%2A> property of each <xref:System.Diagnostics.Process> does not contain the original file name or arguments used to start the process. On .NET Core, if you did not use the <xref:System.Diagnostics.Process.Start%2A> method to start a process then retrieving the <xref:System.Diagnostics.Process.StartInfo%2A> property will throw an <xref:System.InvalidOperationException>.
```
stephentoub: It's a little more subtle than this.  The key thing that matters is whether the `Process` object itself was `Start`'d.  I say it's subtle because you could have used Process.Start to start the process and then gotten another Process instance for that same process via e.g. Process.GetProcessById, and you wouldn't be able to access StartInfo on the latter.

danmosemsft: How about now

PR #2726, opened 2019-07-08
coreclr: Automatic port of triple slash from System.Numerics

carlossanlop:
Found some more triple slash comments that could be automatically ported from dotnet/coreclr
Adding area owners @tannergooding, @ViktorHofer to double check that the descriptions are correct. After this PR is checked-in, please proceed to manually document all the APIs that have missing documentation.

### DIFF
--- a/xml/System.Numerics/Vector`1.xml
+++ b/xml/System.Numerics/Vector`1.xml
@@ -90,8 +90,8 @@
         <Parameter Name="values" Type="System.ReadOnlySpan&lt;System.Byte&gt;" Index="0" FrameworkAlternate="netcore-3.0" />
       </Parameters>
       <Docs>
-        <param name="values">The values to add to the vector, as a read-only span of bytes.</param>
-        <summary>Constructs a vector from the given <see cref="T:System.ReadOnlySpan`1" />. The span must contain at least <see cref="P:System.Numerics.Vector`1.Count" /> elements.</summary>
+        <param name="values">A read-only span of bytes that contains the values to add to the vector. The span must contain at least <see cref="P:System.Numerics.Vector`1.Count" /> elements.</param>
+        <summary>Constructs a vector from the given read-only span of bytes.</summary>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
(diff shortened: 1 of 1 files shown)


xml/System.Numerics/BitOperations.xml
```diff
@@ -42,10 +42,20 @@
         <Parameter Name="value" Type="System.UInt32" />
       </Parameters>
       <Docs>
-        <param name="value">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <param name="value">The mask.</param>
+        <summary>Counts the number of leading zero bits in an unsigned 32-bit integer mask.</summary>
```
tannergooding: The order of these is off. Summary is normally first...

xml/System.Numerics/BitOperations.xml
```diff
@@ -14,7 +14,7 @@
   </Base>
   <Interfaces />
   <Docs>
-    <summary>To be added.</summary>
+    <summary>Utility methods for intrinsic bit-twiddling operations. The methods use hardware intrinsics when available on the underlying platform, otherwise they use optimized software fallbacks.</summary>
     <remarks>To be added.</remarks>
```
tannergooding: Is this staying `To be added.` going to be a problem?

xml/System.Numerics/BitOperations.xml
```diff
@@ -42,10 +42,20 @@
...
+        <param name="value">The mask.</param>
+        <summary>Counts the number of leading zero bits in an unsigned 32-bit integer mask.</summary>
+        <returns>The number of leading zero bits in a mask.</returns>
+        <remarks>
+          <format type="text/markdown">
+            <![CDATA[
+
+### Remarks
+
+Similar in behavior to the x86 instruction LZCNT.
```
tannergooding: It would be nice if this documented the behavior of the edge case handling and indicated it used `LZCNT` on x86 instead...

xml/System.Numerics/BitOperations.xml
carlossanlop: Not in dotnet-api-docs xml files. The format and order are different than the ones you see in xmldocs in corefx/coreclr.
The 'before' lines show you that param is first, and that's how the skeleton was automatically generated for the API by the Docs repo.

xml/System.Numerics/BitOperations.xml
carlossanlop: @tannergooding since you're the area owner, would you mind suggesting the correct text? Press Control + G to add a code suggestion for this line.

xml/System.Numerics/BitOperations.xml
carlossanlop: Remarks are optional. We only use them when it's necessary to explain something additional that doesn't fit in a summary, returns, param or typeparam field.
"To be added" is the default string for what dotnet-api-docs considers empty or undocumented. But in the case of remarks, since they are optional, it won't be a problem.

rpetrusha: "To be added." is automatically ignored at documentation build time.

xml/System.Numerics/BitOperations.xml
tannergooding:
```suggestion
Returns 32 when <paramref name="value" /> is 0.

This API utilizes System.Runtime.Intrinsics.X86.Lzcnt.LeadingZeroCount when System.Runtime.Intrinsics.X86.Lzcnt.IsSupported returns true.
```

carlossanlop:
It's ready to merge, @rpetrusha. Tanner addressed the comment and you already added the commit :)

PR #2359, opened 2019-04-24
Port System.Runtime.Intrinsics source comments to Docs

carlossanlop:
## Summary

Automatically ported triple slash source code comments found in System.Runtime.Intrinsics (coreclr) that were not yet documented in dotnet-api-docs.

Fixes #Issue_Number (if available)

PR #2737, opened 2019-07-09
Automatically port System.Text.Json triple slash comments

carlossanlop:
I synced my dotnet-api-docs clone to the latest bits to ensure I had the latest xml API structure from the most recent reference assembly update. I also synced my corefx repo to the latest bits and built them to ensure I had the latest comments.

Area owners @ahsonkhan @steveharter please help double check that the descriptions are all accurate.

### DIFF
--- a/xml/System.Text.Json.Serialization/JsonExtensionDataAttribute.xml
+++ b/xml/System.Text.Json.Serialization/JsonExtensionDataAttribute.xml
@@ -28,7 +28,7 @@
 
 The dictionary's `TKey` value must be <xref:System.String>, and `TValue` must be <xref:System.Text.Json.JsonElement> or <xref:System.Object>.
 
-During deserialization, when using <xref:System.Object>, a "null" JSON value is treated as a `null` object reference, and when using <xref:System.Text.Json.JsonElement>, a "null" is treated as a JsonElement with <xref:System.Text.Json.JsonElement.Type> set to <xref:System.Text.Json.JsonValueKind.Null?displayProperty=nameWithType>.
+During deserialization, when using <xref:System.Object>, a "null" JSON value is treated as a `null` object reference, and when using <xref:System.Text.Json.JsonElement>, a "null" is treated as a JsonElement with <xref:System.Text.Json.JsonElement.ValueKind> set to <xref:System.Text.Json.JsonValueKind.Null?displayProperty=nameWithType>.
             
 During serialization, the name of the extension data property is not included in the JSON; the data contained within the extension data is serialized as properties of the JSON object.
             


carlossanlop:
@rpetrusha do you mind checking this PR next?

carlossanlop:
@ahsonkhan @steveharter please help review these ported comments.

rpetrusha:
I'm checking it now, @carlossanlop.

rpetrusha:
I've finally finished reviewing this PR, @carlossanlop. I'll approve it, but please let me know if it's OK before I merge.

carlossanlop:
@ahsonkhan @steveharter if you want to take a quick look, that would be great too.
I took a look at your latest commits, @rpetrusha. I didn't find any issues. Let's wait for the build to finish.

ahsonkhan:
> if you want to take a quick look, that would be great too.

Yep, will do. It will take some time though, so give me ~1 hour. Tyvm.

xml/System.Text.Json.Serialization/JsonAttribute.xml
```diff
@@ -31,9 +31,9 @@
       </AssemblyInfo>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Creates a new instance of the <see cref="T:System.Text.Json.Serialization.JsonAttribute" /> instance.</summary>
```
ahsonkhan: Do we generally repeat `instance` twice like this?

`Create an instance of the <FOO> instance.`

xml/System.Text.Json.Serialization/JsonExtensionDataAttribute.xml
```diff
@@ -19,8 +19,24 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>To be added.</summary>
-    <remarks>To be added.</remarks>
+    <summary>When placed on a property of type <see cref="T:System.Collections.Generic.IDictionary`2" />, provides that any properties that do not have a matching member are added to that dictionary during deserialization and written during serialization.</summary>
```
ahsonkhan: Should we specify the generic parameters are TKey, TValue (since we refer to them in the remarks), or is it will understood that IDictionary'2 is `IDictionary<TKey,TValue>`?

https://github.com/dotnet/corefx/blob/8e2f9a765f6533d17e711365ce2a3064b8d1dfc6/src/System.Text.Json/src/System/Text/Json/Serialization/JsonExtensionDataAttribute.cs#L8

ahsonkhan: This sentence doesn't make sense. `... provides that any properties that ...` ???

```suggestion
    <summary>When placed on a property of type <see cref="T:System.Collections.Generic.IDictionary`2" />, any properties that do not have a matching member are added to that dictionary during deserialization and written during serialization.</summary>
```

xml/System.Text.Json.Serialization/JsonExtensionDataAttribute.xml
```diff
@@ -19,8 +19,24 @@
...
+
+## Remarks
+
+The dictionary's `TKey` value must be <xref:System.String>, and `TValue` must be <xref:System.Text.Json.JsonElement> or <xref:System.Object>.
+
+During deserialization, when using <xref:System.Object>, a "null" JSON value is treated as a `null` object reference, and when using <xref:System.Text.Json.JsonElement>, a "null" is treated as a JsonElement with <xref:System.Text.Json.JsonElement.Type> set to <xref:System.Text.Json.JsonValueType.Null?displayProperty=nameWithType>.
+            
+During serialization, the name of the extension data property is not included in the JSON; the data contained within the extension data is serialized as properties of the JSON object.
+            
+If there is more than one extension property on a type, or it the property is not of the correct type, an <xref:System.InvalidOperationException> is thrown during the first serialization or deserialization of that type.
```
ahsonkhan: it -> if
```suggestion
If there is more than one extension property on a type, or if the property is not of the correct type, an <xref:System.InvalidOperationException> is thrown during the first serialization or deserialization of that type.
```

ahsonkhan: I think it would be more clear to say if more than one property has this attribute on it.

```suggestion
If there is more than one property on a type with this extension data attribute, or if the property itself is not of the correct <see cref="T:System.Collections.Generic.IDictionary`2" /> type, an <xref:System.InvalidOperationException> is thrown during the first serialization or deserialization of that type.
```

xml/System.Text.Json/JsonElement.xml
```diff
@@ -170,9 +170,22 @@ This method does not parse the contents of a JSON string value.
...
-        <remarks>To be added.</remarks>
+        <summary>Gets the value of the element as a byte array.</summary>
+        <returns>The value decoded as a byte array.</returns>
+        <remarks>
+          <format type="text/markdown">
+            <![CDATA[
+
+## Remarks
+
+This method does not create a byte representation of values other than base 64-encoded JSON strings.
```
ahsonkhan:
```suggestion
This method does not create a byte[] representation of values other than base 64-encoded JSON strings.
```

xml/System.Text.Json/JsonElement.xml
```diff
@@ -741,10 +752,24 @@ For other types, the value of <xref:System.Text.Json.JsonElement.GetRawText> is
         <Parameter Name="value" Type="System.Byte[]" RefType="out" />
       </Parameters>
       <Docs>
-        <param name="value">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <param name="value">When the method returns, contains the byte representation of <paramref name="value" /> if the method succeeded.</param>
```
ahsonkhan: byte[] or byte array

```suggestion
        <param name="value">When the method returns, contains the byte[] representation of <paramref name="value" /> if the method succeeded.</param>
```

ahsonkhan: Also, this comment doesn't make sense. value doesn't contain the byte[] representation of value, it contains the byte[] representation of the json element.

```suggestion
        <param name="value">When the method returns, contains the byte[] representation of the current JSON string if the method succeeded.</param>
```

ahsonkhan: Also, we may want the Try and the non-Try method to have similar looking documentation.

ahsonkhan: cc @bartonjs

xml/System.Text.Json/JsonEncodedText.xml
```diff
@@ -46,10 +56,11 @@
         <Parameter Name="utf8Value" Type="System.ReadOnlySpan&lt;System.Byte&gt;" />
       </Parameters>
       <Docs>
-        <param name="utf8Value">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="utf8Value">The UTF-8 encoded value to be transformed as JSON encoded text.</param>
```
ahsonkhan: Note: The APIs changed recently to add an extra optional parameter: 
https://github.com/dotnet/corefx/blob/ba2d62a24ea8574f40c43cd27cc384f079958a0c/src/System.Text.Json/src/System/Text/Json/JsonEncodedText.cs#L95-L103

@carlossanlop, would your tool be able to detect that and update it in the next run (post-preview8) or will we have to update this manually?

xml/System.Text.Json/JsonReaderOptions.xml
```diff
@@ -84,10 +95,21 @@ By default, the reader throws a <xref:System.Text.Json.JsonReaderException> if i
         <ReturnType>System.Int32</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets or sets the maximum depth allowed when reading JSON, with the default (i.e. 0) indicating a max depth of 64. Reading past this depth will throw a <exception cref="T:System.Text.Json.JsonException" />.</summary>
-        <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <summary>Gets or sets the maximum depth allowed when reading JSON, with the default (i.e. 0) indicating a maximum depth of 64. </summary>
```
ahsonkhan: I learned recently that we shouldn't use i.e., from @mairaw https://github.com/dotnet/corefx/pull/39092#discussion_r304071760

xml/System.Text.Json/JsonTokenType.xml
```diff
@@ -34,7 +34,7 @@
       </ReturnValue>
       <MemberValue>11</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The token type is the comment string.</summary>
```
ahsonkhan: We use `a` for the other enum values. Seems odd to use `the` here.

```suggestion
        <summary>The token type is a comment string.</summary>
```

xml/System.Text.Json/JsonTokenType.xml
```diff
@@ -94,7 +94,7 @@
       </ReturnValue>
       <MemberValue>9</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The token type is the JSON literal <see langword="false" />.</summary>
```
ahsonkhan: This should **not** be the .NET language keyword, false. This refers to JSON literal "false" which is subtly different.
The same applies to true/null.

Can we use some annotation to convey that? Maybe `<c></c>`? `langword` is not technically accurate.

@bartonjs, any suggestions?

xml/System.Text.Json/JsonTokenType.xml
```diff
@@ -134,7 +145,7 @@
       </ReturnValue>
       <MemberValue>10</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The token type is the JSON literal <see langword="null" />.</summary>
```
ahsonkhan: Same issue here. This isn't the .NET `null`.

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -235,9 +240,19 @@ in more data asynchronously before continuing with a new instance of the <xref:S
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <summary>Parses the current JSON token value from the source and decodes the base 64 encoded JSON string as a byte array.</summary>
+        <returns>The byte array that represents the current JSON token value.</returns>
         <remarks>To be added.</remarks>
+        <exception cref="T:System.InvalidOperationException">The type of the JSON token is not a <see cref="F:System.Text.Json.JsonTokenType.String" />.</exception>
+        <exception cref="T:System.FormatException">The JSON string contains data outside of the expected base 64 range.
```
ahsonkhan: Should we also add the similar level of detail to the `JsonElement.GetBytesFromBase64` API?

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -257,9 +272,10 @@ in more data asynchronously before continuing with a new instance of the <xref:S
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <summary>Parses the current JSON token value from the source as a comment, transcoded it as a <see cref="T:System.String" />.</summary>
+        <returns>The comment that represnets the current JSON token value.</returns>
```
ahsonkhan:
```suggestion
        <returns>The comment that represents the current JSON token value.</returns>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -257,9 +272,10 @@ in more data asynchronously before continuing with a new instance of the <xref:S
...
         <remarks>To be added.</remarks>
+        <exception cref="T:System.InvalidOperationException">The JSON token that is not a comment.</exception>
```
ahsonkhan: Remove `that`.

```suggestion
        <exception cref="T:System.InvalidOperationException">The JSON token is not a comment.</exception>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -651,8 +667,8 @@ For input data within a `ReadOnlySpan<byte>`, this always returns `false`. For i
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Gets the mode of this instance of the <see cref="T:System.Text.Json.Utf8JsonReader" />.</summary>
+        <value><see langword="true" /> if the reader was constructed with the input span containing the entire data to process. <see langword="false" /> if the reader was constructed with an input span that may contain partial data with more data to follow.</value>
```
ahsonkhan: Re-reading this, using the term `span` is incomplete because the reader can be given JSON data contained within a `ReadOnlySequence<byte>` as well.

Maybe:
```suggestion
        <value><see langword="true" /> if the reader was constructed with the input span or sequence containing the entire JSON data to process. <see langword="false" /> if the reader was constructed with an input span or sequence that may contain partial JSON data with more data to follow.</value>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -651,8 +667,8 @@ For input data within a `ReadOnlySpan<byte>`, this always returns `false`. For i
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Gets the mode of this instance of the <see cref="T:System.Text.Json.Utf8JsonReader" />.</summary>
```
ahsonkhan: What will this look like in Intellisense? Will the value tag be visible? With just reading the summary, it isn't clear what "mode" means.

If only the summary tag is visible, we may want to add more details:
```suggestion
        <summary>Gets the mode of this instance of the <see cref="T:System.Text.Json.Utf8JsonReader" /> which indicates whether all the JSON data was provided or there is more data to come.</summary>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -808,9 +844,21 @@ If the look up text is invalid or incomplete UTF-16 text (such as unpaired surro
         <ReturnType>System.Int64</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <summary>Gets the index that the last processed JSON token within the given UTF-8 encoded input text starts at, skipping any white space.</summary>
```
ahsonkhan: The thing being referred to by `starts at` is too far away, making it harder to understand.

How about:
```suggestion
        <summary>Gets the index that the last processed JSON token starts at (within the given UTF-8 encoded input text), skipping any white space.</summary>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -853,10 +901,11 @@ If the look up text is invalid or incomplete UTF-16 text (such as unpaired surro
         <Parameter Name="value" Type="System.Byte[]" RefType="out" />
       </Parameters>
       <Docs>
-        <param name="value">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="value">When the method returns, a byte array that contains the base 64-encoded JSON string if the method succeeds.</param>
```
ahsonkhan: This might be more clear
```suggestion
        <param name="value">If the method succeeds, a byte array that contains the base 64-encoded JSON string.</param>
```

xml/System.Text.Json.Serialization/JsonAttribute.xml
ahsonkhan: Maybe:

```suggestion
        <summary>Creates a new instance of the <see cref="T:System.Text.Json.Serialization.JsonAttribute" />.</summary>
```

Also, JsonAttribute is an abstract class. What ctor?

https://github.com/dotnet/corefx/blob/ba2d62a24ea8574f40c43cd27cc384f079958a0c/src/System.Text.Json/src/System/Text/Json/Serialization/JsonAttribute.cs#L10

I guess the auto-generated protected ctor. Is that the general pattern for abstract classes?

https://github.com/dotnet/corefx/blob/ba2d62a24ea8574f40c43cd27cc384f079958a0c/src/System.Text.Json/ref/System.Text.Json.cs#L453-L456

carlossanlop: This is a very good question. Why did a protected constructor get added by mdoc to the xml structure? @mairaw is this a bug? I don't think we should be documenting protected elements.

bartonjs: >  I don't think we should be documenting protected elements.

Sure we should.  If you override the type you need to know what the protected members do...

xml/System.Text.Json.Serialization/JsonExtensionDataAttribute.xml
carlossanlop: This question has arised in [other PRs](https://github.com/dotnet/dotnet-api-docs/pull/2359/files) that clearly specify `{something}` at the end of the DocId. It seems that triple slash comments with such elements do not get properly converted to the xml files in artifacts/bin. So when I automatically port them, they look like in this comment you mentioned.

I couldn't find an easy fix, so I opted for something that looks like:

> `An <see cref="T:System.Collections.Generic.IDictionary`2" /> of <typeparamref name="TKey" /> and <typeparamref name="TValue" />`

But I haven't heard back from the Docs team for confirmation.

There may be another solution: to parse the html entities for the curly brackets. I don't know if that will build successfully in the Docs repo.

carlossanlop: More info: I learned about this limitation in triple slash comments from [this StackOverflow question](https://stackoverflow.com/questions/532166/how-to-reference-generic-classes-and-methods-in-xml-documentation/41208166#41208166).

xml/System.Text.Json.Serialization/JsonAttribute.xml
carlossanlop: I'll commit it but leave the conversation unresolved.

xml/System.Text.Json.Serialization/JsonExtensionDataAttribute.xml
carlossanlop: Ok I'll commit this second suggestion.

xml/System.Text.Json/JsonElement.xml
carlossanlop: I slightly modified your suggestion: Needs to be wrapped by back tick characters.

xml/System.Text.Json/JsonElement.xml
carlossanlop: I think in this case, using natural language sounds a bit better: byte array. I'll modify your suggestion and commit it.

carlossanlop: I used "byte array representation"

xml/System.Text.Json/JsonEncodedText.xml
carlossanlop: No. The tool only imports new comments. Any modified comments need to be addressed by area owners directly in the Docs repo for APIs that were already previously commented.

xml/System.Text.Json/JsonTokenType.xml
carlossanlop: I am not sure if we allow `<c>` in Docs files. It's a good question for @mairaw and @rpetrusha.

We can simply convert this to:

```suggestion
        <summary>The token type is the JSON literal "false".</summary>
```

xml/System.Text.Json/JsonTokenType.xml
carlossanlop: Let's use "null" then:

```suggestion
        <summary>The token type is the JSON literal "null".</summary>
```

xml/System.Text.Json/Utf8JsonReader.xml
carlossanlop: `<value>` is not shown in IntelliSense, I just tested it. I think this is only for the MS Docs website, just like remarks.

carlossanlop: We still need to fill it, otherwise the API will show up as undocumented.

xml/System.Text.Json/Utf8JsonReader.xml
carlossanlop: I'm not sure what you mean. Can you please elaborate? Or was the question directed at the other Json owners?

xml/System.Text.Json/JsonTokenType.xml
```diff
@@ -94,7 +94,7 @@
       </ReturnValue>
       <MemberValue>9</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The token type is the JSON literal "false".</summary>
```
bartonjs: This is tricky... it isn't the literal `"false"` (which is a string), it's the literal `false` (which is a false, since JSON doesn't really have a boolean type per se).  So this ideally should be unquoted and in a different display mode... but I'll go with whatever Ron / Maira think is best.

xml/System.Text.Json.Serialization/JsonAttribute.xml
bartonjs: https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Constructor says the formula for a constructor in an abstract class is `Called from constructors in derived classes to initialize the <Class> class.`

xml/System.Text.Json.Serialization/JsonExtensionDataAttribute.xml
rpetrusha: I'm not quite sure that I understand the issue. At build time, Dictionary`2 renders as:
image

xml/System.Text.Json/JsonTokenType.xml
rpetrusha: Perhaps remove the quotation marks and boldface it (**false**), @bartonjs?

xml/System.Text.Json.Serialization/JsonExtensionDataAttribute.xml
ahsonkhan: In that case, all's good. Thanks @rpetrusha.

@carlossanlop - in our case here, we are not passing a custom (or specific TKey, TValue) type in the comments, so the issue you mentioned isn't relevant here.

If, for example, if the xml comment specified something like `IDictionary{string,object}`, we would need to do something different to  display that.

xml/System.Text.Json/JsonTokenType.xml
bartonjs: > Perhaps remove the quotation marks and boldface it (false), @bartonjs?

Seems fine to me, not that I know what syntax that would be in a summary tag :smile:.

xml/System.Text.Json/JsonElement.xml
ahsonkhan: That's fine, but you missed the other feedback which was change `<paramref name="value" />` to `the current JSON string`

xml/System.Text.Json/JsonElement.xml
```diff
@@ -741,10 +752,24 @@ For other types, the value of <xref:System.Text.Json.JsonElement.GetRawText> is
...
+        <returns>
+          <see langword="true" /> if the entire token value is encoded as valid base 64 text and can be successfully decoded to bytes.
+              <see langword="false" /> otherwise.</returns>
+        <remarks>
+          <format type="text/markdown">
+            <![CDATA[
+
+## Remarks
+
+This method does not create a byte array representation of values other than bsae 64-encoded JSON strings.
```
bartonjs:
```suggestion
This method does not create a byte array representation of values other than base 64-encoded JSON strings.
```

xml/System.Text.Json/JsonReaderOptions.xml
ahsonkhan: Change to `that is,`, like what @safern did here: https://github.com/dotnet/corefx/pull/39092/commits/f20cfb8e1ab78790426fda2d7efbee763d3398c1

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: The `JsonElement.GetBytesFromBase64` exception documentation doesn't show all the cases where it could fail (i.e. where the data would be invalid and cannot be decoded as Base64 string). The `Utf8JsonReader.GetBytesFromBase64` provides more details.

I am asking if we should add the same info to  `JsonElement.GetBytesFromBase64`, which just says:
`The value is not encoded as base 64 text and hence cannot be decoded to bytes.`
https://github.com/dotnet/dotnet-api-docs/blob/771baa41a5149ce706b3c99d6cd698683a5616fe/xml/System.Text.Json/JsonElement.xml#L187

I say, yes.
@bartonjs?

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -901,7 +901,7 @@ For comments, this value points to before the first comment delimiter (that is,
         <Parameter Name="value" Type="System.Byte[]" RefType="out" />
       </Parameters>
       <Docs>
-        <param name="value">When the method returns, a byte array that contains the base 64-encoded JSON string if the method succeeds.</param>
+        <param name="value">If the method succeeds, a byte array that contains the base 64-encoded JSON string.</param>
```
ahsonkhan: Can we make the same change to the other doc comment, for `JsonElement.TryGetBytesFromBase64`? 
https://github.com/dotnet/dotnet-api-docs/pull/2737/files#diff-5b79fe7599d21bf8a1436d585fe965eaR755

xml/System.Text.Json/JsonElement.xml
ahsonkhan:
```suggestion
        <param name="value">If the method succeeds, contains the decoded binary representation of the base 64 text.</param>
```

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: Maybe:
```suggestion
        <param name="value">If the method succeeds, contains the decoded binary representation of the base 64 text.</param>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -853,10 +901,11 @@ If the look up text is invalid or incomplete UTF-16 text (such as unpaired surro
         <Parameter Name="value" Type="System.Byte[]" RefType="out" />
       </Parameters>
       <Docs>
-        <param name="value">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="value">If the method succeeds, a byte array that contains the base 64-encoded JSON string.</param>
+        <summary>Tries to parse the current JSON token value from the source and decodes the base 64 encoded JSON string as a byte array.</summary>
+        <returns><see langword="true" /> if the entire token value is encoded as valid base 64 text and can be successfully decoded; <see langword="false" /> otherwise.</returns>
```
ahsonkhan: To be consistent with the JsonElement API doc:
```suggestion
        <returns><see langword="true" /> if the entire token value is encoded as valid base 64 text and can be successfully decoded to bytes; <see langword="false" /> otherwise.</returns>
```

xml/System.Text.Json/Utf8JsonReader.xml
bartonjs: I think Utf8JsonReader is overly descriptive.  They all just mean "The JSON string is not a valid base 64-encoded payload." which is what I would use there, and here.

ahsonkhan: The info about what is considered "invalid" is useful, imo, but I can see it not being relevant to JSON (since its primarily about Base64). I would prefer to keep this detail as I don't see a down-side for having it, but don't feel too strongly (since the Base64 API already sufficiently document what is considered "invalid" - https://docs.microsoft.com/en-us/dotnet/api/system.buffers.text.base64.decodefromutf8?view=netstandard-2.1#remarks).

xml/System.Text.Json/Utf8JsonReader.xml
carlossanlop: Sure, I'll add it manually.

xml/System.Text.Json/JsonTokenType.xml
carlossanlop: @rpetrusha can we use html tags?

> `<b>false</b>`

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -663,10 +719,21 @@ The property name is escaped before writing.
         <Parameter Name="value" Type="System.Boolean" />
       </Parameters>
       <Docs>
-        <param name="propertyName">To be added.</param>
-        <param name="value">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="propertyName">The JSON encoded property name of the JSON object to be transcoded and written as UTF-8.</param>
+        <param name="value">The value to be written as a JSON literal "true" or "false" as part of the name/value pair.</param>
```
carlossanlop:
```suggestion
        <param name="value">The value to be written as a JSON literal <b>true</b> or <b>false</b> as part of the name/value pair.</param>
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -663,10 +719,21 @@ The property name is escaped before writing.
...
+        <summary>Writes the pre-encoded property name and <see cref="T:System.Boolean" /> value (as a JSON literal "true" or "false") as part of a name/value pair of a JSON object.</summary>
```
carlossanlop:
```suggestion
        <summary>Writes the pre-encoded property name and <see cref="T:System.Boolean" /> value (as a JSON literal <b>true</b> or <b>false</b>) as part of a name/value pair of a JSON object.</summary>
```

xml/System.Text.Json/JsonTokenType.xml
rpetrusha: Yes, we can use HTML tags.

xml/System.Text.Json.Serialization/JsonExtensionDataAttribute.xml
carlossanlop: Resolving since I misunderstood the problem and it all looks good.

xml/System.Text.Json/Utf8JsonReader.xml
carlossanlop: Should we do something about this or not? Can you provide a code suggestion if yes? @bartonjs @ahsonkhan

carlossanlop: Chatted with Ahson and decided to substitute the long exception message with the shorter one. Fixed.

carlossanlop:
@mairaw @rpetrusha I addressed all pending comments in this review. If the build passes, can we get it merged?

mairaw:
Can you please resolve the merge conflicts here @carlossanlop? Given that Ron has approved earlier, I think it can be merged. Unless you guys wants to re-review given some changes happened since his approval?

rpetrusha:
I've resolved the merge conflict, but I can't account for one of the conflicts. Did you merge your CI update, @mairaw?

carlossanlop:
@rpetrusha the conflicts are weird. Half of the changes were correct on the right side, the other half were correct on the left side.
Example: Did you find out that the function `GetByte` in JsonElement was deleted in main? It still [exists in code](https://github.com/dotnet/corefx/blob/722ab9b7303ef14db74a1d09d464d180f18f512b/src/System.Text.Json/src/System/Text/Json/Document/JsonElement.cs#L494).

rpetrusha:
I didn't see a conflict with GetByte -- there were four conflicts, three of them trivial (an exception on one line versus multiple lines). The fourth was that two overloads of TextEqual were present in your branch but not master -- and source shows a single internal TextEqual method, which suggests a change because of a CI update. Unless I missed it, I didn't see a merge conflict with GetByte.

carlossanlop:
@rpetrusha Oh okay. It looked weird in Visual Studio. I may have done something wrong. Thanks for merging the conflicts.

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -895,72 +895,6 @@ For all other token types, the reader does not move. After the next call to <xre
 The current depth exceeds the recursive limit set by the maximum depth.</exception>
       </Docs>
     </Member>
-    <Member MemberName="TextEquals">
```
ahsonkhan: Since we are updating this to remove `TextEquals` (and losing the documentation that came with it), can we add the documentation for all the `ValueTextEquals` overloads? They are there in source. But I think that can be done separately in a subsequent PR. Presumably, your tool should pick it up, right @carlossanlop 

https://github.com/dotnet/corefx/blob/fc8bd8ec1c622dd27f96d05f1bf05909f3b3cde0/src/System.Text.Json/src/System/Text/Json/Reader/Utf8JsonReader.cs#L412-L432

xml/System.Text.Json/JsonTokenType.xml
```diff
@@ -94,7 +94,7 @@
       </ReturnValue>
       <MemberValue>9</MemberValue>
       <Docs>
-        <summary>The token type is the JSON literal "false".</summary>
+        <summary>The token type is the JSON literal <b>false</b>.</summary>
```
ahsonkhan: We should fix "null" and "true" as well to follow this change to "false" (`TokenType.True`, `TokenType.Null`).

xml/System.Text.Json/JsonTokenType.xml
```diff
@@ -134,7 +145,7 @@
       </ReturnValue>
       <MemberValue>11</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The token type is the JSON literal "null".</summary>
```
ahsonkhan:
```suggestion
        <summary>The token type is the JSON literal <b>null</b>.</summary>
```

xml/System.Text.Json/JsonTokenType.xml
```diff
@@ -254,8 +265,8 @@
       </ReturnValue>
       <MemberValue>9</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The token type is the JSON literal <see langword="true" />.</summary>
```
ahsonkhan:
```suggestion
        <summary>The token type is the JSON literal <see langword=<b>true</b> />.</summary>
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -958,9 +1025,20 @@ The property name is escaped before writing.
         <Parameter Name="propertyName" Type="System.Text.Json.JsonEncodedText" />
       </Parameters>
       <Docs>
-        <param name="propertyName">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="propertyName">The JSON encoded property name of the JSON object to be transcoded and written as UTF-8.</param>
+        <summary>Writes the pre-encoded property name and the JSON literal "null" as part of a name/value pair of a JSON object.</summary>
```
ahsonkhan:
```suggestion
        <summary>Writes the pre-encoded property name and the JSON literal <b>null</b> as part of a name/value pair of a JSON object.</summary>
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -958,9 +1025,20 @@ The property name is escaped before writing.
...
+
+The property name should already be escaped when the instance of <xref:System.Text.Json.JsonEncodedText> was created.
+
+            ]]>
+          </format>
+        </remarks>
+        <exception cref="T:System.InvalidOperationException">Validation is enabled, and this method would result in writing invalid JSON.</exception>
       </Docs>
     </Member>
     <Member MemberName="WriteNullValue">
```
ahsonkhan: If quotation doesn't work, fix the summary of WriteNullValue (and `WriteBooleanValue`) as well to be consistent:
```diff
- <summary>Writes the JSON literal "null" as an element of a JSON array.</summary>
+ <summary>Writes the JSON literal <b>null</b> as an element of a JSON array.</summary>

File: xml/System.Text.Json/Utf8JsonWriter.xml [hunk 30]
```diff
@@ -433,10 +429,22 @@ The <xref:System.Text.Json.Utf8JsonWriter> will continue to use the original wri
         <Parameter Name="bytes" Type="System.ReadOnlySpan&lt;System.Byte&gt;" />
       </Parameters>
       <Docs>
-        <param name="utf8PropertyName">To be added.</param>
-        <param name="bytes">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="utf8PropertyName">The UTF-8 encoded property name of the JSON object to be written.</param>
```
ahsonkhan: @safern recently fixed the wording of a lot of these APIs (in https://github.com/dotnet/corefx/pull/39092). Could you please separate the Utf8JsonWriter.xml change out to a separate PR and refresh it against latest? If doing it in this PR is easiest, I would be fine with that too.

For example, this was re-worded as:
```suggestion
        <param name="utf8PropertyName">The UTF-8 encoded name of the property to write.</param>
```

Otherwise, we are undoing the work from that PR and re-reviewing out-dated text.

ahsonkhan: However, looking now, we already have some feedback/fixes in this PR that we want to accept and re-apply to & merge with the new text:
https://github.com/dotnet/dotnet-api-docs/pull/2737/commits/eee88860199c0b7b31b3084df04845397c61b7d8
https://github.com/dotnet/dotnet-api-docs/pull/2737/commits/33f597f7eaf256c8fe7c6dda330a3e964c6e57ad
https://github.com/dotnet/dotnet-api-docs/pull/2737/commits/5ed4fd191fb97499b878eba7e9c8c5b482b8c5dc
https://github.com/dotnet/dotnet-api-docs/pull/2737/commits/b9d48be98362f274223aa8ebb7402033d4017b3d

@carlossanlop, what do you think is the easiest way to move forward here?

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -511,10 +544,21 @@ The <xref:System.Text.Json.Utf8JsonWriter> will continue to use the original wri
         <Parameter Name="bytes" Type="System.ReadOnlySpan&lt;System.Byte&gt;" />
       </Parameters>
       <Docs>
-        <param name="propertyName">To be added.</param>
-        <param name="bytes">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="propertyName">The JSON encoded property name of the JSON object to be transcoded and written as UTF-8.</param>
```
ahsonkhan: Same here/elsewhere:
```suggestion
        <param name="propertyName">The JSON-encoded name of the property to write.</param>
```

https://github.com/dotnet/corefx/blob/306fe9d60cc6db02a69d8ea6b88dde2b98fe8a7c/src/System.Text.Json/src/System/Text/Json/Writer/Utf8JsonWriter.WriteProperties.Bytes.cs#L16

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -719,14 +786,14 @@ The property name is escaped before writing.
...
-          
+
           ]]></format>
         </remarks>
         <exception cref="T:System.ArgumentException">The specified value is too large.
         
 -or-
 
-<paramref name="utf8Value" /> contains a comment delimiter (i.e. */).</exception>
+<paramref name="utf8Value" /> contains a comment delimiter (such as */).</exception>
```
ahsonkhan: I don't think this should be "such as". There is only one valid comment delimiter that we look for.

```suggestion
<paramref name="utf8Value" /> contains a comment delimiter (that is, */).</exception>
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -754,14 +821,14 @@ The comment value is not escaped before writing.
...
-          
+
           ]]></format>
         </remarks>
         <exception cref="T:System.ArgumentException">The specified value is too large OR.
         
 -or-
 
-<paramref name="value" /> contains a comment delimiter (i.e. */).</exception>
+<paramref name="value" /> contains a comment delimiter (such as */).</exception>
```
ahsonkhan: Same here, elsewhere.

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1045,7 +1123,7 @@ Writes the <xref:System.Decimal> using the default <xref:System.Buffers.Standard
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Double> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'). The property name is escaped before writing.
+Writes the <xref:System.Double> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'). The property name is escaped before writing.
```
ahsonkhan: `The property name is escaped before writing.` should be on its own line since its a separate remark.

```suggestion
Writes the <xref:System.Double> using the default <xref:System.Buffers.StandardFormat> (that is, 'G').

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1079,7 +1157,7 @@ Writes the <xref:System.Double> using the default <xref:System.Buffers.StandardF
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Int32> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'), for example: 32767. The property name is escaped before writing.
+Writes the <xref:System.Int32> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767. The property name is escaped before writing.
```
ahsonkhan: The same applies here/elsewhere.

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -3348,10 +3628,24 @@ The property name and value is escaped before writing.
...
+        <summary>Writes the pre-encoded property name and UTF-8 text value (as a JSON string) as part of a name/value pair of a JSON object.</summary>
+        <remarks>
+          <format type="text/markdown">
+            <![CDATA[
+
+## Remarks
+
+The property name should already be escaped when the instance of <xref:System.Text.Json.JsonEncodedText> was created.
+
+The value is escaped before writing.
```
ahsonkhan: This remark section looks much better (we should follow the same structure in other similar cases that I mentioned above).

Just to clarify, this is an issue/mistake with the comments coming from source (and nothing to do with tools or this port itself). They all weren't properly formatted with `<para>` for multiple remarks to show up nicely (though some were fix, like this one).

https://github.com/dotnet/corefx/blob/306fe9d60cc6db02a69d8ea6b88dde2b98fe8a7c/src/System.Text.Json/src/System/Text/Json/Writer/Utf8JsonWriter.WriteProperties.SignedNumber.cs#L51-L54

However, note that the first remark was removed in https://github.com/dotnet/corefx/pull/39092/files#diff-71ab5a4f550eede8ad13896c013094cfL501 and we should remove it here as well (this feedback could be resolved by refreshing the port, for `Utf8JsonWriter.xml` specifically).

xml/System.Text.Json/JsonTokenType.xml
carlossanlop: @ahsonkhan can you please elaborate? What would you like me to do here?

xml/System.Text.Json/JsonTokenType.xml
carlossanlop: Changed your suggestion to the correct format:

```suggestion
        <summary>The token type is the JSON literal <b>true</b>.</summary>
```

xml/System.Text.Json/Utf8JsonWriter.xml
carlossanlop: I will remove the file from this PR.

carlossanlop: But I will apply all the latest suggestions, save a copy, and I'll send it to you.

xml/System.Text.Json/Utf8JsonWriter.xml
carlossanlop:
```suggestion
<paramref name="value" /> contains a comment delimiter (that is */).</exception>
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -789,14 +856,14 @@ The comment value is not escaped before writing.
...
-          
+
           ]]></format>
         </remarks>
         <exception cref="T:System.ArgumentException">The specified value is too large.
         
 -or-
 
-<paramref name="value" /> contains a comment delimiter (i.e. */).</exception>
+<paramref name="value" /> contains a comment delimiter (such as */).</exception>
```
carlossanlop:
```suggestion
<paramref name="value" /> contains a comment delimiter (that is, */).</exception>
```

xml/System.Text.Json/Utf8JsonWriter.xml
carlossanlop:
```suggestion
Writes the <xref:System.Int32> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767.

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1011,7 +1089,7 @@ The property name is escaped before writing.
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Decimal> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'). The property name is escaped before writing.
+Writes the <xref:System.Decimal> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'). The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Decimal> using the default <xref:System.Buffers.StandardFormat> (that is, 'G').

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1113,7 +1191,7 @@ Writes the <xref:System.Int32> using the default <xref:System.Buffers.StandardFo
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Int64> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'), for example: 32767. The property name is escaped before writing.
+Writes the <xref:System.Int64> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767. The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Int64> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767.

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1147,7 +1225,7 @@ Writes the <xref:System.Int64> using the default <xref:System.Buffers.StandardFo
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Single> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'). The property name is escaped before writing.
+Writes the <xref:System.Single> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'). The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Single> using the default <xref:System.Buffers.StandardFormat> (that is, 'G').

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1186,7 +1264,7 @@ Writes the <xref:System.Single> using the default <xref:System.Buffers.StandardF
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.UInt32> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'), for example: 32767. The property name is escaped before writing.
+Writes the <xref:System.UInt32> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767. The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.UInt32> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767.

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1225,7 +1303,7 @@ Writes the <xref:System.UInt32> using the default <xref:System.Buffers.StandardF
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.UInt64> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'), for example: 32767. The property name is escaped before writing.
+Writes the <xref:System.UInt64> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767. The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.UInt64> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767.

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1259,7 +1337,7 @@ Writes the <xref:System.UInt64> using the default <xref:System.Buffers.StandardF
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Decimal> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'). The property name is escaped before writing.
+Writes the <xref:System.Decimal> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'). The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Decimal> using the default <xref:System.Buffers.StandardFormat> (that is, 'G').

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1293,7 +1371,7 @@ Writes the <xref:System.Decimal> using the default <xref:System.Buffers.Standard
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Double> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'). The property name is escaped before writing.
+Writes the <xref:System.Double> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'). The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Double> using the default <xref:System.Buffers.StandardFormat> (that is, 'G').

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1327,7 +1405,7 @@ Writes the <xref:System.Double> using the default <xref:System.Buffers.StandardF
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Int32> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'), for example: 32767. The property name is escaped before writing.
+Writes the <xref:System.Int32> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767. The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Int32> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767.

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1361,7 +1439,7 @@ Writes the <xref:System.Int32> using the default <xref:System.Buffers.StandardFo
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Int64> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'), for example: 32767. The property name is escaped before writing.
+Writes the <xref:System.Int64> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767. The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Int64> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767.

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1395,7 +1473,7 @@ Writes the <xref:System.Int64> using the default <xref:System.Buffers.StandardFo
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Single> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'). The property name is escaped before writing.
+Writes the <xref:System.Single> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'). The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Single> using the default <xref:System.Buffers.StandardFormat> (that is, 'G').

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1434,7 +1512,7 @@ Writes the <xref:System.Single> using the default <xref:System.Buffers.StandardF
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.UInt32> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'), for example: 32767. The property name is escaped before writing.
+Writes the <xref:System.UInt32> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767. The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.UInt32> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767.

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1473,7 +1551,7 @@ Writes the <xref:System.UInt32> using the default <xref:System.Buffers.StandardF
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.UInt64> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'), for example: 32767. The property name is escaped before writing.
+Writes the <xref:System.UInt64> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767. The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.UInt64> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767.

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1507,7 +1585,7 @@ Writes the <xref:System.UInt64> using the default <xref:System.Buffers.StandardF
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Decimal> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'). The property name is escaped before writing.
+Writes the <xref:System.Decimal> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'). The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Decimal> using the default <xref:System.Buffers.StandardFormat> (that is, 'G').

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1541,7 +1619,7 @@ Writes the <xref:System.Decimal> using the default <xref:System.Buffers.Standard
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Double> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'). The property name is escaped before writing.
+Writes the <xref:System.Double> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'). The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Double> using the default <xref:System.Buffers.StandardFormat> (that is, 'G').

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1575,7 +1653,7 @@ Writes the <xref:System.Double> using the default <xref:System.Buffers.StandardF
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Int32> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'), for example: 32767. The property name is escaped before writing.
+Writes the <xref:System.Int32> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767. The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Int32> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767.

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1609,7 +1687,7 @@ Writes the <xref:System.Int32> using the default <xref:System.Buffers.StandardFo
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Int64> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'), for example: 32767. The property name is escaped before writing.
+Writes the <xref:System.Int64> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767. The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Int64> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767.

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1643,7 +1721,7 @@ Writes the <xref:System.Int64> using the default <xref:System.Buffers.StandardFo
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.Single> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'). The property name is escaped before writing.
+Writes the <xref:System.Single> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'). The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.Single> using the default <xref:System.Buffers.StandardFormat> (that is, 'G').

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1682,7 +1760,7 @@ Writes the <xref:System.Single> using the default <xref:System.Buffers.StandardF
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.UInt32> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'), for example: 32767. The property name is escaped before writing.
+Writes the <xref:System.UInt32> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767. The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.UInt32> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767.

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -1721,7 +1799,7 @@ Writes the <xref:System.UInt32> using the default <xref:System.Buffers.StandardF
         <remarks>
           <format><![CDATA[  
 
-Writes the <xref:System.UInt64> using the default <xref:System.Buffers.StandardFormat> (i.e. 'G'), for example: 32767. The property name is escaped before writing.
+Writes the <xref:System.UInt64> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767. The property name is escaped before writing.
```
carlossanlop:
```suggestion
Writes the <xref:System.UInt64> using the default <xref:System.Buffers.StandardFormat> (that is, 'G'); for example, 32767.

The property name is escaped before writing.
```

xml/System.Text.Json/Utf8JsonWriter.xml
carlossanlop: @ahsonkhan On a second thought, I think it will be easier to get these changes introduced first, since as you said, we already have several suggestions applied, and then in a new PR apply the latest modifications introduced in dotnet/corefx#39092 by @safern.

xml/System.Text.Json/Utf8JsonWriter.xml
carlossanlop: @ahsonkhan The remark is still there in your second hyperlink. It has not been removed from source. Can you clarify this?

xml/System.Text.Json/Utf8JsonReader.xml
carlossanlop: I think we should get all the complex suggestions addressed in a separate PR.

xml/System.Text.Json/Utf8JsonWriter.xml
carlossanlop: Let's address this in a separate PR.

carlossanlop:
@ahsonkhan there are multiple complex requests that should be addressed separately due to recent changes in source. Let's get this PR merged first, then update the documentation in a new PR.

@rpetrusha @mairaw I addressed the last comments. Can you please take a last look to see if we can get it merged?

xml/System.Text.Json/Utf8JsonWriter.xml
ahsonkhan: It was removed: https://github.com/dotnet/corefx/blob/64a9e2ccf5e590e0a3a3d6dc9c2a7c828850c912/src/System.Text.Json/src/System/Text/Json/Writer/Utf8JsonWriter.WriteProperties.String.cs#L499-L503

image

Addressing it separately works for me. Tyvm for your help with this.

xml/System.Text.Json/Utf8JsonWriter.xml
```diff
@@ -663,10 +719,21 @@ The property name is escaped before writing.
...
+
+The property name should already be escaped when the instance of <xref:System.Text.Json.JsonEncodedText> was created.
+
+            ]]>
+          </format>
+        </remarks>
+        <exception cref="T:System.InvalidOperationException">Validation is enabled, and this method would result in writing invalid JSON.</exception>
       </Docs>
     </Member>
     <Member MemberName="WriteBooleanValue">
```
ahsonkhan: Just like `WriteNullValue`, we should fix this one too:

```diff
- The value to be written as a JSON literal "true" or "false" as an element of a JSON array.
+ The value to be written as a JSON literal <b>true</b> or <b>false</b> as an element of a JSON array.
```

carlossanlop: I fixed all the JSON literals I found in that file.

carlossanlop:
@Thanks @rpetrusha. I fixed additional JSON literals that needed to be changed from `"word"` to `<b>word</b>`

ahsonkhan:
A bunch of APIs changed between preview6/preview7. @rpetrusha - thanks for fixing some of the links.
Are you confident we reacted to all of them?
https://github.com/dotnet/core/blob/master/release-notes/3.0/preview/api-diff/preview7/3.0-preview7_System.Text.Json.md

rpetrusha:
I think so, @ahsonkhan. We've updated the API reference to reflect Preview 7 APIs, and the build report isn't showing any additional broken links.

xml/System.Text.Json/JsonElement.xml
```diff
@@ -777,9 +789,8 @@ This method does not parse the contents of a JSON string value.
         <summary>Gets the value at a specified index when the current value is an <see cref="F:System.Text.Json.JsonValueKind.Array" />.</summary>
```
ahsonkhan: Looks like we are missing the docs from source for the the other property, ValueKind:
https://github.com/dotnet/dotnet-api-docs/blob/c7324dff76601c7fdfeb798358fe63a43e61db3a/xml/System.Text.Json/JsonElement.xml#L1527-L1547

https://github.com/dotnet/corefx/blob/9e994a5923d426f0e64ed3b8f9b93df6f214a42a/src/System.Text.Json/src/System/Text/Json/Document/JsonElement.cs#L32-L38

xml/System.Text.Json/JsonElement.xml
```diff
@@ -777,9 +789,8 @@ This method does not parse the contents of a JSON string value.
...
         <value>To be added.</value>
         <remarks>To be added.</remarks>
-        <exception cref="T:System.InvalidOperationException">This value's <see cref="P:System.Text.Json.JsonElement.Type" /> is not <see cref="F:System.Text.Json.JsonValueKind.Array" />.</exception>
-        <exception cref="T:System.IndexOutOfRangeException">
-          <paramref name="index" /> is not in the range [0, <see cref="M:System.Text.Json.JsonElement.GetArrayLength" />()).</exception>
+        <exception cref="T:System.InvalidOperationException">This value's <see cref="P:System.Text.Json.JsonElement.ValueKind" /> is not <see cref="F:System.Text.Json.JsonValueKind.Array" />.</exception>
```
ahsonkhan: We are missing parameter description of this indexer:
https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.item?view=netcore-3.0#System_Text_Json_JsonElement_Item_System_Int32_

https://github.com/dotnet/dotnet-api-docs/blob/c7324dff76601c7fdfeb798358fe63a43e61db3a/xml/System.Text.Json/JsonElement.xml#L787-L795

It doesn't exist in source either.

@bartonjs, can you please wordsmith something here to describe the `JsonValueKind.Array` indexer?
https://github.com/dotnet/corefx/blob/9e994a5923d426f0e64ed3b8f9b93df6f214a42a/src/System.Text.Json/src/System/Text/Json/Document/JsonElement.cs#L40-L61

PR #2752, opened 2019-07-12
Manually added docs for System.IO.Path

carlossanlop:
No triple slash comments to import.

@JeremyKuhne and I are the area owners.

### DIFF
--- a/xml/System.IO/Path.xml
+++ b/xml/System.IO/Path.xml
@@ -2148,7 +2148,7 @@ Note that rooted paths can be either absolute (that is, fully qualified) or rela
 
 ## Remarks
 
-This method simply concatenates all the strings in `paths` and adds a directory separator character between any of the path components if one is not already present. If the <xref:System.ReadOnlySpan%601.Length> of any of the paths in `paths` is zero, the method concatenates the remaining arguments. If the resulting concatenated string's length is zero, the method returns <xref:System.String.Empty?displayProperty=nameWithType>.
+This method simply concatenates all the strings in `paths` and adds a directory separator character between any of the path components if one is not already present. If the <xref:System.String.Length> of any of the paths in `paths` is zero, the method concatenates the remaining arguments. If the resulting concatenated string's length is zero, the method returns <xref:System.String.Empty?displayProperty=nameWithType>.
 
 If any of the paths in `paths`, except for the last one, ends in a path separator character that is not appropriate for the target platform, the `Join` method preserves the original path separator character and appends the supported one. This issue arises in hard-coded paths that use the Windows backslash ("\") character, which is not recognized as a path separator on Unix-based systems. To work around this issue, you can:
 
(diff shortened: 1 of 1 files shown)


xml/System.IO/Path.xml
```diff
@@ -2328,11 +2368,11 @@ The following example illustrates the difference in the paths returned by the <x
...
-        <param name="path2">To be added.</param>
-        <param name="path3">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="path1">The first path to join.</param>
+        <param name="path2">The second path to join.</param>
+        <param name="path3">The third path to join.</param>
+        <summary>Concatenates three paths into a single path.</summary>
+        <returns>The concatenated path.</returns>
         <remarks>To be added.</remarks>
```
rpetrusha: The comments that appear for other Join overloads could be modified and added here.

xml/System.IO/Path.xml
```diff
@@ -2139,10 +2139,30 @@ Note that rooted paths can be either absolute (that is, fully qualified) or rela
...
+        <param name="paths">An array of paths.</param>
+        <summary>Concatenates an array of paths into a single path.</summary>
+        <returns>The concatenated path.</returns>
+        <remarks>
+          <format type="text/markdown">
+            <![CDATA[
+
+## Remarks
+
+This method simply concatenates all the strings in `paths` and adds a directory separator character between any of the path components if one is not already present. If the <xref:System.ReadOnlySpan%601.Length> of any of the paths in `paths` is zero, the method concatenates the remaining arguments. If the resulting concatenated string's length is zero, the method returns <xref:System.String.Empty?displayProperty=nameWithType>.
```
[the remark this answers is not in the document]
carlossanlop: Oops thanks for catching this one.

xml/System.IO/Path.xml
carlossanlop: Yes you're right, I missed this.

```suggestion
        <remarks>
          <format type="text/markdown">
            <![CDATA[

## Remarks

This method simply concatenates `path`, `path2`, and `path3` and adds a directory separator character between any of the path components if one is not already present. If the length of any of `path1`, `path2` or `path3` argument is zero, the method concatenates the remaining arguments. If the length of the resulting concatenated string is zero, the method returns <xref:System.String.Empty?displayProperty=nameWithType>.

If `path1` or `path2` ends in a path separator character that is not appropriate for the target platform, the `Join` method preserves the original path separator character and appends the supported one. This issue arises in hard-coded paths that use the Windows backslash ("\") character, which is not recognized as a path separator on Unix-based systems. To work around this issue, you can:

- Retrieve the value of the <xref:System.IO.Path.DirectorySeparatorChar?displayProperty=nameWithType> property rather than hard-coding a directory separator character.

- Use a forward slash ("/") as the directory separator character. This character is returned by the <xref:System.IO.Path.DirectorySeparatorChar?displayProperty=nameWithType> property on Unix-based systems and by the <xref:System.IO.Path.AltDirectorySeparatorChar?displayProperty=nameWithType> property on Windows systems.

Unlike the <xref:System.IO.Path.Combine%2A> method, the <xref:System.IO.Path.Join%2A> method does not attempt to root the returned path. (That is, if `path2` or `path2` is an absolute path, the `Join` method does not discard the previous paths as the <xref:System.IO.Path.Combine%2A> method does.

Not all invalid characters for directory and file names are interpreted as unacceptable by the `Join` method, because you can use these characters for search wildcard characters. For example, while `Path.Join("c:\\", "temp", "*.txt")` might be invalid when creating a file, it is valid as a search string. The `Join` method therefore successfully interprets it.

          ]]>
          </format>
        </remarks>
```

rpetrusha:
Thanks for the additional changes, @carlossanlop. I'll merge this PR now.

PR #2767, opened 2019-07-17
Added documentation for System.Net.Sockets APIs targeted for 3.0

jozkee:
List of APIs documented:
* AddresFamily - Added support for new Linux protocols dotnet/corefx#37315
   * ControllerAreaNetwork
   * Netlink
   * Packet
* ProtocolFamily
   * ControllerAreaNetwork
   * Netlink
   * Packet
 
* NetworkStream - Added a bunch of span overloads. Also did some corrections to current documentation.
   * Read(System.Span{System.Byte})
   * ReadAsync(System.Memory{System.Byte},System.Threading.CancellationToken)
   * ReadByte
   * Write(System.ReadOnlySpan{System.Byte})
   * WriteAsync(System.ReadOnlyMemory{System.Byte},System.Threading.CancellationToken)
   * WriteByte(System.Byte)
 
* SafeSocketHandle - Added Socket.SafeHandle dotnet/corefx#13470
   * #ctor(System.IntPtr,System.Boolean)
* Socket
   * SafeHandle
 
* SendPacketsElement - Added FileStream and Int64 OffsetLong support dotnet/corefx#25354
   * #ctor(System.IO.FileStream)
   * #ctor(System.IO.FileStream,System.Int64,System.Int32)
   * #ctor(System.IO.FileStream,System.Int64,System.Int32,System.Boolean)
   * #ctor(System.String,System.Int64,System.Int32)
   * #ctor(System.String,System.Int64,System.Int32,System.Boolean)
   * FileStream
   * OffsetLong
   * Offset (Updated)
 
* SocketOptionName - Added support for TcpKeepAlive options dotnet/corefx#25040 
I literally copied the summaries from https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-tcp-socket-options
   * TcpKeepAliveInterval
   * TcpKeepAliveRetryCount
   * TcpKeepAliveCount

### DIFF
--- a/xml/System.Net.Sockets/SendPacketsElement.xml
+++ b/xml/System.Net.Sockets/SendPacketsElement.xml
@@ -86,7 +86,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Net.Sockets.SendPacketsElement.%23ctor%28System.Byte%5B%5D%29> constructor initializes a new instance of the <xref:System.Net.Sockets.SendPacketsElement> class with a byte array of data. The <xref:System.Net.Sockets.SendPacketsElement> class is used with the <xref:System.Net.Sockets.SocketAsyncEventArgs.SendPacketsElements%2A?displayProperty=nameWithType> property to get or set a data buffer or file to be sent using the <xref:System.Net.Sockets.Socket.SendPacketsAsync%2A?displayProperty=nameWithType> method.  
+ The <xref:System.Net.Sockets.SendPacketsElement> class is used with the <xref:System.Net.Sockets.SocketAsyncEventArgs.SendPacketsElements%2A?displayProperty=nameWithType> property to get or set a data buffer or file to be sent using the <xref:System.Net.Sockets.Socket.SendPacketsAsync%2A?displayProperty=nameWithType> method.  
   
  ]]></format>
         </remarks>
@@ -118,12 +118,12 @@
       </Parameters>
       <Docs>
         <param name="fileStream">The file to be transmitted using the <see cref="M:System.Net.Sockets.Socket.SendPacketsAsync(System.Net.Sockets.SocketAsyncEventArgs)" /> method.</param>
-        <summary>Initializes a new instance of the <see cref="T:System.Net.Sockets.SendPacketsElement" /> class using the specified <see cref="T:System.IO.FileStream" />.</summary>
+        <summary>Initializes a new instance of the <see cref="T:System.Net.Sockets.SendPacketsElement" /> class using the specified <see cref="T:System.IO.FileStream" /> object.</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- This constructor initializes a new instance of the <xref:System.Net.Sockets.SendPacketsElement> class using a <xref:System.IO.FileStream> object. The <xref:System.Net.Sockets.SendPacketsElement> class is used with the <xref:System.Net.Sockets.SocketAsyncEventArgs.SendPacketsElements?displayProperty=nameWithType> property to get or set a data buffer or file to be sent using the <xref:System.Net.Sockets.Socket.SendPacketsAsync%2A?displayProperty=nameWithType> method.  
+ The <xref:System.Net.Sockets.SendPacketsElement> class is used with the <xref:System.Net.Sockets.SocketAsyncEventArgs.SendPacketsElements?displayProperty=nameWithType> property to get or set a data buffer or file to be sent using the <xref:System.Net.Sockets.Socket.SendPacketsAsync%2A?displayProperty=nameWithType> method.  
   
  ]]></format>
         </remarks>
@@ -171,7 +171,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Net.Sockets.SendPacketsElement.%23ctor%28System.String%29> constructor initializes a new instance of the <xref:System.Net.Sockets.SendPacketsElement> class with a filename of a file. The <xref:System.Net.Sockets.SendPacketsElement> class is used with the <xref:System.Net.Sockets.SocketAsyncEventArgs.SendPacketsElements%2A?displayProperty=nameWithType> property to get or set a data buffer or file to be sent using the <xref:System.Net.Sockets.Socket.SendPacketsAsync%2A?displayProperty=nameWithType> method.  
+ The <xref:System.Net.Sockets.SendPacketsElement> class is used with the <xref:System.Net.Sockets.SocketAsyncEventArgs.SendPacketsElements%2A?displayProperty=nameWithType> property to get or set a data buffer or file to be sent using the <xref:System.Net.Sockets.Socket.SendPacketsAsync%2A?displayProperty=nameWithType> method.  
   
  Universal Naming Convention (UNC) paths are supported by the `filepath` parameter. If the file is in the current working directory, no path information needs to be specified.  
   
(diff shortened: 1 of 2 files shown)


Jozkee:
Comments about changes to docs of existing APIs:

* NetworkStream.Read:
   * Removed ", or 0 if the socket was closed" from Summary (ln 1378) and Remarks (ln 1383), as far as I know, and asking @cory, Read does not return zero if the socket is closed.
   * On Note 1, replaced IOException for InvalidOperationException since that is the correct exception type thrown when CanRead is false (ln 1386).
   * Moved "There is a failure reading from the network." from ObjectDisposedException to IOException as the exception has nothing to do with the object disposal (ln 1422).
   * On Remarks, removed "If the remote host shuts down the connection, and all available data has been received, this method completes immediately and return zero bytes." (ln 1383) for the same reason of bullet point 1.
   * Moved "An error occurred when accessing the socket" from ArgumentOutOfRangeException to IOException (ln 1418).
   * In IOException, removed "The underlying socket is closed" I think this is already covered in the message "An error occurred when accessing the socket" (ln 1418).

* NetworkStream.Write:
   * Moved text in Remarks section to a separate Note in order to align with Read method's Remarks. (ln 1996)
   * Moved "There is a failure reading from the network." from ObjectDisposedException to IOException (ln 2031).


* SendPacketElem:
   * .ctor(string filepath, int32 offset, int32 count) 
   * .ctor(string filepath, int32 offset, int32 count, bool endOfPacket)
Removed  "The offset and count must be less than the size of the file indicated by the filepath parameter." from ArgumentOutOfRangeException since is actually not being validated on filepath overloads (ln 335 & 539); that validation is only implemented on byte[] overloads.
	
* SendPacketElem:
   * Buffer
   * Filepath
    Remarks now points out that these properties default values will be null if you use a constructor other than the one that populates this property (ln 630 & 767)

Jozkee:
@dotnet/ncl @mairaw @rpetrusha

scalablecory:
> Removed ", or 0 if the socket was closed" from Summary (ln 1378) and Remarks (ln 1383), as far as I know, and asking @scalablecory, Read does not return zero if the socket closes.

As we discussed it does not return 0 *when you close the socket*, but it does return 0 -- this happens once you've exhausted all of the data from the socket after the other side does a `shutdown(send)`.

xml/System.Net.Sockets/SendPacketsElement.xml
```diff
@@ -117,9 +117,18 @@
...
+        <summary>Initializes a new instance of the <see cref="T:System.Net.Sockets.SendPacketsElement" /> class using the specified <paramref name="fileStream" />.</summary>
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+  
+## Remarks  
+ This constructor initializes a new instance of the <xref:System.Net.Sockets.SendPacketsElement> class using a <xref:System.IO.FileStream> object. The <xref:System.Net.Sockets.SendPacketsElement> class is used with the <xref:System.Net.Sockets.SocketAsyncEventArgs.SendPacketsElements%2A?displayProperty=nameWithType> property to get or set a data buffer or file to be sent using the <xref:System.Net.Sockets.Socket.SendPacketsAsync%2A?displayProperty=nameWithType> method.  
+  
+ ]]></format>
+        </remarks>
+        <exception cref="T:System.ArgumentException">The <paramref name="fileStream" /> parameter must have been opened for asynchronous reading and writing.</exception>
```
rpetrusha: must have been? was?

xml/System.Net.Sockets/SendPacketsElement.xml
```diff
@@ -255,11 +264,21 @@
...
       <Docs>
-        <param name="fileStream">To be added.</param>
-        <param name="offset">To be added.</param>
-        <param name="count">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="fileStream">The file to be transmitted using the <see cref="M:System.Net.Sockets.Socket.SendPacketsAsync(System.Net.Sockets.SocketAsyncEventArgs)" /> method.</param>
+        <param name="offset">The offset, in bytes, from the beginning of the file to the location in the file to start sending the bytes in the file.</param>
+        <param name="count">The number of bytes to send starting from the <paramref name="offset" />. If <paramref name="count" /> is zero, the entire file is sent.</param>
+        <summary>Initializes a new instance of the <see cref="T:System.Net.Sockets.SendPacketsElement" /> class using the specified <paramref name="fileStream" />, buffer <paramref name="offset" /> and <paramref name="count" />.</summary>
```
rpetrusha:
```suggestion
        <summary>Initializes a new instance of the <see cref="T:System.Net.Sockets.SendPacketsElement" /> class using the specified <range of a <see cref="T:System.IO.FileStream" /> object.</summary>
```

xml/System.Net.Sockets/SendPacketsElement.xml
```diff
@@ -255,11 +264,21 @@
...
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+  
+## Remarks  
+ This constructor initializes a new instance of the <xref:System.Net.Sockets.SendPacketsElement> class using a <xref:System.IO.FileStream> object. The <xref:System.Net.Sockets.SendPacketsElement> class is used with the <xref:System.Net.Sockets.SocketAsyncEventArgs.SendPacketsElements%2A?displayProperty=nameWithType> property to get or set a data buffer or file to be sent using the <xref:System.Net.Sockets.Socket.SendPacketsAsync%2A?displayProperty=nameWithType> method.
```
rpetrusha:
```suggestion
The <xref:System.Net.Sockets.SendPacketsElement> class is used with the <xref:System.Net.Sockets.SocketAsyncEventArgs.SendPacketsElements%2A?displayProperty=nameWithType> property to get or set a data buffer or file to be sent using the <xref:System.Net.Sockets.Socket.SendPacketsAsync%2A?displayProperty=nameWithType> method.  
```

xml/System.Net.Sockets/SendPacketsElement.xml
Jozkee: What about this?
```suggestion
        <exception cref="T:System.ArgumentException">The <paramref name="fileStream" /> is not open for asynchronous reading and writing.</exception>
```

xml/System.Net.Sockets/SendPacketsElement.xml
Jozkee: If you want me to replace 
> the specified filestream, buffer offset and count.

for 

> the specified range of a filestream object.

Shouldn't we also apply that to ALL the rest of existing overloads, including the ones taking a `byte[]` and a `string` filename path? i.e: look at summary in the above overload:
https://github.com/dotnet/dotnet-api-docs/blob/54c925d4c4b0708415db50abfd0ea0bd36cc151a/xml/System.Net.Sockets/SendPacketsElement.xml#L225

xml/System.Net.Sockets/SendPacketsElement.xml
Jozkee: Same question as [above](https://github.com/dotnet/dotnet-api-docs/pull/2767/files#r305053915), this paragraph is in every single overload.
https://github.com/dotnet/dotnet-api-docs/blob/54c925d4c4b0708415db50abfd0ea0bd36cc151a/xml/System.Net.Sockets/SendPacketsElement.xml#L230

Jozkee:
I've addresssed Ron comments. This is ready for re-review.
@rpetrusha @mairaw

xml/System.Net.Sockets/Socket.xml
```diff
@@ -8341,8 +8341,8 @@ The GetHashCode method returns a hash code of this instance. This value can be u
         <ReturnType>System.Net.Sockets.SafeSocketHandle</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Gets a <see cref="T:System.Net.Sockets.SafeSocketHandle" /> that represents the socket handle that the current <see cref="T:System.Net.Sockets.Socket" /> object encapsulates.</summary>
+        <value>A socket handle exposed in a safe manner for the socket that the current <see cref="T:System.Net.Sockets.Socket" /> object encapsulates.</value>
         <remarks>To be added.</remarks>
```
scalablecory: Lets add something warning how using the safe handle to operate on the socket directly may result in an undocumented state. We just hit this in https://github.com/dotnet/corefx/issues/39677.

xml/System.Net.Sockets/SendPacketsElement.xml
rpetrusha: Yes. "is not open" is good, if you can make that change.

xml/System.Net.Sockets/SendPacketsElement.xml
rpetrusha: Yes, we should make the replacement elsewhere, @Jozkee.

xml/System.Net.Sockets/SendPacketsElement.xml
rpetrusha: We should try to disambiguate. For example, buffer offsets are 32-bit and 64-bit, and both are represented. The remaining param name in the constriuctor list can also be replaced.

xml/System.Net.Sockets/Socket.xml
rpetrusha: This would take the following form:
```suggestion
        <remarks>  
          <format type="text/markdown"><![CDATA[  

> [!IMPORTANT]
> Using the <xref:System.Net.Sockets.SafeSocketHandle> to operate on the socket directly may result in an undocumented state. 

         ]]></format>
         </remarks>
```

rpetrusha:
This looks good, @Jozkee. Thanks for making the additional changes. I'll merge your PR now.

PR #2769, opened 2019-07-17
LOCBLOCKER L104991: Removing spaces to avoid issues in localization process

v-maudel:
Description:  The OPS builds for cs-CZ, pl-PL and tr-TR failed. Extra spaces need to be removed.

@mairaw @sdmaclea Could you please help by reviewing the changes and merge the PR if agreed? It is blocking the localization process. 

FYI @JasonCard


v-maudel:
Thanks @rpetrusha

Issue #2795, opened 2019-07-22
These members should not exist


carlossanlop:
I found several members that should not exist.

They were all added in this PR: [CI Update - .NET Core 3.0 Preview 2 #1866](https://github.com/dotnet/dotnet-api-docs/pull/1866)

## Microsoft.VisualBasic.FileIO/TextFieldParser.xml

But it already existed. The problem is that it was added with incorrect formatting, so it wasn't detected as a duplicate of a previously existing member:

https://github.com/dotnet/dotnet-api-docs/blob/1c7f6a8120aaa3ce0fcab7ea526f6883c3ecbc25/xml/Microsoft.VisualBasic.FileIO/TextFieldParser.xml#L1436-L1461

The original method (with the correct formatting) already existed and was already documented:

https://github.com/dotnet/dotnet-api-docs/blob/1c7f6a8120aaa3ce0fcab7ea526f6883c3ecbc25/xml/Microsoft.VisualBasic.FileIO/TextFieldParser.xml#L620-L649

Here's the change that introduced it: https://github.com/dotnet/dotnet-api-docs/commit/40032d55704974a9ea9a7ff3b65e67a91fc9b1bb#diff-3030e8394a77f7dd05fa74086d06b528R1433

---

## System.Printing/PrintQueueCollection.xml

Another one that should not exist:

https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System.Printing/PrintQueueCollection.xml#L322-L346

It's already documented here:

https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System.Printing/PrintJobInfoCollection.xml#L153-L185

It was introduced by the same PR: https://github.com/dotnet/dotnet-api-docs/pull/1866/files#diff-79ce4a68db15163cee852a1015ee4f7eR320

---

## System.Printing/PrintJobInfoCollection.xml

Similar to the previous one:

https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System.Printing/PrintJobInfoCollection.xml#L218-L241

It's already documented here:

https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System.Printing/PrintQueueCollection.xml#L221-L262

It was introduced by the same PR: https://github.com/dotnet/dotnet-api-docs/pull/1866/files#diff-66c6bdde4fe609d882f255e22fa75726R216

--- 

## System.Runtime.DeserializationBlockedException

[...]

PR #2806, opened 2019-07-23
Remove members that should not exist

carlossanlop:
Fixes https://github.com/dotnet/dotnet-api-docs/issues/2795

@mairaw @rpetrusha

There are 3 members that are duplicated and should not exist. Their "MemberName" has a strange format with the type full name prepended to them. I found they are already documented within the same file.

There is another case of a file that I am wondering if I should remove, because it got deleted from code: xml/System.Runtime.Serialization/DeserializationBlockedException.xml

But I assume we should not remove files because they exist in previous .NET versions. Let me know if this is correct or not.

### DIFF
--- a/xml/Microsoft.VisualBasic.FileIO/TextFieldParser.xml
+++ b/xml/Microsoft.VisualBasic.FileIO/TextFieldParser.xml
@@ -1433,32 +1433,6 @@
         <related type="Article" href="~/docs/visual-basic/developing-apps/programming/drives-directories-files/parsing-text-files-with-the-textfieldparser-object.md">Parsing Text Files with the TextFieldParser Object</related>
       </Docs>
     </Member>
-    <Member MemberName="System.IDisposable.Dispose">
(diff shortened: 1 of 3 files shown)


carlossanlop:
@dend @joelmartinez from what I've seen in other PRs, you folks seem to be the experts I should contact for help to clarify weirdness in the dotnet-api-docs xml files. Can you please also take a look?

joelmartinez:
Hi @carlossanlop ... these members should not be removed. They are explicitly implemented members, so the member name format you see (eg `System.IDisposable.Dispose`) will correlate to a **non-public** explicit member implementation:

```csharp
void IDisposable.Dispose ();
```
Also ... these members will be re-introduced the next time CI is run :)

carlossanlop:
@joelmartinez Thanks for taking a look. It seems weird to me that they show up multiple times. In the original issue I reported, I shared the location of what I think are duplicate signatures.
If I still need to keep them, then I will duplicate the documentation from their existing cloned signature.

PR #2822, opened 2019-07-24
Adding documentation for new CipherSuitesPolicy, SslStream.NegotiatedCipherSuite and TlsCipherSuites

jozkee:
This completes System.Net.Security namespace by documenting the following:

* CipherSuitesPolicy
  * .ctor
  * AllowedCipherSuites
* SslStream
  * NegotiatedCipherSuite
* TlsCipherSuites 
  * ~340 enums representing cipher suites

### DIFF
--- a/xml/System.Net.Security/CipherSuitesPolicy.xml
+++ b/xml/System.Net.Security/CipherSuitesPolicy.xml
@@ -40,15 +40,16 @@
       </Parameters>
       <Docs>
         <param name="allowedCipherSuites">The collection of cipher suites allowed in this policy for negotiation.</param>
-        <summary>Initializes a new instance of the <see cref="T:System.Net.Security.CipherSuitesPolicy" /> class that can be used on a client or server authentication.</summary>
+        <summary>Initializes a new instance of the <see cref="T:System.Net.Security.CipherSuitesPolicy" /> class that can be used for client or server authentication.</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[  
 > [!NOTE]
-> Defining a cipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS of the ability to decide which are the best cipher suites to negotiate with and will need of your attention to manually check and update this code. It is strongly recommended to avoid the use of this property and rely on your constantly updated OS policy.
+> Defining a cipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS from deciding which are the best cipher suites to negotiate with and requires that you manually check and update this code. We strongly recommend that you avoid using this feature and rely on your constantly updated OS policy.
 
 ## Remarks
 
-Unlike the OS underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in `allowedCipherSuites` doesn't guarantee their priority in the client-server negotiation. You can only initialize an instance of this class if you're using Linux systems with OpenSSL 1.1.1 or higher or macOS .
+Unlike the operating system's underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in `allowedCipherSuites` doesn't guarantee their priority in client-server negotiation. You can only initialize an instance of this class if you're using Linux systems with OpenSSL 1.1.1 or higher or macOS.
+
 
  ]]></format>
         </remarks>
@@ -81,13 +82,12 @@ Unlike the OS underlying SSL/TLS implementation, the order of the <xref:System.N
         <remarks>
           <format type="text/markdown"><![CDATA[
 
-## Remarks
 
 > [!NOTE]
-> Defining a cipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS of the ability to decide which are the best cipher suites to negotiate with and will need of your attention to manually check and update this code. It is strongly recommended to avoid the use of this property and rely on your constantly updated OS policy.
+> Defining a cipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS from deciding which are the best cipher suites to negotiate with and requires that you manually check and update this code. We strongly recommend that you avoid using this feature and rely on your constantly updated OS policy.
 
 ## Remarks
-Unlike the OS underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in this property does not determine their priority in the client-server negotiation. It is required to use OSX or Unix-based systems with OpenSSL 1.1.1 in order to initialize an instance of this class.
+Unlike the operating system's underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in `allowedCipherSuites` doesn't guarantee their priority in client-server negotiation. You can only initialize an instance of this class if you're using Linux systems with OpenSSL 1.1.1 or higher or macOS.
 
  ]]></format>
         </remarks>
--- a/xml/System.Net.Security/SslStream.xml
+++ b/xml/System.Net.Security/SslStream.xml
@@ -3460,7 +3460,7 @@ Authentication has not occurred.</exception>
   
 ## Remarks
 
-This property gets the cipher suite that is going to be used in the communication between a client and server by the underlying SSL/TLS implementation. This property is only available after a successful call to <xref:System.Net.Security.SslStream.AuthenticateAsServer%2A> or <xref:System.Net.Security.SslStream.AuthenticateAsClient%2A>.
+This property gets the cipher suite that is going to be used in the communication between a client and server by the underlying SSL/TLS implementation. It is only available after a successful call to <xref:System.Net.Security.SslStream.AuthenticateAsServer%2A> or <xref:System.Net.Security.SslStream.AuthenticateAsClient%2A>.
 
  ]]></format>
         </remarks>
--- a/xml/System.Net.Security/TlsCipherSuite.xml
+++ b/xml/System.Net.Security/TlsCipherSuite.xml
@@ -24,7 +24,7 @@
   
 ## Remarks
 Both field names and values are based on the [TLS Cipher Suites list](https://www.iana.org/assignments/tls-parameters/tls-parameters.xml#tls-parameters-4) from the Internet Assigned Numbers Authority (IANA).
-This enumeration represents values that were known at the time a specific version of .NET was released. Unmapped values may be returned by properties or methods, and methods should generally not verify that a given input was defined within the enum, as the value may be transparently passed from or to the system underlying TLS implementation.
+This enumeration represents values that were known at the time a specific version of .NET was released. Unmapped values may be returned by properties or methods, and methods should generally not verify that a given input was defined within the enum, as the value may be transparently passed from or to the system's underlying TLS implementation.
 Inclusion in this enumeration does not indicate that a .NET platform supports a particular cipher suite.
 
  ]]></format>


Jozkee:
@dotnet/ncl @krwq @bartonjs 
@mairaw @rpetrusha @carlossanlop

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -39,9 +39,19 @@
...
       <Docs>
-        <param name="allowedCipherSuites">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="allowedCipherSuites">The collection of cipher suites allowed in this policy for negotiation.</param>
+        <summary>Initializes a cipher suite policy that can be used on a client or server authentication.</summary>
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+> [!NOTE]
+> Defining a chipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS of the ability to decide which are the best cipher suites to negotiate with and will need of your attention to manually check and update this code. It is strongly recommended to hold off the use of this property and rely on your constantly updated OS policy.
```
bartonjs: A) typo: chipher vs cipher
B) I suggest "avoid" instead of "hold off on"

```suggestion
> Defining a cipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS of the ability to decide which are the best cipher suites to negotiate with and will need of your attention to manually check and update this code. It is strongly recommended to avoid the use of this property and rely on your constantly updated OS policy.
```

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -39,9 +39,19 @@
...
+
+## Remarks
+Unlike the OS underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in `allowedCipherSuites` does not determine their priority in the client-server negotiation. It is required to use OSX or Unix-based systems with OpenSSL 1.1.1 in order to initialize an instance of this class.
```
bartonjs: "macOS" instead of "OSX" (the OS rebranded)

```suggestion
Unlike the OS underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in `allowedCipherSuites` does not determine their priority in the client-server negotiation. It is required to use macOS or Unix-based systems with OpenSSL 1.1.1 in order to initialize an instance of this class.
```

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -39,9 +39,19 @@
         <Parameter Name="allowedCipherSuites" Type="System.Collections.Generic.IEnumerable&lt;System.Net.Security.TlsCipherSuite&gt;" />
       </Parameters>
       <Docs>
-        <param name="allowedCipherSuites">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="allowedCipherSuites">The collection of cipher suites allowed in this policy for negotiation.</param>
```
krwq: should we also have a corresponding change updating comments in corefx?

xml/System.Net.Security/CipherSuitesPolicy.xml
bartonjs: I think the "or" here is lexically ambiguous.  IIRC it requires `((macOS) or (Linux with OpenSSL 1.1.1))`, but it's possible to parse as `((macOS or Linux) with OpenSSL 1.1.1)`.

It probably needs some sort of style bubble for OS-specific requirements.  Something like

```
[!IMPORTANT]
This feature is dependent on platform support.
* Microsoft Windows: This feature is not available.
* macOS: This feature is available.
* Linux: This feature is available on systems with OpenSSL 1.1.1 or higher.
```

But I'd wait for a writer before changing the format.

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -65,9 +75,19 @@
...
-        <summary>To be added.</summary>
-        <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <summary>Gets the collection of cipher suites passed through <see cref="M:System.Net.Security.CipherSuitesPolicy.#ctor(System.Collections.Generic.IEnumerable{System.Net.Security.TlsCipherSuite})" />, which are the ones allowed to negotiate with on a client or server <see cref="T:System.Net.Security.SslStream" /> authentication.</summary>
+        <value>The collection of cipher suites allowed in this policy for negotiation.</value>
+        <remarks>
+          <format type="text/markdown">
+  <![CDATA[  
+> [!NOTE]
+> Defining a chipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS of the ability to decide which are the best cipher suites to negotiate with and will need of your attention to manually check and update this code. It is strongly recommended to hold off the use of this property and rely on your constantly updated OS policy.
```
bartonjs: Same comments as previous.

xml/System.Net.Security/CipherSuitesPolicy.xml
bartonjs: Does it actually not determine the order?  I thought it did on macOS and Linux (but on Windows it won't).

xml/System.Net.Security/SslStream.xml
```diff
@@ -3453,9 +3453,17 @@ Authentication has not occurred.</exception>
         <ReturnType>System.Net.Security.TlsCipherSuite</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <summary>Gets the cipher suite resolved to use by this <see cref="T:System.Net.Security.SslStream"/>.</summary>
```
bartonjs:
```suggestion
        <summary>Gets the cipher suite which was negotiated for this connection.</summary>
```

https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Property says the formula here is `Gets <summary without specifying the type>.`

xml/System.Net.Security/SslStream.xml
```diff
@@ -3453,9 +3453,17 @@ Authentication has not occurred.</exception>
...
+        <value>The cipher suite resolved to use by this <see cref="T:System.Net.Security.SslStream"/>.</value>
```
bartonjs:
```suggestion
        <value>One of the enumeration values that identifies the cipher suite which was negotiated for this connection.</value>
```

https://github.com/dotnet/dotnet-api-docs/wiki/Property-Value says `One of the enumeration values that <additional information>.` (and then some stuff about defaults that don't apply here)

xml/System.Net.Security/CipherSuitesPolicy.xml
krwq: What about `The list of cipher suites allowed to be negotiated`

xml/System.Net.Security/CipherSuitesPolicy.xml
krwq: typo: chipher

xml/System.Net.Security/CipherSuitesPolicy.xml
carlossanlop: It's not required, and it's not part of the effort to backport the documentation to triple slash in code. The source of truth is dotnet-api-docs.

xml/System.Net.Security/SslStream.xml
```diff
@@ -3453,9 +3453,17 @@ Authentication has not occurred.</exception>
...
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+  
+## Remarks
+This property gets the cipher suite that is going to be used in the communication between a client and server by the underlying SSL/TLS implementation. This property is only available after a successful call to <xref:System.Net.Security.SslStream.AuthenticateAsServer%2A> or <xref:System.Net.Security.SslStream.AuthenticateAsClient%2A>.
```
bartonjs: There's a tense problem with this... data may have already been transferred, the cipher suite may already be in use.  I'd just delete the first sentence.

```suggestion
This property is only available after a successful call to <xref:System.Net.Security.SslStream.AuthenticateAsServer%2A> or <xref:System.Net.Security.SslStream.AuthenticateAsClient%2A>.
```

(I don't know if it needs to pedantically talk about the Async variants as well)

xml/System.Net.Security/TlsCipherSuite.xml
```diff
@@ -18,8 +18,15 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>To be added.</summary>
-    <remarks>To be added.</remarks>
+    <summary>Specifies the possible cipher suite values to negotiate with in a client-server authentication that uses the TLS/SSL protocols.</summary>
```
bartonjs:
```suggestion
    <summary>Represents cipher suite values for the TLS (formerly SSL) protocol.</summary>
```

xml/System.Net.Security/CipherSuitesPolicy.xml
krwq: OS is allowed to further restrict the list so it doesn't strictly speaking prevent OS from deciding (implementation will make the last call) but if new cipher suite shows up in the updated version of the underlying implementation a user won't get it by default (which in some cases might or might not be expected).

Also OS should be replaced with `underlying implementation`

xml/System.Net.Security/CipherSuitesPolicy.xml
krwq: I'd drop `Unlike the OS underlying SSL/TLS implementation` - not sure what that part means

xml/System.Net.Security/TlsCipherSuite.xml
```diff
@@ -18,8 +18,15 @@
...
+    <remarks>
+          <format type="text/markdown"><![CDATA[  
+  
+## Remarks
+Both, field names and values are based on the [TLS Cipher Suites list](https://www.iana.org/assignments/tls-parameters/tls-parameters.xml#tls-parameters-4) posted by IANA.
```
bartonjs:
```suggestion
Both field names and values are based on the [TLS Cipher Suites list](https://www.iana.org/assignments/tls-parameters/tls-parameters.xml#tls-parameters-4) from the Internet Assigned Numbers Authority (IANA).
```

We might also want a sentence like

```
This enumeration represents values that were known at the time a specific version of .NET was released.  
Unmapped values may be returned by properties or methods, and methods should generally not verify that a given input was defined within the enum, as the value may be transparently passed from or to the system TLS libraries.

File: xml/System.Net.Security/CipherSuitesPolicy.xml (hunk 2 above)
Comment (krwq) [reply xml/System.Net.Security/CipherSuitesPolicy.xml]: `does not determine their priority` => `does not guarantee their priority`

File: xml/System.Net.Security/CipherSuitesPolicy.xml [hunk 10]
```diff
@@ -39,9 +39,19 @@
...
+          <format type="text/markdown"><![CDATA[  
+> [!NOTE]
+> Defining a chipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS of the ability to decide which are the best cipher suites to negotiate with and will need of your attention to manually check and update this code. It is strongly recommended to hold off the use of this property and rely on your constantly updated OS policy.
+
+## Remarks
+Unlike the OS underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in `allowedCipherSuites` does not determine their priority in the client-server negotiation. It is required to use OSX or Unix-based systems with OpenSSL 1.1.1 in order to initialize an instance of this class.
+
+ ]]></format>
+        </remarks>
+        <exception cref="T:System.PlatformNotSupportedException">Platform does not meet the requirements to setup a policy.</exception>
```
krwq: I'd move `It is required to use OSX or Unix-based systems with OpenSSL 1.1.1 in order to initialize an instance of this class.` here

xml/System.Net.Security/TlsCipherSuite.xml
bartonjs: And maybe another one

```
Inclusion in this enumeration does not indicate that a .NET platform supports a particular cipher suite.
```

xml/System.Net.Security/CipherSuitesPolicy.xml
bartonjs: @krwq/@jozkee We might want to update the `.tt` files to include the summary value for the enum members, so when we run it in the future we can just use Carlos' tooling to merge the corefx comments with the placeholder docs here.

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -65,9 +75,19 @@
         <ReturnType>System.Collections.Generic.IEnumerable&lt;System.Net.Security.TlsCipherSuite&gt;</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <summary>Gets the collection of cipher suites passed through <see cref="M:System.Net.Security.CipherSuitesPolicy.#ctor(System.Collections.Generic.IEnumerable{System.Net.Security.TlsCipherSuite})" />, which are the ones allowed to negotiate with on a client or server <see cref="T:System.Net.Security.SslStream" /> authentication.</summary>
```
krwq: `List of the cipher suites passed into the underlying implementation. It is a subset of passed in allowedCipherSuites`.

(for OSX this will equal to passed in list, for OpenSSL it might be smaller than passed in - implementation on OSX might be later improved so I wouldn't document that implementation detail)

bartonjs: @krwq Style guidelines are that a summary is one sentence, and for a get-only property starts with "Gets ".  (get/set also start with Gets, but "Gets or sets" :smile:)

xml/System.Net.Security/SslStream.xml
krwq: This is a cipher suite which was already negotiated and is already being used. Comment could be something like:

`Cipher suite picked by the server from the client supplied list of allowed cipher suites`

xml/System.Net.Security/SslStream.xml
krwq: When this property is available the channel already uses the selected cipher suite (so `is going to be used` should be re-phrased to something like `being used` or just `used`). Also part about `underlying implementation` is rather confusing. I'd keep it simple to something like:
`Cipher suite negotiated during SSL/TLS handshake. Property is only available after the handshake is complete (after successful call to ...<whatever you already have>)`

krwq: what's the `%2A` for in the xref?

xml/System.Net.Security/SslStream.xml
```diff
@@ -3453,9 +3453,17 @@ Authentication has not occurred.</exception>
...
+
+ ]]></format>
+        </remarks>
+        <exception cref="T:System.InvalidOperationException">The property was accessed before the completion of the authentication process or the authentication process failed.</exception>
```
krwq: Try to match description with this https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslstream.keyexchangealgorithm?view=netframework-4.8 which is related property (KeyExchange is implied by the negotiated cipher suite) as much as possible at least

xml/System.Net.Security/TlsCipherSuite.xml
```diff
@@ -39,7 +46,7 @@
       </ReturnValue>
       <MemberValue>4869</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Represents the TLS_AES_128_CCM_8_SHA256 cipher suite.</summary>
```
krwq: not checking anything after here, let me know if there is something there to check

xml/System.Net.Security/CipherSuitesPolicy.xml
krwq: on OSX I think it does, on Linux it will do some re-ordering (TLS1.3 cipher suites will go first, possibly some null ciphers will go to the back - can't remember the details) but mostly preserves the order. I'd just keep the sentence to `order is not guaranteed` (or we will do best effort or something along these lines)

xml/System.Net.Security/SslStream.xml
krwq: yes, I think everything after client/server hello will be encrypted (minus some edge cases around re-negotiation and stuff)

Jozkee: URL-encoded star (*), I think is used to refer to the method's name without referring to a specific signature.

xml/System.Net.Security/CipherSuitesPolicy.xml
krwq: Didn't know that 😄 my comment was mostly about the second part that the list is guaranteed to be a subset of a list but might not be exactly passed in list.

xml/System.Net.Security/SslStream.xml
krwq: with `Gets` prefix per @bartonjs comment

xml/System.Net.Security/SslStream.xml
bartonjs: > >what's the %2A for in the xref?
>
>  I think is used to refer to the method's name without referring to a specific signature.

Yeah, targets the method group (full overload set) rather than a specific method.

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -39,9 +39,19 @@
         <Parameter Name="allowedCipherSuites" Type="System.Collections.Generic.IEnumerable&lt;System.Net.Security.TlsCipherSuite&gt;" />
       </Parameters>
       <Docs>
-        <param name="allowedCipherSuites">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="allowedCipherSuites">The collection of cipher suites allowed in this policy for negotiation.</param>
+        <summary>Initializes a cipher suite policy that can be used on a client or server authentication.</summary>
```
mairaw: We have sort of a boilerplate on how we describe constructors (see https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Constructor). Perhaps something like this?

```suggestion
        <summary>Initializes a new instance of the <see cref="T:System.Net.Security.CipherSuitesPolicy" /> class that can be used on a client or server authentication.</summary>
```

xml/System.Net.Security/SslStream.xml
Jozkee: I was trying to match against [`HashAlgorithm`](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslstream.hashalgorithm?view=netcore-3.0) which is also part of the negotiation.

krwq: that works too

xml/System.Net.Security/CipherSuitesPolicy.xml
mairaw: I have some extra suggestions but I'll let you add @bartonjs and @krwq suggestions first and then I can review the final text.

xml/System.Net.Security/TlsCipherSuite.xml
```diff
@@ -5019,7 +5026,7 @@
       </ReturnValue>
       <MemberValue>0</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Represents the TLS_NULL_WITH_NULL_NULL cipher suite; this is the default value.</summary>
```
Jozkee: @krwq this might worth reviewing. In this field I am stating `TLS_NULL_WITH_NULL_NULL` is enum's default value; not sure if it should really be included.

xml/System.Net.Security/CipherSuitesPolicy.xml
mairaw:
```suggestion

Unlike the OS underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in `allowedCipherSuites` doesn't guarantee their priority in the client-server negotiation. You can only initialize an instance of this class if you're using Linux systems with OpenSSL 1.1.1 or higher or macOS .
```

mairaw: I left a suggestion above. See what you think. I've changed from Unix-based to Linux since macOS it's also Unix-based.

xml/System.Net.Security/CipherSuitesPolicy.xml
mairaw: You have think of the text here as the \<condition> as in `<Exception> is thrown if <condition>.`
So maybe the exception text could be:

```suggestion
        <exception cref="T:System.PlatformNotSupportedException">Platform is not a Linux system with OpenSSL 1.1.1 or higher or a macOS.</exception>
```

I think it's fine to have that sentence in the remarks as well.

xml/System.Net.Security/CipherSuitesPolicy.xml
mairaw: And you can also use the Insert a suggestion button to provide the exact change that you want that they can commit directly:

image

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -65,9 +75,19 @@
         <ReturnType>System.Collections.Generic.IEnumerable&lt;System.Net.Security.TlsCipherSuite&gt;</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <summary>Gets the collection of cipher suites passed through <see cref="M:System.Net.Security.CipherSuitesPolicy.#ctor(System.Collections.Generic.IEnumerable{System.Net.Security.TlsCipherSuite})" />, which are the ones allowed to negotiate with on a client or server <see cref="T:System.Net.Security.SslStream" /> authentication.</summary>
+        <value>The collection of cipher suites allowed in this policy for negotiation.</value>
+        <remarks>
+          <format type="text/markdown">
```
mairaw:
```suggestion
          <format type="text/markdown"><![CDATA[
```

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -65,9 +75,19 @@
...
+  <![CDATA[
```
mairaw:
```suggestion

## Remarks

```

xml/System.Net.Security/TlsCipherSuite.xml
krwq: it is default when you do `default(TlsCipherSuite)` solely because enums always default to 0 which happens to map to this cipher suite. For the AllowedCipherSuites you will get whatever you pass in and there is only `custom` or `OS default` options

xml/System.Net.Security/SslStream.xml
mairaw:
```suggestion

This property gets the cipher suite that is going to be used in the communication between a client and server by the underlying SSL/TLS implementation. This property is only available after a successful call to <xref:System.Net.Security.SslStream.AuthenticateAsServer%2A> or <xref:System.Net.Security.SslStream.AuthenticateAsClient%2A>.
```

xml/System.Net.Security/SslStream.xml
mairaw: This rule hasn't been enforced in the past but I think it increases readability a lot if we can avoid mentioning types and use more text instead.

xml/System.Net.Security/SslStream.xml
mairaw: It's when you want to link to a method in general and not a specific overload. More info at https://github.com/dotnet/docs/blob/master/styleguide/template.md#links-to-apis

xml/System.Net.Security/CipherSuitesPolicy.xml
Jozkee: What about this? 
Also, in the Remarks section, we may want to explain what's the criteria used to filter `allowedCipherSuites`.
```suggestion
        <summary>Gets the subset of valid cipher suites passed into <see cref="M:System.Net.Security.CipherSuitesPolicy.#ctor(System.Collections.Generic.IEnumerable{System.Net.Security.TlsCipherSuite})" /> constructor.</summary>
```

xml/System.Net.Security/TlsCipherSuite.xml
bartonjs:
```suggestion
        <summary>Represents the TLS_NULL_WITH_NULL_NULL cipher suite.</summary>
```

xml/System.Net.Security/CipherSuitesPolicy.xml
Jozkee:
```suggestion
> Defining a cipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS of the ability to decide which are the best cipher suites to negotiate with and will need of your attention to manually check and update this code. It is strongly recommended to avoid the use of this property and rely on your constantly updated OS policy.
```

xml/System.Net.Security/TlsCipherSuite.xml
Jozkee:
```suggestion
Both field names and values are based on the [TLS Cipher Suites list](https://www.iana.org/assignments/tls-parameters/tls-parameters.xml#tls-parameters-4) from the Internet Assigned Numbers Authority (IANA).
This enumeration represents values that were known at the time a specific version of .NET was released. Unmapped values may be returned by properties or methods, and methods should generally not verify that a given input was defined within the enum, as the value may be transparently passed from or to the system underlying implementation.
Inclusion in this enumeration does not indicate that a .NET platform supports a particular cipher suite.
```

Jozkee: > We might also want a sentence like
> 
> ```
> This enumeration represents values that were known at the time a specific version of .NET was released.  
> Unmapped values may be returned by properties or methods, and methods should generally not verify that a given input was defined within the enum, as the value may be transparently passed from or to the system TLS libraries.
> ```

I would replace `system TLS libraries` for `system underlying TLS implementation`

Jozkee:
@mairaw @rpetrusha @bartonjs @krwq @carlossanlop 
I committed the suggestions a few days ago, this should be ready for re-review or probably merge.

krwq:
@Jozkee I haven't seen corefx PR updating xml doc comments

Jozkee:
@krwq I will do it soon, but that shouldn't stop this PR from merging.
@mairaw @rpetrusha can we merge this?

carlossanlop:
@mairaw @rpetrusha can we get this PR merged?

rpetrusha:
I'll review it quickly now.

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -39,9 +39,20 @@
         <Parameter Name="allowedCipherSuites" Type="System.Collections.Generic.IEnumerable&lt;System.Net.Security.TlsCipherSuite&gt;" />
       </Parameters>
       <Docs>
-        <param name="allowedCipherSuites">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="allowedCipherSuites">The collection of cipher suites allowed in this policy for negotiation.</param>
+        <summary>Initializes a new instance of the <see cref="T:System.Net.Security.CipherSuitesPolicy" /> class that can be used on a client or server authentication.</summary>
```
rpetrusha:
```suggestion
        <summary>Initializes a new instance of the <see cref="T:System.Net.Security.CipherSuitesPolicy" /> class that can be used on a client or server for authentication.</summary>
```

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -39,9 +39,20 @@
...
+> [!NOTE]
+> Defining a cipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS of the ability to decide which are the best cipher suites to negotiate with and will need of your attention to manually check and update this code. It is strongly recommended to avoid the use of this property and rely on your constantly updated OS policy.
+
+## Remarks
+
+Unlike the OS underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in `allowedCipherSuites` doesn't guarantee their priority in the client-server negotiation. You can only initialize an instance of this class if you're using Linux systems with OpenSSL 1.1.1 or higher or macOS .
+
+ ]]></format>
+        </remarks>
+        <exception cref="T:System.PlatformNotSupportedException">Platform is not a Linux system with OpenSSL 1.1.1 or higher or a macOS.</exception>
```
rpetrusha: Is this the right wording? If the exception can be thrown on macOS for have an earlier version than OpenSSL 1.1.1,f it should read, "Platform is not a Linux system or macOS with OpenSSL 1.1.1 or higher." The same question applies to the Remarks section.

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -39,9 +39,20 @@
...
       <Docs>
-        <param name="allowedCipherSuites">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="allowedCipherSuites">The collection of cipher suites allowed in this policy for negotiation.</param>
+        <summary>Initializes a new instance of the <see cref="T:System.Net.Security.CipherSuitesPolicy" /> class that can be used on a client or server authentication.</summary>
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+> [!NOTE]
+> Defining a cipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS of the ability to decide which are the best cipher suites to negotiate with and will need of your attention to manually check and update this code. It is strongly recommended to avoid the use of this property and rely on your constantly updated OS policy.
```
rpetrusha:
```suggestion
> Defining a cipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS from deciding which are the best cipher suites to negotiate with and requires that you manually check and update this code. We strongly recommend that you avoid using this feature and rely on your constantly updated OS policy.
```

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -39,9 +39,20 @@
...
+
+## Remarks
+
+Unlike the OS underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in `allowedCipherSuites` doesn't guarantee their priority in the client-server negotiation. You can only initialize an instance of this class if you're using Linux systems with OpenSSL 1.1.1 or higher or macOS .
```
rpetrusha:
```suggestion
Unlike the operating system's underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in `allowedCipherSuites` doesn't guarantee their priority in client-server negotiation. You can only initialize an instance of this class if you're using Linux systems with OpenSSL 1.1.1 or higher or macOS .
```

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -65,10 +76,22 @@
...
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <summary>Gets a subset of valid cipher suites passed into <see cref="M:System.Net.Security.CipherSuitesPolicy.#ctor(System.Collections.Generic.IEnumerable{System.Net.Security.TlsCipherSuite})" /> constructor as a collection of cipher suites allowed in this policy for negotiation.</summary>
+        <value>The collection of cipher suites allowed in this policy for negotiation.</value>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
```
rpetrusha: The same comments that applied to the constructor apply here as well.

xml/System.Net.Security/SslStream.xml
```diff
@@ -3453,9 +3453,18 @@ Authentication has not occurred.</exception>
...
-        <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <summary>Gets the cipher suite which was negotiated for this connection.</summary>
+        <value>One of the enumeration values that identifies the cipher suite which was negotiated for this connection.</value>
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+  
+## Remarks
+
+This property gets the cipher suite that is going to be used in the communication between a client and server by the underlying SSL/TLS implementation. This property is only available after a successful call to <xref:System.Net.Security.SslStream.AuthenticateAsServer%2A> or <xref:System.Net.Security.SslStream.AuthenticateAsClient%2A>.
```
rpetrusha:
```suggestion
This property gets the cipher suite that is going to be used in the communication between a client and server by the underlying SSL/TLS implementation. It is only available after a successful call to <xref:System.Net.Security.SslStream.AuthenticateAsServer%2A> or <xref:System.Net.Security.SslStream.AuthenticateAsClient%2A>.
```

xml/System.Net.Security/TlsCipherSuite.xml
```diff
@@ -18,8 +18,17 @@
...
   <Docs>
-    <summary>To be added.</summary>
-    <remarks>To be added.</remarks>
+    <summary>Represents cipher suite values for the TLS (formerly SSL) protocol.</summary>
+    <remarks>
+          <format type="text/markdown"><![CDATA[  
+  
+## Remarks
+Both field names and values are based on the [TLS Cipher Suites list](https://www.iana.org/assignments/tls-parameters/tls-parameters.xml#tls-parameters-4) from the Internet Assigned Numbers Authority (IANA).
+This enumeration represents values that were known at the time a specific version of .NET was released. Unmapped values may be returned by properties or methods, and methods should generally not verify that a given input was defined within the enum, as the value may be transparently passed from or to the system underlying TLS implementation.
```
rpetrusha:
```suggestion
This enumeration represents values that were known at the time a specific version of .NET was released. Unmapped values may be returned by properties or methods, and methods should generally not verify that a given input was defined within the enum, as the value may be transparently passed from or to the system and the underlying TLS implementation.
```

xml/System.Net.Security/TlsCipherSuite.xml
```diff
@@ -59,7 +68,7 @@
       </ReturnValue>
       <MemberValue>4868</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Represents the TLS_AES_128_CCM_SHA256 cipher suite.</summary>
```
rpetrusha: Although these descriptions are OK as it, it would be nice to shorten them by removing "Represents the ".

xml/System.Net.Security/CipherSuitesPolicy.xml
Jozkee: We actually try to say (Linux with Open SSL 1.1.1) or (macOS). 
`Open SSL v1.1.1` is only required on linux; please see https://github.com/dotnet/dotnet-api-docs/pull/2822#discussion_r307029105.

rpetrusha: It is the right wording, then. Thanks for clarifying, @Jozkee.

xml/System.Net.Security/CipherSuitesPolicy.xml
Jozkee: I don't think `for` is missing here, since I am trying to refer to `client authentication` and `server authentication` mechanisms.

xml/System.Net.Security/CipherSuitesPolicy.xml
Jozkee: I am taking your suggestion but I also spotted a space before the final dot.
```suggestion
Unlike the operating system's underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in `allowedCipherSuites` doesn't guarantee their priority in client-server negotiation. You can only initialize an instance of this class if you're using Linux systems with OpenSSL 1.1.1 or higher or macOS.

```

xml/System.Net.Security/TlsCipherSuite.xml
Jozkee: I was trying to refer to the `underlying TLS implementation` within the system.
This might be more appropriate perhaps?

```suggestion
This enumeration represents values that were known at the time a specific version of .NET was released. Unmapped values may be returned by properties or methods, and methods should generally not verify that a given input was defined within the enum, as the value may be transparently passed from or to the system's underlying TLS implementation.
```

xml/System.Net.Security/TlsCipherSuite.xml
carlossanlop: @Jozkee used this same description in a template in dotnet/corefx, so modifying these descriptions will mean modifying the ones in the template too.

If it's not blocking, I'd say we keep the current description as it is now.

xml/System.Net.Security/CipherSuitesPolicy.xml
rpetrusha: In that case, "for client or server authentication" might be better.

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -65,10 +76,22 @@
...
+
+> [!NOTE]
+> Defining a cipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS of the ability to decide which are the best cipher suites to negotiate with and will need of your attention to manually check and update this code. It is strongly recommended to avoid the use of this property and rely on your constantly updated OS policy.
```
Jozkee:
```suggestion
> Defining a cipher suite policy on <xref:System.Net.Security.SslStream> authentication will prevent the OS from deciding which are the best cipher suites to negotiate with and requires that you manually check and update this code. We strongly recommend that you avoid using this feature and rely on your constantly updated OS policy.
```

xml/System.Net.Security/CipherSuitesPolicy.xml
```diff
@@ -65,10 +76,22 @@
...
+
+## Remarks
+Unlike the OS underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in this property does not determine their priority in the client-server negotiation. It is required to use OSX or Unix-based systems with OpenSSL 1.1.1 in order to initialize an instance of this class.
```
Jozkee:
```suggestion
Unlike the operating system's underlying SSL/TLS implementation, the order of the <xref:System.Net.Security.TlsCipherSuite> elements in `allowedCipherSuites` doesn't guarantee their priority in client-server negotiation. You can only initialize an instance of this class if you're using Linux systems with OpenSSL 1.1.1 or higher or macOS.
```

xml/System.Net.Security/TlsCipherSuite.xml
rpetrusha: Yes.

xml/System.Net.Security/CipherSuitesPolicy.xml
Jozkee: Remorks section starts 5 lines below.
```suggestion
## Remarks
```

xml/System.Net.Security/TlsCipherSuite.xml
rpetrusha: That's fine.

xml/System.Net.Security/CipherSuitesPolicy.xml
Jozkee: All right, that sounds good to me.

```suggestion
        <summary>Initializes a new instance of the <see cref="T:System.Net.Security.CipherSuitesPolicy" /> class that can be used for client or server authentication.</summary>
```

Jozkee:
@rpetrusha I just addressed your latest suggestions. I think this should be ready to merge once the build passes.

carlossanlop:
@rpetrusha the build passed. Is this good to merge?

rpetrusha:
Yes, this is ready to merge, @Jozkee and @carlossanlop. I'll do it now.

PR #2840, opened 2019-07-26
Recommendations for SHA1.

TimShererWithAquent:
@Thraka  

Adding this text or similar where relevant:

Due to collision problems with SHA1, Microsoft recommends SHA256.

Fixes https://dev.azure.com/mseng/TechnicalContent/_workitems/edit/1572540

Andy - I went through eight topics from the list of about 90 as a pilot. Although we have worked on an item like this before, these xml files were not like the ones in the previous batch. Some of these had narrative sections that were appropriate for changes. 

I checked with the team previously and nobody here has worked on this kind of xml formatted files, so please evaluate carefully whether I'm addressing these the right way.

### DIFF
--- a/xml/System.ServiceModel.Security/Basic128SecurityAlgorithmSuite.xml
+++ b/xml/System.ServiceModel.Security/Basic128SecurityAlgorithmSuite.xml
@@ -121,7 +121,7 @@
       <Docs>
         <summary>Gets the default digest algorithm, Sha1Digest.</summary>
         <value>The default digest algorithm, Sha1Digest.</value>
-        <remarks>Due to collision problems with SHA1, Microsoft recommends SHA256.</remarks>
+        <remarks>Due to collision problems with SHA1, Microsoft recommends a security model based on SHA256 or better.</remarks>
       </Docs>
     </Member>
     <Member MemberName="DefaultEncryptionAlgorithm">
@@ -247,7 +247,7 @@
       <Docs>
         <summary>Gets the default symmetric signature algorithm, HmacSha1Signature.</summary>
         <value>The default symmetric signature algorithm, HmacSha1Signature.</value>
-        <remarks>Due to collision problems with SHA1, Microsoft recommends SHA256.</remarks>
+        <remarks>Due to collision problems with SHA1, Microsoft recommends a security model based on SHA256 or better.</remarks>
       </Docs>
     </Member>
     <Member MemberName="IsAsymmetricKeyLengthSupported">


xml/System.Security.Cryptography/MD5.xml
```diff
@@ -47,7 +47,7 @@
...
   
 > [!NOTE]
->  Newer hash functions, such as the Secure Hash Algorithms SHA-256 and SHA-512, are available. Consider using the <xref:System.Security.Cryptography.SHA256> class or the <xref:System.Security.Cryptography.SHA512> class instead of the <xref:System.Security.Cryptography.MD5> class. Use <xref:System.Security.Cryptography.MD5> only for compatibility with legacy applications and data.  
+>  Due to collision problems with MD5/SHA1, Microsoft recommends SHA256 or and SHA512. Consider using the <xref:System.Security.Cryptography.SHA256> class or the <xref:System.Security.Cryptography.SHA512> class instead of the <xref:System.Security.Cryptography.MD5> class. Use <xref:System.Security.Cryptography.MD5> only for compatibility with legacy applications and data.
```
bartonjs:
```suggestion
>  Due to collision problems with MD5/SHA1, Microsoft recommends SHA256 or SHA512. Consider using the <xref:System.Security.Cryptography.SHA256> class or the <xref:System.Security.Cryptography.SHA512> class instead of the <xref:System.Security.Cryptography.MD5> class. Use <xref:System.Security.Cryptography.MD5> only for compatibility with legacy applications and data.  
```

xml/System.ServiceModel.Security/Basic128SecurityAlgorithmSuite.xml
```diff
@@ -79,7 +79,7 @@
       <Docs>
         <summary>Gets the default asymmetric signature algorithm, RsaSha1Signature.</summary>
         <value>The default asymmetric signature algorithm, RsaSha1Signature.</value>
-        <remarks>To be added.</remarks>
+        <remarks>Due to collision problems with SHA1, Microsoft recommends SHA256.</remarks>
```
bartonjs: "SHA256" isn't a legal input here.  Presumably it would be something like `RsaSha256Signature`; but I don't know what type this is.  Maybe it's fine as a remark, but it seems like the extra bit of followup would be helpful)

bartonjs: (This applies similarly to the other changes in this file)

Thraka: What about changing it to read 

```suggestion
        <remarks>Due to collision problems with SHA1, Microsoft recommends a security model based on SHA256 or better.</remarks>
```

This would make it not specifically talk about "this is SHA1, put in a SHA256-type value" and instead just notes that this entire type is insecure?

Thraka: @bartonjs Does this suggestion make sense?

bartonjs: I like "Due to collision problems with SHA1, Microsoft recommends a security model based on SHA256 or better.", yeah.  But it should also be done to the HMAC entry (the third change in this file).

Thraka: Should we just do it for all of them?

TimShererWithAquent:
@Thraka 

I implemented the change discussed in all three places in that file. This was the language about "security model". 

How should we deal with this going forward? I'm not sure how I can tell from down in the API xml comments whether there is an option for SHA256. I worry that putting "security model" in the boilerplate text will lead some users to stop wonder if they have to set up a security model somewhere.

Thraka:
I think with reference we'll have to infer the name of the method/property to determine if it's something that a user can choose (like i want md5 or sha256) or just (as in this case) describing something that actually IS related to 128/md5. For example, the file Basic128SecurityAlgorithmSuite.xml  is obviously all about md5/sha128 so there is probably no option in there for a sha256.

PR #2842, opened 2019-07-26
Document System.IO.Pipelines Platform Extensions

carlossanlop:
Found some System.IO APIs that were not under the netcore-3.0 moniker but were under dotnet-plat-ext-3.0.

@JeremyKuhne and I are owners.

@davidfowl wrote most of the code here. Can you please take a quick look?

API Proposal: https://github.com/dotnet/corefx/issues/27246
API PR: https://github.com/dotnet/corefx/pull/35212

Source code files:
https://github.com/dotnet/corefx/blob/master/src/System.IO.Pipelines/src/System/IO/Pipelines/StreamPipeReaderOptions.cs
https://github.com/dotnet/corefx/blob/master/src/System.IO.Pipelines/src/System/IO/Pipelines/StreamPipeWriterOptions.cs
https://github.com/dotnet/corefx/blob/master/src/System.IO.Pipelines/src/System/IO/Pipelines/StreamPipeExtensions.cs
https://github.com/dotnet/corefx/blob/master/src/System.IO.Pipelines/src/System/IO/Pipelines/PipeReader.cs
https://github.com/dotnet/corefx/blob/master/src/System.IO.Pipelines/src/System/IO/Pipelines/PipeWriter.cs

### DIFF
--- a/xml/System.IO.Pipelines/PipeWriter.xml
+++ b/xml/System.IO.Pipelines/PipeWriter.xml
@@ -91,7 +91,7 @@
         <Parameter Name="leaveOpen" Type="System.Boolean" Index="0" FrameworkAlternate="dotnet-plat-ext-3.0" />
       </Parameters>
       <Docs>
-        <param name="leaveOpen">An optional flag that indicates whether disposing the returned <see cref="T:System.IO.Stream"/> leaves <see cref="T:System.IO.Pipelines.PipeReader" /> open (<see langword="true" /> or completes <see cref="T:System.IO.Pipelines.PipeReader" /> (<see langword="false" />.</param>
+        <param name="leaveOpen">An optional flag that indicates whether disposing the returned <see cref="T:System.IO.Stream"/> leaves <see cref="T:System.IO.Pipelines.PipeReader" /> open (<see langword="true" />) or completes <see cref="T:System.IO.Pipelines.PipeReader" /> (<see langword="false" />).</param>
         <summary> Returns the underlying <see cref="T:System.IO.Stream" /> wrapped by the <see cref="T:System.IO.Pipelines.PipeWriter" />.</summary>
         <returns>The underlying stream.</returns>
         <remarks>To be added.</remarks>


xml/System.IO.Pipelines/PipeReader.xml
```diff
@@ -114,9 +114,9 @@
         <Parameter Name="leaveOpen" Type="System.Boolean" Index="0" FrameworkAlternate="dotnet-plat-ext-3.0" />
       </Parameters>
       <Docs>
-        <param name="leaveOpen">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="leaveOpen">Optional flag indicating disposing returned <see cref="T:System.IO.Stream"/> won't complete <see cref="T:System.IO.Pipelines.PipeReader" />.</param>
```
rpetrusha:
```suggestion
        <param name="leaveOpen">An optional flag that indicates whether disposing the returned <see cref="T:System.IO.Stream"/> leaves <see cref="T:System.IO.Pipelines.PipeReader" /> open (<see langword="true" /> or completes <see cref="T:System.IO.Pipelines.PipeReader" /> (<see langword="false" />.</param>
```

xml/System.IO.Pipelines/PipeReader.xml
```diff
@@ -264,10 +264,19 @@
         <Parameter Name="state" Type="System.Object" />
       </Parameters>
       <Docs>
-        <param name="callback">To be added.</param>
-        <param name="state">To be added.</param>
-        <summary>Cancels the pending <see cref="M:System.IO.Pipelines.PipeReader.ReadAsync(System.Threading.CancellationToken)" /> operation. If there is none, cancels next <see cref="M:System.IO.Pipelines.PipeReader.ReadAsync(System.Threading.CancellationToken)" /> operation, without completing the <see cref="T:System.IO.Pipelines.PipeWriter" />.</summary>
-        <remarks>To be added.</remarks>
+        <param name="callback">The callback to register.</param>
+        <param name="state">The state object to pass to <paramref name="callback" /> when it's invoked.</param>
+        <summary>Registers a callback that gets executed when the <see cref="T:System.IO.Pipelines.PipeWriter" /> side of the pipe is completed.</summary>
```
rpetrusha:
```suggestion
        <summary>Registers a callback that executes when the <see cref="T:System.IO.Pipelines.PipeWriter" /> side of the pipe is completed.</summary>
```

xml/System.IO.Pipelines/PipeReader.xml
```diff
@@ -264,10 +264,19 @@
...
+         <remarks>
+           <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+> [!IMPORTANT]
+> `OnWriterCompleted` may not be invoked on all implementations of <xref:System.IO.Pipelines.PipeWriter>. This will be removed in a future release.
```
rpetrusha:
```suggestion
> `OnWriterCompleted` may not be invoked on all implementations of <xref:System.IO.Pipelines.PipeWriter>. This limitation will be removed in a future release.
```
Here and below, can we specify what implementations don't support it? If it throws PlatformNotSupportedException or NotImplementedException, that should be noted in exceptions.

xml/System.IO.Pipelines/PipeWriter.xml
```diff
@@ -91,9 +91,9 @@
         <Parameter Name="leaveOpen" Type="System.Boolean" Index="0" FrameworkAlternate="dotnet-plat-ext-3.0" />
       </Parameters>
       <Docs>
-        <param name="leaveOpen">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="leaveOpen">Optional flag indicating disposing returned <see cref="T:System.IO.Stream"/> won't complete <see cref="T:System.IO.Pipelines.PipeWriter" />.</param>
```
rpetrusha:
```suggestion
        <param name="leaveOpen">An optional flag that indicates whether disposing the returned <see cref="T:System.IO.Stream"/> leaves <see cref="T:System.IO.Pipelines.PipeReader" /> open (<see langword="true" /> or completes <see cref="T:System.IO.Pipelines.PipeReader" /> (<see langword="false" />.</param>
```

xml/System.IO.Pipelines/PipeWriter.xml
```diff
@@ -323,10 +323,19 @@ You must request a new buffer after calling <xref:System.IO.Pipelines.PipeWriter
         <Parameter Name="state" Type="System.Object" />
       </Parameters>
       <Docs>
-        <param name="callback">To be added.</param>
-        <param name="state">To be added.</param>
+        <param name="callback">The callback to register.</param>
+        <param name="state">The state object to pass to <paramref name="callback" /> when it's invoked.</param>
         <summary>Registers a callback that gets executed when the <see cref="T:System.IO.Pipelines.PipeReader" /> side of the pipe is completed.</summary>
```
rpetrusha:
```suggestion
        <summary>Registers a callback that executes when the <see cref="T:System.IO.Pipelines.PipeReader" /> side of the pipe is completed.</summary>
```

xml/System.IO.Pipelines/PipeWriter.xml
```diff
@@ -323,10 +323,19 @@ You must request a new buffer after calling <xref:System.IO.Pipelines.PipeWriter
...
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+> [!IMPORTANT]
+> `OnReaderCompleted` may not be invoked on all implementations of <xref:System.IO.Pipelines.PipeWriter>. This will be removed in a future release.
```
rpetrusha:
```suggestion
> `OnReaderCompleted` may not be invoked on all implementations of <xref:System.IO.Pipelines.PipeWriter>. This limitation will be removed in a future release.
```

xml/System.IO.Pipelines/PipeWriter.xml
```diff
@@ -349,10 +358,10 @@ You must request a new buffer after calling <xref:System.IO.Pipelines.PipeWriter
         <Parameter Name="cancellationToken" Type="System.Threading.CancellationToken" />
       </Parameters>
       <Docs>
-        <param name="source">To be added.</param>
-        <param name="cancellationToken">To be added.</param>
-        <summary>Writes <paramref name="source" /> to the pipe and makes data accessible to <see cref="T:System.IO.Pipelines.PipeReader" /></summary>
-        <returns>To be added.</returns>
+        <param name="source">The read-only byte span to write.</param>
+        <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="T:System.Threading.CancellationToken.None"/>.</param>
+        <summary>Writes <paramref name="source" /> to the pipe and makes data accessible to <see cref="T:System.IO.Pipelines.PipeReader" />.</summary>
```
rpetrusha:
```suggestion
        <summary>Writes the specified byte memory range to the pipe and makes data accessible to the <see cref="T:System.IO.Pipelines.PipeReader" />.</summary>
```

xml/System.IO.Pipelines/PipeWriter.xml
```diff
@@ -349,10 +358,10 @@ You must request a new buffer after calling <xref:System.IO.Pipelines.PipeWriter
         <Parameter Name="cancellationToken" Type="System.Threading.CancellationToken" />
       </Parameters>
       <Docs>
-        <param name="source">To be added.</param>
-        <param name="cancellationToken">To be added.</param>
-        <summary>Writes <paramref name="source" /> to the pipe and makes data accessible to <see cref="T:System.IO.Pipelines.PipeReader" /></summary>
-        <returns>To be added.</returns>
+        <param name="source">The read-only byte span to write.</param>
```
rpetrusha:
```suggestion
        <param name="source">The read-only byte memory region to write.</param>
```

xml/System.IO.Pipelines/PipeWriter.xml
```diff
@@ -349,10 +358,10 @@ You must request a new buffer after calling <xref:System.IO.Pipelines.PipeWriter
...
+        <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="T:System.Threading.CancellationToken.None"/>.</param>
+        <summary>Writes <paramref name="source" /> to the pipe and makes data accessible to <see cref="T:System.IO.Pipelines.PipeReader" />.</summary>
+        <returns>A task that represents the asynchronous write operation, and wraps the flush asynchronous operation.</returns>
```
rpetrusha: Is this correct? Does it both write and flush?

xml/System.IO.Pipelines/StreamPipeExtensions.xml
```diff
@@ -14,7 +14,7 @@
   </Base>
   <Interfaces />
   <Docs>
-    <summary>To be added.</summary>
+    <summary>Provides extension methods for <see cref="T:System.IO.Stream" /> that allow performing read and write operations directly into pipes.</summary>
```
rpetrusha:
```suggestion
    <summary>Provides extension methods for <see cref="T:System.IO.Stream" /> that support read and write operations directly into pipes.</summary>
```

xml/System.IO.Pipelines/StreamPipeReaderOptions.xml
```diff
@@ -36,11 +36,11 @@
         <Parameter Name="leaveOpen" Type="System.Boolean" />
       </Parameters>
       <Docs>
-        <param name="pool">To be added.</param>
-        <param name="bufferSize">To be added.</param>
-        <param name="minimumReadSize">To be added.</param>
-        <param name="leaveOpen">To be added.</param>
-        <summary>To be added.</summary>
+        <param name="pool">The <see cref="T:System.Buffers.MemoryPool%2A" /> to use when allocating memory. The default value is <see langword="null" />.</param>
```
rpetrusha:
```suggestion
        <param name="pool">The memory pool to use when allocating memory. The default value is <see langword="null" />.</param>
```

xml/System.IO.Pipelines/StreamPipeReaderOptions.xml
```diff
@@ -36,11 +36,11 @@
...
+        <param name="bufferSize">The minimum buffer size to use when renting memory from the <paramref name="pool" />. The default value is 4096.</param>
+        <param name="minimumReadSize">The threshold of remaining bytes int he buffer before a new buffer is allocated. The default value is 1024.</param>
```
rpetrusha:
```suggestion
        <param name="minimumReadSize">The threshold of remaining bytes in the buffer before a new buffer is allocated. The default value is 1024.</param>
```

xml/System.IO.Pipelines/StreamPipeReaderOptions.xml
```diff
@@ -36,11 +36,11 @@
...
+        <param name="leaveOpen">Set to <see langword="true" /> to leave the underlying stream open after the <see cref="T:System.IO.Pipelines.PipeReader" /> completes, <see langword="false" /> to close it. The default is <see langword="false" />.</param>
```
rpetrusha:
```suggestion
        <param name="leaveOpen"><see langword="true" /> to leave the underlying stream open after the <see cref="T:System.IO.Pipelines.PipeReader" /> completes; <see langword="false" /> to close it. The default is <see langword="false" />.</param>
```

xml/System.IO.Pipelines/StreamPipeReaderOptions.xml
```diff
@@ -36,11 +36,11 @@
...
+        <summary>Initializes a <see cref="T:System.IO.Pipelines.StreamPipeReaderOptions" /> instance, optionally allowing to provide a memory pool, a minimum buffer size, a minimum read size, and whether the underlying stream should be left open after the <see cref="T:System.IO.Pipelines.PipeReader" /> completes.</summary>
```
rpetrusha:
```suggestion
        <summary>Initializes a <see cref="T:System.IO.Pipelines.StreamPipeReaderOptions" /> instance, optionally specifying a memory pool, a minimum buffer size, a minimum read size, and whether the underlying stream should be left open after the <see cref="T:System.IO.Pipelines.PipeReader" /> completes.</summary>
```

xml/System.IO.Pipelines/StreamPipeReaderOptions.xml
```diff
@@ -81,8 +81,8 @@
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Gets the value that indicates if the underlying stream should be left open after the <see cref="T:System.IO.Pipelines.PipeReader" /> completes.</summary>
+        <value>A boolean that indicates if the underlying stream should be left open after the <see cref="T:System.IO.Pipelines.PipeReader" /> completes.</value>
```
rpetrusha:
```suggestion
        <value><see langword="true" /> if the underlying stream should be left open after the <see cref="T:System.IO.Pipelines.PipeReader" /> completes; otherwise, <see langword="false" />.</value>
```

xml/System.IO.Pipelines/StreamPipeWriterOptions.xml
```diff
@@ -35,10 +35,10 @@
         <Parameter Name="leaveOpen" Type="System.Boolean" />
       </Parameters>
       <Docs>
-        <param name="pool">To be added.</param>
-        <param name="minimumBufferSize">To be added.</param>
-        <param name="leaveOpen">To be added.</param>
-        <summary>To be added.</summary>
+        <param name="pool">The <see cref="T:System.Buffers.MemoryPool%2A" /> to use when allocating memory. The default value is <see langword="null" />.</param>
```
rpetrusha:
```suggestion
        <param name="pool">The memory pool to use when allocating memory. The default value is <see langword="null" />.</param>
```

xml/System.IO.Pipelines/StreamPipeWriterOptions.xml
```diff
@@ -35,10 +35,10 @@
...
+        <param name="minimumBufferSize">The minimum buffer size to use when renting memory from the <paramref name="pool" />. The default value is 4096.</param>
+        <param name="leaveOpen">Set to <see langword="true" /> to leave the underlying stream open after the <see cref="T:System.IO.Pipelines.PipeWriter" /> completes, <see langword="false" /> to close it. The default is <see langword="false" />.</param>
```
rpetrusha:
```suggestion
        <param name="leaveOpen"><see langword="true" /> to leave the underlying stream open after the <see cref="T:System.IO.Pipelines.PipeWriter" /> completes; <see langword="false" /> to close it. The default is <see langword="false" />.</param>
```

xml/System.IO.Pipelines/StreamPipeWriterOptions.xml
```diff
@@ -35,10 +35,10 @@
...
+        <summary>Initializes a <see cref="T:System.IO.Pipelines.StreamPipeWriterOptions" /> instance, optionally allowing to provide a memory pool, a minimum buffer size, and whether the underlying stream should be left open after the <see cref="T:System.IO.Pipelines.PipeWriter" /> completes.</summary>
```
rpetrusha:
```suggestion
        <summary>Initializes a <see cref="T:System.IO.Pipelines.StreamPipeWriterOptions" /> instance, optionally specifying a memory pool, a minimum buffer size, and whether the underlying stream should be left open after the <see cref="T:System.IO.Pipelines.PipeWriter" /> completes.</summary>
```

xml/System.IO.Pipelines/StreamPipeWriterOptions.xml
```diff
@@ -58,8 +58,8 @@
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Gets the value that indicates if the underlying stream should be left open after the <see cref="T:System.IO.Pipelines.PipeWriter" /> completes.</summary>
+        <value>A boolean that indicates if the underlying stream should be left open after the <see cref="T:System.IO.Pipelines.PipeWriter" /> completes.</value>
```
rpetrusha:
```suggestion
        <value><see langword="true" /> if the underlying stream should be left open after the <see cref="T:System.IO.Pipelines.PipeWriter" /> completes; otherwise, <see langword="false" />.</value>
```

davidfowl:
Are these preview7 docs?

xml/System.IO.Pipelines/PipeReader.xml
carlossanlop: @rpetrusha:

Here is [the source code](https://github.com/dotnet/corefx/blob/51cd78434953683040f93e69ed2a889e476f5cac/src/System.IO.Pipelines/src/System/IO/Pipelines/PipeWriter.cs) for this function.

I looked at the git blame for the "Obsolete" attribute line, and it was added [11 days ago in this PR](https://github.com/dotnet/corefx/pull/39526/files). That change changed both the `OnWriterCompleted` and `OnReaderCompleted` methods from abstract to virtual, added an empty body, and added the Obsolete attribute on top of both methods.

But the actual function was added 2 years ago [in this commit](https://github.com/dotnet/corefx/commit/5573f26fc34ba040b03f022a5892a5e172573002). So the function is now new, but I suppose that since it was abstract, and now it's virtual, is the reason why it showed up in the Docs spreadsheet with the dotnet-plat-ext-3.0 moniker (but I'm not entirely sure this is the actual reason).

Based on this, would this message make more sense?:

```suggestion
> `OnWriterCompleted` may not be invoked on .NET Core 3.0 implementations of <xref:System.IO.Pipelines.PipeWriter>. This limitation will be removed in a future release.
```

@davidfowl can you chime in, since you made the change to this API? Am I understanding this correctly?

xml/System.IO.Pipelines/PipeWriter.xml
carlossanlop: See my comment for OnWriterCompleted.

xml/System.IO.Pipelines/PipeWriter.xml
carlossanlop: That seems to be the case. If you look at the code, it writes, then asynchronously flushes, and returns the flush task:

https://github.com/dotnet/corefx/blame/bb0516a4f8d0c4d84eff2ba2a3ad0a4764ece9d2/src/System.IO.Pipelines/src/System/IO/Pipelines/PipeWriter.cs#L98

carlossanlop:
@davidfowl they are meant to be used for when we release 3.0. These should be in before August 16th so we can also use them for IntelliSense. Is there any concern? Or why do you ask?

davidfowl:
There are some missing APIs `CompleteAsync` but those are in preview8. I also changed the doc comment for Complete and CompleteAsync in the same change that added CompleteAsync.

xml/System.IO.Pipelines/PipeReader.xml
davidfowl: No that's not correct. We actually want to stop people from using this API because it cannot be implemented in some cases.

carlossanlop: Understood. @davidfowl so would you prefer us to keep the original sentence? Would you like us to reword it a bit?

davidfowl: Feel free to word smith it a bit. That was the best i could come up with 😄

carlossanlop: Okay, so we should not use `limitation` as it was originally suggested. If the intention is to stop people using the API, I think this would clarify the last sentence:

```suggestion
> `OnWriterCompleted` may not be invoked on all implementations of <xref:System.IO.Pipelines.PipeWriter>. This method will be removed in a future release.
```

xml/System.IO.Pipelines/PipeWriter.xml
carlossanlop: I think we should use "method" instead of "limitation" since the idea is to stop using that API altogether:

```suggestion
> `OnReaderCompleted` may not be invoked on all implementations of <xref:System.IO.Pipelines.PipeWriter>. This method will be removed in a future release.
```

xml/System.IO.Pipelines/PipeReader.xml
davidfowl: Yep

rpetrusha:
Is this ready to merge, @carlossanlop?

xml/System.IO.Pipelines/PipeReader.xml
rpetrusha: I think that this could be made stronger, @carlossanlop and @davidfowl. First, it could go in an IMPORTANT note box. And it could read, "We recommend that you not call this method, since it may not be invoked on all implementations of \<xref:System.IO.PipeLines.PipeWriter>. The method will be removed in a future release of .NET Core.

carlossanlop: @rpetrusha I like your suggestion, it's stronger as recommended, while preserving the original idea @davidfowl intended in the comment. If David's ok with the change, we should change it:

```suggestion
We recommend that you not call this method, since it may not be invoked on all implementations of <xref:System.IO.PipeLines.PipeWriter>. The method will be removed in a future release of .NET Core.
```

PR #2872, opened 2019-07-30
Document System.Reflection.Metadata.ImageFormatLimitationException

carlossanlop:
Documenting the 3.0 serialization API added recently by @joperezr.
Added also a general summary. There were no comments in the source, and the exception is older than .NET Core, so I based my summary on the way it's being used in corefx. The area owners are @tmat, @nguerrera. Can you please help make sure the summary is correct?

### DIFF
--- a/xml/System.Reflection.Metadata/ImageFormatLimitationException.xml
+++ b/xml/System.Reflection.Metadata/ImageFormatLimitationException.xml
@@ -17,7 +17,7 @@
   </Base>
   <Interfaces />
   <Docs>
-    <summary>The exception that is thrown when the heap size limit is exceeded.</summary>
+    <summary>The exception that is thrown when an attempt to write metadata exceeds a limit given by the format specification. For example, when the heap size limit is exceeded.</summary>
     <remarks>To be added.</remarks>
   </Docs>
   <Members>


xml/System.Reflection.Metadata/ImageFormatLimitationException.xml
```diff
@@ -17,7 +17,7 @@
   </Base>
   <Interfaces />
   <Docs>
-    <summary>To be added.</summary>
+    <summary>The exception that is thrown when the heap size limit is exceeded.</summary>
```
tmat: I'd say more generally, that the exception is thrown when an attempt to write metadata exceeds a limit given by the format specification. Heap size limit is one of the limits.

carlossanlop: Ah great, thanks for the clarification. Does this sound good?:

```suggestion
    <summary>The exception that is thrown when an attempt to write metadata exceeds a limit given by the format specification. For example, when the heap size limit is exceeded.</summary>
```

carlossanlop: I can remove the example at the end if you don't feel like it should be included.

carlossanlop: @tmat I'll merge the above suggestion based on your comment, but feel free to let us know if you'd like us to change it.

PR #2873, opened 2019-07-30
Document the new Pkcs12-Mode enumerations

bartonjs:
cc: @carlossanlop

### DIFF
--- a/xml/System.Security.Cryptography.Pkcs/Pkcs12ConfidentialityMode.xml
+++ b/xml/System.Security.Cryptography.Pkcs/Pkcs12ConfidentialityMode.xml
@@ -14,7 +14,7 @@
     <BaseTypeName>System.Enum</BaseTypeName>
   </Base>
   <Docs>
-    <summary>To be added.</summary>
+    <summary>Represents the kind of encryption associated with a PKCS#12 SafeContents value.</summary>
     <remarks>To be added.</remarks>
   </Docs>
   <Members>
@@ -36,7 +36,7 @@
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The SafeContents value is not encrypted.</summary>
       </Docs>
     </Member>
     <Member MemberName="Password">
@@ -57,7 +57,7 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The SafeContents value is encrypted with a password.</summary>
       </Docs>
     </Member>
     <Member MemberName="PublicKey">
@@ -78,7 +78,7 @@
       </ReturnValue>
       <MemberValue>3</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The SafeContents value is encrypted using public key cryptography.</summary>
       </Docs>
     </Member>
     <Member MemberName="Unknown">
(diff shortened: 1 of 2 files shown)


xml/System.Security.Cryptography.Pkcs/Pkcs12ConfidentialityMode.xml
```diff
@@ -36,7 +36,7 @@
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The SafeContents value is not encrypted.</summary>
```
Thraka: Would this be better `P:System.Security.Cryptography.Pkcs.Pkcs12SafeContentsBag.SafeContents` ? Is this the correct one?
```suggestion
        <summary>The <see cref="P:System.Security.Cryptography.Pkcs.Pkcs12SafeContentsBag.SafeContents"/> value is not encrypted.</summary>
```

xml/System.Security.Cryptography.Pkcs/Pkcs12ConfidentialityMode.xml
```diff
@@ -57,7 +57,7 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The SafeContents value is encrypted with a password.</summary>
```
Thraka:
```suggestion
        <summary>The <see cref="P:System.Security.Cryptography.Pkcs.Pkcs12SafeContentsBag.SafeContents"/> value is encrypted with a password.</summary>
```

xml/System.Security.Cryptography.Pkcs/Pkcs12ConfidentialityMode.xml
```diff
@@ -78,7 +78,7 @@
       </ReturnValue>
       <MemberValue>3</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The SafeContents value is encrypted using public key cryptography.</summary>
```
Thraka:
```suggestion
        <summary>The <see cref="P:System.Security.Cryptography.Pkcs.Pkcs12SafeContentsBag.SafeContents"/> value is encrypted using public key cryptography.</summary>
```

xml/System.Security.Cryptography.Pkcs/Pkcs12ConfidentialityMode.xml
```diff
@@ -99,8 +99,8 @@
       </ReturnValue>
       <MemberValue>0</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>The kind of encryption applied to the SafeContents is unknown or could not be determined.</summary>
```
Thraka:
```suggestion
        <summary>The kind of encryption applied to the <see cref="P:System.Security.Cryptography.Pkcs.Pkcs12SafeContentsBag.SafeContents"/> is unknown or could not be determined.</summary>
```

xml/System.Security.Cryptography.Pkcs/Pkcs12ConfidentialityMode.xml
bartonjs: Currently it's about the type Pkcs12SafeContents, not any particular property (the more likely place is the enumerator from Pkcs12Info).  I don't think the enum members should really reference another type directly in their documentation, because we almost certainly won't remember to come clean them up in a timely manner if the API scope increases (while maintaining the same concept).

Thraka: @bartonjs What about breaking the word so that it reads "safe contents" and therefor doesn't look like a specific property?

bartonjs: It's referencing "SafeContents" data from RFC7292/PKCS#12.

If that seems too weird and we're trying to generalize it then "The data is not encrypted." is probably the best it'll get.

PR #2897, opened 2019-07-31
Adding missing param description to NetworkStream.Write(System.ReadOnlySpan{System.Byte})

jozkee:
After adding documentation for System.Net.Sockets in #2767, we found out that one param went away with no proper description.
This change adds the missing description to it.


Jozkee:
@mairaw @rpetrusha @carlossanlop

PR #2906, opened 2019-08-01
Document Json namespaces

carlossanlop:
As suggested by @ahsonkhan

### DIFF
--- a/xml/ns-System.Text.Json.xml
+++ b/xml/ns-System.Text.Json.xml
@@ -1,6 +1,6 @@
 <Namespace Name="System.Text.Json">
   <Docs>
-    <summary>The System.Text.Json namespace provides high-performance, low-allocating, and RFC-compliant capabilities to process JavaScript Object Notation (JSON), which includes serializing objects to JSON text and deserializing JSON text to objects, with UTF-8 support built-in. It also provides types to read and write JSON text encoded as UTF-8, and to create a read-only in-memory document object model (DOM) for random access of the JSON elements within a structured view of the data.</summary>
+    <summary>The System.Text.Json namespace provides high-performance, low-allocating, and standards-compliant capabilities to process JavaScript Object Notation (JSON), which includes serializing objects to JSON text and deserializing JSON text to objects, with UTF-8 support built-in. It also provides types to read and write JSON text encoded as UTF-8, and to create an in-memory document object model (DOM) for random access of the JSON elements within a structured view of the data.</summary>
     <remarks>To be added.</remarks>
   </Docs>
 </Namespace>


xml/ns-System.Text.Json.xml
```diff
@@ -1,6 +1,6 @@
 <Namespace Name="System.Text.Json">
   <Docs>
-    <summary>To be added.</summary>
+    <summary>The System.Text.Json namespace provides high-performance, low-allocating, and RFC-compliant capabilities to process JavaScript Object Notation (JSON) which includes serializing objects to JSON text and deserialize JSON text to objects, with UTF-8 support built-in. It also provides types to read and write JSON text encoded as UTF-8, and to create an in-memory document object model (DOM), that is read-only, for random access of the JSON elements within a structured view of the data.</summary>
```
rpetrusha:
```suggestion
    <summary>The System.Text.Json namespace provides high-performance, low-allocating, and RFC-compliant capabilities to process JavaScript Object Notation (JSON), which includes serializing objects to JSON text and deserializing JSON text to objects, with UTF-8 support built-in. It also provides types to read and write JSON text encoded as UTF-8, and to create a read-only in-memory document object model (DOM) for random access of the JSON elements within a structured view of the data.</summary>
```

ahsonkhan: Given we are considering adding modifiable Json DOM APIs in vNext, maybe we can drop the "readonly" specifier. Avoids having the problem of the namespace description being too (current implementation) specific and requiring updating when more things get added.

xml/ns-System.Text.Json.xml
```diff
@@ -1,6 +1,6 @@
 <Namespace Name="System.Text.Json">
   <Docs>
-    <summary>To be added.</summary>
+    <summary>The System.Text.Json namespace provides high-performance, low-allocating, and RFC-compliant capabilities to process JavaScript Object Notation (JSON), which includes serializing objects to JSON text and deserializing JSON text to objects, with UTF-8 support built-in. It also provides types to read and write JSON text encoded as UTF-8, and to create a read-only in-memory document object model (DOM) for random access of the JSON elements within a structured view of the data.</summary>
```
ahsonkhan:
```suggestion
    <summary>The System.Text.Json namespace provides high-performance, low-allocating, and RFC-compliant capabilities to process JavaScript Object Notation (JSON), which includes serializing objects to JSON text and deserializing JSON text to objects, with UTF-8 support built-in. It also provides types to read and write JSON text encoded as UTF-8, and to create an in-memory document object model (DOM) for random access of the JSON elements within a structured view of the data.</summary>
```

ahsonkhan: Re-reading this, I am not sure if we should say "RFC-compliant" or standards-compliant (similar to XML).

"json standard" gets a lot more hits when searching than "json rfc" although we are following RFC 8529: https://tools.ietf.org/html/rfc8259

ahsonkhan: Discussed this a bit more with @GrabYourPitchforks and he brought up a good point about using a term like standard that is more generic/wider reach (and friendlier for reading in different languages).

Let's switch to that instead.

```suggestion
    <summary>The System.Text.Json namespace provides high-performance, low-allocating, and standards-compliant capabilities to process JavaScript Object Notation (JSON), which includes serializing objects to JSON text and deserializing JSON text to objects, with UTF-8 support built-in. It also provides types to read and write JSON text encoded as UTF-8, and to create an in-memory document object model (DOM) for random access of the JSON elements within a structured view of the data.</summary>
```

ahsonkhan:
> Thanks for remembering that this needs to be done, @ahsonkhan, and for opening the PR, @carlossanlop

I merely helped wordsmith the description for the namespaces. @carlossanlop actually detected and brought this up as missing doc. Thanks for addressing this! I appreciate your effort and help getting S.T.Json documented overall.

xml/ns-System.Text.Json.xml
carlossanlop: Thanks for the suggestions. I'll apply it.

PR #2908, opened 2019-08-01
Document System.Text.Json.JsonElement+ArrayEnumerator

carlossanlop:
inheritdoc

### DIFF
--- a/xml/System.Text.Json/JsonElement+ArrayEnumerator.xml
+++ b/xml/System.Text.Json/JsonElement+ArrayEnumerator.xml
@@ -173,7 +173,7 @@
       <Parameters />
       <Docs>
         <summary>Returns an enumerator that iterates through a collection.</summary>
-        <returns>An enumerator for <see cref="T:System.Text.Json.JsonElement" /> objects that can be used to iterate through the collection.</returns>
+        <returns>An enumerator for an array of <see cref="T:System.Text.Json.JsonElement" /> that can be used to iterate through the collection.</returns>
         <remarks>  
           <format type="text/markdown"><![CDATA[  
 


xml/System.Text.Json/JsonElement+ArrayEnumerator.xml
```diff
@@ -196,8 +196,8 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <summary>Returns an enumerator that iterates through a collection.</summary>
+        <returns>An <see cref="T:System.Collections.Generic.IEnumerator" /> object that can be used to iterate through the collection.</returns>
         <remarks>To be added.</remarks>
```
rpetrusha:
```suggestion
        <remarks>  
          <format type="text/markdown"><![CDATA[  

This member is an explicit interface member implementation. It can be used only when the <xref:System.Text.Json.JsonElement.ArrayEnumerator> instance is cast to an <xref:System.Collections.IEnumerable> interface. 

         ]]></format>
         </remarks>
```

xml/System.Text.Json/JsonElement+ArrayEnumerator.xml
```diff
@@ -172,8 +172,8 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <summary>Returns an enumerator that iterates through a collection.</summary>
+        <returns>An <see cref="T:System.Collections.Generic.IEnumerator" /> object for <see cref="T:System.Text.Json.JsonElement" /> objects that can be used to iterate through the collection.</returns>
```
rpetrusha:
```suggestion
        <returns>An enumerator for <see cref="T:System.Text.Json.JsonElement" /> objects that can be used to iterate through the collection.</returns>
```

xml/System.Text.Json/JsonElement+ArrayEnumerator.xml
```diff
@@ -172,8 +172,8 @@
...
         <remarks>To be added.</remarks>
```
rpetrusha:
```suggestion
        <remarks>  
          <format type="text/markdown"><![CDATA[  

This member is an explicit interface member implementation. It can be used only when the <xref:System.Text.Json.JsonElement.ArrayEnumerator> instance is cast to an [IEnumerable\<JsonElement>](<xref:System.Collections.Generic.IEnumerable%601) interface. 

         ]]></format>
         </remarks>
```

carlossanlop: Added the `## Remarks` header to your suggestion.

carlossanlop:
Thanks for your suggestions, @rpetrusha. I applied them. If there are no more comments, we may be able to merge it after the build completes.

xml/System.Text.Json/JsonElement+ArrayEnumerator.xml
```diff
@@ -124,8 +124,8 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <summary>Advances the enumerator to the next element of the collection.</summary>
```
ahsonkhan: It's interesting that we are using "element" which is what other enumerators use, but it is also overloaded in this context (to mean `JsonElement` since each "element" is of that type).

xml/System.Text.Json/JsonElement+ArrayEnumerator.xml
```diff
@@ -172,9 +172,17 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Returns an enumerator that iterates through a collection.</summary>
+        <returns>An enumerator for <see cref="T:System.Text.Json.JsonElement" /> objects that can be used to iterate through the collection.</returns>
```
ahsonkhan: Maybe we should avoid using "objects" here, since it is an overloaded term, JSON object or .NET object and this is meant for "JSON arrays".

```suggestion
        <returns>An enumerator for an array of <see cref="T:System.Text.Json.JsonElement" /> that can be used to iterate through the collection.</returns>
```

bartonjs:
I don't have any strong feelings.  It's just another enumerator, so 100% copied from elsewhere satisfies me.

Ideally, if the docs data used a slightly higher level language, we'd just do something like

```xml
<Docs>
  <Template Name="StructEnumerator" ElementType="System.Text.Json.JsonElement" />
</Docs>
```
and call it a day :smile:.  (And that would fail if the enumerator defined anything outside the template bounds)

xml/System.Text.Json/JsonElement+ArrayEnumerator.xml
carlossanlop: If you would prefer me to change the language, please provide a suggestion and I will be happy to apply it. For now, I will mark the conversation as resolved since there is no action item.

rpetrusha:
@dend, I thought that you might like to see the last comment from @bartonjs.

rpetrusha:
All comments have been addressed and the build has succeeded, so I'll merge.

xml/System.Text.Json/JsonElement+ArrayEnumerator.xml
ahsonkhan: Was mainly talking out loud. Not worth re-wording. Looks good :)

PR #2930, opened 2019-08-05
System.IO.TextReader is cleared behaviors difference in summary between ReadAsync and ReadBlockAsync. (#1027)

kekyo:
# Summary

Added TextReader different behavior in summary tags between ReadAsync and ReadBlockAsync. (#1027)


rpetrusha:
@JeremyKuhne, could you take a look at this PR, please?

JeremyKuhne:
The remarks go into this, and that is where we need to fix things first. ReadAsync isn't right. If we want to clarify further we should probably try and match the verbiage for Read and ReadBlock.

rpetrusha:
Given the feedback from @JeremyKuhne, I'll close this PR, @kekyo. Would you like to submit a PR to align the documentation for ReadAsync with Read, and with ReadBlockAsync with ReadBlock?

PR #2964, opened 2019-08-08
Missing params in System.IO.Path

carlossanlop:
A couple of params I missed.


ericstj:
@layomia can you have a look?

PR #2979, opened 2019-08-12
Remove dead MSDN link to details

ite-jankl:
## Summary

The link for details goes to MSDN, which redirects to the generic and unhelpful URL https://docs.microsoft.com/en-us/previous-versions/ instead.

Searching the web for the term “Metadata Workspace” does not yield results, nor searching it on docs. Trying to open the URL on the web archive does not yield results.

So maybe someone can point out where that content is now so the link can be fixed up. Otherwise and until then, the broken link should be removed.

This change removes the dead link.


ite-klass:
Maybe it is https://docs.microsoft.com/en-us/previous-versions/bb399600(v=vs.90)?

Not sure if that’s helpful and should be linked though? It is old, not integrated documentation.

PR #2983, opened 2019-08-12
Document new constructor for Windows.UI.Xaml.LayoutCycleException

carlossanlop:
Not owned by my immediate team, but I figured it would be quick to document. It's the only undoc API in this namespace.

Based on this:

https://docs.microsoft.com/en-us/dotnet/api/system.exception.-ctor?view=netcore-3.0#System_Exception__ctor_System_Runtime_Serialization_SerializationInfo_System_Runtime_Serialization_StreamingContext_


stevenbrix:
This looks like it's documentation for the WinUI version of this API, which I'm not sure is what we want?

rpetrusha:
The API browser shows the constructor as supported in .NET Platform Extensions for 3.0.

PR #2989, opened 2019-08-13
Removed duplicate member group

rpetrusha:

### DIFF
--- a/xml/System.Numerics/Complex.xml
+++ b/xml/System.Numerics/Complex.xml
@@ -1635,6 +1635,32 @@ The addition of a complex number (a + bi) and a real number (which can be regard
         <altmember cref="P:System.Numerics.Complex.Phase" />
       </Docs>
     </Member>
+    <MemberGroup MemberName="Multiply">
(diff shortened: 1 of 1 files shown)


xml/System.Numerics/Complex.xml
```diff
@@ -1635,31 +1635,6 @@ The addition of a complex number (a + bi) and a real number (which can be regard
         <altmember cref="P:System.Numerics.Complex.Phase" />
       </Docs>
     </Member>
-    <MemberGroup MemberName="op_Multiply">
```
carlossanlop: This was supposed to be a MemberGroup for "Multiply", which is also a copy of the MemberGroup for "op_Multiply".
@rpetrusha Can you bring it back and leave the group, but with the MemberName fixed?

rpetrusha:
Thanks for pointing this out, @carlossanlop. I've renamed the duplicated member group.

carlossanlop:
The build passed so this is ready to be merged, @rpetrusha.

PR #3006, opened 2019-08-15
Correctly explained narrowing and widening casts in MemoryMarshal

svick:
Fixes https://github.com/dotnet/dotnet-api-docs/issues/1076.

### DIFF
--- a/xml/System.Runtime.InteropServices/MemoryMarshal.xml
+++ b/xml/System.Runtime.InteropServices/MemoryMarshal.xml
@@ -385,9 +385,9 @@ Neither `TFrom` nor `TTo` can contain pointers or references. `TFrom` and `TTo` 
 
 If the sizes of the two types are different, the cast combines or splits values, which leads to a change in length.
 
-For example, if `TFrom` is <xref:System.Int64>, the `ReadOnlySpan<Int64>` contains a single value, 0x100001111110F0F, and `TTo` is <xref:System.Int32>, the resulting `ReadOnlySpan<Int32>` contains two values: 0x11110F0F and 0x01000011, assuming little-endian architecture, such as x86.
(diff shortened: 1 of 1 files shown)


carlossanlop:
Adding the System.Runtime.InteropServices area owners @AaronRobinsonMSFT @jkoritzinsky for review.

xml/System.Runtime.InteropServices/MemoryMarshal.xml
```diff
@@ -383,7 +383,11 @@ This method is supported only on platforms that support misaligned memory access
 
 Neither `TFrom` nor `TTo` can contain pointers or references. `TFrom` and `TTo` are checked at runtime in order to preserve type safety.  
 
-The cast occurs in an unchecked context. If the range of `TTo` is less than the range of `TFrom`, data loss can occur. For example, if `TFrom` is <xref:System.Int64>, one of the `ReadOnlySpan<Int32>` values is 0x100001111110F0F, and `TTo` is <xref:System.Int32>, the resulting value is 11110F0F.
+If the sizes of the two types are different, the cast combines or splits values, which leads to a change in length.
+
+For example, if `TFrom` is <xref:System.Int64>, the `ReadOnlySpan<Int64>` contains a single value, 0x100001111110F0F, and `TTo` is <xref:System.Int32>, the resulting `ReadOnlySpan<Int32>` contains two values: 0x11110F0F and 0x01000011, assuming little-endian architecture, such as x86.
```
AaronRobinsonMSFT: I don't see how the endianness of the platform is relevant in this case. The value is being split into `int`s and the endianness of the platform shouldn't enter into the equation. In fact, I would say this example actually demonstrates the endianness isn't involved because the `long` would be stored as `0x0F, 0x0F, 0x11, 0x11, 0x11, 0x00, 0x00, 0x01` on a little endian platform, but is split as expected/illustrated.

Please correct me if I am missing something here.

AaronRobinsonMSFT: Okay, upon chatting with someone, I see what you are saying. Please clarify the contents of the `Span<T>` are ordered based on endianness not the values themselves. I would actually show the two examples:

```
Little endian: [ 0x11110F0F, 0x01000011]
   Big endian: [ 0x01000011, 0x11110F0F]
```

svick: @AaronRobinsonMSFT That's a good point, updated. Does that clarify it enough for you?

PR #3009, opened 2019-08-15
Automatically port System.Text.Json.Serialization.JsonConverterFactory

carlossanlop:


carlossanlop:
@mairaw @rpetrusha I noticed the files are being saved by Visual Studio as UTF-8 by default, but the latest sync of the repo is telling me that all the dotnet-api-docs xml files have the `Western European (Windows)` formatting. This is why the first line of my most recent PRs shows up as modified. Is this something we should worry about?

rpetrusha:
UTF-8 and the BOM doesn't matter, @carlossanlop, except that it is tracked as a change, and also means that every new CI will indicate a change to every single file in the repo. So it's best if the encoding continue to be Western European (Windows).

//cc @mairaw

PR #3015, opened 2019-08-15
Automatically port System.Text.Json.JsonSerializerOptions

carlossanlop:


xml/System.Text.Json/JsonSerializerOptions.xml
```diff
@@ -175,8 +175,8 @@ It is not used when deserializing.
         <ReturnType>System.Text.Encodings.Web.JavaScriptEncoder</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Gets or sets the encoder to use when escaping strings, or <see langword="null" /> to use the default encoder.</summary>
+        <value>The JavaScript character encoding.</value>
```
ahsonkhan: encod**ing** is a bit odd since the return type is an encod**er** that is responsible for doing the encoding, but its fine. The existing JavascriptEncoder docs also say "encoding".
https://docs.microsoft.com/en-us/dotnet/api/system.text.encodings.web.javascriptencoder?view=dotnet-plat-ext-2.1
> Represents a JavaScript character encoding.

PR #3041, opened 2019-08-16
Add WPF's System.Windows.Documents.PresentationUIStyleResources to .NET Core documentation

ryalanms:
## Summary

Add System.Windows.Documents.PresentationUIStyleResources to .NET Core documentation.

This is an empty static class in a cycle-breaker reference assembly.

Fixes dotnet/wpf#1172.


carlossanlop:
Because this API was added in Preview 9, it shouldn't show up in our Docs yet (we are at Preview 8).
New APIs get added automatically to the dotnet-api-docs repo via reflection, not manually.
This API will have to be documented later.

PR #3052, opened 2019-08-19
Update Attribute.xml

john-par:
##  Summary

Minor update related to User Story 1585685: https://mseng.visualstudio.com/TechnicalContent/_workitems/edit/1585685


john-par:
Thanks for the merge @mairaw. And also, thanks for your inquiry.

I've checked with my superiors and unfortunately, we don't know the specific reason this msdn link caused a localization issue. If you'd like, you can reach out to the requestor directly: Mauricio De los Santos (v-maudel@microsoft.com). He may be able to shed more light on your inquiry.

PR #3066, opened 2019-08-22
Fix spacing in System.Net.Sockets remarks

tatokis:
## Summary

There were some whitespaces missing in the `Socket.NoDelay Property` page, resulting in words concatenated with each other. This PR fixes that issue.


tatokis:
Thank you for your quick response, and glad to hear it. Much appreciated!

PR #3075, opened 2019-08-24
"Unix" -> "Unicode" in Char.GetUnicodeCategory() Remarks

srutzky:
The term "the Unix Standard" is incorrect and should instead be "the Unicode Standard". I fixed 4 instances of this mistake, all in "GetUnicodeCategory". This mistake appears to have been introduced in the following commit on March 21, 2017:

https://github.com/dotnet/dotnet-api-docs/commit/cd8c56990ee150d9500f36c39dabae965d29a4b5#diff-206ae69d9a4bfeaf7052220b8e8cde2a


srutzky:
@rpetrusha You're welcome 😺

PR #3084, opened 2019-08-26
Random and identical sequences

### LINKED ISSUE #3074: Update System.Random documentation regarding seeds
The current documentation for [`System.Random`](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Random.xml#L89) says:
> Initializing two random number generators in a tight loop or in rapid succession creates two random number generators that can produce identical sequences of random numbers.

This is true in [.NET Framework](https://referencesource.microsoft.com/#mscorlib/system/random.cs,5d22f8880fc9f8d9,references) but doesn't seem to be true in [.NET Core](https://source.dot.net/#System.Private.CoreLib/shared/System/Random.cs,5d22f8880fc9f8d9,references).

Can we update the documentation to highlight this difference? It doesn't matter too much, of course, because the general advice is to not generate lots of `System.Random`s, but instead use a single one but 🤷‍♂.

rpetrusha:
## Random and identical sequences

Fixes #3074

### DIFF
--- a/xml/System/Random.xml
+++ b/xml/System/Random.xml
@@ -370,8 +370,9 @@ Random.NextDouble() * (maxValue - minValue) + minValue
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
-## Remarks  
- The default seed value is derived from the system clock and has finite resolution. As a result, on the .NET Framework only, different <xref:System.Random> objects that are created in close succession by a call to the parameterless constructor will have identical default seed values and, therefore, will produce identical sets of random numbers. This problem can be avoided by using a single <xref:System.Random> object to generate all random numbers. You can also work around it by calling this constructor on .NET Core, or by modifying the seed value returned by the system clock and then explicitly providing this new seed value to the <xref:System.Random.%23ctor%28System.Int32%29> constructor. For more information, see the <xref:System.Random.%23ctor%28System.Int32%29> constructor.  
+## Remarks
+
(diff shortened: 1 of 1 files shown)


xml/System/Random.xml
```diff
@@ -371,7 +371,7 @@ Random.NextDouble() * (maxValue - minValue) + minValue
...
+ The default seed value is derived from the system clock and has finite resolution. As a result, on the .NET Framework only, different <xref:System.Random> objects that are created in close succession by a call to the parameterless constructor will have identical default seed values and, therefore, will produce identical sets of random numbers. This problem can be avoided by using a single <xref:System.Random> object to generate all random numbers. You can also work around it by calling this constructor on .NET Core, or by modifying the seed value returned by the system clock and then explicitly providing this new seed value to the <xref:System.Random.%23ctor%28System.Int32%29> constructor. For more information, see the <xref:System.Random.%23ctor%28System.Int32%29> constructor.
```
svick: > You can also work around it by calling this constructor on .NET Core […]

This doesn't make much sense to me. If you're suggesting that the workaround is switching to .Net Core, then I think that should be phrased more clearly, but I'm not sure it's actually reasonable. People are not going to switch to .Net Core just to avoid few extra lines of code.

> You can also work around it […] by modifying the seed value returned by the system clock

This text wasn't changed by this PR, but I find it hard to understand and I think it should be clarified too: the clock returns time, not a seed value and there's no way to modify the value that's returned by the clock.

xml/System/Random.xml
```diff
@@ -380,7 +380,7 @@ Random.NextDouble() * (maxValue - minValue) + minValue
...
+ The following example uses the parameterless constructor to instantiate three <xref:System.Random> objects and displays a sequence of five random integers for each. If it is run on .NET Core, because the first two <xref:System.Random> objects are created in close succession, they are instantiated using identical seed values based on the system clock and, therefore, they produce an identical sequence of random numbers. On the other hand, the parameterless constructor of the third <xref:System.Random> object is called after a two-second delay caused by calling the <xref:System.Threading.Thread.Sleep%2A?displayProperty=nameWithType> method. Because this produces a different seed value for the third <xref:System.Random> object, it produces a different sequence of random numbers.
```
svick: > If it is run on .NET Core, […]

Did you mean ".NET Framework" here?

rpetrusha: Yes, I did mean .NET Framework.

rpetrusha:
Thanks for the comments and the suggestions, @svick. I've addressed all of your feedback.

PR #3090, opened 2019-08-27
Change description of 'state' object passed to ProgressUpdate

jgscherber:
Match description of the parameter to the description of the property on the event generated from the function call.

## Summary

The description of the user state returned in [ProgressChangedEventArgs.UserState](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.progresschangedeventargs.userstate?view=netframework-4.8#System_ComponentModel_ProgressChangedEventArgs_UserState) describe the object generically as 'A unique Object indicating the user state.' whereas the `ProgressUpdate` function that generates the event describes it as being the same object that's provided `RunWorkerAsync`. That may be the case - but isn't a requirement and the documentation makes it confusing that it's a requirement. The documentation also doesn't provide information about where the object exists on the event created.

N/A

### DIFF
--- a/xml/System.ComponentModel/BackgroundWorker.xml
+++ b/xml/System.ComponentModel/BackgroundWorker.xml
@@ -745,7 +745,7 @@
       </Parameters>
       <Docs>
         <param name="percentProgress">The percentage, from 0 to 100, of the background operation that is complete.</param>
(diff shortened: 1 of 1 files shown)


rpetrusha:
@maryamariyan @safern could you review this PR?

PR #3118, opened 2019-09-04
Use of …" messing up code highlighting

gabeluci:
## Summary
The unmatched quotes in the code examples were causing everything after them to be considered part of the string, which can be seen in the [live page](https://docs.microsoft.com/en-us/dotnet/api/system.directoryservices.directorysynchronization?view=netframework-4.8).

I just updated `…"` to `"…"` so the quotes are matched.


gabeluci:
Absolutely. I almost did it in the first place but wasn't sure if it was standard practice for the docs in places where someone would usually put their own code.

rpetrusha:
Thanks, @gabeluci! I'll merge your PR now.

Issue #2898, opened 2019-08-01
Utf8JsonReader docs should mention that some properties are non-validating


GrabYourPitchforks:
/cc @ahsonkhan @bartonjs to get their opinions on this wording.

The properties [`Utf8JsonReader.ValueSpan`](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.valuespan) and [`Utf8JsonReader.ValueSequence`](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.valuesequence) do not always contain well-formed data. For example, if the binary input to the `Utf8JsonReader` type contains ill-formed UTF-8 strings, the APIs will not detect this error until the caller attempts to materialize the string via the [`Utf8JsonReader.GetString`](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.getstring) method. If the consumer does not query this method and instead relies solely on `ValueSequence` / `ValueSpan`, there's the possibility that ill-formed UTF-8 data could make its way into the application.

This behavior is by-design. However, since it has potential security and reliability implications, it should probably be called out in the _Remarks_ section for the `ValueSequence` / `ValueSpan` properties. If there's other validation that the `Get*` APIs would normally perform that are skipped by these raw property getters, that should also be called out in the same section.

PR #3142, opened 2019-09-09
Update Utf8JsonReader ValueSpan, Sequence, and ValueTextEquals docs.

ahsonkhan:
Fixes https://github.com/dotnet/dotnet-api-docs/issues/2898

Also, other fixes around consistency and correctness.

cc @mairaw, @GrabYourPitchforks, @carlossanlop

### DIFF
--- a/xml/System.Text.Json/Utf8JsonReader.xml
+++ b/xml/System.Text.Json/Utf8JsonReader.xml
@@ -386,7 +386,7 @@ This method only creates a DateTime representation of JSON strings that conform 
         
 -or-
         
-The entire UTF-8 encoded token value cannot be parsed to a <xref:System.DateTime> value.</exception>
+The entire UTF-8 encoded token value cannot be parsed to a <see cref="System.DateTime" /> value.</exception>
         <altmember cref="P:System.Text.Json.Utf8JsonReader.TokenType" />
       </Docs>
     </Member>
@@ -423,7 +423,7 @@ This method only creates a DateTimeOffset representation of JSON strings that co
 
 -or-
         
-The entire UTF-8 encoded token value cannot be parsed to a <xref:System.DateTimeOffset> value.</exception>
+The entire UTF-8 encoded token value cannot be parsed to a <see cref="System.DateTimeOffset" /> value.</exception>
         <altmember cref="P:System.Text.Json.Utf8JsonReader.TokenType" />
       </Docs>
     </Member>
@@ -502,7 +502,7 @@ The entire UTF-8 encoded token value cannot be parsed to a <xref:System.DateTime
         
 -or-
 
-The entire UTF-8 encoded token value cannot be parsed to a <xref:System.Guid> value.</exception>
+The entire UTF-8 encoded token value cannot be parsed to a <see cref="System.Guid" /> value.</exception>
         <altmember cref="P:System.Text.Json.Utf8JsonReader.TokenType" />
       </Docs>
     </Member>
@@ -786,11 +786,7 @@ The JSON token value represents a number less than <see cref="F:System.UInt32.Mi
             
 -or-
 
-The JSON token value represents a number less than <see cref="F:System.UInt64.MinValue" /> or greater than <see cref="F:System.UInt64.MaxValue" />
-
--or-
-
-The entire UTF-8 encoded token value cannot be parsed to a <xref:System.UInt64> value.</exception>
+The JSON token value represents a number less than <see cref="F:System.UInt64.MinValue" /> or greater than <see cref="F:System.UInt64.MaxValue" />.</exception>
         <altmember cref="P:System.Text.Json.Utf8JsonReader.TokenType" />
       </Docs>
     </Member>
@@ -1516,7 +1512,7 @@ The current depth exceeds the recursive limit set by the maximum depth.</excepti
 
 ## Remarks
 
-If the JSON is provided within a ReadOnlySequence&lt;byte&gt; and the slice that represents the token value fits in a single segment, then <see cref="P:System.Text.Json.Utf8JsonReader.ValueSpan" /> contains the sliced value since it can be represented as a span. Otherwise, `ValueSequence` contains the token value.
+If the JSON is provided within a ReadOnlySequence&lt;byte&gt; and the slice that represents the token value fits in a single segment, then <xref:System.Text.Json.Utf8JsonReader.ValueSpan> contains the sliced value since it can be represented as a span. Otherwise, `ValueSequence` contains the token value.
 
 If <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `false`, `ValueSequence` is empty. 
 Therefore, only access `ValueSequence` if <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `true`.
@@ -1559,7 +1555,7 @@ Since this property returns the raw bytes, avoid using it for text comparison. I
 
 ## Remarks
 
-If the JSON is provided within a ReadOnlySequence&lt;byte&gt; and the slice that represents the token value fits in a single segment, then `ValueSpan` contains the sliced value since it can be represented as a span. Otherwise, <see cref="P:System.Text.Json.Utf8JsonReader.ValueSequence" /> contains the token value.
+If the JSON is provided within a ReadOnlySequence&lt;byte&gt; and the slice that represents the token value fits in a single segment, then `ValueSpan` contains the sliced value since it can be represented as a span. Otherwise, <xref:System.Text.Json.Utf8JsonReader.ValueSequence> contains the token value.
 
 If <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `true`, `ValueSpan` is empty. 
 Therefore, only access `ValueSpan` if <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `false`.


xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1596,13 +1715,13 @@ The comparison of the JSON token value in the source and the lookup text is done
 
 ## Remarks
 
-If the lookup text is invalid UTF-8 text, the method returns `false` since you can't have invalid UTF-8 within the JSON payload.
+The lookup text must be valid UTF-8 text. Otherwise, this method could return `true` if the source has a string token containing invalid UTF-8 text that matches.
```
ahsonkhan: **Question:** Why do we have a preamble description/remarks section for the ValueTextEquals overloads?

https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.valuetextequals?view=netcore-3.0#System_Text_Json_Utf8JsonReader_ValueTextEquals_System_String_

image

Other overloads don't have that. Should we remove it? Specifically, the following:
https://github.com/dotnet/dotnet-api-docs/blob/1a26fca381874610894aa51a7c749dbbbdc94a6b/xml/System.Text.Json/Utf8JsonReader.xml#L1555-L1570

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -716,13 +792,22 @@ The JSON string contains invalid UTF-8 bytes or invalid UTF-16 surrogates.</exce
       <Docs>
         <summary>Parses the current JSON token value from the source as a <see cref="T:System.UInt16" />.</summary>
         <returns>The UTF-8 encoded token value parsed to a <see cref="T:System.UInt16" />.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
```
ahsonkhan: Some of the overloads had this text already (like GetDecimal), along with a link to `See also: TokenType` (which were added in https://github.com/dotnet/dotnet-api-docs/pull/1880).
So, I added it to all the other Get* APIs for consistency.
https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.getdecimal?view=netcore-3.0

image

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -31,7 +31,7 @@ When `Utf8JsonReader` encounters invalid JSON, it throws a <xref:System.Text.Jso
 
 Since this type is a ref struct, it doesn't directly support async. However, it does provide support for reentrancy to read incomplete data and to continue reading once more data is presented.
 
-To be able to set max depth while reading OR allow skipping comments, create an instance of <xref:System.Text.Json.JsonReaderState> and pass it to the reader.
+To be able to set max depth while reading OR allow skipping comments, create an instance of <xref:System.Text.Json.JsonReaderOptions> and pass it to the reader.
```
ahsonkhan: When do we use `<see cref=...` vs `<xref:...`?

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -54,7 +54,7 @@ To be able to set max depth while reading OR allow skipping comments, create an
       </Parameters>
       <Docs>
         <param name="jsonData">The UTF-8 encoded JSON text to process.</param>
-        <param name="options">An object that defines customized behavior of the <see cref="T:System.Text.Json.Utf8JsonReader" /> that differs from the JSON RFC (for example how to handle comments or maximum depth allowed when reading). By default, the <see cref="T:System.Text.Json.Utf8JsonReader" /> follows the JSON RFC strictly; comments within the JSON are invalid, and the maximum depth is 64.</param>
+        <param name="options">Defines customized behavior of the <see cref="T:System.Text.Json.Utf8JsonReader" /> that differs from the JSON RFC (for example how to handle comments or maximum depth allowed when reading). By default, the <see cref="T:System.Text.Json.Utf8JsonReader" /> follows the JSON RFC strictly; comments within the JSON are invalid, and the maximum depth is 64.</param>
```
ahsonkhan: `JsonReaderOptions` is not a class, but rather a struct. Hence, removed the use of "object".

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -122,7 +122,7 @@ This constructor assumes that the entire JSON payload is contained in `jsonData`
         <param name="jsonData">The UTF-8 encoded JSON text to process.</param>
         <param name="isFinalBlock">
           <see langword="true" /> if the input sequence contains the entire data to process; <see langword="false" /> only the input span contains partial data with more data to follow.</param>
-        <param name="state">An object that contains the reader state. If this is the first call to the constructor, pass in a default state. Otherwise, capture the state from the previous instance of the <see cref="T:System.Text.Json.Utf8JsonReader" /> and pass that back.</param>
+        <param name="state">A structure that contains the reader state. If this is the first call to the constructor, pass in a default state. Otherwise, capture the state from the previous instance of the <see cref="T:System.Text.Json.Utf8JsonReader" /> by calling <see cref="T:System.Text.Json.Utf8JsonReader.CurrentState" /> and pass that back.</param>
```
ahsonkhan: Similarly here, JsonReaderState is a struct, not an class. Changed "object" to "structure".

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1500,18 +1609,23 @@ The current depth exceeds the recursive limit set by the maximum depth.</excepti
         <ReturnType>System.Buffers.ReadOnlySequence&lt;System.Byte&gt;</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets the value of the last processed token as a <see cref="T:System.ReadOnlySpan`1" /> slice of the input payload. If the JSON is provided within a <see cref="T:System.Buffers.ReadOnlySequence`1" /> and the slice that represents the token value fits in a single segment, then <see cref="P:System.Text.Json.Utf8JsonReader.ValueSpan" /> contains the sliced value since it can be represented as a span. Otherwise, <see cref="P:System.Text.Json.Utf8JsonReader.ValueSequence" /> contains the token value.</summary>
+        <summary>Gets the raw value of the last processed token as a ReadOnlySequence&lt;byte&gt; slice of the input payload. If the JSON is provided within a ReadOnlySequence&lt;byte&gt; and the slice that represents the token value fits in a single segment, then <see cref="P:System.Text.Json.Utf8JsonReader.ValueSpan" /> contains the sliced value since it can be represented as a span. Otherwise, <see cref="P:System.Text.Json.Utf8JsonReader.ValueSequence" /> contains the token value.</summary>
```
ahsonkhan: This was incorrect before. ValueSequence returns ReadOnly**Sequence**, not ReadOnly**Span**.

Also, changed it from generic `ROSequence<T>` to the specific `ROSequence<byte>`. However, that loses the link to the ROSequence docs page. Any suggestions?

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1624,10 +1743,10 @@ The comparison of the JSON token value in the source and the lookup text is done
         <Parameter Name="text" Type="System.ReadOnlySpan&lt;System.Char&gt;" />
       </Parameters>
       <Docs>
-        <param name="text">The UTF-8 encoded text to compare against.</param>
-        <summary>Compares the UTF-8 encoded text in a read-only character span to the unescaped JSON token value in the source and returns a value that indicates whether they match.</summary>
+        <param name="text">The text to compare against.</param>
```
ahsonkhan: This was incorrect. `ROS<char>` doesn't contain UTF-8 text, but rather UTF-16 text (like the string overload).

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1662,21 +1781,21 @@ The comparison of the JSON token value in the source and the lookup text is done
       </Parameters>
       <Docs>
         <param name="text">The text to compare against.</param>
-        <summary>Compares the text in a character span to the unescaped JSON token value in the source and returns a value that indicates whether they match.</summary>
+        <summary>Compares the string text to the unescaped JSON token value in the source and returns a value that indicates whether they match.</summary>
```
ahsonkhan: This takes a string parameter, not `ROS<char>`.

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: cc @rpetrusha

carlossanlop: @ahsonkhan xref is markdown language, so it is used in remarks, cref is used in normal xml tags.

xml/System.Text.Json/Utf8JsonReader.xml
carlossanlop: Not really necessary but the result sounds ok to me.

If you look at the MS documentation for "C# Objects", struct instances can also be called objects:

> _[...] Because structs are value types, a variable of a struct object holds a copy of the entire object. Instances of structs can also be created by using the new operator, but this is not required [...]_

Source: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/objects

xml/System.Text.Json/Utf8JsonReader.xml
carlossanlop: potato vs potato 😄

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -368,10 +395,21 @@ The JSON token value represents a number less than <see cref="F:System.Byte.MinV
...
         <summary>Reads the next JSON token value from the source and parses it to a <see cref="T:System.DateTime" />.</summary>
-        <returns>The date and time, if the entire UTF-8 encoded token value can be successfully parsed.</returns>
-        <remarks>This method only creates a DateTime representation of JSON strings that conform to the ISO 8601-1 extended format (see [DateTime and DateTimeOffset support in System.Text.Json](~/docs/standard/datetime/system-text-json-support.md)).</remarks>
+        <returns>The date and time offset, if the entire UTF-8 encoded token value can be successfully parsed.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+This method only creates a DateTime representation of JSON strings that conform to the ISO 8601-1 extended format (see [DateTime and DateTimeOffset support in System.Text.Json](~/docs/standard/datetime/system-text-json-support.md)).
```
carlossanlop: @ahsonkhan is DateTime here supposed to be treated as a "JSON DateTime" as opposed to a "C# System.DateTime instance"? Similar to the JSON boolean?

If that's the case, then we could make it bold, like we've been doing with the JSON booleans:

```suggestion
This method only creates a **DateTime** representation of JSON strings that conform to the ISO 8601-1 extended format (see [DateTime and DateTimeOffset support in System.Text.Json](~/docs/standard/datetime/system-text-json-support.md)).
```

If I'm wrong, feel free to ignore my suggestion.

carlossanlop: Same question applies to DateTimeOffset below. I suspect we are using C# objects, not JSON objects, because that's the returned value by the methods.

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -483,9 +532,18 @@ The `GetDouble` method throws an exception if the entire UTF-8 encoded token val
       <Docs>
         <summary>Reads the next JSON token value from the source and parses it to a <see cref="T:System.Guid" />.</summary>
         <returns>The GUID value, if the entire UTF-8 encoded token value can be successfully parsed.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+The `GetGuid` method throws an exception if the entire UTF-8 encoded token value cannnot be parsed to a <xref:System.Guid> value.
```
carlossanlop: I think this explanation should be added to the exception itself, not to the remark. Also, which of the two exceptions are you referring to?

carlossanlop: I'm seeing a few similar instances of this remark about the exception being thrown, so please consider moving their explanations there too.

xml/System.Text.Json/Utf8JsonReader.xml
carlossanlop: We had a similar problem in Intrinsics, where we needed the documentation to have a link to the API, but the link would not show up if you added a specific type, so what I did was this (look at the summary):

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.intrinsics.vector256.asuint16?view=netcore-3.0#System_Runtime_Intrinsics_Vector256_AsUInt16__1_System_Runtime_Intrinsics_Vector256___0__

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1500,18 +1609,23 @@ The current depth exceeds the recursive limit set by the maximum depth.</excepti
...
 
 ## Remarks
 
 
-If <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `false`, `ValueSequence` is empty. 
+If <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `false`, <xref:System.Text.Json.Utf8JsonReader.ValueSequence> is empty. 
 Therefore, only access `ValueSequence` if <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `true`.
 Otherwise, the token value must be accessed from <xref:System.Text.Json.Utf8JsonReader.ValueSpan>.
 
+There is no guarantee that this property will always contain well-formed data. If the input JSON passed in to the Utf8JsonReader contains invalid UTF-8 bytes within JSON string tokens, this property will return back those invalid UTF-8 bytes as is.
```
carlossanlop: Wrapping Utf8JsonReader with backticks (no need to add a link, it's the current type):

```suggestion
There is no guarantee that this property will always contain well-formed data. If the input JSON passed in to the `Utf8JsonReader` contains invalid UTF-8 bytes within JSON string tokens, this property will return back those invalid UTF-8 bytes as is.
```

xml/System.Text.Json/Utf8JsonReader.xml
rpetrusha: Or more precisely, <xref> can only be used in a CDATA section, as on line 24. Otherwise, <cref> should be used.

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1537,7 +1651,7 @@ Otherwise, the token value must be accessed from <xref:System.Text.Json.Utf8Json
         <ReturnType>System.ReadOnlySpan&lt;System.Byte&gt;</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets the value of the last processed token as a ReadOnlySpan&lt;byte&gt; slice of the input payload. If the JSON is provided within a ReadOnlySequence&lt;byte&gt; and the slice that represents the token value fits in a single segment, then <see cref="P:System.Text.Json.Utf8JsonReader.ValueSpan" /> contains the sliced value since it can be represented as a span. Otherwise, the <see cref="P:System.Text.Json.Utf8JsonReader.ValueSequence" /> contains the token value.</summary>
+        <summary>Gets the raw value of the last processed token as a ReadOnlySpan&lt;byte&gt; slice of the input payload. If the JSON is provided within a ReadOnlySequence&lt;byte&gt; and the slice that represents the token value fits in a single segment, then <see cref="P:System.Text.Json.Utf8JsonReader.ValueSpan" /> contains the sliced value since it can be represented as a span. Otherwise, <see cref="P:System.Text.Json.Utf8JsonReader.ValueSequence" /> contains the token value.</summary>
```
carlossanlop: See my suggestion above (for intrinsics) about mentioning the type and adding a link.

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1548,6 +1662,11 @@ If <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `true`, <xref:Syst
 Therefore, only access `ValueSpan` if <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `false`.
 Otherwise, the token value must be accessed from <xref:System.Text.Json.Utf8JsonReader.ValueSequence>.
 
+There is no guarantee that this property will always contain well-formed data. If the input JSON passed in to the Utf8JsonReader contains invalid UTF-8 bytes within JSON string tokens, this property will return back those invalid UTF-8 bytes as is.
```
carlossanlop:
```suggestion
There is no guarantee that this property will always contain well-formed data. If the input JSON passed in to the `Utf8JsonReader` contains invalid UTF-8 bytes within JSON string tokens, this property will return back those invalid UTF-8 bytes as is.
```

xml/System.Text.Json/Utf8JsonReader.xml
carlossanlop: It's because the overloads are wrapped inside a MemberGroup, to avoid repeating information.
The group has a general summary and a general remarks section that applies to all the overloads equally.
If you think the information is not relevant or accurate in the member group, feel free to alter it.

xml/System.Text.Json/Utf8JsonReader.xml
carlossanlop: I left you a comment regarding this. I think the exception information should contain this text, not the remark.

xml/System.Text.Json/Utf8JsonReader.xml
rpetrusha: An alternative is to just leave it as is, since the property value section links to both ROSequence and Byte: 
image

But I'd also move all of the text starting with "If the JSON is provided..." to the Remarks section.

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1500,18 +1609,23 @@ The current depth exceeds the recursive limit set by the maximum depth.</excepti
...
+        <summary>Gets the raw value of the last processed token as a ReadOnlySequence&lt;byte&gt; slice of the input payload. If the JSON is provided within a ReadOnlySequence&lt;byte&gt; and the slice that represents the token value fits in a single segment, then <see cref="P:System.Text.Json.Utf8JsonReader.ValueSpan" /> contains the sliced value since it can be represented as a span. Otherwise, <see cref="P:System.Text.Json.Utf8JsonReader.ValueSequence" /> contains the token value.</summary>
         <value>A byte read-only sequence.</value>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
 
-If <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `false`, `ValueSequence` is empty. 
+If <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `false`, <xref:System.Text.Json.Utf8JsonReader.ValueSequence> is empty.
```
rpetrusha: Is best to leave this in backticks. The xref creates a self-referencing link, which causes a page reload if the user clicks on it. And page view data generally shows that users frequently click on self-referencing links.

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1500,18 +1609,23 @@ The current depth exceeds the recursive limit set by the maximum depth.</excepti
...
 
-If <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `false`, `ValueSequence` is empty. 
+If <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `false`, <xref:System.Text.Json.Utf8JsonReader.ValueSequence> is empty. 
 Therefore, only access `ValueSequence` if <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `true`.
 Otherwise, the token value must be accessed from <xref:System.Text.Json.Utf8JsonReader.ValueSpan>.
 
+There is no guarantee that this property will always contain well-formed data. If the input JSON passed in to the Utf8JsonReader contains invalid UTF-8 bytes within JSON string tokens, this property will return back those invalid UTF-8 bytes as is.
+Therefore, if the input is untrusted or not previously validated, call <xref:System.Text.Json.Utf8JsonReader.GetString> to get the JSON string token.
+
+Since this property returns the raw bytes, avoid using it for text comparison. Instead call <xref:System.Text.Json.Utf8JsonReader.ValueTextEquals> which will unescape the text if necessary.
```
rpetrusha:
```suggestion
Since this property returns the raw bytes, avoid using it for text comparison. Instead call <xref:System.Text.Json.Utf8JsonReader.ValueTextEquals>, which will unescape the text if necessary.
```

xml/System.Text.Json/Utf8JsonReader.xml
rpetrusha: Here again, everything from "If the JSON" on could be moved to the Remarks section.

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1548,6 +1662,11 @@ If <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `true`, <xref:Syst
 Therefore, only access `ValueSpan` if <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `false`.
 Otherwise, the token value must be accessed from <xref:System.Text.Json.Utf8JsonReader.ValueSequence>.
 
+There is no guarantee that this property will always contain well-formed data. If the input JSON passed in to the Utf8JsonReader contains invalid UTF-8 bytes within JSON string tokens, this property will return back those invalid UTF-8 bytes as is.
+Therefore, if the input is untrusted or not previously validated, call <xref:System.Text.Json.Utf8JsonReader.GetString> to get the JSON string token.
+
+Since this property returns the raw bytes, avoid using it for text comparison. Instead call <xref:System.Text.Json.Utf8JsonReader.ValueTextEquals> which will unescape the text if necessary.
```
rpetrusha:
```suggestion
Since this property returns the raw bytes, avoid using it for text comparison. Instead call <xref:System.Text.Json.Utf8JsonReader.ValueTextEquals>, which will unescape the text if necessary.
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -122,7 +122,7 @@ This constructor assumes that the entire JSON payload is contained in `jsonData`
         <param name="jsonData">The UTF-8 encoded JSON text to process.</param>
         <param name="isFinalBlock">
           <see langword="true" /> if the input sequence contains the entire data to process; <see langword="false" /> only the input span contains partial data with more data to follow.</param>
```
mairaw: boolean params have a special entry in the writing guidelines: https://github.com/dotnet/dotnet-api-docs/wiki/Parameters%3A-Methods-and-constructors

maybe something like this?

```suggestion
          <see langword="true" /> to indicate that the input sequence contains the entire data to process; <see langword="false" /> to indicate the input span contains partial data with more data to follow.</param>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -157,7 +157,7 @@ This is the reason why the constructor accepts a <xref:System.Text.Json.JsonRead
         <param name="jsonData">The UTF-8 encoded JSON text to process.</param>
         <param name="isFinalBlock">
           <see langword="true" /> if the input sequence contains the entire data to process; <see langword="false" /> if the input span contains partial data with more data to follow.</param>
-        <param name="state">An object that contains the reader state. If it's the first call to the constructor, it should reflect a default state; otherwise, it should capture the state from the previous instance instance of the <see cref="T:System.Text.Json.Utf8JsonReader" /> and pass that back.</param>
+        <param name="state">A structure that contains the reader state. If it's the first call to the constructor, it should reflect a default state; otherwise, it should capture the state from the previous instance instance of the <see cref="T:System.Text.Json.Utf8JsonReader" /> by calling <see cref="T:System.Text.Json.Utf8JsonReader.CurrentState" /> and pass that back.</param>
```
mairaw:
```suggestion
        <param name="state">A structure that contains the reader state. If it's the first call to the constructor, it should reflect a default state; otherwise, it should capture the state from the previous instance instance of the <see cref="T:System.Text.Json.Utf8JsonReader" /> by calling the <see cref="T:System.Text.Json.Utf8JsonReader.CurrentState" /> method and pass that back.</param>
```

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: Can I just remove it all together?

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: I'll fix this instance.

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: Right, JSON doesn't have a concept of DateTime/DateTimeOffset. These are referring to C# objects, unlike JSON boolean).

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -368,10 +395,21 @@ The JSON token value represents a number less than <see cref="F:System.Byte.MinV
       <Parameters />
       <Docs>
         <summary>Reads the next JSON token value from the source and parses it to a <see cref="T:System.DateTime" />.</summary>
-        <returns>The date and time, if the entire UTF-8 encoded token value can be successfully parsed.</returns>
-        <remarks>This method only creates a DateTime representation of JSON strings that conform to the ISO 8601-1 extended format (see [DateTime and DateTimeOffset support in System.Text.Json](~/docs/standard/datetime/system-text-json-support.md)).</remarks>
+        <returns>The date and time offset, if the entire UTF-8 encoded token value can be successfully parsed.</returns>
```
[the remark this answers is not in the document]
ahsonkhan: It returns DateTimeOffset. I don't know why we say "data and time value" for that (or for DateTime). I can revert this back to `date and time value` if you prefer. The reason I chose `date and time offset` was to try and differentiate between the two types (DateTime and DateTimeOffset).

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: How could I see this and know to use %2A to fix it? I had seen something similar for generics (%601). How do I interpret these custom escaped characters?

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: I could do that, particularly because this is getting lengthy. However, the concern of putting it in remarks is that it won't show up in IntelliSense, and currently this information highlights which property to use in which case.

Is it worth instead adding a "see also" in its place? Or maybe I can condense it a bit and put the details in the remarks.

image

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: CurrentState is a property though, so should I add "property" instead?

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: > We had a similar problem in Intrinsics, where we needed the documentation to have a link to the API, but the link would not show up if you added a specific type, so what I did was this (look at the summary):

I looked at the summary. I don't understand what you are suggesting here. You are still using the generic type:
image

xml/System.Text.Json/Utf8JsonReader.xml
mairaw: Good catch! The calling verb threw me off since we use that for method. Made some changes to my suggestion. Please take a look.

xml/System.Text.Json/Utf8JsonReader.xml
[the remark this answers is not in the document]
Jozkee: @ahsonkhan, I think you confused `GetDateTime` and `GetDateTimeOffset` methods when writing this.
```suggestion
        <returns>The date and time, if the entire UTF-8 encoded token value can be successfully parsed.</returns>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -392,10 +430,21 @@ The JSON token value represents a number less than <see cref="F:System.Byte.MinV
       <Parameters />
       <Docs>
         <summary>Reads the next JSON token value from the source and parses it to a <see cref="T:System.DateTimeOffset" />.</summary>
-        <returns>The date and time value, if the entire UTF-8 encoded token value can be successfully parsed.</returns>
-        <remarks>This method only creates a DateTimeOffset representation of JSON strings that conform to the ISO 8601-1 extended format (see [DateTime and DateTimeOffset support in System.Text.Json](~/docs/standard/datetime/system-text-json-support.md)).</remarks>
+        <returns>The date and time, if the entire UTF-8 encoded token value can be successfully parsed.</returns>
```
Jozkee: See [comment above](https://github.com/dotnet/dotnet-api-docs/pull/3142/files#r322483039)
```suggestion
        <returns>The date and time offset, if the entire UTF-8 encoded token value can be successfully parsed.</returns>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -970,7 +1071,7 @@ The current depth exceeds the recursive limit set by the maximum depth.</excepti
 
 For JSON strings (including property names), this value points to before the start quote.
 
-For comments, this value points to before the first comment delimiter (that is, '/').
+For comments, this value points to before the first comment delimiter (that is, '/'). This is only applicable when the reader is constructed using the <see cref="F:System.Text.Json.JsonCommentHandling.Allow" /> option.
```
Jozkee:
```suggestion
For comments, this value points to before the first comment delimiter (that is, '/'). This is only applicable when the reader is constructed using the <xref:System.Text.Json.JsonCommentHandling.Allow?displayProperty=nameWithType> option.
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -270,7 +270,15 @@ Unlike the <xref:System.Text.Json.Utf8JsonReader>, which is a ref struct, the st
         <summary>Reads the next JSON token value from the source as a <see cref="T:System.Boolean" />.</summary>
         <returns>
           <see langword="true" /> if the <see cref="P:System.Text.Json.Utf8JsonReader.TokenType" /> is <see cref="F:System.Text.Json.JsonTokenType.True" />; <see langword="false" /> if the <see cref="P:System.Text.Json.Utf8JsonReader.TokenType" /> is <see cref="F:System.Text.Json.JsonTokenType.False" />.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+The `GetBoolean` method throws an exception if the entire UTF-8 encoded token value cannnot be parsed to a <xref:System.Boolean> value.
```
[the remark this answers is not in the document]
carlossanlop: Or maybe as I suggested in my other comment, we should probably mention this case in the exception itself. What do you think, @rpetrusha?

Oh by the way, typo:

```suggestion
The `GetBoolean` method throws an exception if the entire UTF-8 encoded token value cannot be parsed to a <xref:System.Boolean> value.
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -294,13 +302,22 @@ Unlike the <xref:System.Text.Json.Utf8JsonReader>, which is a ref struct, the st
       <Docs>
         <summary>Parses the current JSON token value from the source as a <see cref="T:System.Byte" />.</summary>
         <returns>The value of the UTF-8 encoded token.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+The `GetByte` method throws an exception if the entire UTF-8 encoded token value cannnot be parsed to a <xref:System.Byte> value.
```
carlossanlop: Typo here too, but please consider mentioning this behavior in the exception summary itself, instead of the remark.
```suggestion
The `GetByte` method throws an exception if the entire UTF-8 encoded token value cannot be parsed to a <xref:System.Byte> value.
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -322,9 +339,18 @@ The JSON token value represents a number less than <see cref="F:System.Byte.MinV
       <Docs>
         <summary>Parses the current JSON token value from the source and decodes the Base64 encoded JSON string as a byte array.</summary>
         <returns>The byte array that represents the current JSON token value.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+The `GetBytesFromBase64` method throws an exception if the entire UTF-8 encoded token value cannnot be decoded to a <xref:System.Byte[]> value.
```
carlossanlop:
```suggestion
The `GetBytesFromBase64` method throws an exception if the entire UTF-8 encoded token value cannot be decoded to a <xref:System.Byte[]> value.
```

carlossanlop: We should see how System.Byte[] renders after the build finishes.

xml/System.Text.Json/Utf8JsonReader.xml
Jozkee: @ahsonkhan, there is no link for classes/structs that define the type used in the generic (i.e: there is no link to ReadOnlySequence<byte>). Did you consider using "byte sequence" or "sequence of bytes" as plain text, rather than forcing a link to render? I think that is what @rpetrusha is suggesting.

xml/System.Text.Json/Utf8JsonReader.xml
Jozkee: @ahsonkhan, those are URL encoded characters, %2A ("*") is used to refer to any signature of the `ValueTextEquals` method.
%60 ("\`") is the tilde which is used to refer to the generic-type members i.e: \<xref:System.Span`1\>; this tag is equivalent to \<xref:System.Span%601\>.
I have seen both ways used within the docs, with no actual difference between each other when they render.

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: Since the remarks are not all the same (and there are subtle differences in the summary too), I am removing the member group (especially in the interest of time since this doesn't affect intellisense anyway).

Reverting feedback from https://github.com/dotnet/dotnet-api-docs/pull/2890#discussion_r309999765

cc @mairaw

xml/System.Text.Json/Utf8JsonReader.xml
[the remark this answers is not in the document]
ahsonkhan: Ah yes, thanks for clarifying.

ahsonkhan:
I have addressed all the feedback that I thought was actionable and made judgement calls for some others. Please take a look. Tyvm for the review :)

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: Doesn't render well:
image

Any suggestions?

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1537,37 +1552,28 @@ Otherwise, the token value must be accessed from <xref:System.Text.Json.Utf8Json
...
+        <summary>Gets the raw value of the last processed token as a ReadOnlySpan&lt;byte&gt; slice of the input payload, if the token fits in a single segment or if the reader was constructed with a JSON payload contained in a ReadOnlySpan&lt;byte&gt;.</summary>
         <value>A read-only span of bytes.</value>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
-If <xref:System.Text.Json.Utf8JsonReader.HasValueSequence> is `true`, <xref:System.Text.Json.Utf8JsonReader.ValueSpan> is empty. 
+If the JSON is provided within a ReadOnlySequence&lt;byte&gt; and the slice that represents the token value fits in a single segment, then `ValueSpan` contains the sliced value since it can be represented as a span. Otherwise, <see cref="P:System.Text.Json.Utf8JsonReader.ValueSequence" /> contains the token value.
```
carlossanlop:
```suggestion
If the JSON is provided within a ReadOnlySequence&lt;byte&gt; and the slice that represents the token value fits in a single segment, then `ValueSpan` contains the sliced value since it can be represented as a span. Otherwise, <xref:System.Text.Json.Utf8JsonReader.ValueSequence> contains the token value.
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1500,18 +1509,24 @@ The current depth exceeds the recursive limit set by the maximum depth.</excepti
...
       <Docs>
-        <summary>Gets the value of the last processed token as a <see cref="T:System.ReadOnlySpan`1" /> slice of the input payload. If the JSON is provided within a <see cref="T:System.Buffers.ReadOnlySequence`1" /> and the slice that represents the token value fits in a single segment, then <see cref="P:System.Text.Json.Utf8JsonReader.ValueSpan" /> contains the sliced value since it can be represented as a span. Otherwise, <see cref="P:System.Text.Json.Utf8JsonReader.ValueSequence" /> contains the token value.</summary>
+        <summary>Gets the raw value of the last processed token as a ReadOnlySequence&lt;byte&gt; slice of the input payload, only if the token is contained within multiple segments.</summary>
         <value>A byte read-only sequence.</value>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
+If the JSON is provided within a ReadOnlySequence&lt;byte&gt; and the slice that represents the token value fits in a single segment, then <see cref="P:System.Text.Json.Utf8JsonReader.ValueSpan" /> contains the sliced value since it can be represented as a span. Otherwise, `ValueSequence` contains the token value.
```
carlossanlop:
```suggestion
If the JSON is provided within a ReadOnlySequence&lt;byte&gt; and the slice that represents the token value fits in a single segment, then <xref:System.Text.Json.Utf8JsonReader.ValueSpan> contains the sliced value since it can be represented as a span. Otherwise, `ValueSequence` contains the token value.
```

xml/System.Text.Json/Utf8JsonReader.xml
carlossanlop: Then maybe we should simply use
```suggestion
The `GetBytesFromBase64` method throws an exception if the entire UTF-8 encoded token value cannnot be decoded to a byte array.
```

carlossanlop: @ahsonkhan I added this suggestion after approving. You might want to commit this too.

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -368,10 +371,23 @@ The JSON token value represents a number less than <see cref="F:System.Byte.MinV
...
+
+ ]]></format>
+        </remarks>
         <exception cref="T:System.InvalidOperationException">The value of the JSON token isn't a <see cref="F:System.Text.Json.JsonTokenType.String" />.</exception>
-        <exception cref="T:System.FormatException">The JSON token value cannot be read as a <see cref="T:System.DateTime" />.</exception>
+        <exception cref="T:System.FormatException">The JSON token value cannot be read as a <see cref="T:System.DateTime" />.
+        
+-or-
+        
+The entire UTF-8 encoded token value cannot be parsed to a <xref:System.DateTime> value.</exception>
```
Jozkee: Fixes build error.
```suggestion
The entire UTF-8 encoded token value cannot be parsed to a <see cref="System.DateTime" /> value.</exception>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -392,10 +408,23 @@ The JSON token value represents a number less than <see cref="F:System.Byte.MinV
...
+
+ ]]></format>
+        </remarks>
         <exception cref="T:System.InvalidOperationException">The value of the JSON token isn't a <see cref="F:System.Text.Json.JsonTokenType.String" />.</exception>
-        <exception cref="T:System.FormatException">The JSON token value cannot be read as a <see cref="T:System.DateTimeOffset" />.</exception>
+        <exception cref="T:System.FormatException">The JSON token value cannot be read as a <see cref="T:System.DateTimeOffset" />.
+
+-or-
+        
+The entire UTF-8 encoded token value cannot be parsed to a <xref:System.DateTimeOffset> value.</exception>
```
Jozkee:
```suggestion
The entire UTF-8 encoded token value cannot be parsed to a <see cref="System.DateTimeOffset" /> value.</exception>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -485,7 +498,12 @@ The `GetDouble` method throws an exception if the entire UTF-8 encoded token val
         <returns>The GUID value, if the entire UTF-8 encoded token value can be successfully parsed.</returns>
         <remarks>To be added.</remarks>
         <exception cref="T:System.InvalidOperationException">The value of the JSON token isn't a <see cref="F:System.Text.Json.JsonTokenType.String" />.</exception>
-        <exception cref="T:System.FormatException">The JSON token value is in an unsupported format for a Guid.</exception>
+        <exception cref="T:System.FormatException">The JSON token value is in an unsupported format for a Guid.
+        
+-or-
+
+The entire UTF-8 encoded token value cannot be parsed to a <xref:System.Guid> value.</exception>
```
Jozkee:
```suggestion
The entire UTF-8 encoded token value cannot be parsed to a <see cref="System.Guid" /> value.</exception>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -789,7 +786,11 @@ The JSON token value represents a number less than <see cref="F:System.UInt32.Mi
             
 -or-
 
-The JSON token value represents a number less than <see cref="F:System.UInt64.MinValue" /> or greater than <see cref="F:System.UInt64.MaxValue" />.</exception>
+The JSON token value represents a number less than <see cref="F:System.UInt64.MinValue" /> or greater than <see cref="F:System.UInt64.MaxValue" />
+
+-or-
+
+The entire UTF-8 encoded token value cannot be parsed to a <xref:System.UInt64> value.</exception>
```
Jozkee:
```suggestion
The entire UTF-8 encoded token value cannot be parsed to a <see cref="System.UInt64" /> value.</exception>
```

ahsonkhan:
Anything left to do or is this good to merge?

xml/System.Text.Json/Utf8JsonReader.xml
mairaw: I still prefer the old version here.

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -121,8 +121,8 @@ This constructor assumes that the entire JSON payload is contained in `jsonData`
...
         <param name="jsonData">The UTF-8 encoded JSON text to process.</param>
         <param name="isFinalBlock">
-          <see langword="true" /> if the input sequence contains the entire data to process; <see langword="false" /> only the input span contains partial data with more data to follow.</param>
-        <param name="state">An object that contains the reader state. If this is the first call to the constructor, pass in a default state. Otherwise, capture the state from the previous instance of the <see cref="T:System.Text.Json.Utf8JsonReader" /> and pass that back.</param>
+          <see langword="true" /> to indicate that the input sequence contains the entire data to process; <see langword="false" /> to indicate that the input span contains partial data with more data to follow.</param>
+        <param name="state">An object that contains the reader state. If this is the first call to the constructor, pass in a default state. Otherwise, capture the state from the previous instance of the <see cref="T:System.Text.Json.Utf8JsonReader" /> by getting the value of the <see cref="P:System.Text.Json.Utf8JsonReader.CurrentState" /> property and pass that back.</param>
```
mairaw: I was thinking on my way home on this one 

```suggestion
        <param name="state">An object that contains the reader state. If this is the first call to the constructor, pass the default state; otherwise, pass the value of the <see cref="P:System.Text.Json.Utf8JsonReader.CurrentState" /> property from the previous instance of the <see cref="T:System.Text.Json.Utf8JsonReader" />.</param>
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -368,10 +371,23 @@ The JSON token value represents a number less than <see cref="F:System.Byte.MinV
...
         <summary>Reads the next JSON token value from the source and parses it to a <see cref="T:System.DateTime" />.</summary>
-        <returns>The date and time, if the entire UTF-8 encoded token value can be successfully parsed.</returns>
-        <remarks>This method only creates a DateTime representation of JSON strings that conform to the ISO 8601-1 extended format (see [DateTime and DateTimeOffset support in System.Text.Json](~/docs/standard/datetime/system-text-json-support.md)).</remarks>
+        <returns>The date and time value, if the entire UTF-8 encoded token value can be successfully parsed.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+This method only creates a DateTime representation of JSON strings that conform to the ISO 8601-1 extended format (see [DateTime and DateTimeOffset support in System.Text.Json](~/docs/standard/datetime/system-text-json-support.md)).
```
mairaw:
```suggestion
This method only creates a <xref:System.DateTime> representation of JSON strings that conform to the ISO 8601-1 extended format (see [DateTime and DateTimeOffset support in System.Text.Json](~/docs/standard/datetime/system-text-json-support.md)).
```

xml/System.Text.Json/Utf8JsonReader.xml
```diff
@@ -1537,37 +1548,28 @@ Otherwise, the token value must be accessed from <xref:System.Text.Json.Utf8Json
...
-    <MemberGroup MemberName="ValueTextEquals">
-      <Docs>
-        <summary>Compares the specified text to the unescaped JSON token value in the source and returns a value that indicates whether they match.</summary>
-        <remarks>
-          <format type="text/markdown"><![CDATA[  
+There is no guarantee that this property will always contain well-formed data. If the input JSON passed in to the `Utf8JsonReader` contains invalid UTF-8 bytes within JSON string tokens, this property will return back those invalid UTF-8 bytes as is.
+Therefore, if the input is untrusted or not previously validated, call <xref:System.Text.Json.Utf8JsonReader.GetString> to get the JSON string token.
 
-## Remarks
+Since this property returns the raw bytes, avoid using it for text comparison. Instead call <xref:System.Text.Json.Utf8JsonReader.ValueTextEquals%2A> which will unescape the text if necessary.
```
mairaw:
```suggestion
Since this property returns the raw bytes, avoid using it for text comparison. Instead call <xref:System.Text.Json.Utf8JsonReader.ValueTextEquals%2A>, which unescapes the text if necessary.
```

ahsonkhan:
Changes look good.

xml/System.Text.Json/Utf8JsonReader.xml
ahsonkhan: Looks like I missed this.

PR #3165, opened 2019-09-10
Updating IsControl Char page

IndigoShock:
## Summary
The goal is to make this page more concise and readable
- Deleted an extraneous T in the third Remarks section.
- Deleted the Remarks section at the top because under each example has their own Remarks section.

### DIFF
--- a/xml/System/Char.xml
+++ b/xml/System/Char.xml
@@ -1068,6 +1068,9 @@ When a managed <xref:System.Char> type, which is represented as a Unicode UTF-16
         <remarks>
           <format type="text/markdown"><![CDATA[  
    
+## Remarks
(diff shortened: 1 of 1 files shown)


xml/System/Char.xml
```diff
@@ -1067,10 +1067,7 @@ When a managed <xref:System.Char> type, which is represented as a Unicode UTF-16
         <summary>Indicates whether a specified Unicode character is categorized as a control character.</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[  
-  
-## Remarks
```
rpetrusha: Since this paragraph applies to both overloads, a better option than deleting it is to keep this instance and delete the remaining two. That allows a Remarks section to continue to appear in the mini-TOC, where it's more likely to be noticed. (Sections of the documentation for individual overloads do not appear in the mini-TOC.)

rpetrusha:
Thanks both for contributing to making the documentation more concise, @IndigoShock. I'll merge your PR now.

PR #3179, opened 2019-09-14
Automatically port ReadOnlySpan.GetPinnableReference

carlossanlop:
Summary and return.

### DIFF
--- a/xml/System/ReadOnlySpan`1.xml
+++ b/xml/System/ReadOnlySpan`1.xml
@@ -380,7 +380,7 @@ This method returns `default` when `array` is `null`.
       <Parameters />
       <Docs>
         <summary>A read-only reference to an object of type T that can be used for pinning, and is required to support the use of span within a fixed statement.</summary>
-        <returns>Returns a reference to the 0th element of the span. If the span is empty, returns <see langword="null" /> reference.</returns>
+        <returns>A reference to the element of the span at index 0, or <see langword="null" /> if the span is empty.</returns>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>


xml/System/ReadOnlySpan`1.xml
```diff
@@ -379,8 +379,8 @@ This method returns `default` when `array` is `null`.
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <summary>A read-only reference to an object of type T that can be used for pinning, and is required to support the use of span within a fixed statement.</summary>
```
rpetrusha:
```suggestion
        <summary>A read-only reference to an object of type <typeparamref>T</typeparamref> that can be used for pinning. It is required to support the use of a span within a <code>fixed</code> statement.</summary>
```

xml/System/ReadOnlySpan`1.xml
```diff
@@ -379,8 +379,8 @@ This method returns `default` when `array` is `null`.
...
+        <returns>Returns a reference to the 0th element of the span. If the span is empty, returns <see langword="null" /> reference.</returns>
```
rpetrusha:
```suggestion
        <returns>A reference to the element of the span at index 0, or <see langword="null" /> if the span is empty.</returns>
```

xml/System/ReadOnlySpan`1.xml
carlossanlop: @rpetrusha the build passed, but shouldn't we be using `<typeparamref name="T" />` instead? Or are both valid? I ask because this is the first time I see this format.

rpetrusha: Sorry, @carlossanlop. Yes, we should be using `<typeparmref name="T" />`. I'm not sure what I was thinking.

carlossanlop: no worries!

rpetrusha:
Thanks for the additional changes, @carlossanlop. I'll merge your PR now.

PR #3234, opened 2019-09-24
Noted change of behavior or Parse/TryParse

rpetrusha:
## Noted change of behavior or Parse/TryParse

//cc @tannergooding

### DIFF
--- a/xml/System/Double.xml
+++ b/xml/System/Double.xml
@@ -1,4 +1,4 @@
-<Type Name="Double" FullName="System.Double">	<Type Name="Double" FullName="System
+<Type Name="Double" FullName="System.Double">
   <TypeSignature Language="C#" Value="public struct Double : IComparable, IComparable&lt;double&gt;, IConvertible, IEquatable&lt;double&gt;, IFormattable" />	
   <TypeSignature Language="ILAsm" Value=".class public sequential ansi serializable sealed beforefieldinit float64 extends System.ValueType implements class System.IComparable, class System.IComparable`1&lt;float64&gt;, class System.IConvertible, class System.IEquatable`1&lt;float64&gt;, class System.IFormattable" />	
   <TypeSignature Language="DocId" Value="T:System.Double" />	


xml/System/Double.xml
```diff
@@ -1855,12 +1855,14 @@
...
+ Ordinarily, if you pass the <xref:System.Double.Parse%2A?displayProperty=nameWithType> method a string that is created by calling the <xref:System.Double.ToString%2A?displayProperty=nameWithType> method, the original <xref:System.Double> value is returned. However, because of a loss of precision, the values may not be equal. In addition, attempting to parse the string representation of either <xref:System.Double.MinValue> or <xref:System.Double.MaxValue> fails to round-trip. On .NET Framework and .NET Core 2.2 and previous versions, it throws an <xref:System.OverflowException>. On .NET Core 3.0 and later versions, it returns <xref:System.Double.NegativeInfinity?displayProperty=nameWithType> if you attempt to parse <xref:System.Double.MinValue?displayProperty=nameWithtype> or <xref:System.Double.PositiveInfinity?displayProperty=nameWithType> if you attempt to parse <xref:System.Double.MaxValue?displayProperty=nameWithtype>. The following example provides an illustration.
```
tannergooding: This first part of this statement is no longer correct for the default implementation of `ToString()`.

On .NET Core 3.0 and above, the result of `value.ToString()` or `value.ToString("R")` will always roundtrip. The result of specific format specifiers (e.g. `value.ToString("G15")`) may not roundtrip successfully.

tannergooding: So, namely, the following statements don't apply to .NET Core 3.0 (given the above conditions).
> However, because of a loss of precision, the values may not be equal. In addition, attempting to parse the string representation of either <xref:System.Double.MinValue> or <xref:System.Double.MaxValue> fails to round-trip.

tannergooding: >  On .NET Core 3.0 and later versions, it returns <xref:System.Double.NegativeInfinity?displayProperty=nameWithType> if you attempt to parse <xref:System.Double.MinValue?displayProperty=nameWithtype> or <xref:System.Double.PositiveInfinity?displayProperty=nameWithType> if you attempt to parse <xref:System.Double.MaxValue?displayProperty=nameWithtype>.

This is also not correct. On .NET Core 3.0 these values successfully roundtrip by default. Values greater or lesser than these will round to positive/negative infinity (but not strictly speaking all; there is a small subset of values greater/lesser than min/max that are considered "closer" to max/min value than +/- infinity and which will return min/max instead).

xml/System/Double.xml
```diff
@@ -1855,12 +1855,14 @@
...
-  
- If a separator is encountered in the `s` parameter during a parse operation, and the applicable currency or number decimal and group separators are the same, the parse operation assumes that the separator is a decimal separator rather than a group separator. For more information about separators, see <xref:System.Globalization.NumberFormatInfo.CurrencyDecimalSeparator%2A>, <xref:System.Globalization.NumberFormatInfo.NumberDecimalSeparator%2A>, <xref:System.Globalization.NumberFormatInfo.CurrencyGroupSeparator%2A>, and <xref:System.Globalization.NumberFormatInfo.NumberGroupSeparator%2A>.  
+
+If `s` is out of range of the <xref:System.Double> data type, the method throws an <xref:System.OverflowException> on .NET Framework and .NET Core 2.2 and earlier versions. On .NET Core 3.0 and later versions, it returns <xref:System.Double.NegativeInfinity?displayProperty=nameWithType> if `s` is less than <xref:System.Double.MinValue?displayProperty=nameWithType> and <xref:System.Double.PositiveInfinity?displayProperty=nameWithType> if `s` is greater than <xref:System.Double.MaxValue?displayProperty=nameWithType>.
```
tannergooding: The cutoff is actually `double.MaxValue + (double.MaxValue - Math.BitDecrement(double.MaxValue))`, which is approx: `1.7976931348623159E+308`

Values less than this will return `double.MaxValue`. Values greater than this will return `double.PositiveInfinity`.

The same, but inverse values (so less than approx. `1.7976931348623159E+308`) apply to `double.MinValue` and `double.NegativeInfinity`.

tannergooding: Should there be a similar comment about underflow to `0`? That is, values parsed to be less than `double.Epsilon` will round to exactly zero (with some similar math as given above to account for closer to epsilon vs closer to zero).

gewarren: Since Ron has retired and in the interest of getting this PR merged, I'm not planning to add this underflow explanation. If you feel strongly that it should be included, please let me know.

tannergooding: Could you please log a bug for it and assign it to me. Since this falls under my ownership area, I can track getting a PR up to fixup the documentation further.

gewarren: Sure, done at https://github.com/dotnet/dotnet-api-docs/issues/3538.

PR #3243, opened 2019-09-26
L124989: Fixing CDATA formatting

v-maudel:
Description: the Markdown CDATA content is incorrectly formatted in source files, moreover it is breaking the target publishing builds since the content inside those tags are being processed as XML tag instead of Markdown content.


v-maudel:
Hi @rpetrusha, @mairaw , could you please approve this PR?
Thanks in advance.

v-maudel:
Thanks @mairaw for approve

PR #3313, opened 2019-10-09
Update BaseType (line 30)

billchungiii:
According to IL code, EventHandler's base type should be System.MulticastDelegate

IL Code as below:
.class public auto ansi sealed System.EventHandler

### DIFF
--- a/xml/System/EventHandler.xml
+++ b/xml/System/EventHandler.xml
@@ -27,7 +27,7 @@
     <AssemblyVersion>2.1.0.0</AssemblyVersion>
   </AssemblyInfo>
   <Base>
-    <BaseTypeName>System.Delegate</BaseTypeName>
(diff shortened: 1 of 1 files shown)


rpetrusha:
Thank you, @billchungiii, for pointing this out and opening this PR. 

This is an issue that applies not only to the Predicate\<T> type and the two other delegates for which you've opened PRs, but to all delegates in the .NET API reference as a whole. Information on inheritance is also not authored content, but is gathered using reflection. That means that any changes made to inheritance information will be overwritten when the API reference is next updated using reflection. 

@mimisasouvanh @dend and @joelmartinez, could you take a look at this issue, please?

We'll investigate why MulticastDelegate is skiipped in the inheritance hierarchy, @billchungiii. But given that a manual modification of a delegate's base type does not offer a long-term solution, I'm going to close this and your two other PRs.

PR #3314, opened 2019-10-09
Update BaseTypeName (line 32)

billchungiii:
According to IL code,  System.EventHandler`1<TEventArgs>'s base type should be System.MulticastDelegate

IL Code as below:
.class public auto ansi sealed System.EventHandler`1<TEventArgs>


rpetrusha:
See #3313.

PR #3315, opened 2019-10-09
Update BaseTypeName (line 36)

billchungiii:
According to IL code, System.Predicate`1<- T>'s base type should be System.MulticastDelegate.

IL Code as below:
.class public auto ansi sealed System.Predicate`1<- T>


rpetrusha:
See #3313.

PR #3338, opened 2019-10-13
Update Array.CreateInstance docs to mention language support of non-zero based arrays

Gnbrkm41:
## Summary

* Non-zero based arrays are not supported in all major languages (C#, VB, F# etc...). This commit adds a note to the `System.Array.CreateInstance(System.Type,System.Int32[],System.Int32[])` to mention that not all languages support non-zero based arrays with a brief C# example.

Fixes https://github.com/dotnet/dotnet-api-docs/issues/901.

### DIFF
--- a/xml/System/Array.xml
+++ b/xml/System/Array.xml
@@ -2464,8 +2464,8 @@
   
  This method is an O(`n`) operation, where `n` is the product of all values in `lengths`.  
   
-[!NOTE]
- Not all languages support arrays with non-zero lower bounds, and therefore you may not be able to cast instances of non-zero based <xref:System.Array> to the language's array type. For example, you cannot cast a 1-dimensional integer array with the lower bound 6 to C#'s `int[]` type. This results in <xref:System.InvalidCastException> during runtime with the message "Unable to cast object of type 'System.Int32[*]' to type 'System.Int32[]'.", where the asterisk (*) means non-zero based index. However, you can cast zero based arrays of any rank created with <xref:System.Array.CreateInstance(System.Type,System.Int32[],System.Int32[])> to the language's array. For example, you can cast a 2-dimensional zero based integer array created with this method to C#'s `int[,]` type.
+> [!NOTE]
+> Not all languages support arrays with non-zero lower bounds, and therefore you may not be able to cast instances of non-zero based <xref:System.Array> to the language's array type. For example, you cannot cast a 1-dimensional integer array with the lower bound 6 to C#'s `int[]` type. This results in <xref:System.InvalidCastException> during runtime with the message "Unable to cast object of type 'System.Int32[*]' to type 'System.Int32[]'.", where the asterisk (*) means non-zero based index. However, you can cast zero based arrays of any rank created with <xref:System.Array.CreateInstance(System.Type,System.Int32[],System.Int32[])> to the language's array. For example, you can cast a 2-dimensional zero based integer array created with this method to C#'s `int[,]` type.
 
 ## Examples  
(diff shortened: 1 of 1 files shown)


xml/System/Array.xml
```diff
@@ -2464,8 +2464,9 @@
   
  This method is an O(`n`) operation, where `n` is the product of all values in `lengths`.  
   
-   
-  
+[!NOTE]
+ Not all languages support arrays with non-zero lower bounds, and therefore you may not be able to cast instances of non-zero based <xref:System.Array> to the language's array type. For example, you cannot cast a 1-dimensional integer array with the lower bound 6 to C#'s `int[]` type. This results in <xref:System.InvalidCastException> during runtime with the message "Unable to cast object of type 'System.Int32[*]' to type 'System.Int32[]'.", where the asterisk (*) means non-zero based index. However, you can cast zero based arrays of any rank created with <xref:System.Array.CreateInstance(System.Type,System.Int32[],System.Int32[])> to the language's array. For example, you can cast a 2-dimensional zero based integer array created with this method to C#'s `int[,]` type.
```
Gnbrkm41: I'm not sure if `<xref:System.Array.CreateInstance(System.Type,System.Int32[],System.Int32[])>` in middle of markdowns will be interpreted as reference to the specific method. Can someone confirm?

Gnbrkm41: In addition, I also wasn't sure if line breaks in the source code will affect how this paragraph will be rendered; would it show up as one paragraph with no line breaks even if I add some line breaks in this?

carlossanlop:
@stephentoub @jkotas @marek-safar since you've been the main modifiers of [Array.CoreCLR.cs](https://github.com/dotnet/runtime/blob/master/src/coreclr/src/System.Private.CoreLib/src/System/Array.CoreCLR.cs), can you please help review these new remarks for this overload of `CreateInstance`?

```cs
public static Array CreateInstance (Type elementType, int[] lengths, int[] lowerBounds)
```

xml/System/Array.xml
```diff
@@ -2464,8 +2464,9 @@
   
  This method is an O(`n`) operation, where `n` is the product of all values in `lengths`.  
   
-   
-  
+[!NOTE]
```
jkotas: The paragraph above says "Generally, the .NET Framework class library and many programming languages do not handle nonzero lower bounds.". This is rehearsing the same fact and adds an example.

Are we happy with the duplication?

Thraka: Missing the `>` syntax for the note.

```suggestion
> [!NOTE]
```

xml/System/Array.xml
Thraka: Missing the `>` syntax for the note.

```suggestion
> Not all languages support arrays with non-zero lower bounds, and therefore you may not be able to cast instances of non-zero based <xref:System.Array> to the language's array type. For example, you cannot cast a 1-dimensional integer array with the lower bound 6 to C#'s `int[]` type. This results in <xref:System.InvalidCastException> during runtime with the message "Unable to cast object of type 'System.Int32[*]' to type 'System.Int32[]'.", where the asterisk (*) means non-zero based index. However, you can cast zero based arrays of any rank created with <xref:System.Array.CreateInstance(System.Type,System.Int32[],System.Int32[])> to the language's array. For example, you can cast a 2-dimensional zero based integer array created with this method to C#'s `int[,]` type.
```

Thraka:
@Gnbrkm41 some fixes for you to merge into your copy. The xref should resolve correctly, we can check that in the build after you commit my suggestions.

PR #3355, opened 2019-10-16
Update Docs for DateTime.Ticks to explain it depends on the time zone of the object instance.

AWilco:
## Summary

Corrected documentation of Ticks property to include information that it is related to the number of ticks since 0:00:00 on January 1, 0001 in the time zone of the the DateTime instance, and not always UTC. Thus using ToUniversalTime() and ToLocalTime() can change the number of ticks, even though it represents the same instant in time.

As discussed here: https://social.msdn.microsoft.com/Forums/vstudio/en-US/fde7e5b0-e2b9-4d3b-8a63-c2ae75e316d8/datetimeticks-bug?forum=netfxbcl
And tested on .NET 4.5.1.

If this has changed in the latest .NET runtimes, the documentation says that it has not and is the same functionality as 4.5.1

Not known to be an open issue.

### DIFF
--- a/xml/System/DateTime.xml
+++ b/xml/System/DateTime.xml
@@ -6827,7 +6827,7 @@ The following example demonstrates the <xref:System.DateTime.Parse%28System.Stri
   
 The value of this property represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001 in the Gregorian calendar, which represents <xref:System.DateTime.MinValue>. It does not include the number of ticks that are attributable to leap seconds.
 If the DateTime object has its `Kind` property set to `Local`, its ticks represent the time elapsed time since 12:00:00 midnight, January 1, 0001 in the local time as specified by the current time zone setting.
-If the DateTime object has Kind property set to Utc, that means the ticks inside this object is representing the time elapsed time since 12:00:00 midnight, January 1, 0001 in the Coordinated Universal Time.
+If the `DateTime` object has its `Kind` property set to `Utc`, its ticks represent the time elapsed time since 12:00:00 midnight, January 1, 0001 in the Coordinated Universal Time.
 If the `DateTime` object has its `Kind` property set to `Unspecified`, its ticks represent the time elapsed time since 12:00:00 midnight, January 1, 0001 in the unknown time zone.
 
 In general, the ticks represent the time according to the time zone specified by the `Kind` property.


Thraka:
@AWilco Thanks a bunch for correcting this. I'll get someone to double check it. 👍

xml/System/DateTime.xml
```diff
@@ -6825,9 +6825,11 @@ The following example demonstrates the <xref:System.DateTime.Parse%28System.Stri
 ## Remarks  
  A single tick represents one hundred nanoseconds or one ten-millionth of a second. There are 10,000 ticks in a millisecond, or 10 million ticks in a second.  
   
- The value of this property represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001 (0:00:00 UTC on January 1, 0001, in the Gregorian calendar), which represents <xref:System.DateTime.MinValue?displayProperty=nameWithType>. It does not include the number of ticks that are attributable to leap seconds.  
+ The value of this property represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001 (0:00:00 on January 1, 0001, in the Gregorian calendar, in the current time zone), which represents <xref:System.DateTime.MinValue?displayProperty=nameWithType>.It does not include the number of ticks that are attributable to leap seconds.
```
tarekgh: >in the current time zone) [](start = 193, length = 25)

I don't think this is true, DateTime.Kind can specify Utc, Local, or Unspecified. so this depend on Kind property inside the DateTiem object

xml/System/DateTime.xml
```diff
@@ -6825,9 +6825,11 @@ The following example demonstrates the <xref:System.DateTime.Parse%28System.Stri
 ## Remarks  
  A single tick represents one hundred nanoseconds or one ten-millionth of a second. There are 10,000 ticks in a millisecond, or 10 million ticks in a second.  
   
- The value of this property represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001 (0:00:00 UTC on January 1, 0001, in the Gregorian calendar), which represents <xref:System.DateTime.MinValue?displayProperty=nameWithType>. It does not include the number of ticks that are attributable to leap seconds.  
+ The value of this property represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001 (0:00:00 on January 1, 0001, in the Gregorian calendar, in the current time zone), which represents <xref:System.DateTime.MinValue?displayProperty=nameWithType>.It does not include the number of ticks that are attributable to leap seconds.  
   
-   
+ As the starting instant of time is dependent on the local time zone, changing time zone using, for example, .ToUniversalTime() will potentially change the number of ticks, although the new value will represent the same instant in time.
```
tarekgh: > As the starting instant of time is dependent on the local time zone [](start = 0, length = 68)

This is not true either, you can create a DateTime with Kind value Unspecified which not dependent at all on the local time zone

AWilco:
@tarekgh I appreciate that what I have written might not be the factual truth of how DateTime works.

What I intended to convey was that the Ticks property is NOT always the number of ticks since 0:00:00 **UTC** on January 1, 0001. As I understand it, given limited testing myself and reviewing the discussion thread I linked to, if a timezone is set then the Ticks property is the number of ticks since 0:00:00 on January 1, 0001 in that timezone.

Practically what I was attempting to solve was that when I read that description, I then assume that if I take a specific instant in time represented in a DateTime Object, e.g. 2019-10-01T00:00:00UTC
and change the timezone used to represent that instant, e.g. to
2019-10-01T01:00:00UTC+1

Then the Ticks property would stay the same, as the elapsed time between either of those dates and  0:00:00 **UTC** on January 1, 0001. However this is not true.

As a test:
```
            var dateInUtc = new DateTime(1, 1, 1, 1, 0, 0, DateTimeKind.Utc);
            var ticksInUtc = dateInUtc.Ticks;
            var dateInLocal = dateInUtc.ToLocalTime(); //My local time zone is London time, as of writing it is UTC+1
            var ticksInLocal = dateInLocal.Ticks;
            Assert(ticksInUtc - ticksInLocal == 0); //This fails
```

My new description may not be 100% accurate, I don't know enough about DateTime to ensure what I write is fully accurate, I would appreciate help in improving my new text.

Regarding what happens if you use DateTime.Unspecified, converting to UTC or to local appears to avoid adjusting the ticks at all, and just changes the DateTimeKind property. This means that the instant of time that is represented may be different, but the Ticks property will stay the same.

tarekgh:
@AWilco Thanks for tying to clarify the docs. what about the following? do you think it is good enough or we need to specify more clarification? 

```
The value of this property represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001 in the Gregorian calendar, which represents DateTime.MinValue. It does not include the number of ticks that are attributable to leap seconds.
If the DateTime object has Kind property set to Local, that means the ticks inside this object is representing the time elapsed time since 12:00:00 midnight, January 1, 0001 in the local time specified by the current time zone setting.
If the DateTime object has Kind property set to Utc, that means the ticks inside this object is representing the time elapsed time since 12:00:00 midnight, January 1, 0001 in the Coordinated Universal Time.
If the DateTime object has Kind property set to Unspecified, that means the ticks inside this object is representing the time elapsed time since 12:00:00 midnight, January 1, 0001 in some time zone time which not necessary match Local not Utc time.

In general, the ticks represent the time according to the time zone specified by the Kind property.
```

AWilco:
> @AWilco Thanks for tying to clarify the docs. what about the following? do you think it is good enough or we need to specify more clarification?
> 
> ```
> The value of this property represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001 in the Gregorian calendar, which represents DateTime.MinValue. It does not include the number of ticks that are attributable to leap seconds.
> If the DateTime object has Kind property set to Local, that means the ticks inside this object is representing the time elapsed time since 12:00:00 midnight, January 1, 0001 in the local time specified by the current time zone setting.
> If the DateTime object has Kind property set to Utc, that means the ticks inside this object is representing the time elapsed time since 12:00:00 midnight, January 1, 0001 in the Coordinated Universal Time.
> If the DateTime object has Kind property set to Unspecified, that means the ticks inside this object is representing the time elapsed time since 12:00:00 midnight, January 1, 0001 **in the same unspecified time zone**
> 
> In general, the ticks represent the time according to the time zone specified by the Kind property.
> ```

HI Tarek,

Yes, I think what you have is good, although your words about an unspecified time zone seemed to suggest it would be random. In actuality, if you had the timezone as unspecified, then the whole issue is moot, because there isn't any way to take into account a time zone. The starting instant of 12:00:00 midnight, January 1, 0001 is in the same unspecified time zone. You could aso say that it assumes the time is in UTC for the purposes of this property.

Either way, what you have proposed is an improvement and doesn't introduce any new inaccuracies.

Thraka:
@AWilco let me know when you make the changes with the new suggestions. Thanks.

tarekgh:
@AWilco thanks for the feedback.

> You could aso say that it assumes the time is in UTC for the purposes of this property.

We cannot be specific about this assumption here. There are some contexts that unspecified treated as UTC and in other contexts treated as local. 
does it make sense if we changed ```in the same unspecified time zone``` to ```in the unknown time zone```?

Feel free to suggest something different. My goal is to have anyone reading to understand how to use it. thanks again.

carlossanlop:
@AWilco do you have an update on the feedback received on this PR?

AWilco:
@tarekgh @carlossanlop I've updated the change with the new text as discussed.

xml/System/DateTime.xml
```diff
@@ -6825,9 +6825,13 @@ The following example demonstrates the <xref:System.DateTime.Parse%28System.Stri
 ## Remarks  
  A single tick represents one hundred nanoseconds or one ten-millionth of a second. There are 10,000 ticks in a millisecond, or 10 million ticks in a second.  
   
- The value of this property represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001 (0:00:00 UTC on January 1, 0001, in the Gregorian calendar), which represents <xref:System.DateTime.MinValue?displayProperty=nameWithType>. It does not include the number of ticks that are attributable to leap seconds.  
-  
-   
+The value of this property represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001 in the Gregorian calendar, which represents DateTime.MinValue. It does not include the number of ticks that are attributable to leap seconds.
```
carlossanlop: MinValue should be an xref:
```suggestion
The value of this property represents the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001 in the Gregorian calendar, which represents <xref:System.DateTime.MinValue>. It does not include the number of ticks that are attributable to leap seconds.
```

xml/System/DateTime.xml
```diff
@@ -6825,9 +6825,13 @@ The following example demonstrates the <xref:System.DateTime.Parse%28System.Stri
...
+If the DateTime object has Kind property set to Local, that means the ticks inside this object is representing the time elapsed time since 12:00:00 midnight, January 1, 0001 in the local time specified by the current time zone setting.
```
carlossanlop: Some minor suggestions to ensure they are formatted correctly:
```suggestion
If the `DateTime` object has its `Kind` property set to `Local`, that means its ticks represent the time elapsed time since 12:00:00 midnight, January 1, 0001 in the local time specified by the current time zone setting.
```
@Thraka @mairaw could double check these language suggestions I am providing.

xml/System/DateTime.xml
```diff
@@ -6825,9 +6825,13 @@ The following example demonstrates the <xref:System.DateTime.Parse%28System.Stri
...
+If the DateTime object has Kind property set to Utc, that means the ticks inside this object is representing the time elapsed time since 12:00:00 midnight, January 1, 0001 in the Coordinated Universal Time.
```
carlossanlop:
```suggestion
If the `DateTime` object has its `Kind` property set to `Utc`, that means its ticks represent the time elapsed time since 12:00:00 midnight, January 1, 0001 in the Coordinated Universal Time.
```

xml/System/DateTime.xml
```diff
@@ -6825,9 +6825,13 @@ The following example demonstrates the <xref:System.DateTime.Parse%28System.Stri
...
+If the DateTime object has Kind property set to Unspecified, that means the ticks inside this object is representing the time elapsed time since 12:00:00 midnight, January 1, 0001 in the unknown time zone.
```
carlossanlop:
```suggestion
If the `DateTime` object has its `Kind` property set to `Unspecified`, that means its ticks represent the time elapsed time since 12:00:00 midnight, January 1, 0001 in the unknown time zone.
```

xml/System/DateTime.xml
```diff
@@ -6825,9 +6825,13 @@ The following example demonstrates the <xref:System.DateTime.Parse%28System.Stri
...
+
+In general, the ticks represent the time according to the time zone specified by the Kind property.
```
carlossanlop:
```suggestion
In general, the ticks represent the time according to the time zone specified by the `Kind` property.
```

xml/System/DateTime.xml
gewarren:
```suggestion
If the DateTime object has its `Kind` property set to `Local`, its ticks represent the time elapsed time since 12:00:00 midnight, January 1, 0001 in the local time as specified by the current time zone setting.
```

xml/System/DateTime.xml
gewarren:
```suggestion
If the `DateTime` object has its `Kind` property set to `Utc`, its ticks represent the time elapsed time since 12:00:00 midnight, January 1, 0001 in the Coordinated Universal Time.
```

xml/System/DateTime.xml
gewarren:
```suggestion
If the `DateTime` object has its `Kind` property set to `Unspecified`, its ticks represent the time elapsed time since 12:00:00 midnight, January 1, 0001 in the unknown time zone.
```

AWilco:
Thanks, I believe I have accepted all the proposed changes.

xml/System/DateTime.xml
Thraka: @awilco this one was resolved but not applied. Can you apply it? I think we can merge this

Thraka:
@AWilco Awesome! Thanks a bunch for the updates.

PR #3379, opened 2019-10-19
Document System.IO.Compression.Brotli.BrotliDecoder

carlossanlop:
As described in the API proposal https://github.com/dotnet/corefx/issues/25785 and the PR that introduced this code https://github.com/dotnet/corefx/pull/26229 .

### DIFF
--- a/xml/System.IO.Compression/BrotliDecoder.xml
+++ b/xml/System.IO.Compression/BrotliDecoder.xml
@@ -22,7 +22,7 @@
     </Interface>
   </Interfaces>
   <Docs>
-    <summary>Provides non-allocating, performant single-pass Brotli decompression methods that are independent from the <see cref="T:System.IO.Compression.Brotli.BrotliDecoder" /> instance.</summary>
+    <summary>Provides non-allocating, performant Brotli decompression methods. The methods decompress in a single pass without using a <see cref="T:System.IO.Compression.Brotli.BrotliStream" /> instance.</summary>
     <remarks>To be added.</remarks>
   </Docs>
   <Members>
@@ -52,13 +52,25 @@
         <Parameter Name="bytesWritten" Type="System.Int32" RefType="out" />
       </Parameters>
       <Docs>
-        <param name="source">The byte read-only span containing the compressed data.</param>
+        <param name="source">A buffer containing the compressed data.</param>
         <param name="destination">When this method returns, a byte span containing the decompressed data.</param>
         <param name="bytesConsumed">The total number of bytes that were read from <paramref name="source" />.</param>
         <param name="bytesWritten">The total number of bytes that were written in the <paramref name="destination" />.</param>
         <summary>Decompresses data that was compressed using the Brotli algorithm.</summary>
-        <returns>One of the enumeration values that can be returned from span-based operations that support processing of input contained in multiple discontiguous buffers.</returns>
-        <remarks>To be added.</remarks>
+        <returns>One of the enumeration values that indicates the status of the decompression operation.</returns>
+         <remarks>
+           <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+The return value can be as follows:
+- <xref:System.Buffers.OperationStatus.Done?displayProperty=nameWithType>: `source` was successfully and completely decompressed into `destination`.
+- <xref:System.Buffers.OperationStatus.DestinationTooSmall?displayProperty=nameWithType>: There is not enough space in `destination` to decompress `source`.
+- <xref:System.Buffers.OperationStatus.NeedMoreData?displayProperty=nameWithType>: The decompression action is partially done at least one more byte is required to complete the decompression task. This method should be called again with more input to decompress.
+- <xref:System.Buffers.OperationStatus.InvalidData?displayProperty=nameWithType>: The data in `source` is invalid and could not be decompressed.
+
+           ]]></format>
+         </remarks>
       </Docs>
     </Member>
     <Member MemberName="Dispose">
@@ -85,7 +97,7 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>Releases all resources used by the current brotli decoder instance.</summary>
+        <summary>Releases all resources used by the current Brotli decoder instance.</summary>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
@@ -114,12 +126,20 @@
         <Parameter Name="bytesWritten" Type="System.Int32" RefType="out" />
       </Parameters>
       <Docs>
-        <param name="source">The byte read-only span containing the compressed data.</param>
+        <param name="source">A buffer containing the compressed data.</param>
         <param name="destination">When this method returns, a byte span containing the decompressed data.</param>
         <param name="bytesWritten">The total number of bytes that were written in the <paramref name="destination" />.</param>
         <summary>Attempts to decompress data that was compressed with the Brotli algorithm.</summary>
         <returns><see langword="true" /> on success; <see langword="false" /> otherwise.</returns>
-        <remarks>To be added.</remarks>
+         <remarks>
+           <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+If this method returns `false`, `destination` may be empty or contain partially decompressed data, with `bytesWritten` being zero or greater than zero but less than the expected total.
+
+           ]]></format>
+         </remarks>
       </Docs>
     </Member>
   </Members>


xml/System.IO.Compression/BrotliDecoder.xml
```diff
@@ -52,12 +52,12 @@
         <Parameter Name="bytesWritten" Type="System.Int32" RefType="out" />
       </Parameters>
       <Docs>
-        <param name="source">To be added.</param>
-        <param name="destination">To be added.</param>
-        <param name="bytesConsumed">To be added.</param>
-        <param name="bytesWritten">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="source">The byte read-only span containing the compressed data.</param>
```
tdykstra: Since the parameter type is already shown, does it need to be repeated here, i.e., could this be simply "The compressed data."? Same comment on the following parameter.

xml/System.IO.Compression/BrotliDecoder.xml
```diff
@@ -85,7 +85,7 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Releases all resources used by the current brotli decoder instance.</summary>
```
tdykstra:
```suggestion
        <summary>Releases all resources used by the current Brotli decoder instance.</summary>
```

xml/System.IO.Compression/BrotliDecoder.xml
```diff
@@ -114,11 +114,11 @@
         <Parameter Name="bytesWritten" Type="System.Int32" RefType="out" />
       </Parameters>
       <Docs>
-        <param name="source">To be added.</param>
-        <param name="destination">To be added.</param>
-        <param name="bytesWritten">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="source">The byte read-only span containing the compressed data.</param>
```
tdykstra: Same comment on parameter type as noted above.

xml/System.IO.Compression/BrotliDecoder.xml
```diff
@@ -22,7 +22,7 @@
     </Interface>
   </Interfaces>
   <Docs>
-    <summary>To be added.</summary>
+    <summary>Provides non-allocating, performant single-pass Brotli decompression methods that are independent from the <see cref="T:System.IO.Compression.Brotli.BrotliDecoder" /> instance.</summary>
```
ahsonkhan: > `that are independent from the <see cref="T:System.IO.Compression.Brotli.BrotliDecoder" /> instance`

What does this mean?

xml/System.IO.Compression/BrotliDecoder.xml
```diff
@@ -52,12 +52,12 @@
...
-        <param name="bytesConsumed">To be added.</param>
-        <param name="bytesWritten">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="source">The byte read-only span containing the compressed data.</param>
+        <param name="destination">When this method returns, a byte span containing the decompressed data.</param>
+        <param name="bytesConsumed">The total number of bytes that were read from <paramref name="source" />.</param>
+        <param name="bytesWritten">The total number of bytes that were written in the <paramref name="destination" />.</param>
+        <summary>Decompresses data that was compressed using the Brotli algorithm.</summary>
+        <returns>One of the enumeration values that can be returned from span-based operations that support processing of input contained in multiple discontiguous buffers.</returns>
```
ahsonkhan: I think we should converge on the documentation for the "OperationStatus" based APIs that process a source into a destination span.

`that support processing of input contained in multiple discontiguous buffers.` <= this seems like unnecessary context.

Personally, I like the Bas64 API/parameter descriptions. Let's use that as a template (including for parameters like bytesWritten, bytesConsumed):
https://docs.microsoft.com/en-us/dotnet/api/system.buffers.text.base64.decodefromutf8?view=netcore-3.0#System_Buffers_Text_Base64_DecodeFromUtf8_System_ReadOnlySpan_System_Byte__System_Span_System_Byte__System_Int32__System_Int32__System_Boolean_

xml/System.IO.Compression/BrotliDecoder.xml
```diff
@@ -52,12 +52,12 @@
...
         <remarks>To be added.</remarks>
```
ahsonkhan: We should add remarks for what happens for the semantics of what it means for different `OperationStatus` being returned.

xml/System.IO.Compression/BrotliDecoder.xml
```diff
@@ -52,12 +52,12 @@
...
       </Parameters>
       <Docs>
-        <param name="source">To be added.</param>
-        <param name="destination">To be added.</param>
-        <param name="bytesConsumed">To be added.</param>
-        <param name="bytesWritten">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="source">The byte read-only span containing the compressed data.</param>
+        <param name="destination">When this method returns, a byte span containing the decompressed data.</param>
```
ahsonkhan: **Question:** Can the source/destination buffers overlap (or even be the same span)? Worth testing and adding a remark.

xml/System.IO.Compression/BrotliDecoder.xml
```diff
@@ -114,11 +114,11 @@
...
-        <param name="destination">To be added.</param>
-        <param name="bytesWritten">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="source">The byte read-only span containing the compressed data.</param>
+        <param name="destination">When this method returns, a byte span containing the decompressed data.</param>
+        <param name="bytesWritten">The total number of bytes that were written in the <paramref name="destination" />.</param>
+        <summary>Attempts to decompress data that was compressed with the Brotli algorithm.</summary>
+        <returns><see langword="true" /> on success; <see langword="false" /> otherwise.</returns>
         <remarks>To be added.</remarks>
```
ahsonkhan: What do the parameters/destination contain when we return false? Are there any guarantees we provide about the contents of source never changing, and the contents of destination only changing if we return true? Or is the destination overwritten even for false?

Work adding those remarks.

xml/System.IO.Compression/BrotliDecoder.xml
carlossanlop: Gahh I made a mistake. Thanks for pointing it out. I meant `BrotliStream`, not `BrotliDecoder`.

I got this explanation from the [API proposal](https://github.com/dotnet/corefx/issues/25785), but shorter:

----
> ### BrotliEncoder & BrotliDecoder
> #### Proposed API
> The goal of the streamless implementation is to provide a non-allocating, performant Brotli implementation free from Streams. It contains simple Compress/Decompress operations that return an enum indicating the success of the operation as well as static CompressFully/DecompressFully operations that allow single-pass compression/decompression without the need for a BrotliEncoder/BrotliDecoder instance.
----

How about:
```suggestion
    <summary>Provides non-allocating, performant and single-pass Brotli decompression methods that are free from streams; that is, they do not depend on creating a <see cref="T:System.IO.Compression.Brotli.BrotliStream " /> instance.</summary>
```

xml/System.IO.Compression/BrotliDecoder.xml
carlossanlop: I see what you mean, makes sense. Other similar methods, like [DeflateStream.BeginWrite](https://docs.microsoft.com/en-us/dotnet/api/system.io.compression.deflatestream.beginwrite?view=netcore-3.0) or [GZipStream.Write](https://docs.microsoft.com/en-us/dotnet/api/system.io.compression.gzipstream.write?view=netcore-3.0#System_IO_Compression_GZipStream_Write_System_Byte___System_Int32_System_Int32_), don't really mention the data type, only that it's a buffer.

How about:

```suggestion
        <param name="source">A buffer containing the compressed data.</param>
```

xml/System.IO.Compression/BrotliDecoder.xml
carlossanlop: Does this sound good?:

```suggestion
        <returns>One of the enumeration values that indicates the status of the decoding operation.</returns>
```

xml/System.IO.Compression/BrotliDecoder.xml
carlossanlop: Does this sound good?:

```suggestion
         <remarks>
           <format type="text/markdown"><![CDATA[

## Remarks

The return value can be as follows:
- <xref:System.Buffers.OperationStatus.Done?displayProperty=nameWithType>: Processing of the entire input span succeeded.
- <xref:System.Buffers.OperationStatus.DestinationTooSmall?displayProperty=nameWithType>: There is not enough space in the output span to write the decoded input.
- <xref:System.Buffers.OperationStatus.NeedMoreData?displayProperty=nameWithType>: `isFinalBlock` is `false` and the input is not a multiple of 4. Otherwise, the partial input is considered <xref:System.Buffers.OperationStatus.InvalidData>.
- <xref:System.Buffers.OperationStatus.InvalidData?displayProperty=nameWithType>: The input contains bytes outside of the expected base 64 range, has invalid or more than two padding characters, or is incomplete (that is, not a multiple of 4) and `isFinalBlock` is `true`.
           ]]></format>
         </remarks>
```

xml/System.IO.Compression/BrotliDecoder.xml
carlossanlop: If you agree with the one above, I hope you agree with this one too, since it's the same text 😄 :

```suggestion
        <param name="source">A buffer containing the compressed data.</param>
```

xml/System.IO.Compression/BrotliDecoder.xml
carlossanlop: > _Are there any guarantees we provide about the contents of source never changing_

`source` is a read-only span, so I'd be really worried if it changed.

> _and the contents of destination only changing if we return true? Or is the destination overwritten even for false?_

Apparently, we can return `false` with a partially decompressed buffer.
See: [corefx/src/Native/AnyOS/brotli/dec/decode.c#L1287](https://github.com/dotnet/corefx/blob/6ad08fec6cd1fcb7348dfe60fda775ca69476a8d/src/Native/AnyOS/brotli/dec/decode.c#L1287)

Here's what I suggest adding to the remarks. Do you agree?:

```suggestion
         <remarks>
           <format type="text/markdown"><![CDATA[

## Remarks

If this method returns `false`, `destination` may be empty or contain partially decompressed data, with `bytesWritten` being zero or greater than zero but less than the expected total.

           ]]></format>
         </remarks>
```

ahsonkhan: > ... with `bytesWritten` being zero or greater than zero but less than the expected total.

If that's the behavior, we should fix it in source (if possible). Can you test it out and verify that bytesWritten is != zero for any case where we return false? That doesn't follow the "try" pattern. Generally, in the "try" pattern, we set the out parameter to default before returning.

For example:
https://github.com/dotnet/corefx/blob/0954a975f972546314bd3ac4eb6e0a6380a85701/src/System.Text.Json/src/System/Text/Json/Reader/Utf8JsonReader.TryGet.cs#L614-L623

Regardless of the behavior, I wouldn't document it as such. Try returning false means the out parameters contain undefined/meaningless info and shouldn't be relied upon.

xml/System.IO.Compression/BrotliDecoder.xml
ahsonkhan: We aren't decoding here, we are decompressing. I would tailor the comments to be specific to this API rather than copying it as-is.

xml/System.IO.Compression/BrotliDecoder.xml
ahsonkhan: There is no is final block parameter in this API and none of the base-64 specific details make sense here. I wouldn't copy it as is, but rather update it to be context specific to the decompression API.

Take a look at the implementation/tests to reason about the behavior.
https://source.dot.net/#System.IO.Compression.Brotli/System/IO/Compression/dec/BrotliDecoder.cs,f2d7661515181718

xml/System.IO.Compression/BrotliDecoder.xml
ahsonkhan: I don't the current summary is correctly phrased (for example, I wouldn't say "free from streams" - reading it from an end-user perspective, I would ask myself, what does this mean?). Taking snippets from the API proposal is exposing design discussion level detail in an API documentation/description which doesn't fit the audience of docs, imo.

When would I use this API over the ones on BrotliStream?

carlossanlop: > _I wouldn't say "free from streams" - reading it from an end-user perspective, I would ask myself, what does this mean?_

I think you're going to the other extreme, which is assuming a user that wants to use this API is completely ignorant/unfamiliar with streams. I think a user that visits a compression-related documentation page may already know a thing or two about using streams to compress/decompress.

> _When would I use this API over the ones on BrotliStream?_

Precisely! This API is intended to be used when the user does not want to depend on streams, that was the main reason why this API was proposed.

@tdykstra thanks for the correction, I will try to keep the sentences short. I need to find a way to explain the API more clearly, but in short sentences.

xml/System.IO.Compression/BrotliDecoder.xml
carlossanlop:
```suggestion
        <returns>One of the enumeration values that indicates the status of the decompression operation.</returns>
```

xml/System.IO.Compression/BrotliDecoder.xml
carlossanlop:
```suggestion
         <remarks>
           <format type="text/markdown"><![CDATA[

## Remarks

The return value can be as follows:
- <xref:System.Buffers.OperationStatus.Done?displayProperty=nameWithType>: `source` was successfully and completely decompressed into `destination`.
- <xref:System.Buffers.OperationStatus.DestinationTooSmall?displayProperty=nameWithType>: There is not enough space in `destination` to decompress `source`.
- <xref:System.Buffers.OperationStatus.NeedMoreData?displayProperty=nameWithType>: The decompression action is partially done at least one more byte is required to complete the decompression task. This method should be called again with more input to decompress.
- <xref:System.Buffers.OperationStatus.InvalidData?displayProperty=nameWithType>: The data in `source` is invalid and could not be decompressed.

           ]]></format>
         </remarks>
```

xml/System.IO.Compression/BrotliDecoder.xml
carlossanlop: I'll use @tdykstra 's suggestion, it makes sense, will contain a link for the user to click it and figure out why BrotliStream is not a dependency:

```suggestion
    <summary>Provides non-allocating, performant Brotli decompression methods. The methods decompress in a single pass without using a <see cref="T:System.IO.Compression.Brotli.BrotliStream" /> instance.</summary>
```

xml/System.IO.Compression/BrotliDecoder.xml
carlossanlop: The destination needs to be larger than the source, since it's a decompression task.
Also, I don't think we should document that since it's common sense to not decompress your input in the same span that contains your input.

carlossanlop:
@tdykstra I addressed the comments. If the build has no warnings and you agree with the applied suggestions, can you please get this merged?

xml/System.IO.Compression/BrotliDecoder.xml
```diff
@@ -52,13 +52,25 @@
...
+        <returns>One of the enumeration values that indicates the status of the decompression operation.</returns>
+         <remarks>
+           <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+The return value can be as follows:
+- <xref:System.Buffers.OperationStatus.Done?displayProperty=nameWithType>: `source` was successfully and completely decompressed into `destination`.
+- <xref:System.Buffers.OperationStatus.DestinationTooSmall?displayProperty=nameWithType>: There is not enough space in `destination` to decompress `source`.
+- <xref:System.Buffers.OperationStatus.NeedMoreData?displayProperty=nameWithType>: The decompression action is partially done at least one more byte is required to complete the decompression task. This method should be called again with more input to decompress.
```
ahsonkhan:
```suggestion
- <xref:System.Buffers.OperationStatus.NeedMoreData?displayProperty=nameWithType>: The decompression action is partially done. At least one more byte is required to complete the decompression task. This method should be called again with more input to decompress.
```

PR #3397, opened 2019-10-23
L133895: Fixing CDATA formatting

v-maudel:
Description: the Markdown CDATA content is incorrectly formatted in source files, moreover it is breaking the target publishing builds since the content inside those tags are being processed as XML tag instead of Markdown content.

Also there is another file that I can't edit that needs the same fix: https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/String.xml

@mairaw these Issues are related to the previou PR I created here: https://github.com/dotnet/dotnet-api-docs/pull/3366

Regards

### DIFF
--- a/xml/System.Reflection.PortableExecutable/PEHeader.xml
+++ b/xml/System.Reflection.PortableExecutable/PEHeader.xml
@@ -492,8 +492,7 @@ Represents `IMAGE_DIRECTORY_ENTRY_EXPORT`.
       <Docs>
         <summary>Gets the alignment factor (in bytes) that is used to align the raw data of sections in the image file.</summary>
         <value>A power of 2 between 512 and 64K, inclusive. The default is 512.</value>
-        <remarks><format type="text/markdown">
-  <![CDATA[
+        <remarks><format type="text/markdown"><![CDATA[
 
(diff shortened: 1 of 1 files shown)


tdykstra:
@v-maudel Is there an issue for the file that was too big to edit in the browser?

v-maudel:
Hi @tdykstra , yes, actually the "edit" icon is not available,

tdykstra:
Sorry, I meant is there a repo issue to track the need to update it?  If not I'll create one.

v-maudel:
Oh, my apologies then :) 
No, there isn't any issue created, we only create PRs and we were encouraged to not create Issues in Repos in our common worklfow.

mairaw:
@v-maudel can you do this not using the GitHub UI? The edit button not being available for large files is a GitHub limitation.

v-maudel:
Hi @mairaw , I'm not sure about it, we only work in the Github page. Although, I know some teams, like localization one, use the Github Desktop app for check files. Maybe I can try this way but I need to do a research on that app first.

mairaw:
I'll take care of that fix for you then.

PR #3408, opened 2019-10-25
Fix Directory.EnumerateFiles documentation is misleading 

### LINKED ISSUE #3356: Directory.EnumerateFiles documentation is misleading
The `Directory.EnumerateFiles` documentation seems to indicate that the return values are Directory Names only. This is metioned and hinted multiple times. Only if you look closer at the Return value definition you actually see that  full names including paths are returned. Maybe I'm the only one that fell into this trap but I think the documentation should be improved to make it clear that full names including path are returned.

image

So my suggestion is to either change the occurrence of _directory names_  to _directory full names_ or to _ full names_ or _full names (including paths) like mentioned in the returns documentation

dsychin:
## Summary

Update Directory.EnumerateFiles to mention directory full names.

Fixes #3356

### DIFF
--- a/xml/System.IO/Directory.xml
+++ b/xml/System.IO/Directory.xml
@@ -834,7 +834,7 @@ Directory::CreateDirectory("Public\\Html");
         <AssemblyVersion>4.1.0.0</AssemblyVersion>
       </AssemblyInfo>
       <Docs>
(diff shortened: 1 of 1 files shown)


danielmarbach:
Volks you all rock! Thanks for caring even about little details

PR #3440, opened 2019-11-03
Add note on invariant culture

ChrisMaddock:
## Summary

The Invariant Culture is a special case for the `CultureInfo.Name` property in that it returns an empty string. This is noted on the [InvariantCulture](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.invariantculture?view=netframework-4.8) API page, but not on the name property page, and caught me out for a little while today.

I've added a note on this. Not sure if there's a better place to put this, it feels a little 'tagged-on' here. Would appreciate any advice. 🙂

### DIFF
--- a/xml/System.Globalization/CultureInfo.xml
+++ b/xml/System.Globalization/CultureInfo.xml
@@ -2911,7 +2911,7 @@ For a list of predefined culture names on Windows systems, see the **Language ta
       </ReturnValue>
       <Docs>
         <summary>Gets the culture name in the format *languagecode2*-*country/regioncode2*.</summary>
-        <value>The culture name in the format *languagecode2*-*country/regioncode2*. *languagecode2* is a lowercase two-letter code derived from ISO 639-1. *country/regioncode2* is derived from ISO 3166 and usually consists of two uppercase letters, or a BCP-47 language tag. The invariant culture has an empty string.</value>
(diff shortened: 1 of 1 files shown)


xml/System.Globalization/CultureInfo.xml
```diff
@@ -2911,7 +2911,7 @@ For a list of predefined culture names on Windows systems, see the **Language ta
       </ReturnValue>
       <Docs>
         <summary>Gets the culture name in the format *languagecode2*-*country/regioncode2*.</summary>
-        <value>The culture name in the format *languagecode2*-*country/regioncode2*. *languagecode2* is a lowercase two-letter code derived from ISO 639-1. *country/regioncode2* is derived from ISO 3166 and usually consists of two uppercase letters, or a BCP-47 language tag.</value>
+        <value>The culture name in the format *languagecode2*-*country/regioncode2*. *languagecode2* is a lowercase two-letter code derived from ISO 639-1. *country/regioncode2* is derived from ISO 3166 and usually consists of two uppercase letters, or a BCP-47 language tag. The invariant culture has an empty string.</value>
```
MSDN-WhiteKnight:
```suggestion
        <value>The culture name in the format *languagecode2*-*country/regioncode2*, if the current <see cref="T:System.Globalization.CultureInfo" /> is culture-dependent; or an empty string, if it's an invariant culture. *languagecode2* is a lowercase two-letter code derived from ISO 639-1. *country/regioncode2* is derived from ISO 3166 and usually consists of two uppercase letters, or a BCP-47 language tag.</value>
```
Perhaps this wording would be better?

BillWagner: This wording does flow better. Great suggestion!

ChrisMaddock:
Thanks both!

PR #3464, opened 2019-11-09
Update ValidationContext.xml to reflect a recent change in this property

HongGit:
## Summary

The doc indicated a change didn't ship with .NET 4.8.  Instead the .NET Framework October Quality Rollup contains the fix to the correct behavior.  Therefore, this doc needs to be updated accordingly.


HongGit:
@HongGit  this PR needs to be updated.

PR #3499, opened 2019-11-15
Document EIIs of integer types (IComparable/IConvertible)

carlossanlop:
Int16, Int32, Int64, UInt16, UInt32, UInt64 all implement these two methods:

- IComparable.CompareTo
- IConvertible.GetTypeCode

### DIFF
--- a/xml/System/Int16.xml
+++ b/xml/System/Int16.xml
@@ -248,8 +248,7 @@
   
  ]]></format>
         </remarks>
-        <exception cref="T:System.ArgumentException">
-          <paramref name="value" /> is not an <see cref="T:System.Int16" />.</exception>
+        <exception cref="T:System.ArgumentException"><paramref name="value" /> is not an <see cref="T:System.Int16" />.</exception>
         <altmember cref="T:System.UInt16" />
       </Docs>
     </Member>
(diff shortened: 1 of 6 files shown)


xml/System/Int16.xml
```diff
@@ -1065,10 +1064,44 @@
...
+              <term>Return Value</term>
+              <description>Description</description>
+            </listheader>
+            <item>
+              <term>Less than zero</term>
+              <description>This instance is less than <paramref name="value" />.</description>
+            </item>
+            <item>
+              <term>Zero</term>
+              <description>This instance is equal to <paramref name="value" />.</description>
```
BillWagner: On the interface, we've described this as follows:

> This current instance occurs in the same position in the sort order as `value`.

For all the integral types, those two descriptions mean the same. I'll approve this, and you can decide if you want to use the same consistent language. (This comment applies to all pages in this PR)

carlossanlop: Hey @BillWagner would you mind elaborating a bit more? I'm not sure where you got that description from, so I'm not understanding your suggestion here.

In case it helps: I got all this text from `<Member MemberName="CompareTo">` in this same file (line 181), and simply added the extra remark about the EII.

BillWagner: @carlossanlop The language I quoted above is from the [`IComparable<T>` interface page](https://docs.microsoft.com/en-us/dotnet/api/system.icomparable-1?view=netcore-3.0#remarks), in the remarks section.

For all the types in this PR, `CompareTo` returns 0 if and only if the two instances are equal. That's not true for all types.  (Imagine ordering events by date. Two events on the same date would be in the same location in the sort order, but may not be equal.)

For these types, that distinction doesn't matter. I just want to make sure we're consistent and explicit on how we make this decision.

carlossanlop: Oh I see what you mean. I think the descriptions are fine for these int types, they were adapter for their specific behavior.

PR #3505, opened 2019-11-15
Document System.IO.DirectoryInfo

carlossanlop:


carlossanlop:
> LGTM. I noticed two unescaped * characters and a double space. I fixed those with find/replace, about 16 matches each.

Thanks. I double checked the preview page and they look good:
https://review.docs.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.enumeratedirectories?view=netcore-3.0&branch=pr-en-us-3505

I wasn't aware that asterisks had to be escaped in markdown (for example, this is unescaped: *).

PR #3543, opened 2019-11-20
Clarify LazyInitialize.EnsureInitialized

### LINKED ISSUE #3541: Docs for LazyInitializer.EnsureInitialize<T>(ref T) are wrong
In the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.threading.lazyinitializer.ensureinitialized?view=netstandard-2.0#System_Threading_LazyInitializer_EnsureInitialized__1___0__) for `LazyInitializer.EnsureInitialize<T>(ref T target)`, it states,

> In the event that multiple threads access this method concurrently, multiple instances of T may be created, but only one will be stored into target. In such an occurrence, this method will not dispose of the objects that were not stored. If such objects must be disposed, it is up to the caller to determine if an object was not used and to then dispose of the object appropriately.

The code uses `Activator.CreateInstance` if not already null (guarded by `Interlocked.CompareExchanged`) but `target` is only ever returned. The result of `Activator.CreateInstance` is lost if not placed into `target`, so callers can never dispose of it if not using an overload that takes a `Func<T>` and specifically checking for whether it was initialized.

I'm happy to fix the doc comments, but wanted confirmation first that this was actually the intended behavior and that the docs are wrong and should change.

Opened in lieu of dotnet/corefx#42643. I'll are a PR. /cc @danmosemsft

heaths:
## Summary

Documentation for `LazyInitialize.EnsureInitialized<T>(ref T)` make it sound like you can and should dispose of any objects created by other threads that were not stored; however, without a `valueFactory` that's not even possible.

Fixes #3541

### DIFF
--- a/xml/System.Threading/LazyInitializer.xml
+++ b/xml/System.Threading/LazyInitializer.xml
@@ -125,7 +125,7 @@ ExpensiveData dataToUse = LazyInitializer.EnsureInitialized(ref _data, ref _data
   
  This method may be used concurrently by multiple threads to initialize `target`.  
   
- In the event that multiple threads access this method concurrently, multiple instances of `T` may be created, but only one will be stored into `target`. In such an occurrence, this method will not dispose of the objects that were not stored. If such objects must be disposed, it is up to the caller to determine if an object was not used and to then dispose of the object appropriately.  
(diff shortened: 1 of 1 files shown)


danmosemsft:
Thank you.

PR #3550, opened 2019-11-21
InitializesServerApplication required for calls to interface methods

GuacoIV:
Clarify that setting InitializesServerApplication to true is actually necessary/prerequisite to receiving calls to the interface methods


GuacoIV:
Ping :)

PR #3559, opened 2019-11-24
Improve System.GC documentation

MSDN-WhiteKnight:
## Summary
- Format identifiers as xrefs to avoid localizing them
- Remove "Notes for inheritors" block (it makes no sence for GC which is static class you can't inherit from), and move its contents to remarks
- Replace MSDN links to MS Docs

### DIFF
--- a/xml/System/GC.xml
+++ b/xml/System/GC.xml
@@ -93,7 +93,7 @@
     </remarks>
     <altmember cref="T:System.Runtime.GCSettings" />
     <related type="Article" href="~/docs/standard/garbage-collection/index.md">Garbage Collection</related>
-    <related type="Article" href="https://msdn.microsoft.com/library/67c5a20d-1be1-4ea7-8a9a-92b0b08658d2">Fundamentals of Garbage Collection</related>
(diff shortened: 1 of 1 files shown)


xml/System/GC.xml
```diff
@@ -218,7 +218,7 @@
         <exception cref="T:System.InvalidOperationException">This member is not available when concurrent garbage collection is enabled. See the [&lt;gcConcurrent&gt;](~/docs/framework/configure-apps/file-schema/runtime/gcconcurrent-element.md) runtime setting for information about how to disable concurrent garbage collection.</exception>
         <permission cref="F:System.Security.Permissions.SecurityAction.LinkDemand">for full trust for the immediate caller. This member cannot be used by partially trusted code.</permission>
```
gewarren:
```suggestion
        <permission cref="F:System.Security.Permissions.SecurityAction.LinkDemand">requires full trust for the immediate caller. This member cannot be used by partially trusted code.</permission>
```

gewarren: I'll go ahead and merge as-is.

PR #3637, opened 2019-12-06
Document System.TimeSpan

carlossanlop:
There were 3 APIs left to document in this type:
- CompareTo EII
- TryParseExact
- op_Multiply

### DIFF
--- a/xml/System/TimeSpan.xml
+++ b/xml/System/TimeSpan.xml
@@ -4877,7 +4877,7 @@ In some cases, when a time interval component in the string to be parsed contain
         <param name="format">A span containing the characters that represent a standard or custom format string that defines the acceptable format of <paramref name="input" />.</param>
         <param name="formatProvider">An object that supplies culture-specific formatting information.</param>
         <param name="result">When this method returns, contains an object that represents the time interval specified by <paramref name="input" />, or <see cref="F:System.TimeSpan.Zero" /> if the conversion failed. This parameter is passed uninitialized.</param>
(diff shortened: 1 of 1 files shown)


xml/System/TimeSpan.xml
```diff
@@ -5083,8 +5115,8 @@ In some cases, when a time interval component in the string to be parsed contain
...
         <summary>Converts the specified span representation of a time interval to its <see cref="T:System.TimeSpan" /> equivalent by using the specified format, culture-specific format information, and styles, and returns a value that indicates whether the conversion succeeded. The format of the string representation must match the specified format exactly.</summary>
         <returns><see langword="true" /> if <paramref name="input" /> was converted successfully; otherwise, <see langword="false" />.</returns>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <summary>Converts the specified span representation of a time interval to its <see cref="T:System.TimeSpan" /> equivalent by using the specified formats, culture-specific format information, and styles, and returns a value that indicates whether the conversion succeeded. The format of the string representation must match one of the specified formats exactly.</summary>
```
Thraka: I don't think we need to specify what the return is in the summary.

```suggestion
        <summary>Converts the specified span representation of a time interval to its <see cref="T:System.TimeSpan" /> equivalent by using the specified formats, culture-specific format information, and styles. The format of the string representation must match one of the specified formats exactly.</summary>
```

carlossanlop: Good point. In that case, I'll also remove that text from the other TryParseExact overloads.

PR #3659, opened 2019-12-11
Document caution regarding AdvanceTo's side effect

AArnott:
After calling `PipeReader.AdvanceTo`, the `ReadResult.Buffer` property's backing data is recycled, leading to indeterministic changes to not only the content but its `Length` property. This isn't particular predictable as a consumer and the docs didn't make it clear either.

This updates the docs to match the updated xml doc comments from https://github.com/dotnet/runtime/pull/691

### DIFF
--- a/xml/System.IO.Pipelines/PipeReader.xml
+++ b/xml/System.IO.Pipelines/PipeReader.xml
@@ -53,7 +53,7 @@
 
 ## Remarks
 
-The memory for the consumed data will be released and no longer available. The `ReadResult.Buffer` previously returned from `ReadAsync(CancellationToken)` must not be accessed after this call. The examined data communicates to the pipeline when it should signal more data is available.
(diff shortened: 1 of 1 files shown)


xml/System.IO.Pipelines/PipeReader.xml
```diff
@@ -53,7 +53,7 @@
 
 ## Remarks
 
-The memory for the consumed data will be released and no longer available. The examined data communicates to the pipeline when it should signal more data is available.
+The memory for the consumed data will be released and no longer available. The `ReadResult.Buffer` previously returned from `ReadAsync(CancellationToken)` must not be accessed after this call. The examined data communicates to the pipeline when it should signal more data is available.
```
BillWagner: These could be xref links:

```suggestion
The memory for the consumed data will be released and no longer available. The <xref:System.IO.Pipelines.ReadResult.Buffer%2A> previously returned from <xref:System.IO.Pipelines.PipeReader.ReadAsync(System.Threading.CancellationToken)> must not be accessed after this call. The examined data communicates to the pipeline when it should signal more data is available.
```

AArnott: I like that. I didn't know how to do it in markdown, but this is the same kind of thing I did in the original xml doc comments.

BillWagner:
Thanks for the update @AArnott 

I'll :shipit: now.

AArnott:
Thanks.

PR #3660, opened 2019-12-11
Add missing, and update existing, documentation on the RSA classes

bartonjs:
cc: @carlossanlop

### DIFF
--- a/xml/System.Security.Cryptography/RSACng.xml
+++ b/xml/System.Security.Cryptography/RSACng.xml
@@ -132,7 +132,7 @@ If a key is loaded via the <xref:System.Security.Cryptography.RSACng.ImportParam
 Valid key sizes range from 384 to 16,384 bits, in increments of 8. We recommend that a minimum size of 2,048 bits be used for all keys.  
   
 This constructor does not generate a new public/private keypair immediately.
-This constructor sets the <xref:System.Security.Cryptography.RSACng.KeySize> property to `keySize` and when a key is needed one is generated using the property value.
+This constructor sets the <xref:System.Security.Cryptography.AsymmetricAlgorithm.KeySize> property to `keySize` and when a key is needed one is generated using the property value.
 If a key is loaded via the <xref:System.Security.Cryptography.RSACng.ImportParameters%2A> method, or other key import method, the key size from this constructor has no meaning.
   
  ]]></format>
--- a/xml/System.Security.Cryptography/RSAOpenSsl.xml
+++ b/xml/System.Security.Cryptography/RSAOpenSsl.xml
@@ -244,7 +244,7 @@ This instance represents only a public key.
 -or-
 
 The decryption operation failed.</exception>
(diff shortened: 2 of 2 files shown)


xml/System.Security.Cryptography/RSACng.xml
```diff
@@ -861,14 +922,36 @@
...
+        <param name="hash">The hash value of the data to be signed.</param>
         <param name="destination">To be added.</param>
-        <param name="hashAlgorithm">To be added.</param>
-        <param name="padding">To be added.</param>
-        <param name="bytesWritten">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <param name="hashAlgorithm">The hash algorithm used to create the hash value of the data.</param>
+        <param name="padding">The padding.</param>
```
carlossanlop: This parameter could use a better description.

xml/System.Security.Cryptography/RSAOpenSsl.xml
```diff
@@ -448,12 +556,26 @@
...
       </Parameters>
       <Docs>
-        <param name="hash">To be added.</param>
-        <param name="hashAlgorithm">To be added.</param>
-        <param name="padding">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="hash">The hash value of the data to be signed.</param>
+        <param name="hashAlgorithm">The hash algorithm used to create the hash value of the data.</param>
+        <param name="padding">The padding.</param>
```
carlossanlop: This parameter could use a better description.

xml/System.Security.Cryptography/RSAOpenSsl.xml
```diff
@@ -481,13 +603,18 @@
...
-        <param name="hash">To be added.</param>
-        <param name="signature">To be added.</param>
-        <param name="hashAlgorithm">To be added.</param>
-        <param name="padding">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="hash">The hash value of the signed data.</param>
+        <param name="signature">The signature data to be verified.</param>
+        <param name="hashAlgorithm">The hash algorithm used to create the hash value.</param>
+        <param name="padding">The padding mode.</param>
```
carlossanlop: This parameter has one extra word, but I think it could also be described better.

carlossanlop: Same suggestion applies to the rest of the padding params.

bartonjs: The "padding" parameter descriptions are the ones that are already present on the documented overloads and variants (e.g. https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.rsa.signdata?view=netframework-4.8).

They can be long winded, such as "The padding mode to use for the {encryption|decryption|signing|verification} process.", but there's really nothing else that I can think of to say about them.  I'm open to suggestions, but absent a specific suggestion I'm going to leave them as-is.

bartonjs:
Looks like there were some bad cref/xrefs, I'll get them after lunch (assuming no one does it before then).

BillWagner:
@bartonjs This now has merge conflicts. Can you resolve those?

bartonjs:
@BillWagner Conflicts resolved, bad crefs/xrefs resolved.

PR #3666, opened 2019-12-12
process_refresh.cs has a bug, not sure how to fix

mczarnek:
This line: [!code-csharp[process_refresh#1](~/samples/snippets/csharp/VS_Snippets_CLR/process_refresh/CS/process_refresh.cs#1)]

This function
 private void myProcess_Exited(object sender, System.EventArgs e)
    {
        eventHandled = true;
        Console.WriteLine(
            $"Exit time    : {myProcess.ExitTime}\n" +
            $"Exit code    : {myProcess.ExitCode}\n" +
            $"Elapsed time : {elapsedTime}");
    }

myProcess doesn't exist within that scope. I believe missing this line of code before Console.WriteLine:
Process myProcess = (Process)sender;

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Diagnostics/Process.xml
+++ b/xml/System.Diagnostics/Process.xml
@@ -719,7 +719,8 @@ process.BeginOutputReadLine();
  [!code-cpp[process_refresh#1](~/samples/snippets/cpp/VS_Snippets_CLR/process_refresh/CPP/process_refresh.cpp#1)]
  [!code-csharp[process_refresh#1](~/samples/snippets/csharp/VS_Snippets_CLR/process_refresh/CS/process_refresh.cs#1)]
  [!code-vb[process_refresh#1](~/samples/snippets/visualbasic/VS_Snippets_CLR/process_refresh/VB/process_refresh.vb#1)]  
-  
+^Not sure how to edit code? I'll add description of change in the "Propose file change" textbox^   
+
  ]]></format>
         </remarks>
         <exception cref="T:System.InvalidOperationException">The process has already exited.  


MSDN-WhiteKnight:
You can suggest edits to samples by submitting PR into [samples](https://github.com/dotnet/samples) repository. I don't see the mentioned lines in process_refresh.cs file, though. What is the actual page on Microsoft Docs site where you see this code?

mczarnek:
Right here, in the first code block under **Examples**
https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.exited?view=netframework-4.8

C# code

MSDN-WhiteKnight:
This sample is actually in the file [processexitedevent.cs](https://github.com/dotnet/samples/blob/master/snippets/csharp/VS_Snippets_CLR_System/system.Diagnostics.Process.EnableExited/CS/processexitedevent.cs) (https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System.Diagnostics/Process.xml#L1175). Anyway, i don't think it has the mentioned issue: there's `Process myProcess` instance field in the PrintProcessClass, which is accessible to the myProcess_Exited event handler. No need to read the sender artument. 

However, the sample is still incorrect, due to how it performs thread syncronization. It uses while loop to check when background thread sets the boolean field, but that field is not even volatile. The proper way to implement this would be to wait on the proper syncronisation object, like ManualResetEvent.

BillWagner:
closing in favor of the fix in dotnet/samples#1900

PR #3670, opened 2019-12-12
Document System.Configuration.Internal.DelegatingConfigHost

carlossanlop:
@JeremyKuhne wrote these APIs. They seem to be used by ASP.NET, and even though they are public, they are meant for internal consumption. They still need to be documented because they are exposed.

PR that introduced them: https://github.com/dotnet/corefx/pull/14293

Source: https://source.dot.net/#System.Configuration.ConfigurationManager/System/Configuration/Internal/DelegatingConfigHost.cs

Current area owner: @maryamariyan

### DIFF
--- a/xml/System.Configuration.Internal/DelegatingConfigHost.xml
+++ b/xml/System.Configuration.Internal/DelegatingConfigHost.xml
@@ -833,7 +833,7 @@
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Determines if the application configuration file comes from a non-local URI (that is, <c>http://<c>) or is a local file.</summary>
+        <summary>Determines if the application configuration file comes from a non-local URI (that is, <c>http://</c>) or is a local file.</summary>
         <value>
           <see langword="true" /> if the delegated host's sections belong to <c>System.Net</c> and the configuration file for the application is downloaded via HTTP using <see cref="T:System.Net.WebClient" />; otherwise, <see langword="false" />.</value>
          <remarks>To be added.</remarks>


xml/System.Configuration.Internal/DelegatingConfigHost.xml
```diff
@@ -192,9 +192,9 @@
       </Parameters>
       <Docs>
         <param name="encryptedXml">An encrypted section of a configuration file.</param>
-        <param name="protectionProvider">A <see cref="T:System.Configuration.ProtectedConfigurationProvider" /> object.</param>
-        <param name="protectedConfigSection">A <see cref="T:System.Configuration.ProtectedConfigurationSection" /> object.</param>
-        <param name="protectedSection">To be added.</param>
+        <param name="protectionProvider">An object containing the providers that encrypt and decrypt protected configuration data.</param>
+        <param name="protectedConfigSection">An object that provides programatic access to the <c>configProtectedData</c> configuration section.</param>
```
carlossanlop: This parameter had a ref vs src mismatch at some point, causing this xml to be generated with both parameter names.

Same problem applies to the next two diffs in this file.

xml/System.Configuration.Internal/DelegatingConfigHost.xml
```diff
@@ -829,8 +831,9 @@
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Determines if the delegated host's sections belong to <c>System.Net</c> and the configuration file for the application is downloaded via HTTP using <see cref="T:System.Net.WebClient" />.</summary>
```
carlossanlop: I am not really sure about this property summary. It was not documented in triple slash, the PR did not describe it and there's no API proposal explaining it. But I found a comment where this property was being used, describing its purpose:
https://github.com/dotnet/runtime/blob/4f9ae42d861fcb4be2fcd5d3d55d5f227d30e723/src/libraries/System.Configuration.ConfigurationManager/src/System/Configuration/ClientConfigurationSystem.cs#L75

JeremyKuhne: It is meant to tell you whether or not the application config file is coming from a non-local URI (i.e. `http://`) or is a local file. That is all I would say here.

xml/System.Configuration.Internal/DelegatingConfigHost.xml
JeremyKuhne: So I presume this will be fixed?

carlossanlop: I don't think these get fixed because I it seems we published them in a particular .NET version, so we still need to keep that old parameter name there for when the user selects that particular .NET version from the dropdown in MS Docs.

xml/System.Configuration.Internal/DelegatingConfigHost.xml
carlossanlop: Okay thanks. I'll use your suggested text:
```suggestion
        <summary>Determines if the application configuration file comes from a non-local URI (that is, <c>http://<c>) or is a local file.</summary>
```

BillWagner:
@carlossanlop @maryamariyan I gave this a quick read this morning, and I have one initial question: Should we add remarks, or a summary that these APIs are intended for internal use? We've done that with other APIs where that is the case.

maryamariyan:
> @carlossanlop @maryamariyan I gave this a quick read this morning, and I have one initial question: Should we add remarks, or a summary that these APIs are intended for internal use? We've done that with other APIs where that is the case.

Thanks @BillWagner. I found similar examples in the repo such as the following. Perhaps we can also add the `forInternalUseOnly` xml element as shown here.
https://github.com/dotnet/dotnet-api-docs/blob/f37171b1ee7912fad473703ce48a91f96df382b8/xml/System.Runtime.CompilerServices/TaskAwaiter%601.xml#L54-L59

carlossanlop:
I am not sure if the `<forInternalUseOnly />` should be manually added. @mairaw can you confirm that we can do that?

@BillWagner @maryamariyan I will add the remark. Thanks for noticing.

carlossanlop:
Update: Apparently we can add them manually. I found the PR that introduced them, they were not autogenerated:

xml/System.Configuration.Internal/DelegatingConfigHost.xml
```diff
@@ -1510,7 +1513,7 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Invokes the delegated host's method that refreshes configuration paths.</summary>
         <remarks>To be added.</remarks>
```
carlossanlop:
```suggestion
        <remarks>To be added.</remarks>
        <forInternalUseOnly />
```

xml/System.Configuration.Internal/DelegatingConfigHost.xml
```diff
@@ -829,8 +831,9 @@
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Determines if the application configuration file comes from a non-local URI (that is, <c>http://<c>) or is a local file.</summary>
+        <value>
+          <see langword="true" /> if the delegated host's sections belong to <c>System.Net</c> and the configuration file for the application is downloaded via HTTP using <see cref="T:System.Net.WebClient" />; otherwise, <see langword="false" />.</value>
         <remarks>To be added.</remarks>
```
carlossanlop:
```suggestion
         <remarks>To be added.</remarks>
         <forInternalUseOnly />
```

xml/System.Configuration.Internal/DelegatingConfigHost.xml
```diff
@@ -599,8 +600,9 @@
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Determines if the delegated host has a roaming configuration.</summary>
+        <value>
+          <see langword="true" /> if the current instance has a roaming configuration; otherwise, <see langword="false" />.</value>
         <remarks>To be added.</remarks>
```
carlossanlop:
```suggestion
         <remarks>To be added.</remarks>
         <forInternalUseOnly />
```

xml/System.Configuration.Internal/DelegatingConfigHost.xml
```diff
@@ -574,8 +574,9 @@
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Determines if the delegated host has a local configuration.</summary>
+        <value>
+          <see langword="true" /> if the current instance has a local configuration; otherwise, <see langword="false" />.</value>
         <remarks>To be added.</remarks>
```
carlossanlop:
```suggestion
         <remarks>To be added.</remarks>
         <forInternalUseOnly />
```

xml/System.Configuration.Internal/DelegatingConfigHost.xml
```diff
@@ -192,9 +192,9 @@
...
         <param name="encryptedXml">An encrypted section of a configuration file.</param>
-        <param name="protectionProvider">A <see cref="T:System.Configuration.ProtectedConfigurationProvider" /> object.</param>
-        <param name="protectedConfigSection">A <see cref="T:System.Configuration.ProtectedConfigurationSection" /> object.</param>
-        <param name="protectedSection">To be added.</param>
+        <param name="protectionProvider">An object containing the providers that encrypt and decrypt protected configuration data.</param>
+        <param name="protectedConfigSection">An object that provides programatic access to the <c>configProtectedData</c> configuration section.</param>
+        <param name="protectedSection">An object that provides programatic access to the <c>configProtectedData</c> configuration section.</param>
         <summary>Decrypts an encrypted configuration section.</summary>
         <returns>A string representing a decrypted configuration section.</returns>
         <remarks>To be added.</remarks>
```
carlossanlop:
```suggestion
         <remarks>
           <format type="text/markdown"><![CDATA[

## Remarks

This type and its members are intended for use by the compiler. 

           ]]></format>
         </remarks>
```

xml/System.Configuration.Internal/DelegatingConfigHost.xml
```diff
@@ -829,9 +833,11 @@
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <summary>Determines if the application configuration file comes from a non-local URI (that is, <c>http://<c>) or is a local file.</summary>
```
carlossanlop:
```suggestion
        <summary>Determines if the application configuration file comes from a non-local URI (that is, <c>http://</c>) or is a local file.</summary>
```

carlossanlop:
@BillWagner is it ready to 🚢 ?

PR #3673, opened 2019-12-13
Document System.Buffers.Text namespace

carlossanlop:
Thank you @layomia for providing the description.

### DIFF
--- a/xml/ns-System.Buffers.Text.xml
+++ b/xml/ns-System.Buffers.Text.xml
@@ -1,6 +1,6 @@
 <Namespace Name="System.Buffers.Text">
   <Docs>
-    <summary>The namespace contains types that can be used to parse and format common data types from and to UTF-8 text representations.</summary>
+    <summary>This namespace contains types that can be used to parse and format common data types to and from UTF-8 text representations.</summary>
     <remarks>To be added.</remarks>
   </Docs>
 </Namespace>


xml/ns-System.Buffers.Text.xml
```diff
@@ -1,6 +1,6 @@
 <Namespace Name="System.Buffers.Text">
   <Docs>
-    <summary>To be added.</summary>
+    <summary>The namespace contains types that can be used to parse and format common data types to and from UTF-8 representations.</summary>
```
layomia:
```suggestion
    <summary>The namespace contains types that can be used to parse and format common data types from and to UTF-8 representations.</summary>
```

xml/ns-System.Buffers.Text.xml
```diff
@@ -1,6 +1,6 @@
 <Namespace Name="System.Buffers.Text">
   <Docs>
-    <summary>To be added.</summary>
+    <summary>The namespace contains types that can be used to parse and format common data types from and to UTF-8 representations.</summary>
```
ahsonkhan: nit:
```suggestion
    <summary>The namespace contains types that can be used to parse and format common data types from and to UTF-8 text representations.</summary>
```

xml/ns-System.Buffers.Text.xml
```diff
@@ -1,6 +1,6 @@
 <Namespace Name="System.Buffers.Text">
   <Docs>
-    <summary>To be added.</summary>
+    <summary>The namespace contains types that can be used to parse and format common data types from and to UTF-8 text representations.</summary>
```
gewarren:
```suggestion
    <summary>This namespace contains types that can be used to parse and format common data types to and from UTF-8 text representations.</summary>
```

ahsonkhan: Out of curiosity, what's the reason for using `to and from` rather than `from and to` here? Convention? Preference?

gewarren: "To and from" is the way those words are usually ordered in English, so "from and to" sounded weird to me. Maybe it's just me though.

tdykstra: It's not just you -- "from and to" sounds weird to me too.

ahsonkhan: > sounded weird to me. Maybe it's just me though.

Agreed. Same here.

The rationale @layomia had when suggesting `from and to` was related to the `parse and format` snippet. You `parse from` UTF-8 text and you `format to` UTF-8 text. But then again, changing the ordering to `format and parse` also seems a bit odd to me (compared to `parse and format`).

The way I read the sentence (the way it is written now/merged), the "to and from" refers to going from common data types **to** UTF-8 text and **from** UTF-8 text to common data types (and no the parse/format verbs). That makes sense to me :)

PR #3682, opened 2019-12-17
Replace `+` with `.`

Youssef1313:


MSDN-WhiteKnight:
Are you sure this change is correct? I think this mdoc XML metadata uses the same syntax as reflection, i.e. methods like `Type.GetType()`, which expect nested types to be separated by `+`. For example, `Type.GetType("System.Environment.SpecialFolder");` returns null, but `Type.GetType("System.Environment+SpecialFolder");` returns the correct value. It matches how the type itself is declared: https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Environment+SpecialFolder.xml#L1 The fact that this plus sign shows up on the docs site in overloads list does not seem right, though.

Youssef1313:
@MSDN-WhiteKnight, thanks for your clarification. I didn't know that information. My change was in attempt to solve the `+` sign appearing in docs site which is weird. Do you know what is the correct change to solve it ?

MSDN-WhiteKnight:
This is probably issue with Docs site engine rather then with XML metadata. The nested types are displayed correctly in xrefs, type headings etc., and only display with plus sign in [overloads list](https://docs.microsoft.com/en-us/dotnet/api/system.environment.getfolderpath?view=netcore-3.0#overloads) and corresponding overload headings. They should display dot in all cases, or, ideally,  honor current user language setting (i.e. display `::` for C++/CLI). You can report this as bug in <https://github.com/MicrosoftDocs/feedback>.

Youssef1313:
@MSDN-WhiteKnight Opened https://github.com/MicrosoftDocs/feedback/issues/2286.
Closing.

PR #3722, opened 2019-12-30
Update JavaScriptEncoder.xml

joshlangs:
The docs had a missing backslash.  

I have no idea if this will fix the issue.  It may need \\\".  But there's no way for me to tell.  Sorry lol.


joshlang:
Yep, that was what was intended!  Looks good to me.

PR #3758, opened 2020-01-10
Change Json.NET to System.Text.Json

tdykstra:
[Internal review URL](https://review.docs.microsoft.com/en-us/dotnet/api/system.web.script.serialization?branch=pr-en-us-3758&view=netframework-4.8)

### DIFF
--- a/xml/System.Web.Script.Serialization/JavaScriptSerializer.xml
+++ b/xml/System.Web.Script.Serialization/JavaScriptSerializer.xml
@@ -16,7 +16,7 @@
   <Interfaces />
   <Docs>
     <summary>
-      For .NET Framework 4.7.2 and later versions, use the APIs in the <see cref="N:System.Text.Json" /> namespace for serialization and deserialization. For earlier versions of .NET Framework, use <see href="https://www.newtonsoft.com/json">Newtonsoft.Json</see>.
+      For .NET Framework 4.7.2 and later versions, use the APIs in the <see cref="N:System.Text.Json" /> namespace for serialization and deserialization. For earlier versions of .NET Framework, use <see href="https://www.newtonsoft.com/json">Newtonsoft.Json</see>. This type was intended to provide serialization and deserialization functionality for AJAX-enabled applications.
     </summary>
     <remarks>
       <format type="text/markdown"><![CDATA[  


xml/System.Web.Script.Serialization/JavaScriptSerializer.xml
```diff
@@ -16,7 +16,7 @@
   <Interfaces />
   <Docs>
     <summary>
-      <see href="https://www.newtonsoft.com/json">Json.NET</see> should be used for serialization and deserialization. Provides serialization and deserialization functionality for AJAX-enabled applications.</summary>
+      The APIs in the <see cref="N:System.Text.Json" /> namespace should be used for serialization and deserialization. Provides serialization and deserialization functionality for AJAX-enabled applications.</summary>
```
scottaddie: Isn't this statement true ONLY for .NET Framework 4.6.1 or later?

tdykstra: Wherever we were recommending Json.NET we should  be recommending STJ, so I only changed that. That said, the structure of this section doesn't really make sense. If I'm reading it correctly, the "Provides..." sentence applies to this class, and so the first sentence was probably added later to say "Use Json.NET instead of this class" but the way it was added doesn't make that intention clear.

scottaddie: > Use Json.NET instead of this class

Yes, I recall switching from `JavaScriptSerializer` to JSON.NET years ago for performance reasons. At one point, there was messaging from Microsoft that completely discouraged the use of `JavaScriptSerializer`.

tdykstra: @ahsonkhan In this doc should we say use Newtonsoft.json for earlier than 4.6.1 and System.Text.Json for 4.6.1 and later?

carlossanlop: @layomia @Jozkee @steveharter can you please help answer this question?

carlossanlop: @steveharter @layomia @Jozkee ping. What is the correct sentence to use here?

Jozkee: Yes, `System.Text.Json` targets `netstandard2.0` which makes it compatible only with .NET Framework 4.6.1 or later. 

> should we say use Newtonsoft.json for earlier than 4.6.1 and System.Text.Json for 4.6.1 and later

That seems to be the most accurate. Why is that advice on the summary and not as a warning in the `JavaScriptSerializer` page?

tdykstra: > Why is that advice on the summary and not as a warning in the JavaScriptSerializer page?

No way to know why it's there or who put it there -- it's already there in the earliest commit in the repo.

I agree it's out of place where it is, I'll move it to Remarks.

xml/System.Web.Script.Serialization/JavaScriptSerializer.xml
```diff
@@ -16,14 +16,19 @@
...
+      Provides serialization and deserialization functionality for AJAX-enabled applications.
+    </summary>
     <remarks>
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Web.Script.Serialization.JavaScriptSerializer> class is used internally by the asynchronous communication layer to serialize and deserialize the data that is passed between the browser and the Web server. You cannot access that instance of the serializer. However, this class exposes a public API. Therefore, you can use the class when you want to work with JavaScript Object Notation (JSON) in managed code.  
+
+> [!IMPORTANT]
+> For .NET Framework 4.6.1 and later versions, the APIs in the <xref:System.Text.Json> namespace should be used for serialization and deserialization. For earlier versions of .NET Framework, use [Newtonsoft.Json](https://www.newtonsoft.com/json).
```
ahsonkhan: This should be .NET Framework 4.7.2 or later. Even though S.T.Json is ns2.0 compatible, 4.6.1 support has an asterisk/footnote, so I think if we are going to recommend something, let's be a bit conservative and suggest 4.7.2 instead.

https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support
> 2 The versions listed here represent the rules that NuGet uses to determine whether a given .NET Standard library is applicable. While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.

xml/System.Web.Script.Serialization/JavaScriptSerializer.xml
ahsonkhan: > I'll move it to Remarks.

If we only have it in remarks, it won't show up in IntelliSense for folks using IDEs like VS. We may want to keep the note in the summary as well as the remarks.

xml/System.Web.Script.Serialization/JavaScriptSerializer.xml
```diff
@@ -16,14 +16,19 @@
...
-      <see href="https://www.newtonsoft.com/json">Json.NET</see> should be used for serialization and deserialization. Provides serialization and deserialization functionality for AJAX-enabled applications.</summary>
+      Provides serialization and deserialization functionality for AJAX-enabled applications.
+    </summary>
     <remarks>
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Web.Script.Serialization.JavaScriptSerializer> class is used internally by the asynchronous communication layer to serialize and deserialize the data that is passed between the browser and the Web server. You cannot access that instance of the serializer. However, this class exposes a public API. Therefore, you can use the class when you want to work with JavaScript Object Notation (JSON) in managed code.  
+
+> [!IMPORTANT]
```
ahsonkhan: Can you please share the internal docs link so I can take a quick glance at how this will end up looking like on the docs page.

xml/System.Web.Script.Serialization/JavaScriptSerializer.xml
```diff
@@ -16,14 +16,19 @@
   <Interfaces />
   <Docs>
     <summary>
-      <see href="https://www.newtonsoft.com/json">Json.NET</see> should be used for serialization and deserialization. Provides serialization and deserialization functionality for AJAX-enabled applications.</summary>
+      Provides serialization and deserialization functionality for AJAX-enabled applications.
```
ahsonkhan: If we are truly recommending against this in favor of  `Newontosft.Json` for < netfx 4.7.2, should we consider obsoleting the API in .NET Core going forward or is there no need for that/the usage too high? Of course for net4.7.2+ and .net core `S.T.Json` is the way to go :)

cc @terrajobst, @GrabYourPitchforks

xml/System.Web.Script.Serialization/JavaScriptSerializer.xml
carlossanlop: @ahsonkhan When the CI finishes, click on "Details", then scroll down to the preview link for this API.

ahsonkhan: I don't see it:
https://opbuildstorageprod.blob.core.windows.net/report/2020%5C3%5C25%5C500c749d-de78-3d4a-00e4-269d599bf9ec%5CPullRequest%5C202003250031355427-3758%5Cworkflow_report.html?sv=2016-05-31&sr=b&sig=NFFwGrzp7innzIOjl3ukNRilz7HD1L0hOknTawYzb98%3D&st=2020-03-25T01%3A43%3A15Z&se=2020-04-25T01%3A48%3A15Z&sp=r

image

ahsonkhan: OK, I got it. I took one of the other links and just navigated to the `JavaScriptSerializer` type:
https://review.docs.microsoft.com/en-us/dotnet/api/system.web.script.serialization.javascriptserializer?view=netframework-4.8&branch=pr-en-us-3758

ahsonkhan: Why does it say that this page isn't available for netfx 4.8?
https://review.docs.microsoft.com/en-us/dotnet/api/system.text.json?view=netcore-3.1&branch=pr-en-us-3758&viewFallbackFrom=netframework-4.8
image

That could confuse users when we just said that for netfx 4.7.2+ use `S.T.J`. Is the platform support matrix in the docs incomplete? I would have expected anything that is netstanard2.0 compatible to have an entry in the appropriate netfx versions of the page (rather than redirecting to the .NET core version).

carlossanlop: > I don't see it
That's odd. But I'm glad you found it by navigating into another API.
> Why does it say that this page isn't available for netfx 4.8?
It's also not showing up [in NetStandard2.0](https://review.docs.microsoft.com/en-us/dotnet/api/system.text.json?view=netcore-3.1&branch=pr-en-us-3758&viewFallbackFrom=netstandard-2.0). Do you see the same problem in the production MS Docs site?

ahsonkhan: > Do you see the same problem in the production MS Docs site?

Yes.

carlossanlop: @joelmartinez can you help us figure out why System.Text.Json is not showing up in the expected monikers?

ahsonkhan: I think @mimisasouvanh might be better point of contact here.

xml/System.Web.Script.Serialization/JavaScriptSerializer.xml
tdykstra:
```suggestion
> For .NET Framework 4.7.2 and later versions, the APIs in the <xref:System.Text.Json> namespace should be used for serialization and deserialization. For earlier versions of .NET Framework, use [Newtonsoft.Json](https://www.newtonsoft.com/json).
```

xml/System.Web.Script.Serialization/JavaScriptSerializer.xml
```diff
@@ -16,14 +16,19 @@
   <Interfaces />
   <Docs>
     <summary>
-      <see href="https://www.newtonsoft.com/json">Json.NET</see> should be used for serialization and deserialization. Provides serialization and deserialization functionality for AJAX-enabled applications.</summary>
+      For .NET Framework 4.7.2 and later versions, use the APIs in the <see cref="N:System.Text.Json" /> namespace for serialization and deserialization. For earlier veresions of .NET Framework, use <see href="https://www.newtonsoft.com/json">Newtonsoft.Json</see>.
```
tdykstra:
```suggestion
      For .NET Framework 4.7.2 and later versions, use the APIs in the <see cref="N:System.Text.Json" /> namespace for serialization and deserialization. For earlier versions of .NET Framework, use <see href="https://www.newtonsoft.com/json">Newtonsoft.Json</see>.
```

xml/System.Web.Script.Serialization/JavaScriptSerializer.xml
```diff
@@ -16,14 +16,19 @@
   <Interfaces />
   <Docs>
     <summary>
-      <see href="https://www.newtonsoft.com/json">Json.NET</see> should be used for serialization and deserialization. Provides serialization and deserialization functionality for AJAX-enabled applications.</summary>
+      For .NET Framework 4.7.2 and later versions, use the APIs in the <see cref="N:System.Text.Json" /> namespace for serialization and deserialization. For earlier versions of .NET Framework, use <see href="https://www.newtonsoft.com/json">Newtonsoft.Json</see>.
```
ahsonkhan: Should we still keep the one-liner description of the type at the end? Up to you.

```suggestion
      For .NET Framework 4.7.2 and later versions, use the APIs in the <see cref="N:System.Text.Json" /> namespace for serialization and deserialization. For earlier versions of .NET Framework, use <see href="https://www.newtonsoft.com/json">Newtonsoft.Json</see>. This type was intended to provide serialization and deserialization functionality for AJAX-enabled applications.
```

PR #3767, opened 2020-01-13
Clarify WaitHandle.WaitAny/WaitAll behavior on STA threads

jkotas:
- Update the documentation to capture maximum handle limits for STA threads (it is 63 vs. 64 for non-STA threads)
- Change references STAThreadAttribute to a more general STA state. STAThreadAttribute is just one of the several ways to create thread in STA state.
- Delete references to Windows 95 and Windows Millennium

Related to https://github.com/dotnet/runtime/issues/1646


jkotas:
cc @stephentoub

PR #3770, opened 2020-01-14
Add remark on result range to `Thread.GetCurrentProcessorId()`

CoffeeFlux:
This qualifier should be noted somewhere other than the source code.

See: https://github.com/dotnet/runtime/pull/467/files/1618787fa4471c84bbddc86980ba3efdcae9f0d4#r366041987

cc: @jkotas @VSadov

### DIFF
--- a/xml/System.Threading/Thread.xml
+++ b/xml/System.Threading/Thread.xml
@@ -1858,7 +1858,7 @@ Main thread: ThreadProc.Join has returned.  Press Enter to end program.
 
 This value is cached and periodically refreshed. It may be stale by the time it's used: consuming code must not rely on its accuracy for correctness.
 
-There is also no guarantee the result will be 0-based, and consuming code should not rely on that behavior.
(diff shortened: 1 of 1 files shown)


xml/System.Threading/Thread.xml
```diff
@@ -1858,6 +1858,8 @@ Main thread: ThreadProc.Join has returned.  Press Enter to end program.
 
 This value is cached and periodically refreshed. It may be stale by the time it's used: consuming code must not rely on its accuracy for correctness.
 
+There is also no guarantee the result will be 0-based, and consuming code should not rely on that behavior.
```
jkotas:
```suggestion
The value is not guaranteed to be zero-based processor number.
```

jkotas: I do not think we need to say that the code should not rely on something that is not guaranteed.

CoffeeFlux: Works for me.

PR #3855, opened 2020-02-07
Update Random.xml

### LINKED ISSUE #3764: Document changes in System.Random in .NET Core
**Issue description**

There is a substantial change in the behaviour of `System.Random` in .NET Core. This change is however not reflected in the documentation, so the claims provided by the documentation are incorrect.

**Target framework**

- [x] .NET Core 2.x, 3.x, perhaps 1.x too
- [ ] .NET Framework
- [ ] .NET Standard

**Details**

The behaviour of the `System.Random` class in .NET Framework is that the parameterless constuctor  takes the seed value [from the current system time](https://referencesource.microsoft.com/#mscorlib/system/random.cs,53) (`Environment.TickCount`). This has been documented here: https://docs.microsoft.com/en-us/dotnet/api/system.random?view=netframework-4.8#instantiating-the-random-number-generator. This behaviour leads to the well-known issue: several random generators created in quick succession produce the same value sequences.

The behaviour has changed in .NET Core: now the initial seed value [is randomized as well](https://source.dot.net/#System.Private.CoreLib/Random.cs,124), so several random generator's instances produce different sequences even if created in quick succession. However the documentation doesn't reflect these changes: https://docs.microsoft.com/en-us/dotnet/api/system.random?view=netcore-3.0#instantiating-the-random-number-generator.

I think, `System.Random` is a very often-used tool, so refuting the traditional wisdom about its behaviour is important.

MSDN-WhiteKnight:
Add information that the default seed value in .NET Core is randomized

Fixes #3764

### DIFF
--- a/xml/System/Random.xml
+++ b/xml/System/Random.xml
@@ -75,7 +75,7 @@
   
 -   The <xref:System.Random.%23ctor> constructor uses the default seed value. This is the most common way of instantiating the random number generator.  
   
- In .NET Framework, the default seed value is time-dependent. In .NET Core, the default seed value is produced by the thread-static pseudo-random number generator.
+ In .NET Framework, the default seed value is time-dependent. In .NET Core, the default seed value is produced by the thread-static, pseudo-random number generator.
   
  If the same seed is used for separate <xref:System.Random> objects, they will generate the same series of random numbers. This can be useful for creating a test suite that processes random values, or for replaying games that derive their data from random numbers. However, note that <xref:System.Random> objects in processes running under different versions of the .NET Framework may return different series of random numbers even if they're instantiated with identical seed values.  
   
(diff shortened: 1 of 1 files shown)


xml/System/Random.xml
```diff
@@ -73,7 +73,9 @@
   
 -   The <xref:System.Random.%23ctor%28System.Int32%29> constructor uses an explicit seed value that you supply.  
   
--   The <xref:System.Random.%23ctor> constructor uses the system clock to provide a seed value. This is the most common way of instantiating the random number generator.  
+-   The <xref:System.Random.%23ctor> constructor uses the default seed value. This is the most common way of instantiating the random number generator.  
+  
+ In .NET Framework, the default seed value is time-dependent. In .NET Core, the default seed value is produced by the thread-static pseudo-random number generator.
```
gewarren:
```suggestion
 In .NET Framework, the default seed value is time-dependent. In .NET Core, the default seed value is produced by the thread-static, pseudo-random number generator.
```

xml/System/Random.xml
```diff
@@ -369,13 +371,14 @@ Random.NextDouble() * (maxValue - minValue) + minValue
       </AssemblyInfo>
       <Parameters />
       <Docs>
-        <summary>Initializes a new instance of the <see cref="T:System.Random" /> class, using a time-dependent default seed value.</summary>
+        <summary>Initializes a new instance of the <see cref="T:System.Random" /> class, using a default seed value.</summary>
```
gewarren:
```suggestion
        <summary>Initializes a new instance of the <see cref="T:System.Random" /> class using a default seed value.</summary>
```

xml/System/Random.xml
```diff
@@ -369,13 +371,14 @@ Random.NextDouble() * (maxValue - minValue) + minValue
...
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks
+In .NET Framework, the default seed value is derived from the system clock, which has finite resolution. As a result, different <xref:System.Random> objects that are created in close succession by a call to the parameterless constructor will have identical default seed values and, therefore, will produce identical sets of random numbers. This problem can be avoided by using a single <xref:System.Random> object to generate all random numbers. You can also work around it by generating your own random seed value and passing it to the <xref:System.Random.%23ctor%28System.Int32%29> constructor. For more information, see the <xref:System.Random.%23ctor%28System.Int32%29> constructor.
```
gewarren:
```suggestion
In .NET Framework, the default seed value is derived from the system clock, which has finite resolution. As a result, different <xref:System.Random> objects that are created in close succession by a call to the parameterless constructor have identical default seed values and, therefore, produce identical sets of random numbers. You can avoid this problem by using a single <xref:System.Random> object to generate all random numbers. You can also work around it by generating your own random seed value and passing it to the <xref:System.Random.%23ctor%28System.Int32%29> constructor. For more information, see the <xref:System.Random.%23ctor%28System.Int32%29> constructor. 
```

xml/System/Random.xml
```diff
@@ -369,13 +371,14 @@ Random.NextDouble() * (maxValue - minValue) + minValue
...
 
-The default seed value is derived from the system clock, which has finite resolution. As a result, on the .NET Framework only, different <xref:System.Random> objects that are created in close succession by a call to the parameterless constructor will have identical default seed values and, therefore, will produce identical sets of random numbers. This problem can be avoided by using a single <xref:System.Random> object to generate all random numbers. You can also work around it by generating your own random seed value and passing it to the <xref:System.Random.%23ctor%28System.Int32%29> constructor. For more information, see the <xref:System.Random.%23ctor%28System.Int32%29> constructor. Note that this limitation does not apply to .NET Core.  
+In .NET Core, the default seed value is produced by the thread-static pseudo-random number generator, so the above limitation does not apply. Different <xref:System.Random> objects created in close succession produce different sets of random numbers in .NET Core.
```
gewarren:
```suggestion
In .NET Core, the default seed value is produced by the thread-static, pseudo-random number generator, so the previously described limitation does not apply. Different <xref:System.Random> objects created in close succession produce different sets of random numbers in .NET Core.
```

carlossanlop:
> _@MSDN-WhiteKnight Thanks for adding this. @carlossanlop Can someone just verify the info about how .NET Core seeds the number generator?_

The Core version:
[dotnet/runtime/src/libraries/System.Private.CoreLib/src/System/Random.cs](https://github.com/dotnet/runtime/blob/4f9ae42d861fcb4be2fcd5d3d55d5f227d30e723/src/libraries/System.Private.CoreLib/src/System/Random.cs)

The Framework version:
[referencesource.microsoft.com/#mscorlib/system/random.cs,4c2fc83207c654d1](https://referencesource.microsoft.com/#mscorlib/system/random.cs,4c2fc83207c654d1)

The changes provided in this PR seem to be describing the code correctly, although I would like a second opinion in case the descriptions can be improved.

@stephentoub @tannergooding would you mind taking a quick look?

tannergooding:
The description is correct.

PR #3890, opened 2020-02-19
Rename RenderTable to RenderOuterTable

csharpfritz:
The property name available is RenderOuterTable

## Summary

Changed the property name from RenderTable to RenderOuterTable


scottaddie:
Thank you for fixing this, @csharpfritz!

PR #3891, opened 2020-02-19
Update DeflateStream.xml

AroglDarthu:
The current implementation of `Flush` **does** flush the internal buffer.


BillWagner:
ping @ViktorHofer, @ianhays for review

PR #3893, opened 2020-02-20
Added note for nuget packages in other versions

AdmiralSnyder:
also: when the docs page's version is changed to e.g. 4.7.1 the hint for the nuget package should be visible instead of forwarding to core 3.1 and saying it's not available...

### DIFF
--- a/xml/ns-System.Text.Json.xml
+++ b/xml/ns-System.Text.Json.xml
@@ -1,6 +1,6 @@
 <Namespace Name="System.Text.Json">
   <Docs>
-    <summary>The System.Text.Json namespace provides high-performance, low-allocating, and standards-compliant capabilities to process JavaScript Object Notation (JSON), which includes serializing objects to JSON text and deserializing JSON text to objects, with UTF-8 support built-in. It also provides types to read and write JSON text encoded as UTF-8, and to create an in-memory document object model (DOM) for random access of the JSON elements within a structured view of the data.</summary>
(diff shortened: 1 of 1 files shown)


Thraka:
@docs-product thoughts?

Hi @AdmiralSnyder. Thank you for opening this PR. I know you're trying to solve a problem with how the site displays this information, IE you can't figure out the fact that the NuGet package contains this class and it's not just available on the system by default. We have a much larger problem with displaying this information. This is something we would rather get fixed by the site (such as automatically listing the nuget packages per class) rather than note in text that we would eventually have to remove.

BillWagner:
I'm closing this.

@AdmiralSnyder We do have a plan to display the NuGet package that contains a type throughout the API site. That will fix this issue, and fix it across all types in .NET.

AdmiralSnyder:
thanks :-) . knowing there's gonna be a solution for this general issue is helpful. is there something i can bookmark to get info for when it's in place?

also: should i better have created an issue instead?

Thraka:
@AdmiralSnyder I don't believe there is anything public tracking that work. Just be assured we're well aware and have raised this to the docs team so it's on the plan and in flight. No ETA though.

Issues are really useful when you have a lot of work that you want to discuss and get approved, or just get feedback on before starting the work. Doing a lot of work and then getting it denied sucks 😄 but something like this is minor and doesn't take you a ton of time so it's not really terrible if it gets denied, you're not out a lot of time and effort. We really want to protect contributors time, we really appreciate all the work you do to helping make docs better. So thanks!!

PR #3908, opened 2020-02-23
Fix cache policy typo

georgevanburgh:
# Summary

As with the other constructors, cache level is initialized to `Default`. Pretty sure this one just got missed


BillWagner:
@FireEater64 Did you make this change in the wrong overload? This change is on the overload that takes a `level` argument. So, I think the existing text is correct. I plan on closing this, but I want to make sure I'm not missing something first.

FireEater64:
@BillWagner - you're completely correct - I misread this horribly. Apologies for the confusion, and thanks for checking!

BillWagner:
@FireEater64 That's fine. That's why everything is reviewed. I'll close this.

PR #3919, opened 2020-02-24
Add exception documentation for MulticastOption.Group

eerhardt:
## Summary

Updating the exception documentation on MulticastOption.Group, which now throws `ArgumentNullException` on .NET 5.

See https://github.com/dotnet/runtime/pull/32518, which made the change.

/cc @carlossanlop

### DIFF
--- a/xml/System.Net.Sockets/MulticastOption.xml
+++ b/xml/System.Net.Sockets/MulticastOption.xml
@@ -272,7 +272,7 @@
  ]]></format>
         </remarks>
         <exception cref="T:System.ArgumentNullException">
-          <paramref name="group" /> is <see langword="null" />.</exception>
+          .NET 5.0 and later versions: <paramref name="group" /> is <see langword="null" />.</exception>
       </Docs>
     </Member>
     <Member MemberName="InterfaceIndex">


xml/System.Net.Sockets/MulticastOption.xml
```diff
@@ -271,6 +271,8 @@
   
  ]]></format>
         </remarks>
+        <exception cref="T:System.ArgumentNullException">
+          <paramref name="group" /> is <see langword="null" />.</exception>
```
carlossanlop: If this is specific for .NET 5, then we should add the note prefix:

```suggestion
          .NET 5 and above: <paramref name="group" /> is <see langword="null" />.</exception>
```

@BillWagner should we update documentation for code changes that haven't been released?

eerhardt: There's got to be a better way to version the documentation, isn't there?

carlossanlop: Unfortunately no. AFAIK, we don't have a better way to condition showing an exception for certain versions.

cc @mimisasouvanh @joelmartinez Do we have plans to allow showing exceptions for specific versions/monikers?

BillWagner: > should we update documentation for code changes that haven't been released?

We do have a moniker for ".NET 5.0 alpha".  We aren't showing those yet, so we may want to wait to publish until the first public preview goes out.

mimisasouvanh: @carlossanlop @joelmartinez @BillWagner We currently don't have plans to support versioning for exceptions. Is there a big need for that?

carlossanlop: This is only the second time I see us specifying the version that supports an exception. I think we can consider this a low priority for now, but it could be considered for the future.

eerhardt:
Thanks for updating this PR @gewarren. Can it be merged now?

gewarren:
> Thanks for updating this PR @gewarren. Can it be merged now?

Yes, I'll merge it now.

PR #3936, opened 2020-02-25
For obsolote method alternate suggesting XmlDateTimeSerializationMode

buyaa-n:
## Summary

Method `XmlConvert.ToString(DateTime)` is obsolete, and suggested alternate is `XmlConvert.ToString(DateTime, XmlDateTimeSerializationMode)` customers want to know what  `XmlDateTimeSerializationMode` is best fit for the replacement. Suggested mode is `XmlDateTimeSerializationMode.RoundtripKind` but it will not produce exact matching result in some conditions, and there is no other matching mode which produce exact matching replacement, in that case `ToString(DateTime, String)` can be used with format string  `"yyyy-MM-ddTHH:mm:ss.fffffffzzzzzz"`. Updating notes section for suggested `XmlDateTimeSerializationMode` and exact match atlernate

Fixes https://github.com/dotnet/dotnet-api-docs/issues/2665


tarekgh:
LGTM. can we also suggest to use the overload which takes DateTimeOffset too?

xml/System.Xml/XmlConvert.xml
```diff
@@ -1844,6 +1844,7 @@ writer.WriteElementString(XmlConvert.EncodeName("Order Detail"),"My order");
   
 > [!NOTE]
 >  The <xref:System.Xml.XmlConvert.ToString%28System.DateTime%29?displayProperty=nameWithType> method is obsolete in the 2.0 version of the .NET Framework and has been replaced with the <xref:System.Xml.XmlConvert.ToString%28System.DateTime%2CSystem.Xml.XmlDateTimeSerializationMode%29?displayProperty=nameWithType> method.  
+  Suggested mode is XmlDateTimeSerializationMode.RoundtripKind, if exact matching result expected please use <xref:System.Xml.XmlConvert.ToString%28System.DateTime%2CSystem.String%29?displayProperty=nameWithType> with format string 'yyyy-MM-ddTHH:mm:ss.fffffffzzzzzz'
```
carlossanlop:
```suggestion
  The suggested mode is <xref:System.Xml.XmlDateTimeSerializationMode.RoundtripKind>. If an exact match is expected, please use <xref:System.Xml.XmlConvert.ToString%28System.DateTime%2CSystem.String%29?displayProperty=nameWithType> with the format string `yyyy-MM-ddTHH:mm:ss.fffffffzzzzzz`. 
```

buyaa-n:
> LGTM. can we also suggest to use the overload which takes DateTimeOffset too?

Not sure, for using `ToString(DateTimeOffset, String)` overload current `DateTime` instance  need to be converted or cast to `DateTimeOffset ` type, cannot be used directly i assume. Then next question would come up like which conversion would give closest/exact match etc, not sure if we need to add such info in this doc

PR #3943, opened 2020-02-27
Rename readerOptions to options in JsonDocument exceptions

carlossanlop:
The parameters were renamed. They already have the correct name in `<Parameter>` and in `<param>` tags above, but they were not updated in the exceptions.


carlossanlop:
> How did you notice this?

@layomia My tool that auto ports comments detected those exceptions as completely new because their value was different just by the parameter name.

PR #3960, opened 2020-02-28
Update GetRandomFileName remark

carlossanlop:
Updating the `Path.GetRandomFileName` remark as suggested in https://github.com/dotnet/runtime/issues/31271.


lpereira:
The original wording was pretty strong regarding the unpredictability of the file names this function would generate, even though it could never provide such guarantees (considering it would only provide 8 random characters).  How is the process to make a change like this?  Shouldn't this go into a "what you definitely need to know when updating to .NET 5" doc?

jkotas:
> what you definitely need to know when updating to .NET 5 doc

This edit is not specific to .NET 5. It applies to all versions.

PR #3982, opened 2020-03-10
Explain that CancellationTokenSource.Token is not cancelled when Dispose is called

wazzamatazz:
## Summary

Add note to Dispose documentation to highlight that the `CancellationTokenSource` does not cancel its underlying `CancellationToken` when it is disposed, and that cancellation must be triggered separately.


kouvel:
Looks right to me, CC @tarekgh

stephentoub:
LGTM

PR #3992, opened 2020-03-16
Key sizes can only be certain lengths

tkhadimullin:
The wording could be interpreted as "any value between 128 and 256 bits is valid" leading to confision.
Specifying exact values will hopefully provide more clarity and eliminate confusion.
See https://github.com/microsoft/referencesource/blob/master/System.Core/System/Security/Cryptography/AesCryptoServiceProvider.cs#L323 for implementation details.

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Security.Cryptography/AesCng.xml
+++ b/xml/System.Security.Cryptography/AesCng.xml
@@ -633,6 +633,7 @@
   
 ## Remarks 
  For AES, the legal key sizes are 128, 192, and 256 bits.
+
  If you've created the <xref:System.Security.Cryptography.AesCng> object using an existing persisted key and you set the value of the <xref:System.Security.Cryptography.AesCng.KeySize%2A> property,  the persisted key will no longer be used and an ephemeral key will be used instead. If you need to use the persisted key again, a new instance of <xref:System.Security.Cryptography.AesCng> class needs to be created.  
   
  ]]></format>


tkhadimullin:
Documentation for `Aes` itself does not seem to make any representations regarding legal key sizes. The source code actually defines what sizes are valid: https://github.com/microsoft/referencesource/blob/master/mscorlib/system/security/cryptography/aes.cs#L21. 
As far as I understand it, the logic is "between 128 and 256 with increments of 64 bits. Which still yields the same 3 key sizes. Arguably this mathematics is implementation details and likely does not belong in documentation? 

 `AesCng` actually exposes documentation regarding this property, so I updated it with the same wording.

xml/System.Security.Cryptography/AesCng.xml
```diff
@@ -631,7 +631,8 @@
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
-## Remarks  
+## Remarks 
+ For AES, the legal key sizes are 128, 192, and 256 bits.
  If you've created the <xref:System.Security.Cryptography.AesCng> object using an existing persisted key and you set the value of the <xref:System.Security.Cryptography.AesCng.KeySize%2A> property,  the persisted key will no longer be used and an ephemeral key will be used instead. If you need to use the persisted key again, a new instance of <xref:System.Security.Cryptography.AesCng> class needs to be created.
```
bartonjs: There should be a blank line between these statements so that they become separate paragraphs on the docs site.

tkhadimullin: i'm not sure if I put newline in the right place?

bartonjs: Yep, looks like what I'd expect.

tkhadimullin:
was there anything else I could do to move this PR along?

bartonjs:
@dotnet/docs This is ready for final review

PR #3994, opened 2020-03-16
Add notes about using impersonation with async/await

kouvel:
## Summary

Added a warning about using `WindowsIdentity.Impersonate` with the async/await pattern, recommending `RunImpersonated` instead.


kouvel:
@dotnet/docs, do you happen to know when this change would get published to the docs site?

gewarren:
It should go live by tomorrow morning. We usually publish once a day.

PR #4006, opened 2020-03-19
Distinguish ModifyAccessRule/ModifyAuditRule params from returns

nxtn:
https://github.com/dotnet/runtime/blob/d2f7b6b0d46a45be29ebafca0584142219ffe495/src/libraries/System.Security.AccessControl/src/System/Security/AccessControl/CommonObjectSecurity.cs#L236-L246

The modification request can succeed without actual changes, e.g. a remove request for an empty list.


gewarren:
Thanks @NextTurn

PR #4014, opened 2020-03-20
Add documentation for JsonConstructorAttribute

layomia:
## Summary

Add documentation for JsonConstructorAttribute added in https://github.com/dotnet/runtime/pull/33444.


carlossanlop:
@layomia adding the file yourself is not the correct way to get this file to show up in Docs.

The assembly reference containing this type needs to be sent by our team to the Docs team, and the Docs build system needs to generate this xml file automatically. After that happens, then you can edit the file (add the new documentation).

@safern @BillWagner when was the last ref assembly update we made? Was this type included in that update? If not, can we get another update?

safern:
The last ref drop we provided is preview1 which doesn't contain this change. It seems like this change will be contained in preview3.

layomia:
I see. I'll open a new PR when the changes flow to the docs. Thanks

Issue #3934, opened 2020-02-25
System 1.x undocumented APIs


carlossanlop:
Here are the undocumented API from `System`, introduced in 1.x (except 3 APIs, which are debt from 2.x).

There are [no specific area owners](https://github.com/dotnet/runtime/blob/master/docs/area-owners.md) for these APIs, so I'll mark this issue as `up-for-grabs`.

For more information on the plan, see: #3904

Total: 25
_Note: The table contains 3 really wide columns._

DocId | Source File Path | Docs URL
-- | -- | --
T:System.ArraySegment`1.Enumerator | https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/ArraySegment`1+Enumerator.xml | https://review.docs.microsoft.com/en-us/dotnet/api/System.ArraySegment-1.Enumerator
M:System.Boolean.System#IComparable#CompareTo(System.Object) | https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Boolean.xml | https://review.docs.microsoft.com/en-us/dotnet/api/System.Boolean.System-IComparable-CompareTo
M:System.Boolean.System#IConvertible#GetTypeCode | https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Boolean.xml | https://review.docs.microsoft.com/en-us/dotnet/api/System.Boolean.System-IConvertible-GetTypeCode
M:System.Boolean.System#IConvertible#ToString(System.IFormatProvider) | https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Boolean.xml | https://review.docs.microsoft.com/en-us/dotnet/api/System.Boolean.System-IConvertible-ToString
M:System.Byte.System#IComparable#CompareTo(System.Object) | https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Byte.xml | https://review.docs.microsoft.com/en-us/dotnet/api/System.Byte.System-IComparable-CompareTo
M:System.Byte.System#IConvertible#GetTypeCode | https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Byte.xml | https://review.docs.microsoft.com/en-us/dotnet/api/System.Byte.System-IConvertible-GetTypeCode
[...]

PR #4020, opened 2020-03-21
Automatically port DateTime comments (EII)

carlossanlop:
Fixes some APIs from: https://github.com/dotnet/dotnet-api-docs/issues/3934


xml/System/DateTime.xml
```diff
@@ -6126,9 +6126,25 @@ The following example demonstrates the <xref:System.DateTime.Parse%28System.Stri
       </Parameters>
       <Docs>
         <param name="value">To be added.</param>
```
carlossanlop:
```suggestion
        <param name="value">The object to compare.</param>
```

PR #4028, opened 2020-03-22
Add remarks regarding the initial content of span in string.Create

Gnbrkm41:
Fixes dotnet/runtime#33907.

cc @bartonjs @stephentoub @jkotas

### DIFF
--- a/xml/System/String.xml
+++ b/xml/System/String.xml
@@ -4493,7 +4493,12 @@ Depending on Why you want to call the `Copy` method, there are a number of alter
         <param name="action">A callback to initialize the string.</param>
         <summary>Creates a new string with a specific length and initializes it after creation by using the specified callback.</summary>
         <returns>The created string.</returns>
-        <remarks>The initial content of the destination span passed to <paramref name="action" /> is undefined. Therefore, it is the delegate's responsiblity to ensure that every element of the span is assigned. Otherwise, the resulting string could contain random characters.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+## Remarks
+The initial content of the destination span passed to `action` is undefined. Therefore, it is the delegate's responsibility to ensure that every element of the span is assigned. Otherwise, the resulting string could contain random characters.
+ ]]></format>
+        </remarks>
       </Docs>
     </Member>
     <Member MemberName="Empty">


xml/System/String.xml
```diff
@@ -4493,7 +4493,7 @@ Depending on Why you want to call the `Copy` method, there are a number of alter
         <param name="action">A callback to initialize the string.</param>
         <summary>Creates a new string with a specific length and initializes it after creation by using the specified callback.</summary>
         <returns>The created string.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>The initial values of the span passed into <paramref name="action" /> are undefined. Therefore, it is caller's responsiblity to ensure every characters of the span are assigned to.</remarks>
```
stephentoub:
```suggestion
        <remarks>The initial content of the destination span passed to <paramref name="action" /> is undefined. Therefore, it is the delegate's responsiblity to ensure that every element of the span is assigned. Otherwise, the resulting string could contain random characters.</remarks>
```

Gnbrkm41: Thanks for the suggestion 🙂

xml/System/String.xml
```diff
@@ -4493,7 +4493,7 @@ Depending on Why you want to call the `Copy` method, there are a number of alter
         <param name="action">A callback to initialize the string.</param>
         <summary>Creates a new string with a specific length and initializes it after creation by using the specified callback.</summary>
         <returns>The created string.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>The initial content of the destination span passed to <paramref name="action" /> is undefined. Therefore, it is the delegate's responsiblity to ensure that every element of the span is assigned. Otherwise, the resulting string could contain random characters.</remarks>
```
bartonjs: Remarks are a bit different than normal xmldoc:

```suggestion
        <remarks>
          <format type="text/markdown"><![CDATA[
## Remarks

The initial content of the destination span passed to `action` is undefined. Therefore, it is the delegate's responsibility to ensure that every element of the span is assigned. Otherwise, the resulting string could contain random characters.
 ]]></format>
        </remarks>
```

Gnbrkm41: Ugh, right... forgot about it.

PR #4031, opened 2020-03-23
Document behavior of StartProfile with null argument

jkotas:
Contributes to https://github.com/dotnet/runtime/issues/33426


xml/System.Runtime/ProfileOptimization.xml
```diff
@@ -143,6 +143,8 @@
   
  The code that performs the recording tracks the rate at which methods are compiled, including methods that are compiled on both the background thread and on the main thread. When the rate falls below a predetermined level, recording stops. Recording also stops if the rate of JIT compilation exceeds a predetermined upper limit.  
   
+ Calling this method stops any previous recording in progress. Calling this method with null profile file name stops recording in progress and does not start a new recording.
```
stephentoub:
```suggestion
 Calling this method stops any previous recording in progress. Calling this method with a null profile file name stops any recording in progress and does not start a new recording.
```

PR #4055, opened 2020-03-31
Add clarification to SuppressMessageAttribute

jcouv:
I see `// SuppressMessage attributes do not apply to compiler diagnostics.` in Roslyn, but I think it is useful to make that clear in the documentation.

Tagging @BillWagner


jcouv:
@BillWagner @mairaw Could you help with review or routing this? Thanks

PR #4066, opened 2020-04-03
Add note about HttpClientHandler

gewarren:
Contributes to dotnet/docs#16088.

### DIFF
--- a/xml/System.Net.Http/HttpClientHandler.xml
+++ b/xml/System.Net.Http/HttpClientHandler.xml
@@ -35,7 +35,7 @@ The `HttpClientHandler` class and classes derived from it enable developers to c
 
 ### HttpClientHandler in .NET Core
 
-Starting in .NET Core 2.1, the implementation of the `HttpClientHandler` class was changed to be based on the HTTP protocol stack used by the <xref:System.Net.Http.SocketsHttpHandler?displayProperty=nameWithType> class. The <xref:System.Net.Http.SocketsHttpHandler> class is a cross-platform, HTTP protocol stack. Prior to .NET Core 2.1, the `HttpClientHandler` class used older HTTP protocol stacks (<xref:System.Net.Http.WinHttpHandler> on Windows and `CurlHandler`, an internal class implemented on top of Linux's native libcurl component, on Linux).
+Starting in .NET Core 2.1, the implementation of the `HttpClientHandler` class was changed to be based on the cross-platform HTTP protocol stack used by the <xref:System.Net.Http.SocketsHttpHandler?displayProperty=nameWithType> class. Prior to .NET Core 2.1, the `HttpClientHandler` class used older HTTP protocol stacks (<xref:System.Net.Http.WinHttpHandler> on Windows and `CurlHandler`, an internal class implemented on top of Linux's native libcurl component, on Linux).
 
 You can configure your app to use the older HTTP protocol stacks in one of the following three ways:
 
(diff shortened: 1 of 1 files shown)


xml/System.Net.Http/HttpClientHandler.xml
```diff
@@ -30,11 +30,14 @@
...
 ## Remarks
- The `HttpClientHandler` class and classes derived from it enable developers to configure a variety of options ranging from proxies to authentication.
+
+The `HttpClientHandler` class and classes derived from it enable developers to configure a variety of options ranging from proxies to authentication.
 
 ### HttpClientHandler in .NET Core
 
-Starting with .NET Core 2.1, the <xref:System.Net.Http.SocketsHttpHandler?displayProperty=nameWithType> class instead of `HttpClientHandler` provides the implementation used by higher-level HTTP networking APIs. You can configure your application to use `HttpClientHandler` instead in any of the following ways:
+Starting with .NET Core 2.1, higher-level HTTP networking APIs use the implementation provided by the <xref:System.Net.Http.SocketsHttpHandler?displayProperty=nameWithType> class instead of `HttpClientHandler`.
+
```
davidsh, line 39: This is not really what is happening here.

It's true that by default in .NET Core 2.1, the implementation of the HttpClientHandler class was changed to be based on the implementation of the SocketsHttpHandler class which was a newly written. cross-platform, HTTP protocol stack.

But using the AppContext switches below and setting them to 'false', re-enable the older HTTP protocol stacks.  On Windows, that is based on WinHttpHandler, a public class. On Linux, it is based on CurlHandler, an internal class implemented on top of Linux's native libcurl component.

The current text below "You can configure your application to use HttpClientHandler instead..." doesn't make any sense to me.

I would suggest that be rewritten to use the explanation I've given.

gewarren: Thanks for the feedback; let me know if this makes more sense now.

gewarren: @davidsh Could you also check if the first sentence in this section makes sense? The switch is more about configuring HttpClientHandler than the higher-level APIs, correct?

https://review.docs.microsoft.com/en-us/dotnet/core/run-time-config/networking?branch=pr-en-us-17733#usesocketshttphandler

davidsh: >@davidsh Could you also check if the first sentence in this section makes sense? The switch is more about configuring HttpClientHandler than the higher-level APIs, correct?

Technically, yes, the switch controls what "implementation" is used for the HttpClientHandler class. This handler is used either explicitly when the developer does this:

```c#
var handler = new HttpClientHandler();
var client = new HttpClient(handler);
```

or implicitly (a handler is created internally) when they do this:

```c#
var client = new HttpClient();
```

However, many developers might not create handler directly via `new HttpClientHandler()`. In fact, some developers don't even create the `HttpClient` directly. Instead, they might use HttpClientFactory or other upstack software.

So, either way, it's probably fine for the first sentence to still reference HttpClient. But if you think you can incorporate my comments here in some way to the docs then feel free.

davidsh:
From: https://review.docs.microsoft.com/en-us/dotnet/core/run-time-config/networking?branch=pr-en-us-17733#usesocketshttphandler

the underlined sentences need to change. It shouldn't say "enables the use of HttpClientHandler". Instead it should say something like "enables the use of the X" where X is either WinHttpHandler or libcurl.

image

PR #4072, opened 2020-04-04
HttpListener.IsSupported always returns true

nxtn:
Contributes to dotnet/docs#5199

### DIFF
--- a/xml/System.Net/HttpListener.xml
+++ b/xml/System.Net/HttpListener.xml
@@ -965,7 +965,7 @@ The following code example demonstrates calling the `Close` method:
       <Docs>
         <summary>Gets a value that indicates whether <see cref="T:System.Net.HttpListener" /> can be used with the current operating system.</summary>
         <value>
-          Returns <see langword="true" /> on all supported platforms.</value>
+          <see langword="true" /> on all platforms.</value>
         <remarks>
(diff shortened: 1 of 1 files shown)


xml/System.Net/HttpListener.xml
```diff
@@ -965,15 +965,10 @@ The following code example demonstrates calling the `Close` method:
       <Docs>
         <summary>Gets a value that indicates whether <see cref="T:System.Net.HttpListener" /> can be used with the current operating system.</summary>
         <value>
-          <see langword="true" /> if <see cref="T:System.Net.HttpListener" /> is supported; otherwise, <see langword="false" />.</value>
+          Returns <see langword="true" /> on all supported platforms.</value>
```
gewarren, line 968: To be consistent with other property value descriptions, I'd reword it:

```suggestion
          <see langword="true" /> if the current platform is supported; otherwise, <see langword="false" />.</value>
```

gewarren: Although, according to [this](https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Property), it should be worded as follows:

"Gets a value that indicates whether the current platform is supported."

@mairaw Is that boolean property guidance still valid?

gewarren:
Thank you Next Turn!

PR #4086, opened 2020-04-08
Corrected description of PrinterSettings.SupportsColor property

### LINKED ISSUE #4083: PrinterSettings.SupportsColor property is broken
[Documentation](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.printing.printersettings.supportscolor?view=netframework-4.8) claims that `SupportsColor` property :

> Gets a value indicating whether this printer supports color printing.  

This is not accurate. This [property](https://github.com/dotnet/runtime/blob/1cf033de96a38b286c07b8311edd5f2eabcfb462/src/libraries/System.Drawing.Common/src/System/Drawing/Printing/PrinterSettings.Windows.cs#L516-L522) P/Invokes [GetDeviceCaps(BITSPIXEL..)](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-getdevicecaps), which returns:

> Number of adjacent color bits for each pixel. 

It does not take into account "color planes" or gray-scale rendering. I would not recommend using this API in .NET Framework. However in .NET Core 5, we fixed it(issue #34630). I think we should publish a workaround in the docs:

```
using System;
using System.Drawing.Printing;
using System.Runtime.InteropServices;

namespace ConsoleApp1
{
    class Program
    {
        static void Main(string[] _)
        {
            PrinterSettingsTest();
        }


        static void PrinterSettingsTest()
        {
            var document = new PrintDocument();
            foreach (string name in PrinterSettings.InstalledPrinters)
            {
                document.PrinterSettings.PrinterName = name;
                Console.WriteLine($"{name} SupportsColor = {document.PrinterSettings.SupportsColor}");
                var capability = GetColorDeviceCapability(name);
                Console.WriteLine($"\tDeviceCapabilities(DC_COLORDEVICE) returns {capability}");
            }

            Console.ReadLine();
        }

        [DllImport("winspool.drv", SetLastError = true, CharSet = CharSet.Unicode, ExactSpelling = true)]
        private static extern int DeviceCapabilitiesW(string pDevice, string pPort, short fwCapabilities, IntPtr pOutput, IntPtr pDevMode);
        private const short DC_COLORDEVICE = 32;

        private static bool GetColorDeviceCapability(string printerName)
        {
            return DeviceCapabilitiesW(printerName, null, DC_COLORDEVICE, IntPtr.Zero, IntPtr.Zero) == 1;
        }
    }
}
```

Tanya-Solyanik:
## Summary

* Corrected description of PrinterSettings.SupportsColor property.
* added a CS code snippet that demonstrates how to obtain a correct value using PInvoke

Things to finish:
* [ ] VB sample
* [ ] CPP sample
* [ ] differentiate between framework versions where this issue is fixed and the ones where it's not. It will be fixed in .NET Core 5.0 only. See issue https://github.com/dotnet/runtime/issues/34630

Fixes #4083

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_Winforms/PaperSource and PaperSize Example with Resolution/CS/source.cs
+++ b/samples/snippets/csharp/VS_Snippets_Winforms/PaperSource and PaperSize Example with Resolution/CS/source.cs
@@ -6,6 +6,7 @@ using System.Collections;
 using System.ComponentModel;
 using System.Windows.Forms;
 using System.Data;
+using System.Runtime.InteropServices;
 
 namespace WindowsApplication3
 {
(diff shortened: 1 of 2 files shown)


Tanya-Solyanik:
Want to make this PR a draft, closing

PR #4099, opened 2020-04-11
Fix unsafe Random use in ReaderWriterLock samples

chwarr:
The ReaderWriterLock samples were sharing a single Random instance
across threads. The various .Next methods on Random instances are not
safe to be called concurrently.

Now, each thread has its own Random instance that it uses. Some of the
worker methods needed to be adjusted to take a Random instance as input.

I considered using a thread-local Random instance. However, the sample
is already complex, so I opted to avoid introducing thread-local values,
which would need to be briefly explained.


chwarr:
I don't know whether or not the documentation build process build sample code as well. I have built and run all of these manually already.

PR #4113, opened 2020-04-15
Fix ChangePassword description

uecasm:
The existing text implies that `ChangePassword` will modify the password contained in the `SqlCredential`, which is not at all what it actually does.


carlossanlop:
@cheenamalhotra @david-engel @karinazhou @JRahnama can you please review?

PR #4115, opened 2020-04-15
Document Decoder.Convert exception conditions

AArnott:
The doc incorrectly stated in the remarks section that this method does not throw when the output buffer is too small.
In its exception section it incorrectly indicates that the output buffer must be large enough for the entire input buffer.

The correction indicates that the output buffer need only be **2 chars** in length in order to avoid the exception.

### DIFF
--- a/xml/System.Text/Decoder.xml
+++ b/xml/System.Text/Decoder.xml
@@ -172,7 +172,7 @@ Any remaining processed data that is part of a logical unit, such as the high su
 
 The `Convert` method is designed to be used in a loop to decode an arbitrary amount of input, such as data read from a file or stream. 
 It stores the output of the decoding operation in a fixed-size buffer.
-<xref:System.Text.Decoder.GetChars%2A> will throw an exception if the output buffer isn't large enough, but <xref:System.Text.Decoder.Convert%2A> will fill as much space as possible and return the bytes read and chars written. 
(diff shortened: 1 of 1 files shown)


tarekgh:
@AArnott isn't the behavior depends on the passed `flush` parameter? if it is true, will need to have enough space in the buffer?

tarekgh:
I looked more and it looks the default implementation is behaving as you described it. The only thing here is if any other decoder overrides it and throws. I think the change is reasonable as it describes the default behavior.

GrabYourPitchforks:
@tarekgh You don't need to override the decoder to trigger an exception, even with 2 output chars in the buffer. You could easily trigger an exception by using the built-in `EncoderReplacementFallback` and providing a replacement string of more than one char.

```cs
Encoding encoding = Encoding.GetEncoding("utf-8", EncoderFallback.ExceptionFallback, new DecoderReplacementFallback("???"));
Decoder decoder = encoding.GetDecoder();

char[] chars = new char[2];
int charCount = decoder.GetChars(new byte[] { 0xC0 }, 0, 1, chars, 0, flush: false); // throws
```

The reason for this is that the decoder sees the invalid `[ C0 ]` byte and triggers the fallback mechanism. The fallback mechanism tries to write three chars, but it fails because the output buffer is only 2 chars in length. Since the decoder couldn't make any forward progress at all (it consumed 0 input bytes), it throws an exception.

Is it reasonable to assume that somebody might be using a replacement fallback in this manner?

tarekgh:
> Is it reasonable to assume that somebody might be using a replacement fallback in this manner?

It is possible but very unlikely.

PR #4147, opened 2020-04-21
Auto port System.Threading ACL comments

carlossanlop:
Automatically ported the triple slash comments in the new ACL types and methods in System.Threading.

Ported with [DocsPortingTool](https://github.com/carlossanlop/DocsPortingTool).

### DIFF
--- a/xml/System.Threading/EventWaitHandleAcl.xml
+++ b/xml/System.Threading/EventWaitHandleAcl.xml
@@ -44,11 +44,11 @@
         <param name="initialState">
           <see langword="true" /> to set the initial state to signaled if the named event is created as a result of this call; <see langword="false" /> to set it to non-signaled.</param>
         <param name="mode">One of the enum values that determines whether the event resets automatically or manually.</param>
-        <param name="name">The name, if the event is a system-wide synchronization event; otherwise, <see langword="null" /> or an empty string.</param>
+        <param name="name">The name, if the event is to be shared with other processes; otherwise, <see langword="null" /> or an empty string.</param>
         <param name="createdNew">When this method returns, this argument is always set to <see langword="true" /> if a local event is created; that is, when <paramref name="name" /> is <see langword="null" /> or <see cref="F:System.String.Empty" />. If <paramref name="name" /> has a valid, non-empty value, this argument is set to <see langword="true" /> when the system event is created, or it is set to <see langword="false" /> if an existing system event is found with that name.</param>
         <param name="eventSecurity">The optional Windows access control security to apply.</param>
         <summary>Gets or creates an <see cref="T:System.Threading.EventWaitHandle" /> instance, allowing a <see cref="T:System.Security.AccessControl.EventWaitHandleSecurity" /> instance to be optionally specified to set it during the event creation.</summary>
-        <returns>An object that represents a system event wait handle, if named, or a local event wait handle, if nameless.</returns>
+        <returns>An object that represents an event wait handle as requested.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
@@ -58,7 +58,11 @@ If a `name` is passed and the system event already exists, the existing event is
 
           ]]></format>
         </remarks>
-        <exception cref="T:System.ArgumentNullException">.NET Framework only: The <paramref name="name" /> length is beyond MAX_PATH (260 characters).</exception>
+        <exception cref="T:System.ArgumentException">.NET Framework only: The <paramref name="name" /> length is beyond MAX_PATH (260 characters).
+
+-or-
+
+.NET Framework only: The <paramref name="mode" /> enum value was out of legal range.</exception>
         <exception cref="T:System.ArgumentOutOfRangeException">The <paramref name="mode" /> enum value was out of legal range.</exception>
         <exception cref="T:System.IO.DirectoryNotFoundException">Could not find a part of the path specified in <paramref name="name" />.</exception>
         <exception cref="T:System.Threading.WaitHandleCannotBeOpenedException">A system-wide synchronization event with the provided <paramref name="name" /> was not found.


tarekgh:
I'll delegate this review to @kouvel

xml/System.Threading/EventWaitHandleAcl.xml
```diff
@@ -41,14 +41,28 @@
...
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+If a `name` is passed and the system event already exists, the existing event is returned. If `name` is `null` or <xref:System.String.Empty>, a new local event is always created.
+
+          ]]></format>
+        </remarks>
+        <exception cref="T:System.ArgumentNullException">.NET Framework only: The <paramref name="name" /> length is beyond MAX_PATH (260 characters).</exception>
```
kouvel, line 61: The `MAX_PATH` issue would result in an `ArgumentException`, not an `ArgumentNullException`. From legacy code review anyway, I don't see a case where `ArgumentNullException` would be thrown by this method.

xml/System.Threading/EventWaitHandleAcl.xml
```diff
@@ -41,14 +41,28 @@
...
+        <exception cref="T:System.ArgumentOutOfRangeException">The <paramref name="mode" /> enum value was out of legal range.</exception>
```
kouvel, line 62: Hmm I see an `ArgumentException` thrown by .NET Framework, `ArgumentOutOfRangeException` would make sense in this case but I'm wondering now if I'm seeing different docs on which this port was based (I'm currently looking at .NET Framework docs for the equivalent APIs).

xml/System.Threading/EventWaitHandleAcl.xml
```diff
@@ -41,14 +41,28 @@
...
+        <exception cref="T:System.IO.DirectoryNotFoundException">Could not find a part of the path specified in <paramref name="name" />.</exception>
```
kouvel, line 63: I don't see this exception being raised by this API, the `name` parameter has nothing to do with directories on Windows, and on Unixes it's an implementation detail and probably should not result in such an exception

carlossanlop:
> _What sources were used for the port?_

These are the new types and methods I introduced in System.Threading back in November.

PRs and source files:

- MutexAcl - [PR](https://github.com/dotnet/corefx/pull/42213) -  [runtime/src/libraries/System.Threading.AccessControl/src/System/Threading/MutexAcl.cs](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Threading.AccessControl/src/System/Threading/MutexAcl.cs)
- EventWaitHandleAcl - [PR](https://github.com/dotnet/corefx/pull/42213) - [runtime/src/libraries/System.Threading.AccessControl/src/System/Threading/EventWaitHandleAcl.cs](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Threading.AccessControl/src/System/Threading/EventWaitHandleAcl.cs)
- SemaphoreAcl - [PR](https://github.com/dotnet/corefx/pull/42377) - [runtime/src/libraries/System.Threading.AccessControl/src/System/Threading/SemaphoreAcl.cs](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Threading.AccessControl/src/System/Threading/SemaphoreAcl.cs)

xml/System.Threading/EventWaitHandleAcl.xml
carlossanlop: Good catch.
```suggestion
        <exception cref="T:System.ArgumentException">.NET Framework only: The <paramref name="name" /> length is beyond MAX_PATH (260 characters).</exception>
```

xml/System.Threading/EventWaitHandleAcl.xml
carlossanlop: .NET Core: https://github.com/dotnet/runtime/blob/8dc13298a89d50e64f24d0e755675bb01a33a399/src/libraries/System.Threading.AccessControl/src/System/Threading/EventWaitHandleAcl.cs#L35-L38

You're right, I should indicate another exception for Framework. I'll add it to the suggestion from the previous comment. https://referencesource.microsoft.com/#mscorlib/system/threading/eventwaithandle.cs,136

xml/System.Threading/EventWaitHandleAcl.xml
carlossanlop: When we reach here: https://github.com/dotnet/runtime/blob/8dc13298a89d50e64f24d0e755675bb01a33a399/src/libraries/System.Threading.AccessControl/src/System/Threading/EventWaitHandleAcl.cs#L82

We throw here: https://github.com/dotnet/runtime/blob/25d316cefafc4c3884f38acf7bdef971d2b9fe4b/src/libraries/Common/src/System/IO/Win32Marshal.cs#L39

Unit test verifying this exception added here: https://github.com/dotnet/runtime/blob/25d316cefafc4c3884f38acf7bdef971d2b9fe4b/src/libraries/System.Threading.AccessControl/tests/EventWaitHandleAclTests.cs#L70-L84

kouvel: I see thanks, I didn't know about that. If it can throw this exception for `EventWaitHandle` then it would throw in similar cases for the other types as well. This is fine for now, I'll look into fixing issues, filed issue https://github.com/dotnet/runtime/issues/35518.

xml/System.Threading/EventWaitHandleAcl.xml
kouvel: Hmm there are many inconsistencies between in-code docs, tests, and .NET Framework docs. Filed https://github.com/dotnet/runtime/issues/35519 to fix those, I think following the XML docs is good for this PR.

PR #4153, opened 2020-04-22
Correct unobserved Task exception behavior doc

AArnott:
The doc described .NET 4.0 behavior only without stating how on .NET 4.5+ the behavior changed.

### DIFF
--- a/xml/System.Threading.Tasks/Task.xml
+++ b/xml/System.Threading.Tasks/Task.xml
@@ -2599,7 +2599,7 @@ Task t Status: RanToCompletion, Result: 42
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- Tasks that throw unhandled exceptions store the resulting exception and propagate it wrapped in a <xref:System.AggregateException> in calls to <xref:System.Threading.Tasks.Task.Wait%2A> or in accesses to the <xref:System.Threading.Tasks.Task.Exception%2A> property. On .NET Framework 4.0, any exceptions not observed by the time the task instance is garbage collected will be propagated on the finalizer thread, which crashes the process. On .NET Framework 4.5 and later and on .NET Core, unobserved exceptions are not rethrown from the Finalizer. For more information and an example, see [Exception Handling (Task Parallel Library)](~/docs/standard/parallel-programming/exception-handling-task-parallel-library.md).  
+ Tasks that throw unhandled exceptions store the resulting exception and propagate it wrapped in a <xref:System.AggregateException> in calls to <xref:System.Threading.Tasks.Task.Wait%2A> or in accesses to the <xref:System.Threading.Tasks.Task.Exception%2A> property. On .NET Framework 4.0, any exceptions not observed by the time the task instance is garbage collected will be propagated on the finalizer thread, which crashes the process. On .NET Framework 4.5 and later the default behavior changed so unobserved exceptions are not rethrown from the Finalizer. .NET Core does not rethrow the exception on the Finalizer. For more information and an example, see [Exception Handling (Task Parallel Library)](~/docs/standard/parallel-programming/exception-handling-task-parallel-library.md).  
   
  ]]></format>
         </remarks>


xml/System.Threading.Tasks/Task.xml
```diff
@@ -2599,7 +2599,7 @@ Task t Status: RanToCompletion, Result: 42
...
+ Tasks that throw unhandled exceptions store the resulting exception and propagate it wrapped in a <xref:System.AggregateException> in calls to <xref:System.Threading.Tasks.Task.Wait%2A> or in accesses to the <xref:System.Threading.Tasks.Task.Exception%2A> property. On .NET 4.0, any exceptions not observed by the time the task instance is garbage collected will be propagated on the finalizer thread, which crashes the process. On .NET 4.5 and later, unobserved exceptions are not rethrown from the Finalizer. For more information and an example, see [Exception Handling (Task Parallel Library)](~/docs/standard/parallel-programming/exception-handling-task-parallel-library.md).
```
stephentoub, line 2602: > are not rethrown from the Finalizer

At least not by default:
https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/file-schema/runtime/throwunobservedtaskexceptions-element
Effectively .NET 4.5 changed the default value of that from true to false.

stephentoub, line 2602: Can we say ".NET Framework 4.0" and ".NET Framework 4.5" instead of just ".NET 4.0" and ".NET 4.5"?

stephentoub:
> One small question, should we mentioned .Net Core too

I think it'd be good to clarify.

xml/System.Threading.Tasks/Task.xml
AArnott: Do we need to document that here, or is it enough that it's documented where that xml config is documented and we can describe default behavior here?

AArnott: There's already a link at the end of this line to a topic all about exceptions where the details are given.

stephentoub: I think we should at least add "by default" when describing the new behavior.

AArnott: It's only the 'default' on .NET Framework, right? .NET Core doesn't offer the switch?

AArnott: Done.

stephentoub: Correct.  We're just explicitly listing .NET Framework versions here.

AArnott: Fixed. What do you think?

xml/System.Threading.Tasks/Task.xml
```diff
@@ -2599,7 +2599,7 @@ Task t Status: RanToCompletion, Result: 42
...
+ Tasks that throw unhandled exceptions store the resulting exception and propagate it wrapped in a <xref:System.AggregateException> in calls to <xref:System.Threading.Tasks.Task.Wait%2A> or in accesses to the <xref:System.Threading.Tasks.Task.Exception%2A> property. On .NET Framework 4.0, any exceptions not observed by the time the task instance is garbage collected will be propagated on the finalizer thread, which crashes the process. On .NET Framework 4.5 and later the default behavior changed so unobserved exceptions are not rethrown from the Finalizer. .NET Core does not rethrow the exception on the Finalizer. For more information and an example, see [Exception Handling (Task Parallel Library)](~/docs/standard/parallel-programming/exception-handling-task-parallel-library.md).
```
stephentoub, line 2602: Technically the "which crashes the process" behavior is also the default and changeable, but... it's fine :)

stephentoub:
Thanks

xml/System.Threading.Tasks/Task.xml
AArnott: Interesting. I can configure the Finalizer to not crash the process if it throws? I didn't know that. But if it's fine, I'll leave it. :)

PR #4194, opened 2020-05-05
Corrected and simplified `Array.Sort` and `List<T>.Sort` time complexity remarks.

RelativelyRandom:
Remarks for `List<T>.Sort` incorrectly state that the method's worst-case time complexity is O(n<sup>2</sup>). This is probably a leftover from when it used just the basic quicksort.

`Array.Sort` is technically not wrong:

> For arrays that are sorted by using the Heapsort and Quicksort algorithms, in the worst case, this method is an O(`n` log `n`) operation

However, since insertion sort is limited to small arrays and doesn't influence the big O, it is just as correct to simply state that the method is an O(n log n) operation.


gewarren:
@eiriktsarpalis @layomia Could you review this PR?

PR #4204, opened 2020-05-07
Update docs for IsNNN extension methods on IHost(ing)Environment

analogrelay:
## Summary

Update docs to clarify the purpose of the `IsNNN` extension methods on `IHostEnvironment` and `IHostingEnvironment` as being specifically for end-applications.

Fixes dotnet/extensions#3054


anurse:
Wordsmithing recommendations very welcome! The general idea here is that "Environment" names are completely in the control of an application. We provide three default names (Development, Staging, Production) but it's completely up to the application. The linked bug (above) resolved into a request for some clarity on that in the API docs to guide library authors to avoid coupling behavior to specific names (using these APIs).

anurse:
What's needed to finish this PR? I don't have merge permissions :).

PR #4217, opened 2020-05-12
Exclude DependencyInjectionSpecificationTests in _filter.xml

TianqiZhang:
this is a unit test class which should not be in our api docs.

### DIFF
--- a/xml/_filter.xml
+++ b/xml/_filter.xml
@@ -161,5 +161,9 @@
     <namespaceFilter name="Microsoft.Extensions.Logging.Internal">
       <typeFilter name="*" expose="false" />
     </namespaceFilter>
+    <namespaceFilter name="Microsoft.Extensions.DependencyInjection.Specification">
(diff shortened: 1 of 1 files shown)


TianqiZhang:
Thank you @BillWagner ! Actually, when I dig deeper into this namespace, it made me realize that maybe we should get rid of this assembly `Microsoft.Extensions.DependencyInjection.Specification.Tests.dll` entirely from `apidrop/binaries` repo. This dll looks like a unit test dll and doesn't contain useful APIs.

For example, in [this class](https://review.docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.specification.classwithoptionalargsctor?view=dotnet-plat-ext-2.2&branch=master) we can even see a property named 'Whatever'.

BillWagner:
Thanks for the added investigation @TianqiZhang 

I'm in the process of getting Preview 4 ready, so I'll remove that DLL from the build and test it.

Issue #4038, opened 2020-03-25
Update Refresh documentation when folder is created


carlossanlop:
As requested here:
https://github.com/dotnet/corefx/pull/40677#issuecomment-526671157

The state is invalidated when directory is created, so the behavior/usage of Refresh needs to be documented.

cc @wfurt

PR #4240, opened 2020-05-14
Update data refresh description on DirectoryInfo.Create

carlossanlop:
Fixes https://github.com/dotnet/dotnet-api-docs/issues/4038
Original issue with the behavior change: https://github.com/dotnet/corefx/pull/40677
@wfurt please take a look. Let me know if my modifications were accurate.

### DIFF
--- a/xml/System.IO/FileSystemInfo.xml
+++ b/xml/System.IO/FileSystemInfo.xml
@@ -1282,6 +1282,7 @@ On Unix platforms that do not support creation or birth time, this property retu
   
 ## Remarks  
  `FileSystemInfo.Refresh` takes a snapshot of the file from the current file system. `Refresh` cannot correct the underlying file system even if the file system returns incorrect or outdated information. This can happen on platforms such as Windows 98.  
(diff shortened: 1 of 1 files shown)


xml/System.IO/FileSystemInfo.xml
```diff
@@ -1283,8 +1283,6 @@ On Unix platforms that do not support creation or birth time, this property retu
 ## Remarks  
  `FileSystemInfo.Refresh` takes a snapshot of the file from the current file system. `Refresh` cannot correct the underlying file system even if the file system returns incorrect or outdated information. This can happen on platforms such as Windows 98.
```
carlossanlop, line 1285: I looked at the code again, and I think this line should be reverted.

Both FileInfo and DirectoryInfo get their data refreshed when retrieving attributes, if the data has been invalidated (-1).
We do invalidate our data when we make changes via code, but if the user did not make any invocations that may have changed state, the data could've still been updated externally, so the user should still call Refresh.

```suggestion
  Calls must be made to `Refresh` before attempting to get the attribute information, or the information will be outdated.  
  
```

PR #4247, opened 2020-05-16
Update DockPanel.xml

VaderConsulting:
The default value of the LastChildFill property is actually False, not True as originally stated.

## Summary

Removed ", which is the default setting".


gewarren:
@dotnet/wpf-developers Can you confirm this change?

SamBent:
The published doc is correct.  The default value of LastChildFill is True.  No change is needed.

gewarren:
Thanks @SamBent. I'm going to close this PR then.

PR #4250, opened 2020-05-17
Clarify String.Contains

danmoseley:
## Summary

Clarify that `String.Contains(char)` uses Ordinal comparison. (This may be assumed to most, but it clarifies when one needs to use `String.Contains(char, StringComparison)`.

Note that the extension method version of `String.Contains(string, StringComparison)` is only relevant to .NET Framework. The method is included in .NET Core.

### DIFF
--- a/xml/System/String.xml
+++ b/xml/System/String.xml
@@ -4105,7 +4105,13 @@ Examples of instantiating strings:
         <summary>Returns a value indicating whether a specified character occurs within this string.</summary>
         <returns>
           <see langword="true" /> if the <paramref name="value" /> parameter occurs within this string; otherwise, <see langword="false" />.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+ This method performs an ordinal (case-sensitive and culture-insensitive) comparison.
+ ]]></format>
+        </remarks>
       </Docs>
     </Member>
     <Member MemberName="Contains">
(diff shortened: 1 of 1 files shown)


xml/System/String.xml
```diff
@@ -4105,7 +4105,13 @@ Examples of instantiating strings:
         <summary>Returns a value indicating whether a specified character occurs within this string.</summary>
         <returns>
           <see langword="true" /> if the <paramref name="value" /> parameter occurs within this string; otherwise, <see langword="false" />.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
```
pgovind, line 4108: My understanding is that remarks don't show in intellisense. If we want this to show up in VS, we should move this change up into the summary.

danmosemsft: Just copying `Contains (string value)`. @BillWagner ?

xml/System/String.xml
```diff
@@ -4160,7 +4166,7 @@ Examples of instantiating strings:
...
 
- To determine whether a string contains a specified substring by using something other than ordinal comparison (such as culture-sensitive comparison, or ordinal case-insensitive comparison), you can create a custom method. The following example illustrates one such approach. It defines a <xref:System.String> extension method  that includes a <xref:System.StringComparison> parameter and indicates whether a string contains a substring when using the specified form of string comparison.
+ **.NET Framework only**: To determine whether a string contains a specified substring by using something other than ordinal comparison (such as culture-sensitive comparison, or ordinal case-insensitive comparison), you can create a custom method. The following example illustrates one such approach. It defines a <xref:System.String> extension method  that includes a <xref:System.StringComparison> parameter and indicates whether a string contains a substring when using the specified form of string comparison.
```
pgovind, line 4169: Does writing .NET Framework only at the beginning make it seem like the ability exists only in .NET Framework and not in .NET core(when it actually does)?

danmosemsft: Not sure, I just copied this pattern. I defer to docs folks.

xml/System/String.xml
danmosemsft: @BillWagner could you help with that question?

danmosemsft:
@pgovind since the questions above relate to the existing doc, perhaps it would make sense to merge this as is?

pgovind:
> @pgovind since the questions above relate to the existing doc, perhaps it would make sense to merge this as is?

Agreed. Approved

PR #4259, opened 2020-05-19
CI Update: Add MSBuild APIs

BillWagner:
tagging @ghogen 

This has the updated binaries for MS Build.


BillWagner:
Merging so we can unblock updated mdoc fixes

PR #4264, opened 2020-05-20
Remove using statements in first BlockingCollection example.

gurustron:
## Summary
`using ` statement is not needed when calling `Task.WhenAll` or `Task.WaitAll` [ after .NET 4.5](https://devblogs.microsoft.com/pfxteam/do-i-need-to-dispose-of-tasks/).

Removed extra `using` statements from example.


gurustron:
@BillWagner  thank you!

PR #4285, opened 2020-05-26
Automatically port 5.0 System.Runtime.CompilerServices documentation

carlossanlop:
@Anipik @steveharter

We want area owners to learn the process of automatically porting new documentation using this tool, and we want people to get used to do this once a month for any new APIs, so they are available in the next Preview.

This PR is to show the commands they can use to run [DocsPortingTool](https://github.com/carlossanlop/DocsPortingTool), in case they are not yet familiar with them, so you learn how to port documents in your area for the next preview.

Steps:

1. I used this command first:
`.\DocsPortingTool.exe -Docs D:\dotnet-api-docs\xml\ -TripleSlash D:\runtime\artifacts\bin,D:\runtime\artifacts\bin\coreclr\Windows_NT.x64.Release\IL -Save true -IncludedAssemblies System.Runtime.CompilerServices`

2. Merged a commit for what was found in step 1, making sure it all looks ok.

3. Then I ran this command on top of the already commited changes, to find exceptions (I didn't find any):
`.\DocsPortingTool.exe -Docs D:\dotnet-api-docs\xml\ -TripleSlash D:\runtime\artifacts\bin,D:\runtime\artifacts\bin\coreclr\Windows_NT.x64.Release\IL -Save true -IncludedAssemblies System.Runtime.CompilerServices -SkipExceptions false`
Note: This command is noisy because the tool does not have an easy way to determine if an exception already exists in dotnet-api-docs or not, mainly because tends to change a lot compared to triple slash comments.

### DIFF
--- a/xml/System.Runtime.CompilerServices/RuntimeHelpers.xml
+++ b/xml/System.Runtime.CompilerServices/RuntimeHelpers.xml
@@ -88,7 +88,7 @@
       <Docs>
         <param name="type">The type associated with the allocated memory.</param>
         <param name="size">The amount of memory to allocate, in bytes.</param>
-        <summary>Allocate memory that is associated with the <paramref name="type" /> and will be freed if and when the <see cref="T:System.Type" /> is unloaded.</summary>
+        <summary>Allocates memory that's associated with the <paramref name="type" /> and is freed if and when the <see cref="T:System.Type" /> is unloaded.</summary>
         <returns>The allocated memory.</returns>
         <remarks>To be added.</remarks>
       </Docs>


xml/System.Runtime.CompilerServices/RuntimeHelpers.xml
```diff
@@ -86,10 +86,10 @@
         <Parameter Name="size" Type="System.Int32" Index="1" FrameworkAlternate="net-5.0" />
       </Parameters>
       <Docs>
-        <param name="type">To be added.</param>
-        <param name="size">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="type">Type associated with the allocated memory.</param>
```
gewarren, line 89:
```suggestion
        <param name="type">The type associated with the allocated memory.</param>
```

xml/System.Runtime.CompilerServices/RuntimeHelpers.xml
```diff
@@ -86,10 +86,10 @@
...
+        <param name="size">Amount of memory in bytes to allocate.</param>
```
gewarren, line 90:
```suggestion
        <param name="size">The amount of memory to allocate, in bytes.</param>
```

xml/System.Runtime.CompilerServices/RuntimeHelpers.xml
```diff
@@ -86,10 +86,10 @@
...
+        <summary>Allocate memory that is associated with the <paramref name="type" /> and will be freed if and when the <see cref="T:System.Type" /> is unloaded.</summary>
```
gewarren, line 91:
```suggestion
        <summary>Allocates memory that's associated with the <paramref name="type" /> and is freed if and when the <see cref="T:System.Type" /> is unloaded.</summary>
```

xml/System.Runtime.CompilerServices/SkipLocalsInitAttribute.xml
```diff
@@ -19,8 +19,16 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>To be added.</summary>
-    <remarks>To be added.</remarks>
+    <summary>Used to indicate to the compiler that the .locals init flag should not be set in nested method headers when emitting to metadata.</summary>
```
gewarren, line 22:
```suggestion
    <summary>Indicates to the compiler that the .locals init flag should not be set in nested method headers when emitting to metadata.</summary>
```

xml/System.Runtime.CompilerServices/SkipLocalsInitAttribute.xml
```diff
@@ -19,8 +19,16 @@
...
+    <remarks>
+      <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+This attribute is unsafe because it may reveal uninitialized memory to the application in certain instances (e.g., reading from uninitialized stackalloc'd memory). If applied to a method directly, the attribute applies to that method and all nested functions (lambdas, local functions) below it. If applied to a type or module, it applies to all methods nested inside. This attribute is intentionally not permitted on assemblies. Use at the module level instead to apply to multiple type declarations.
```
gewarren, line 28:
```suggestion
This attribute is unsafe, because it may reveal uninitialized memory to the application in certain instances (for example, reading from uninitialized stack-allocated memory). If applied to a method directly, the attribute applies to that method and all its nested functions, including lambdas and local functions. If applied to a type or module, it applies to all methods nested inside. This attribute is intentionally not permitted on assemblies. To apply the attribute to multiple type declarations, use it at the module level instead.
```

xml/System.Runtime.CompilerServices/RuntimeHelpers.xml
```diff
@@ -86,10 +86,10 @@
...
+        <returns>The allocated memory</returns>
```
carlossanlop, line 92:
```suggestion
        <returns>The allocated memory, in bytes.</returns>
```

carlossanlop:
Thanks for your review, @gewarren. I applied your suggestions.

BillWagner:
@carlossanlop It looks like there is one change still requested. Once that's made, this is ready.

PR #4297, opened 2020-05-28
A small correction on the note about stream size rounded up by page size

shjin404:
The view size cannot be smaller than mapped file size. It is just getting rounded up by page size which is normally 4kb.

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.IO.MemoryMappedFiles/MemoryMappedFile.xml
+++ b/xml/System.IO.MemoryMappedFiles/MemoryMappedFile.xml
@@ -1396,7 +1396,7 @@ Process C says: True
 ## Remarks  
  You can use the view returned by this method for random access to a memory-mapped file.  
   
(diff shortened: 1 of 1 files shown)


shjin404:
From dev comment found from github, 

When MemoryMappedFileOptions.DelayAllocatePages option is used, the view size can be smaller than the mapped file size if memory is not committed. Should this be added as a comment to make this better?

https://github.com/dotnet/runtime/blob/4f9ae42d861fcb4be2fcd5d3d55d5f227d30e723/src/libraries/System.IO.MemoryMappedFiles/src/System/IO/MemoryMappedFiles/MemoryMappedView.Windows.cs#L51-L60

gewarren:
@carlossanlop @jozkee Can you review this PR?

jeffhandley (approved):
@shjin404 Sorry this sat for so long. Thank you for making this improvement. I validated the behavior of rounding up based on the source here: https://github.com/dotnet/runtime/blob/master/src/libraries/System.IO.MemoryMappedFiles/src/System/IO/MemoryMappedFiles/MemoryMappedView.Windows.cs#L20-L25

PR #4298, opened 2020-05-28
fix remarks CDATA format in sql apis

TianqiZhang:
We found some double `<format>` or double `<remarks>` tags in sql apis, this PR is to fix these tags.


carlossanlop:
@TianqiZhang the build should've failed when these were originally merged, no?

PR #4326, opened 2020-06-03
To be added (test only)

gewarren:
Test for how strict the "To be added" string match is.


gewarren:
Answer: Match must be exact, with period, and must be the only text.

PR #4345, opened 2020-06-07
Added important sideeffect behaviour

merarischroeder:
RE: internal call to Advance

## Summary

Added further information in the description

### DIFF
--- a/xml/System.IO.Pipelines/PipeWriter.xml
+++ b/xml/System.IO.Pipelines/PipeWriter.xml
@@ -426,9 +426,9 @@ You must request a new buffer after calling <xref:System.IO.Pipelines.PipeWriter
       <Docs>
         <param name="source">The read-only byte memory region to write.</param>
         <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="P:System.Threading.CancellationToken.None" />.</param>
-        <summary>Writes the specified byte memory range to the pipe and makes data accessible to the <see cref="T:System.IO.Pipelines.PipeReader" />. This function calls Advance internally, if you call Advance after you will get bytes with value 0 in the buffer.</summary>
+        <summary>Writes the specified byte memory range to the pipe and makes data accessible to the <see cref="T:System.IO.Pipelines.PipeReader" />.</summary>
         <returns>A task that represents the asynchronous write operation, and wraps the flush asynchronous operation.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>This method calls <see cref="M:System.IO.Pipelines.PipeWriter.Advance(System.Int32)" /> internally. If you call <see cref="M:System.IO.Pipelines.PipeWriter.Advance(System.Int32)" /> afterwards, the bytes in the buffer will all have a value of 0.</remarks>
       </Docs>
     </Member>
   </Members>


xml/System.IO.Pipelines/PipeWriter.xml
```diff
@@ -426,7 +426,7 @@ You must request a new buffer after calling <xref:System.IO.Pipelines.PipeWriter
       <Docs>
         <param name="source">The read-only byte memory region to write.</param>
         <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="P:System.Threading.CancellationToken.None" />.</param>
-        <summary>Writes the specified byte memory range to the pipe and makes data accessible to the <see cref="T:System.IO.Pipelines.PipeReader" />.</summary>
+        <summary>Writes the specified byte memory range to the pipe and makes data accessible to the <see cref="T:System.IO.Pipelines.PipeReader" />. This function calls Advance internally, if you call Advance after you will get bytes with value 0 in the buffer.</summary>
         <returns>A task that represents the asynchronous write operation, and wraps the flush asynchronous operation.</returns>
         <remarks>To be added.</remarks>
```
gewarren, line 431:
```suggestion
        <remarks>This method calls <see cref="M:System.IO.Pipelines.PipeWriter.Advance(System.Int32)" /> internally. If you call <see cref="M:System.IO.Pipelines.PipeWriter.Advance(System.Int32)" /> afterwards, the bytes in the buffer will all have a value of 0.</remarks>
```

carlossanlop:
Ping @halter73 @davidfowl

xml/System.IO.Pipelines/PipeWriter.xml
```diff
@@ -428,7 +428,7 @@ You must request a new buffer after calling <xref:System.IO.Pipelines.PipeWriter
         <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="P:System.Threading.CancellationToken.None" />.</param>
         <summary>Writes the specified byte memory range to the pipe and makes data accessible to the <see cref="T:System.IO.Pipelines.PipeReader" />.</summary>
         <returns>A task that represents the asynchronous write operation, and wraps the flush asynchronous operation.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>This method calls <see cref="M:System.IO.Pipelines.PipeWriter.Advance(System.Int32)" /> internally. If you call <see cref="M:System.IO.Pipelines.PipeWriter.Advance(System.Int32)" /> afterwards, the bytes in the buffer will all have a value of 0.</remarks>
```
halter73, line 431: PipeWriter.WriteAsync() also generally calls PipeWriter.GetSpan(System.Int32) and PipeWriter.FlushAsync(CancellationToken) internally. Why does PipeWriter.Advance(System.Int32) get a special call out? Custom PipeWriter are free not to call any of these methods.

> afterwards, the bytes in the buffer will all have a value of 0.

What buffer? The one returned from the next call to PipeWriter.GetSpan(System.Int32) or PipeWriter.GetMemory(System.Int32)? The bytes *might* be zero, but we don't make any guarantees about that. Quite the opposite in fact. You should always manually zero out memory (if that's what you want) rather than leave it uninitialized when using PipeWriter.GetSpan(System.Int32) or else risk leaking information.

If this is referring to the buffer returned from the *last* call to GetSpan/GetMemory, that's even worse. Reusing this buffer after writing/flushing/advancing is not allowed. Doing so is a use-after-free bug and could end up corrupting pooled memory that's already being used elsewhere or even segfaulting.

carlossanlop:
@merarischroeder I'll close this PR since there has not been any response since September. Feel free to reopen if you can address the questions and suggestions provided.

PR #4357, opened 2020-06-10
Fix max port number in TcpListener doc

Tornhoof:
The max port number 5000 is not correct on .NET Core, example code:
```csharp
        {
            var tcpListener = new TcpListener(IPAddress.Loopback, 0);
            tcpListener.Start();
            var port  = (tcpListener.LocalEndpoint as IPEndPoint).Port;
            Console.WriteLine(port);
            tcpListener.Stop();
            Console.ReadLine();
        }
```
Output is 62106 (or similar)

## Summary

The max port number 5000 is not correct on .NET Core, changed to 65535.


xml/System.Net.Sockets/TcpListener.xml
```diff
@@ -253,7 +253,7 @@
...
+ This constructor allows you to specify the local IP address and port number on which to listen for incoming connection attempts. Before calling this constructor you must first create an <xref:System.Net.IPAddress> using the desired local address. Pass this <xref:System.Net.IPAddress> to the constructor as the `localaddr` parameter. If you do not care which local address is assigned, specify <xref:System.Net.IPAddress.Any?displayProperty=nameWithType> for the `localaddr` parameter, and the underlying service provider will assign the most appropriate network address. This might help simplify your application if you have multiple network interfaces. If you do not care which local port is used, you can specify 0 for the port number. In this case, the service provider will assign an available port number between 1024 and 65535. If you use this approach, you can discover what local network address and port number has been assigned by using the <xref:System.Net.Sockets.TcpListener.LocalEndpoint%2A> property.
```
halter73, line 256: @wfurt So just strip " number between 1024 and 65535"? Seems reasonable.

```suggestion
 This constructor allows you to specify the local IP address and port number on which to listen for incoming connection attempts. Before calling this constructor you must first create an <xref:System.Net.IPAddress> using the desired local address. Pass this <xref:System.Net.IPAddress> to the constructor as the `localaddr` parameter. If you do not care which local address is assigned, specify <xref:System.Net.IPAddress.Any?displayProperty=nameWithType> for the `localaddr` parameter, and the underlying service provider will assign the most appropriate network address. This might help simplify your application if you have multiple network interfaces. If you do not care which local port is used, you can specify 0 for the port number. In this case, the service provider will assign an available port. If you use this approach, you can discover what local network address and port number has been assigned by using the <xref:System.Net.Sockets.TcpListener.LocalEndpoint%2A> property.  
```

wfurt: yes. but I'm fine will leaving the numbers in as well. Even if I don't really see how that would be interesting to anybody. (assuming people understand port is uint16)

Tornhoof: ``IPEndPoint.Port`` (the one from LocalEndPoint) is integer, so they might not now it's ushort.
One could argue that 1024/65535 is an implementation detail and simply refer to ``IPEndPoint.MinPort`` (const value of 0) and ``IPEndPoint.MaxPort`` (const value of 65535).
e.g.
the service provider will assign an available port number between <xref:System.Net.IPEndPoint.MinPort> and <xref:System.Net.IPEndPoint.MaxPort>.
The "1024" information is lost now, but the values refer to the proper min/max of the relevant type.

gewarren:
@Tornhoof Thanks, I'll merge this now since we seem to be okay with leaving the min/max numbers in there.

PR #4376, opened 2020-06-13
Replaced DateTime.Now with DateTime.UtcNow in Stopwatch.xml

devigo:
## Summary

Replaced `DateTime.Now` with `DateTime.UtcNow`, based on the code [here](https://referencesource.microsoft.com/#System/services/monitoring/system/diagnosticts/Stopwatch.cs,142)

In fact, the following sentence is not relevant starting with .NET Core 3.0 and higher, because the implementation is changed [here](https://github.com/dotnet/corefx/pull/36071/files?diff=split&w=1#diff-c3b10564cf9d302c537d7034001c7712R119-R123)
image

### DIFF
--- a/xml/System.Diagnostics/Stopwatch.xml
+++ b/xml/System.Diagnostics/Stopwatch.xml
@@ -436,7 +436,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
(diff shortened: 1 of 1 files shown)


gewarren:
@tommcdon @sywhang @krwq Could you review this PR and address the comment in the original description?

krwq:
I've just merged https://github.com/dotnet/dotnet-api-docs/pull/4215 which seem identical to this PR

PR #4382, opened 2020-06-15
Simple fix for thread priority sample snippet

kkokosa:
## Summary

Currently the snippet for `Thread.Priority` property does not work in Release mode. The flag `loopSwitch` is enregistered and not shared between threads. The simplest solution is to make it volatile.

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR_System/System.Threading.ThreadPriority/cs/Example1.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR_System/System.Threading.ThreadPriority/cs/Example1.cs
@@ -29,7 +29,7 @@ class Test
 
 class PriorityTest
 {
-    volatile static bool loopSwitch;
+    static volatile bool loopSwitch;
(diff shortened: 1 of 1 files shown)


samples/snippets/csharp/VS_Snippets_CLR_System/System.Threading.ThreadPriority/cs/Example1.cs
```diff
@@ -29,7 +29,7 @@ static void Main()
 
 class PriorityTest
 {
-    static bool loopSwitch;
+    volatile static bool loopSwitch;
```
kouvel, line 32:
```suggestion
    static volatile bool loopSwitch;
```
Think it's more common for `volatile` to be placed before the type, in the sense the type can be thought of as `volatile bool`

kouvel:
Nice catch, thanks for fixing. Just a small suggestion but LGTM.

gewarren:
@kkokosa Do you want to incorporate the suggestion so we can merge this?

PR #4393, opened 2020-06-16
Update DataContext.xml

aodennison:
## Summary

Typo "TextReader" should read "TextWriter"

### DIFF
--- a/xml/System.Data.Linq/DataContext.xml
+++ b/xml/System.Data.Linq/DataContext.xml
@@ -1171,7 +1171,7 @@ Next
       </ReturnValue>
       <Docs>
         <summary>Gets or sets the destination to write the SQL query or command.</summary>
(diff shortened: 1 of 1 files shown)


xml/System.Data.Linq/DataContext.xml
```diff
@@ -1171,7 +1171,7 @@ Next
       </ReturnValue>
       <Docs>
         <summary>Gets or sets the destination to write the SQL query or command.</summary>
-        <value>The <see cref="T:System.IO.TextReader" /> to use for writing the command.</value>
+        <value>The <see cref="T:System.IO.TextWriter" /> to use for writing the command.</value>
```
gewarren, line 1174:
```suggestion
        <value>The writer to use for writing the  command.</value>
```

Since the type is provided automatically, it should actually just be omitted here. Thanks for noticing this.

PR #4412, opened 2020-06-18
.NET Interactive: Enable_try_dotnet_to_batch_12c

v-mepa:
## Enable_try_dotnet_to_batch_12c

URL | Batch | Comment
-- | -- | --
https://docs.microsoft.com/en-us/dotnet/api/system.datetime.year | Batch   12 | Only   example 2
https://docs.microsoft.com/en-us/dotnet/api/system.timespan.tostring | Batch 12 | Remaining examples can use the new syntax
https://docs.microsoft.com/en-us/dotnet/api/system.tuple | Batch 12 |  
https://docs.microsoft.com/en-us/dotnet/api/system.array.reverse | Batch 12 |  
https://docs.microsoft.com/en-us/dotnet/api/system.exception.data | Batch 12 |


v-mepa:
fyi @mairaw

PR #4416, opened 2020-06-19
Improve the accuracy of the way ValueType's GetHashCode is described

### LINKED ISSUE #4365: Documentation error: Object.GetHashCode regarding value types
The documentation for the Object type:

https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Object.xml

...has a section that talks about `GetHashCode` and it makes the following claim:

> If value types do not override GetHashCode, the ValueType.GetHashCode method of the base class uses reflection to compute the hash code based on the values of the type's fields.

In the source code for `ValueType`, there are comments that suggest that this is factually incorrect:

        /*=================================GetHashCode==================================
        **Action: Our algorithm for returning the hashcode is a little bit complex.  We look
        **        for the first non-static field and get its hashcode.  If the type has no
        **        non-static fields, we return the hashcode of the type.  We can't take the
        **        hashcode of a static member because if that member is of the same type as
        **        the original type, we'll end up in an infinite loop.
        **Returns: The hashcode for the type.
        **Arguments: None.
        **Exceptions: None.
        ==============================================================================*/
        [MethodImpl(MethodImplOptions.InternalCall)]
        public extern override int GetHashCode();

I can understand not wanting to document this behaviour exactly, because of course it would be undesirable for it to become contractual, rather than merely an implementation detail, but as far as I can tell, no aspect of this "uses reflection to compute the hash code based on the values of the type's fields". Saying that it uses reflection raises alarm bells with me with regard to performance; upon reading this, I immediately began to worry that not overriding `GetHashCode` on every single `struct` could result in every `GetHashCode()` call dynamically discovering all of the type's properties.

Perhaps a better wording for this might be:

> If value types do not override GetHashCode, the runtime supplies an implementation that computes the hash code based on the values of the type's fields.

logiclrd:
## Summary

Updates the descriptions of `GetHashCode` w.r.t. `ValueType`.

The old description states that it always uses reflection and is always a performance concern. This is not actually true, though. There are conditions that cause reflection to be used:

- A `ValueType` that is not tightly-packed.
- Usage of pointers or floating point numbers (I think because floats have some extra semantics with `Equals`, where certain differing bit patterns might technically be the same number, and `GetHashCode` needs to return the same value for these?).
- Fields whose types are other `ValueType`s that are themselves ineligible.

In addition, the reflection-based path currently inspects only the first non-`null` field that it finds. So, the wording that asserts:

> If value types do not override `GetHashCode`, the `GetHashCode` of the base class uses reflection to compute the hash code based on the values of the type's fields.

...implies that _all_ value types have _all_ their fields incorporated into a hash code via reflection.

The changes here adjust this wording slightly, so that the basic description doesn't state that reflection is always in play, and the "Notes to Implementors" explains a _little_ bit more detail about the performance implications, without actually contractually explaining every detail of the present algorithm.

Fixes #4365


gewarren:
@carlossanlop Who should review this one?

SingleAccretion:
I think stronger wording about the fact that people __should__ override `GetHashCode` for custom value types would be warranted precisely because the default algorithm takes only the first non-null field into account.

carlossanlop:
Seems related to reflection. @steveharter @GrabYourPitchforks can you please review?

carlossanlop:
Ping again @steveharter @GrabYourPitchforks . If you don't think this belongs to your area, can you please point us to the right people?

BillWagner:
ping @carlossanlop to help move this along.

ericstj:
Dropping a couple source pointers here: https://github.com/dotnet/runtime/blob/a5ec8aa173e4bc76b173a70aa7fa3be1867011eb/src/coreclr/vm/comutilnative.cpp#L1981
https://github.com/dotnet/runtime/blob/7a6c8087a006bf6bd85ba43b2626e3b1e539db5a/src/coreclr/System.Private.CoreLib/src/System/ValueType.cs#L25

I think @jkotas and @bartonjs might be better folks to comment about how to tweak this documentation around best practices for ValueTypes.

xml/System/Object.xml
```diff
@@ -685,7 +685,7 @@ and the <xref:System.IDisposable> interface. The <xref:System.IDisposable.Dispos
   
  Otherwise, you might think that the mutable object is lost in the hash table. If you do choose to override <see cref="M:System.Object.GetHashCode" /> for a mutable reference type, your documentation should make it clear that users of your type should not modify object values while the object is stored in a hash table.  
   
- For value types, <see cref="M:System.ValueType.GetHashCode" /> provides a default hash code implementation that uses reflection. You should consider overriding it for better performance.
```
jkotas, line 688: I would keep the existing description.

The fast path in the default implementation is still fundamentally using reflection and it is still significantly slower than overridden implementation. I would not complicate the explanation here with runtime implementation details. Also, these details vary between runtime versions.

bartonjs: I agree with Jan.  Following the chain of guidelines:

> DO implement IEquatable<T> on value types.

> DO override GetHashCode whenever you override Equals.

Our recommendations are that essentially every value type should explicitly override GetHashCode.  For some transient types (like one discussed in API review today) where they're just a custom type representing an N-tuple response from a callback, it might be reasonable to decide that value comparison is irrelevant and skip those... but then the performance implications of reflection don't matter.

gewarren:
Closing this PR based on the feedback.

xml/System/Object.xml
logiclrd: > The fast path in the default implementation is still fundamentally using reflection and it is still significantly slower than overridden implementation.

Hmm, but it's not using reflection on every call to `GetHashCode`, just once as set-up that is then cached. It makes every effort to be as performant as possible:

* Check if it is a `null` reference.
* Wrap object reference in `OBJECTREF` struct on the stack. This struct is just one pointer in size and is a debugging aid. Probably doesn't actually result in actual code.
* Get a pointer to the object's method table. Pretty sure this is O(1).
* Get the current AppDomain. This is O(1).
* Look up the TypeID. This is O(lg(N)) where N is the number of loaded types.
* Create an initial hash code by multiplying the TypeID by a big prime and applying an offset.
* Read the cached `CanCompareBitsOrUseFastGetHashCode` value.
* If `canUseFastGetHashCodeHelper` is true, then `FastGetValueTypeHashCodeHelper` reads the structure's data as though it were an unmanaged `int` array.

The majority of these are reading a single pointer from memory.

The first time you call it, it needs to compute the `CanCompareBitsOrUseFastGetHsahCode` value to cache, but the common path does _not_ use reflection to enumerate the members of the type. The total complexity is a map lookup where the map size is the number of loaded types and walking through the structure's raw data.

It is disingenuous to tell the user that `GetHashCode` calls _will by definition_ always use reflection. The argument seems to be that the user should be told this so that they'll feel more of an incentive to follow guidelines and provide their own `GetHashCode` implementation, but this is literally lying about what the implementation is doing for this purpose.

I just made a brute-force benchmark of it with a `struct` with 256 `int` fields, and I found:

* The default `GetHashCode` implementation takes about 2.7 times longer than a one-liner that XORs all the fields in a hardcoded manner.
* Using reflection to enumerate the fields and obtain each field's value takes about 123 times longer than the default `GetHashCode` implementation, or 338 times longer than the single expression.

image

PR #4418, opened 2020-06-20
Changes "beyond" to "within".

pekspro:
## Summary

Changes the line:

Seeking to any location **beyond** the length of the stream is supported.

to:

Seeking to any location **within** the length of the stream is supported.

Which I think makes more sense, but I am not sure :-)

### DIFF
--- a/xml/System.IO/Stream.xml
+++ b/xml/System.IO/Stream.xml
@@ -1858,7 +1858,7 @@
 ## Remarks  
  The stream must support seeking to get or set the position. Use the <xref:System.IO.Stream.CanSeek%2A> property to determine whether the stream supports seeking.  
   
- Seeking to any location beyond the length of the stream is supported.  
+ Seeking to any location within the length of the stream is supported.  
   
  The `Position` property does not keep track of the number of bytes from the stream that have been consumed, skipped, or both.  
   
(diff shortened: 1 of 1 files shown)


carlossanlop:
Thanks, @pekspro. The text is not wrong.

If you read the text for [FileStream.Position](https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream.position?view=netcore-3.1#remarks), then this generalized description for `Stream.Position` makes a little more sense:

> _Seeking to any location beyond the length of the stream is supported. When you seek beyond the length of the file, the file size grows. In Microsoft Windows NT and newer, any data added to the end of the file is set to zero. In Microsoft Windows 98 or earlier, any data added to the end of the file is not set to zero, which means that previously deleted data is visible to the stream. Setting the position of the stream to a large value beyond the end of the stream in Windows 98 or earlier may result in an exception being raised._

Since `Stream` is the base class, its descriptions tend to be a bit more broad, while classes that inherit from `Stream` usually expand or describe in more detail the API descriptions.

If you think you can expand the `Stream` descriptions a bit more to ensure they are not too ambiguous, I'll be happy to review your suggestions, @pekspro. Otherwise, feel free to close the PR.

PR #4424, opened 2020-06-22
SYSTEM_INFO type in the FieldOffset example should be a struct

Perksey:
## Summary
Alters the FieldOffsetAttribute example to demonstrate its usage on a struct instead of a class.

### DIFF
--- a/xml/System.Runtime.InteropServices/FieldOffsetAttribute.xml
+++ b/xml/System.Runtime.InteropServices/FieldOffsetAttribute.xml
@@ -57,7 +57,7 @@
    
   
 ## Examples  
- The following example demonstrates how to apply the <xref:System.Runtime.InteropServices.FieldOffsetAttribute> to members of a class with an explicit layout.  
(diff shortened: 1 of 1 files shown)


gewarren:
@jkoritzinsky I think that the use of "class" may be intentional. Can you weigh in? If we approve this PR, we'll also need to update the intro text at https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.fieldoffsetattribute?view=netcore-3.1#examples.

jkoritzinsky:
The text should say "struct or class" and the example can be either.

Perksey:
Wow, TIL. It's probably better to use struct as in most cases that's where StructLayout et al will be used, and I'm willing to update the text to reflect this if it's desirable?

jkoritzinsky:
I would be OK updating the example to be a struct as long as the text is also updated.

gewarren:
This example is also using it on a class.

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.structlayoutattribute?view=netcore-3.1#examples

PR #4430, opened 2020-06-23
mdoc updates 5.7.6

gewarren:


xml/System.ServiceModel.Channels/MessageVersion.xml
```diff
@@ -26,8 +26,7 @@
   </AssemblyInfo>
   <TypeForwardingChain>
     <TypeForwarding From="System.ServiceModel" FromVersion="4.0.0.0" To="System.ServiceModel.Primitives" ToVersion="4.6.0.0" FrameworkAlternate="dotnet-plat-ext-2.1;dotnet-plat-ext-2.2;dotnet-plat-ext-3.0" />
-    <TypeForwarding From="System.ServiceModel" FromVersion="4.0.0.0" To="System.ServiceModel.Primitives" ToVersion="4.7.0.0" FrameworkAlternate="dotnet-plat-ext-3.1" />
-    <TypeForwarding From="System.ServiceModel" FromVersion="4.0.0.0" To="System.ServiceModel.Primitives" ToVersion="5.0.0.0" FrameworkAlternate="dotnet-plat-ext-5.0" />
```
gewarren, line 30: Does this look expected?

xml/System.Windows.Shell/WindowChrome.xml
```diff
@@ -480,6 +480,23 @@
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
+    <Member MemberName="IsHitTestVisibleInChrome">
```
gewarren, line 483: This file has a couple new members, which seems odd.

xml/System/IntPtr.xml
```diff
@@ -51,7 +51,19 @@
     <BaseTypeName>System.ValueType</BaseTypeName>
   </Base>
   <Interfaces>
-    <Interface>
+    <Interface FrameworkAlternate="net-5.0">
+      <InterfaceName>System.IComparable</InterfaceName>
```
gewarren, line 55: New interfaces are showing up here.

xml/System.Windows.Shell/WindowChrome.xml
gewarren: I've sent email to Joel/Mimi about these attached properties. They were previously documented on MSDN only.

xml/System/IntPtr.xml
gewarren: I'll assume this is just mdoc being able to differentiate which interfaces are implemented between different versions. The published doc is here - https://docs.microsoft.com/en-us/dotnet/api/system.intptr?view=net-5.0.

gewarren:
The warnings for unresolved types will be fixed by [Multiple Return Type causes build warnings](https://ceapex.visualstudio.com/Engineering/_workitems/edit/242151) and [Handle return types that are private](https://dev.azure.com/ceapex/Engineering/_workitems/edit/242401). I think this is good to merge. @BillWagner do you approve?

PR #4440, opened 2020-06-25
Log Levels - Sort log levels by their integer value

dthompsonza:
The current implementation sorts log levels by their name (Critical, Debug, Error, Info, None, Trace and Warning) when its more useful to the reader to view the log levels in order of their values, ie Trace, Debug, Info, Warning, Error, Critical, None.

Thank you for considering this pull request.


gewarren:
@dthompsonza Howzit! I agree the order you propose is preferable. Unfortunately, the next time we import new APIs, these changes will be overwritten.

dthompsonza:
Darn. Okay thank you

PR #4483, opened 2020-07-08
Fix remarks for Insert overload which takes CacheDependency

mikemccaughan:
## Summary

The Remarks for the overload of `Insert` which takes a `CacheDependency` appears to have been copied and pasted from the overload which does not take that parameter and therefore stated that the object added to the cache would be inserted with "no file or cache dependencies", which is the purpose of the overload. This edit removes that text.

### DIFF
--- a/xml/System.Web.Caching/Cache.xml
+++ b/xml/System.Web.Caching/Cache.xml
@@ -412,7 +412,7 @@
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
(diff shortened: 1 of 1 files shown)


gewarren:
Ping @scottaddie

scottaddie:
Thank you for fixing this, @mikemccaughan

mikemccaughan:
Glad I could be of help.

PR #4484, opened 2020-07-08
Fix wording in the Int32 documentation

ForNeVeR:
Hello! I've noticed weird wording in some of `Int32` method descriptions (namely, "a specified object or `Int32`" for the methods that only take `Int32`) and decided to fix it. I've taken the actual method descriptions from the XML doc, they look more up-to-date.


gewarren:
@ForNeVeR Thank you!

PR #4485, opened 2020-07-08
Improving midpoint rounding documentation

SingleAccretion:
## Summary

These are the minimum improvements I mentioned in the issue below + some interpolated strings in the code snippet. I would personally be fine with this version of the doc, but there could still be some additions:
- Add directed modes to the snippet. I decided to leave this one out for now because the snippet is already kind of big, and it would duplicate information already in the "Directed rounding" table.
- It is not immediately clear that "precision" in the doc's context means "number of decimal places", but I guess there is a reason it's written this way. For reference, the parameter name on `Math.Round` correctly mentions "fractional digits". This is confusing because, to me at least, "precision" == "number of significant (not fractional) digits".
- The explanation for the rounding to nearest modes is quite long, but I did not want to shorten without some feedback if this is even desirable (the "for" argument could be that it makes the reader "drown in letters", the "against" is that clarity > brevity).

Should I add anything from the above?

Fixes dotnet/docs#19314
cc @gewarren


xml/System/MidpointRounding.xml
```diff
@@ -52,15 +52,15 @@
     <summary>Specifies how mathematical rounding methods should process a number that is midway between two numbers.</summary>
     <remarks>
       <format type="text/markdown"><![CDATA[  
+
```
gewarren, line 55:
```suggestion
## Remarks
```

xml/System/MidpointRounding.xml
```diff
@@ -52,15 +52,15 @@
...
+Use `MidpointRounding` with appropriate overloads of <xref:System.Math.Round%2A?displayProperty=nameWithType> to provide more control of the rounding process.  
+
+## Rounding to nearest
```
gewarren, line 58:
```suggestion
### Round to nearest  
```

xml/System/MidpointRounding.xml
```diff
@@ -52,15 +52,15 @@
     <summary>Specifies how mathematical rounding methods should process a number that is midway between two numbers.</summary>
     <remarks>
       <format type="text/markdown"><![CDATA[  
+
+Use `MidpointRounding` with appropriate overloads of <xref:System.Math.Round%2A?displayProperty=nameWithType> to provide more control of the rounding process.  
+
+## Rounding to nearest  
+ A round to nearest operation takes an original number with an implicit or specified precision; examines the next digit, which is at that precision plus one; and returns the nearest number with the same precision as the original number. For positive numbers, if the next digit is from 0 through 4, the nearest number is toward negative infinity. If the next digit is from 6 through 9, the nearest number is toward positive infinity. For negative numbers, if the next digit is from 0 through 4, the nearest number is toward positive infinity. If the next digit is from 6 through 9, the nearest number is toward negative infinity.
```
gewarren, line 59:
```suggestion
 A round-to-nearest operation takes an original number with an implicit or specified precision; examines the next digit, which is at that precision plus one; and returns the nearest number with the same precision as the original number. For positive numbers, if the next digit is from 0 through 4, the nearest number is toward negative infinity. If the next digit is from 6 through 9, the nearest number is toward positive infinity. For negative numbers, if the next digit is from 0 through 4, the nearest number is toward positive infinity. If the next digit is from 6 through 9, the nearest number is toward negative infinity.  
```

xml/System/MidpointRounding.xml
```diff
@@ -52,15 +52,15 @@
...
   
-## Remarks  
- Use `MidpointRounding` with appropriate overloads of <xref:System.Math.Round%2A?displayProperty=nameWithType> to provide more control of the rounding process.  
+ In the previous cases, the `MidpointRounding.AwayFromZero` and `MidpointRounding.ToEven` do not affect the result of the rounding operation. However, if the next digit is 5, which is the midpoint between two possible results, and all remaining digits are zero or there are no remaining digits, the nearest number is ambiguous. In this case, the round to nearest modes in `MidpointRounding` enable you to specify whether the rounding operation returns the nearest number away from zero or the nearest even number.
```
gewarren, line 61:
```suggestion
 In the previous cases, the `MidpointRounding.AwayFromZero` and `MidpointRounding.ToEven` do not affect the result of the rounding operation. However, if the next digit is 5, which is the midpoint between two possible results, and all remaining digits are zero or there are no remaining digits, the nearest number is ambiguous. In this case, the round-to-nearest modes in `MidpointRounding` enable you to specify whether the rounding operation returns the nearest number away from zero or the nearest even number.  
```

xml/System/MidpointRounding.xml
```diff
@@ -52,15 +52,15 @@
...
+ The following table demonstrates the results of rounding some negative and positive numbers in conjunction with round to nearest modes. The precision used to round the numbers is zero, which means the number after the decimal point affects the rounding operation. For example, for the number -2.5, the digit after the decimal point is 5. Because that digit is the midpoint, you can use a `MidpointRounding` value to determine the result of rounding. If `AwayFromZero` is specified, -3 is returned because it is the nearest number away from zero with a precision of zero. If `ToEven` is specified, -2 is returned because it is the nearest even number with a precision of zero.
```
gewarren, line 63:
```suggestion
 The following table demonstrates the results of rounding some negative and positive numbers in conjunction with round-to-nearest modes. The precision used to round the numbers is zero, which means the number after the decimal point affects the rounding operation. For example, for the number -2.5, the digit after the decimal point is 5. Because that digit is the midpoint, you can use a `MidpointRounding` value to determine the result of rounding. If `AwayFromZero` is specified, -3 is returned because it is the nearest number away from zero with a precision of zero. If `ToEven` is specified, -2 is returned because it is the nearest even number with a precision of zero.  
```

gewarren:
@SingleAccretion Thanks for adding this detail. I'll merge it now!

PR #4491, opened 2020-07-09
Fix PipeOptions ctor pauseWriterThreshold

carlossanlop:
Should be zero instead of negative.

Fixes https://github.com/dotnet/runtime/issues/476


carlossanlop:
ping @halter73

halter73:
Sorry for the delay.

PR #4493, opened 2020-07-09
Clarify thread-safety of Socket.Connected

tmds:
Based on reported issue https://github.com/dotnet/runtime/issues/34872, this clarifies the thread-safety of `Socket.Connected`.

@wfurt @stephentoub does it make sense to include this info in the docs?

cc @drieseng

### DIFF
--- a/xml/System.Net.Sockets/Socket.xml
+++ b/xml/System.Net.Sockets/Socket.xml
@@ -3802,7 +3802,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The `Connected` property gets the connection state of the <xref:System.Net.Sockets.Socket> as of the last I/O operation. When it returns `false`, the <xref:System.Net.Sockets.Socket> was either never connected, or is no longer connected. `Connected` is not thread-safe, it may return `true` after an operation is aborted due to an operation that disconnects the <xref:System.Net.Sockets.Socket> on another thread.
+ The `Connected` property gets the connection state of the <xref:System.Net.Sockets.Socket> as of the last I/O operation. When it returns `false`, the <xref:System.Net.Sockets.Socket> was either never connected, or is no longer connected. `Connected` is not thread-safe; it may return `true` after an operation is aborted when the <xref:System.Net.Sockets.Socket> is disconnected from another thread.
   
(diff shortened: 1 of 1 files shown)


xml/System.Net.Sockets/Socket.xml
```diff
@@ -3802,7 +3802,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The `Connected` property gets the connection state of the <xref:System.Net.Sockets.Socket> as of the last I/O operation. When it returns `false`, the <xref:System.Net.Sockets.Socket> was either never connected, or is no longer connected.  
+ The `Connected` property gets the connection state of the <xref:System.Net.Sockets.Socket> as of the last I/O operation. When it returns `false`, the <xref:System.Net.Sockets.Socket> was either never connected, or is no longer connected. `Connected` is not thread-safe, it may return `true` after an operation is aborted due to an operation that disconnects the <xref:System.Net.Sockets.Socket> on another thread.
```
wfurt, line 3805: should we mention this is transient e.g. short time race condition?

tmds: I don't want to imply that people could poll this value. I'm not sure that is guaranteed to always work.

wfurt: ok, I assumed it would converge.

gewarren, line 3805:
```suggestion
 The `Connected` property gets the connection state of the <xref:System.Net.Sockets.Socket> as of the last I/O operation. When it returns `false`, the <xref:System.Net.Sockets.Socket> was either never connected or is no longer connected. `Connected` is not thread-safe; it may return `true` if the <xref:System.Net.Sockets.Socket> is disconnected on another thread.
```

Is the "operation is aborted" piece important? It seemed a bit redundant to me.

tmds: @gewarren 

I slightly updated the wording. I kept the _after an operation is aborted_ piece.
The issue that got reported had code like this:
```
socket.SomeOperation(); // returns SocketError.OperationAborted
if (socket.Connected) { ... }
```
The user expectation was that when SomeOperation returns OperationAborted because the Socket gets disconnected from another thread, that the _if (socket.Connected)_ would be false. This is not guaranteed, and the comment is for clarifying that.

PR #4507, opened 2020-07-12
Update ArrayPool`1.xml

jkotas:
## Summary

Note that arrays returned by ArrayPool.Rent are not zero-initialized

Fixes https://github.com/dotnet/coreclr/pull/24504#issuecomment-657141532


xml/System.Buffers/ArrayPool`1.xml
```diff
@@ -199,7 +199,9 @@ The instance of the <xref:System.Buffers.ArrayPool%601> class created by this me
 ## Remarks  
 
 This buffer is loaned to the caller and should be returned to the same pool using the <xref:System.Buffers.ArrayPool%601.Return%2A> method, so that it can be reused in subsequent calls to the <xref:System.Buffers.ArrayPool%601.Rent%2A> method. Failure to return a rented buffer is not a fatal error. However, it may lead to decreased application performance, as the pool may need to create a new buffer to replace the lost one.
- ]]></format>
+
+The array returned by this method is not zero-initialized.
```
stephentoub, line 203:
```suggestion
The array returned by this method may not be zero-initialized. 
```
Just suggesting that because the default pool doesn't for value types, not references still will be, and other pool implementations still do.

jkotas: Default pool does not guarantee zero-initialization for references either.

stephentoub: I meant it may still zero for references, e.g. when it needs to allocate a new array.

PR #4538, opened 2020-07-17
Add notes to not-recommended algorithm classes

tdykstra:
Fixes https://github.com/dotnet/docs/issues/10202

Most of the affected classes already had the needed warnings. This PR just propagates those notes to derived and related classes and in some cases moves warning text into a note box.

### DIFF
--- a/xml/System.Security.Cryptography/DSACng.xml
+++ b/xml/System.Security.Cryptography/DSACng.xml
@@ -93,7 +93,7 @@
 Valid key sizes range from 512 to 3,072 bits, in increments of 64. We recommend that a minimum size of 2,048 bits be used for all keys.  
   
 > [!IMPORTANT]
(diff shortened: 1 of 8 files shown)


xml/System.Security.Cryptography/DSACng.xml
```diff
@@ -90,8 +90,11 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- Valid key sizes range from 512 to 3,072 bits, in increments of 64. We recommend that a minimum size of 2,048 bits be used for all keys.  
+Valid key sizes range from 512 to 3,072 bits, in increments of 64. We recommend that a minimum size of 2,048 bits be used for all keys.  
   
+> [!IMPORTANT]
+> Newer asymmetric algorithms are available. Consider using the <xref:System.Security.Cryptography.RSA> class or the <xref:System.Security.Cryptography.ECDsa> class instead of the <xref:System.Security.Cryptography.DSA> class. Use <xref:System.Security.Cryptography.DSA> only for compatibility with legacy applications and data.
```
bartonjs, line 96: RSA is from 1977 and DSA is from 1991.  RSA just had better staying power.

So perhaps "Better" instead of "Newer"?  Or "The creators of the DSA algorithm have withdrawn their support for it."?

bartonjs:
Just the one semantic issue, but its resolution needs to be copied to all of the DSA types.

PR #4539, opened 2020-07-17
Fix up broken links (System.Runtime through System.Timers)

gewarren:
Contributes to dotnet/docs#19387.


gewarren:
The remaining warning in these namespaces is beyond my control.

PR #4540, opened 2020-07-18
Fix xref links - System.Web through System.Xml

gewarren:
Contributes to dotnet/docs#19387.


gewarren:
I checked the build report for these namespaces and the warnings are no longer present.

gewarren:
Remaining warnings for these namespaces are beyond my control.

PR #4542, opened 2020-07-18
Replace MD5 and SHA1

tdykstra:
Fixes https://github.com/dotnet/docs/issues/10198

### DIFF
--- a/samples/snippets/cpp/VS_Snippets_CLR/AssemblyName_CodeBase/CPP/assemblyname_codebase.cpp
+++ b/samples/snippets/cpp/VS_Snippets_CLR/AssemblyName_CodeBase/CPP/assemblyname_codebase.cpp
@@ -71,10 +71,8 @@ int main()
    // Set the culture information of the assembly to 'English-American'.
    myAssemblyName->CultureInfo = gcnew CultureInfo( "en-US" );
 
-   // Set the hash algorithm to 'SHA1'.
-   // This example uses the SHA1 algorithm.
-   // Due to collision problems with SHA1, Microsoft recommends SHA256 or better.
-   myAssemblyName->HashAlgorithm = AssemblyHashAlgorithm::SHA1;
+   // Set the hash algorithm to 'SHA256'.
+   myAssemblyName->HashAlgorithm = AssemblyHashAlgorithm::SHA256;
    myAssemblyName->Name = "MyAssembly";
    myAssemblyName->Version = gcnew Version( "1.0.0.2001" );
    MakeAssembly( myAssemblyName, "MyAssembly.exe" );
--- a/samples/snippets/csharp/VS_Snippets_CLR/AssemblyName_CodeBase/CS/assemblyname_codebase.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR/AssemblyName_CodeBase/CS/assemblyname_codebase.cs
@@ -67,10 +67,8 @@ public class AssemblyName_CodeBase
       myAssemblyName.CodeBase = Directory.GetCurrentDirectory();
       // Set the culture information of the assembly to 'English-American'.
       myAssemblyName.CultureInfo = new CultureInfo("en-US");
-      // Set the hash algorithm to 'SHA1'.
-      // This example uses the SHA1 algorithm.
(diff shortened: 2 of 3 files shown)


samples/snippets/cpp/VS_Snippets_CLR/AssemblyName_CodeBase/CPP/assemblyname_codebase.cpp
```diff
@@ -71,8 +71,8 @@ int main()
    // Set the culture information of the assembly to 'English-American'.
    myAssemblyName->CultureInfo = gcnew CultureInfo( "en-US" );
 
-   // Set the hash algoritm to 'SHA1'.
-   myAssemblyName->HashAlgorithm = AssemblyHashAlgorithm::SHA1;
+   // Set the hash algorithm to 'SHA256'.
+   myAssemblyName->HashAlgorithm = AssemblyHashAlgorithm::SHA256;
```
bartonjs, line 75: I don't know that SHA256 actually works here.

samples/snippets/csharp/VS_Snippets_CLR/AssemblyName_CodeBase/CS/assemblyname_codebase.cs
```diff
@@ -67,8 +67,8 @@ public static void Main()
       myAssemblyName.CodeBase = Directory.GetCurrentDirectory();
       // Set the culture information of the assembly to 'English-American'.
       myAssemblyName.CultureInfo = new CultureInfo("en-US");
-      // Set the hash algoritm to 'SHA1'.
-      myAssemblyName.HashAlgorithm = AssemblyHashAlgorithm.SHA1;
+      // Set the hash algorithm to 'SHA256'.
+      myAssemblyName.HashAlgorithm = AssemblyHashAlgorithm.SHA256;
```
bartonjs, line 71: Another triplet of AssemblyName.HashAlgorithm being updated.  It might work, I just know that some parts of this system are effectively limited to MD5 and SHA1... and I don't know if this is one of them or not.  (We could, of course, update them, and if we get feedback it doesn't work, then change them back)

samples/snippets/csharp/VS_Snippets_Remoting/Classic HttpWebRequest.ConnectionGroupName Example/CS/source.cs
```diff
@@ -12,8 +12,8 @@ public static int Main(String[] args)
  {
 // <Snippet1>
   // Create a secure group name.
-  SHA1Managed Sha1 = new SHA1Managed();
-  Byte[] updHash = Sha1.ComputeHash(Encoding.UTF8.GetBytes("username" + "password" +  "domain"));
+  SHA256 Sha256 = SHA256.Create();
+  Byte[] updHash = Sha256.ComputeHash(Encoding.UTF8.GetBytes("username" + "password" +  "domain"));
   String secureGroupName = Encoding.Default.GetString(updHash);
```
bartonjs, line 17: Not really the problem at hand, but: I don't know that this is going to do anything useful.  The hash may contain a zero-byte, which will make the string effectively terminate early.

It's also weird that the "secure group name" is a hash of static data.  Might want to get networking to review this sample to see what it's supposed to be doing.

tdykstra: This is for HttpWebRequest, which has an **Important** alert box warning that "We don't recommend that you use HttpWebRequest for new development. Instead, use the System.Net.Http.HttpClient class." The example is for [ConnectionGroupName](https://docs.microsoft.com/en-us/dotnet/api/system.net.httpwebrequest.connectiongroupname?view=netcore-3.1) property. It may not be worth someone's time to investigate and I should just revert the changes on these files(?)

bartonjs: Yeah, reverting the edit and changing it to the "don't use SHA1" comment may be the short-term best.

PR #4554, opened 2020-07-22
Fix HttpClient Get/Put/Delete/Post exception documentation

glihm:
## Summary

Current implementation of `HttpClient` Get/Put/Delete/Post async
functions are never throwing `ArgumentNullException`.

Instead, they throw `InvalidOperationException`.

However, for the `DeleteAsync` function, `InvalidOperationException`
is thrown in two cases:

1. when the request was already sent.
2. when the `BaseAddress` and `requestUri` can't be combined into a valid
   absolute URI.

For this reason, the current commit is combining two reasons associated
with the `InvalidOperationException`.

@gewarren : I am sorry I don't know if an exception can combined two reasons... but it's important to me 
to have an accurate documentation, and please feel to comment if the sentence of `DeleteAsync` exception's text is not
correct.

I opened an [issue](https://github.com/dotnet/runtime/issues/39758) into the runtime repo to address the subject.

## Example code

```C#
    // This code throws InvalidOperationException, not ArgumentNullException.
    HttpClient hc = new HttpClient();
    hc.DeleteAsync((string)null);
```

### DIFF
--- a/xml/System.Net.Http/HttpClient.xml
+++ b/xml/System.Net.Http/HttpClient.xml
@@ -637,7 +637,11 @@ The <paramref name="requestUri" /> is not an absolute URI.
 
 -or-
 
-The <paramref name="requestUri" /> must be an absolute URI or <see cref="P:System.Net.Http.HttpClient.BaseAddress" /> must be set.</exception>
+The <paramref name="requestUri" /> is not an absolute URI.
+
+-or-
+
+<see cref="P:System.Net.Http.HttpClient.BaseAddress" /> is not set.</exception>
         <exception cref="T:System.Net.Http.HttpRequestException">The request failed due to an underlying issue such as network connectivity, DNS failure, server certificate validation or timeout.</exception>
         <exception cref="T:System.Threading.Tasks.TaskCanceledException">.NET Core and .NET 5.0 and later only: The request failed due to timeout.</exception>
       </Docs>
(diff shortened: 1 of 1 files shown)


glihm:
@gewarren I have found an example with "OR" for exceptions:

img

I will edit the pull request if possible, or closing this one and open a new one to add the correct style.

xml/System.Net.Http/HttpClient.xml
```diff
@@ -574,8 +574,11 @@ The `DefaultRequestVersion` property can be changed as long as the <xref:System.
 > - <xref:System.Threading.Tasks.TaskCanceledException> that nests a <xref:System.TimeoutException> is thrown on .NET 5.0 and later versions.
  ]]></format>
         </remarks>
-        <exception cref="T:System.ArgumentNullException">The <paramref name="requestUri" /> is <see langword="null" />.</exception>
-        <exception cref="T:System.InvalidOperationException">The request message was already sent by the <see cref="T:System.Net.Http.HttpClient" /> instance.</exception>
+        <exception cref="T:System.InvalidOperationException">The request message was already sent by the <see cref="T:System.Net.Http.HttpClient" /> instance.
+
+-or-
+
+The <paramref name="requestUri" /> must be an absolute URI or <see cref="P:System.Net.Http.HttpClient.BaseAddress" /> must be set.</exception>
```
gewarren, line 581:
```suggestion
The <paramref name="requestUri" /> is not an absolute URI.

-or-

<see cref="P:System.Net.Http.HttpClient.BaseAddress" /> is not set.</exception>
```

aik-jahoda:
I don't recommend this change until the https://github.com/dotnet/runtime/issues/40519 is triaged. It seems like we have an regression in HttpClient.

glihm:
> I don't recommend this change until the https://github.com/dotnet/runtime/issues/40519 is triaged. It seems like we have an regression in HttpClient.

Thanks for the Feedback. :)

aik-jahoda:
It seems the wrong exception is not a regression, the same behaviour is on 2.1

xml/System.Net.Http/HttpClient.xml
glihm: I don't think only `is not an absolute URI` is a valid text, because `requestUri` can be a relative URI if a `BaseAddress` is set.

glihm:
> It seems the wrong exception is not a regression, the same behaviour is on 2.1


So, what can we do to enhance the documentation / runtime clarity about this?

The point is that, in the way `HttpClient` works, a `BaseAddress` can be defined, and this makes sense to *not* throw `ArgumentNullException` to me. Indeed, a request can be made to the `BaseAddress` only passing null to `DeleteAsync` for instance.

That is the reason why I have written the PR, to ensure at least that the behavior of the documentation is matching the current runtime behavior. 👍

glihm:
@gewarren thank you for the rewording and formatting. :)

aik-jahoda:
@glihm agree with the change in this PR. LGTM

aik-jahoda (approved):

PR #4558, opened 2020-07-22
Reorder DayOfWeek by enumeration value instead of alphabetized order …

### LINKED ISSUE #4210: List DayOfWeek enum by value instead of alphabetized by name
Fields should not be alphabetized and instead follow the order:
* Sunday - 0
* Monday - 1
* Tuesday - 2
* Wednesday - 3
* Thursday - 4
* Friday - 5
* Saturday - 6

https://github.com/dotnet/dotnet-api-docs/blob/b4ec0e45b0e446a1f632ff4c9eca1753d3c377b0/xml/System/DayOfWeek.xml#L68

glihm:
Fixes #4210 

## Summary

As mentioned in the issue, the enumeration value order looks far better.

### DIFF
--- a/xml/System/DayOfWeek.xml
+++ b/xml/System/DayOfWeek.xml
@@ -67,123 +67,7 @@
     </remarks>
   </Docs>
   <Members>
-    <Member MemberName="Friday">
-      <MemberSignature Language="C#" Value="Friday" />
-      <MemberSignature Language="ILAsm" Value=".field public static literal valuetype System.DayOfWeek Friday = int32(5)" />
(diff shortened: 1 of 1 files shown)


gewarren:
@glihm I appreciate your effort here, and I agree it looks better, but it the order is auto-generated and will just be overwritten the next time we import the API updates, unfortunately.

glihm:
> @glihm I appreciate your effort here, and I agree it looks better, but it the order is auto-generated and will just be overwritten the next time we import the API updates, unfortunately.

@gewarren np, thanks for the detailed answer. :)

gewarren:
This has been fixed!

PR #4574, opened 2020-07-25
Update xor2.cs

JaredKreft:
Cleaner version of Equals. Avoids hiding the execution break in an if(){} and also avoids the unnecessary ! on (obj is Point).

## Summary

Describe your changes here.


samples/snippets/csharp/VS_Snippets_CLR_System/system.object.gethashcode/cs/xor2.cs
```diff
@@ -14,10 +14,15 @@ public Point(int x, int y)
 
     public override bool Equals(Object obj)
     {
-       if (!(obj is Point)) return false;
-
-       Point p = (Point) obj;
-       return x == p.x & y == p.y;
+        if(obj is Point)
```
gewarren, line 17:
```suggestion
        if (obj is Point)

```

PR #4576, opened 2020-07-26
Add note about ValueTask and ValueTask<TResult> default constructors (#4397)

### LINKED ISSUE #4397: Document ValueTask default constructor
https://github.com/dotnet/dotnet-api-docs/blob/892ea1386d69a7a1f1b33ac3011b63df0025f423/xml/System.Threading.Tasks/ValueTask.xml#L48

The behavior of the ValueTask default constructor is not documented, apparently it is equivalent to `new ValueTask(Task.CompletedTask)`

glihm:
Fixes #4397 

## Summary

As described in the issue, `ValueTask` and `ValueTask<TResult>` are lacking of information about default constructors behavior.
This PR is adding a note reworded from the [code itself](https://github.com/stephentoub/coreclr/blob/f9a312befe71a6c68c908ed09234a3ebc2d54b6c/src/mscorlib/shared/System/Threading/Tasks/ValueTask.cs).

### DIFF
--- a/xml/System.Threading.Tasks/ValueTask.xml
+++ b/xml/System.Threading.Tasks/ValueTask.xml
@@ -66,7 +66,7 @@ Using a `ValueTask` instead of a <xref:System.Threading.Tasks.Task> introduces s
 > The use of the `ValueTask` type is supported starting with C# 7.0 and is not supported by any version of Visual Basic.
 
 > [!NOTE]
-> An instance created with the default constructor (a zero initialized structure) represents a synchronously, successfully completed operation.
+> An instance created with the parameterless constructor or by the `default(ValueTask)` syntax (a zero-initialized structure) represents a synchronously, successfully completed operation.
 
  ]]></format>
     </remarks>
--- a/xml/System.Threading.Tasks/ValueTask`1.xml
+++ b/xml/System.Threading.Tasks/ValueTask`1.xml
@@ -79,7 +79,7 @@ As such, the default choice for any asynchronous method should be to return a <x
 >  The use of the <xref:System.Threading.Tasks.ValueTask%601> type is supported starting with C# 7.0, and is not supported by any version of Visual Basic.
 
 > [!NOTE]
-> An instance created with the default constructor (a zero initialized structure) represents a synchronously, successfully completed operation with a result of `default(TResult)`.
+> An instance created with the parameterless constructor or by the `default(ValueTask<TResult>)` syntax (a zero-initialized structure) represents a synchronously, successfully completed operation with a result of `default(TResult)`.
 
  ]]></format>
     </remarks>


tarekgh:
I am not sure it is a good idea to document the default constructor that is automatically generated by the compiler. Usually users shouldn't use the default constructor. and If using the `default` value, should be used to indicate if the instance is initialized or not. This is a generic issue for the whole framework structs. 

I'll let @stephentoub comment on that but his response may be delayed a little.

stephentoub:
Documenting what the status of a zero-initialized `ValueTask`/`ValueTask<T>` is fine, though it's not really using a "default constructor" when it does so, so I'd be careful with such language.

glihm:
@tarekgh @stephentoub thanks for the feedback, I'll try a rewording to avoid such confusion. 👍🏻

glihm:
In `struct` [documentation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct), those lines are written:

```
You can't declare a parameterless constructor.
Every structure type already provides an implicit parameterless constructor that produces the default value of the type.
```

changing `default` to `parameterless` constructor can be less confusing and in the same spirit of struct semantic, what do you think?

tarekgh:
> changing default to parameterless constructor can be less confusing and in the same spirit of struct semantic, what do you think?

I think this can be reasonable but I am not expecting users doing `ValueTask v = default` will always relate to the parameterless constructor. maybe we can have something like:

An instance created with parameterless constructor or by the default(ValueTask) syntax (a zero initialized structure) represents a synchronously, successfully completed operation. I am sure @stephentoub can rephrase it better :-)

carlossanlop:
@glihm any chance you can follow up?

xml/System.Threading.Tasks/ValueTask.xml
```diff
@@ -64,6 +64,10 @@ Using a `ValueTask` instead of a <xref:System.Threading.Tasks.Task> introduces s
 
 > [!NOTE]
 > The use of the `ValueTask` type is supported starting with C# 7.0 and is not supported by any version of Visual Basic.
+
+> [!NOTE]
+> An instance created with the default constructor (a zero initialized structure) represents a synchronously, successfully completed operation.
```
adamsitnik, line 69:
```suggestion
> An instance created with the parameterless constructor or by the `default(ValueTask)` syntax (a zero-initialized structure) represents a synchronously, successfully completed operation.
```

xml/System.Threading.Tasks/ValueTask`1.xml
```diff
@@ -77,6 +77,10 @@ As such, the default choice for any asynchronous method should be to return a <x
 
 > [!NOTE]
 >  The use of the <xref:System.Threading.Tasks.ValueTask%601> type is supported starting with C# 7.0, and is not supported by any version of Visual Basic.
+
+> [!NOTE]
+> An instance created with the default constructor (a zero initialized structure) represents a synchronously, successfully completed operation with a result of `default(TResult)`.
```
adamsitnik, line 82:
```suggestion
> An instance created with the parameterless constructor or by the `default(ValueTask<TResult>)` syntax (a zero-initialized structure) represents a synchronously, successfully completed operation with a result of `default(TResult)`.
```

adamsitnik (approved):
LGTM, thank you @glihm !

PR #4577, opened 2020-07-26
Clarify behavior of Enumerable.Any()-with-predicate on empty sequences

RavuAlHemio:
## Summary

While the documentation for `System.Linq.Enumerable.All<TSource>(this IEnumerable<TSource>, Func<TSource, bool>)` explicitly documents that `true` is returned if the sequence is empty, the documentation for `Enumerable.Any<TSource>(this IEnumerable<TSource>, Func<TSource, bool>)` does not explicitly mention the method's behavior if the sequence is empty.

This pull requests slightly reformulates the _Returns_ section of the `Enumerable.Any()` variant with a predicate argument to clarify that `true` can only be returned if the source sequence is not empty.

### DIFF
--- a/xml/System.Linq/Enumerable.xml
+++ b/xml/System.Linq/Enumerable.xml
@@ -474,7 +474,7 @@
         <param name="predicate">A function to test each element for a condition.</param>
         <summary>Determines whether any element of a sequence satisfies a condition.</summary>
         <returns>
-          <see langword="true" /> if any elements in the source sequence pass the test in the specified predicate; otherwise, <see langword="false" />.</returns>
+          <see langword="true" /> if the source sequence is not empty and at least one of its elements passes the test in the specified predicate; otherwise, <see langword="false" />.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[  
   


gewarren:
@eiriktsarpalis @adamsitnik Could you take a quick look at this small change?

eiriktsarpalis:
I personally don't find any ambiguity in the current documentation. If at least one element is satisfied by the predicate, then the enumerable is necessarily not empty. 

That being said, I wouldn't mind adding a redundant clarification if we know it improves documentation for users, but it should probably be made consistent with the wording in the equivalent `All()` method: does it explicitly mention that empty enumerables return `true`?

RavuAlHemio:
Yes, the documentation of the equivalent `All()` method mentions empty sequences explicitly:

> `true` if every element of the source sequence passes the test in the specified predicate, or if the sequence is empty; otherwise, `false`.

As for the clarity of the documentation for `Any()`, I don't consider it intuitively obvious whether the word "any" implies vacuous truth or not. Despite the equivalence of `All()` and ∀ and the equivalence of `Any()` and ∃, most programming languages implementing similar operations explicitly document their behavior on empty sequences just to make sure.

PR #4580, opened 2020-07-27
Add NotSupportedException to all Vector<T> members

gewarren:
Related to dotnet/docs#19197.


gewarren:
@tannergooding None of the members previously had NotSupportedException listed, but I added it to all of them. Not sure if that's desirable or not.

PR #4590, opened 2020-07-29
Add NSE to Serialize and Deserialize

gewarren:
Contributes to dotnet/docs#19601.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter.deserialize?view=net-5.0&branch=pr-en-us-4590#System_Runtime_Serialization_Formatters_Binary_BinaryFormatter_Deserialize_System_IO_Stream_).


gewarren:
In the XML file, there are two overloads of both Serialize and Deserialize, one that takes a Stream, and one that takes a Stream and a HeaderHandler. Only the overloads with the single parameter appear in the docs though. Any idea why??

gewarren:
> In the XML file, there are two overloads of both Serialize and Deserialize, one that takes a Stream, and one that takes a Stream and a HeaderHandler. Only the overloads with the single parameter appear in the docs though. Any idea why??

The second overload doesn't apply to .NET 5. I'll remove the NSE from those overloads.

PR #4594, opened 2020-07-30
Recommend not to use System.Drawing

benaadams:
## Summary

Recommend not to use System.Drawing, but other libraries instead like the .NET Foundation project SixLabors.ImageSharp instead as it is cross-platform (unlike Windows Imaging Components) and safe to use in a ASP.NET and Windows Service.

The article currently carries this caution, which does not help cross platform:

> Classes within the System.Drawing namespace are not supported for use within a Windows or ASP.NET service. Attempting to use these classes from within one of these application types may produce unexpected problems, such as diminished service performance and run-time exceptions. For a supported alternative, see Windows Imaging Components.

This is similar to the [SmtpClient doc](https://docs.microsoft.com/en-us/dotnet/api/System.Net.Mail.SmtpClient?view=netframework-4.7) which recommends the .NET Foundation project MailKit 

> We don't recommend that you use the SmtpClient class for new development because SmtpClient doesn't support many modern protocols. Use MailKit or other libraries instead. For more information, see SmtpClient shouldn't be used on GitHub.

/cc @richlander @sbwalker @clairernovotny @BillWagner

### DIFF
--- a/includes/drawing.md
+++ b/includes/drawing.md
@@ -1,2 +1,2 @@
 > [!WARNING]
-> We don't recommend that you use the `System.Drawing` namespace for new development as it is  not supported for use within a Windows or ASP.NET service and it is not cross-platform. Use  [ImageSharp](https://github.com/SixLabors/ImageSharp) or other libraries instead.
+> The `System.Drawing` namespace is not recommended for new development, due to not being supported within a Windows or ASP.NET service and it is not cross-platform. [ImageSharp](https://github.com/SixLabors/ImageSharp) and [SkiaSharp](https://github.com/mono/SkiaSharp) are recommended as alternatives.


includes/drawing.md
```diff
@@ -0,0 +1,2 @@
+> [!WARNING]
+> We don't recommend that you use the `System.Drawing` namespace for new development as it is  not supported for use within a Windows or ASP.NET service and it is not cross-platform. Use  [ImageSharp](https://github.com/SixLabors/ImageSharp) or other libraries instead.
```
richlander, line 2:
```suggestion
> The `System.Drawing` namespace is not recommended for new development, due to not being supported within a Windows or ASP.NET service and it is not cross-platform. [ImageSharp](https://github.com/SixLabors/ImageSharp) and [SkiaSharp](https://github.com/mono/SkiaSharp) are recommended as alternatives.
```

PR #4597, opened 2020-07-31
Remove mentions of Itanium

svick:
Itanium has not been relevant for a long time and I believe it's also not supported, so this PR removes mentions of that CPU architecture.

### DIFF
--- a/xml/System.Threading/Interlocked.xml
+++ b/xml/System.Threading/Interlocked.xml
@@ -1971,8 +1971,6 @@ If `comparand` and the object in `location1` are equal by reference, then `value
 ## Remarks  
  This method was added to the <xref:System.Threading.Interlocked> class in the [!INCLUDE[net_v45](~/includes/net-v45-md.md)] as a convenience; it's a wrapper for the <xref:System.Threading.Thread.MemoryBarrier%2A?displayProperty=nameWithType> method.  
   
- <xref:System.Threading.Interlocked.MemoryBarrier%2A> is required only on multiprocessor systems that have weak memory ordering (for example, a system that employs an ARM processor).  
-  
  For most purposes, the C# `lock` statement, the Visual Basic `SyncLock` statement, or the <xref:System.Threading.Monitor> class provide easier ways to synchronize data.  
   
  ]]></format>
--- a/xml/System.Threading/Thread.xml
+++ b/xml/System.Threading/Thread.xml
@@ -2811,8 +2811,6 @@ The value is not guaranteed to be a zero-based processor number.
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- <xref:System.Threading.Thread.MemoryBarrier%2A> is required only on multiprocessor systems with weak memory ordering (for example, a system employing an ARM processor).  
-  
  For most purposes, the C# `lock` statement, the Visual Basic `SyncLock` statement, or the <xref:System.Threading.Monitor> class provide easier ways to synchronize data.  
   
  ]]></format>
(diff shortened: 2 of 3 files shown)


xml/System.Threading/Interlocked.xml
```diff
@@ -1971,7 +1971,7 @@ If `comparand` and the object in `location1` are equal by reference, then `value
 ## Remarks  
  This method was added to the <xref:System.Threading.Interlocked> class in the [!INCLUDE[net_v45](~/includes/net-v45-md.md)] as a convenience; it's a wrapper for the <xref:System.Threading.Thread.MemoryBarrier%2A?displayProperty=nameWithType> method.  
   
- <xref:System.Threading.Interlocked.MemoryBarrier%2A> is required only on multiprocessor systems that have weak memory ordering (for example, a system that employs multiple Intel Itanium processors).  
+ <xref:System.Threading.Interlocked.MemoryBarrier%2A> is required only on multiprocessor systems that have weak memory ordering (for example, a system that employs an ARM processor).
```
kouvel, line 1974: The statement is a bit confusing to me, it seems to suggest that `MemoryBarrier` is not required on x86/x64. Currently a memory barrier may be required for correctness on any supported processor arch, so I would just remove the statement completely.

xml/System.Threading/Thread.xml
```diff
@@ -2811,7 +2811,7 @@ The value is not guaranteed to be a zero-based processor number.
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- <xref:System.Threading.Thread.MemoryBarrier%2A> is required only on multiprocessor systems with weak memory ordering (for example, a system employing multiple Intel Itanium processors).  
+ <xref:System.Threading.Thread.MemoryBarrier%2A> is required only on multiprocessor systems with weak memory ordering (for example, a system employing an ARM processor).
```
kouvel, line 2814: Same as other comment, I suggest removing the statement completely

xml/System/String.xml
```diff
@@ -720,7 +720,7 @@ Examples of instantiating strings:
...
 
-     If the <xref:System.String.%23ctor%28System.Char%2A%29> overload is called and the array is not null-terminated, or if the <xref:System.String.%23ctor%28System.Char%2A%2CSystem.Int32%2CSystem.Int32%29> overload is called and `startIndex` + `length`-1 includes a range that it outside the memory allocated for the sequence of characters, the behavior of the constructor is system-dependent, and an access violation may occur. In addition, on the Intel Itanium processor, calls to the <xref:System.String.%23ctor%28System.Char%2A%2CSystem.Int32%2CSystem.Int32%29> constructor may throw a <xref:System.DataMisalignedException> exception. If this occurs, call the <xref:System.String.%23ctor%28System.Char%5B%5D%2CSystem.Int32%2CSystem.Int32%29> instead.
+     If the <xref:System.String.%23ctor%28System.Char%2A%29> overload is called and the array is not null-terminated, or if the <xref:System.String.%23ctor%28System.Char%2A%2CSystem.Int32%2CSystem.Int32%29> overload is called and `startIndex` + `length`-1 includes a range that it outside the memory allocated for the sequence of characters, the behavior of the constructor is system-dependent, and an access violation may occur.
```
kouvel, line 723:
```suggestion
     If the <xref:System.String.%23ctor%28System.Char%2A%29> overload is called and the array is not null-terminated, or if the <xref:System.String.%23ctor%28System.Char%2A%2CSystem.Int32%2CSystem.Int32%29> overload is called and `startIndex` + `length`-1 includes a range that is outside the memory allocated for the sequence of characters, the behavior of the constructor is system-dependent, and an access violation may occur.
```

svick:
@kouvel Done.

PR #4608, opened 2020-08-04
Remove snippet intro and delete snippet files

### LINKED ISSUE #4368: Missing & incorrect examples in xml/System/Boolean.xml
### Missing example

The following paragraph says that an example follows, but this example is absent:

> When converting from floating-point values to Boolean values, the conversion methods perform an exact comparison with zero. If the floating-point value has lost precision, the result can be unexpected. This is illustrated in the following example, in which a Double variable whose value should be zero is converted to a Boolean value. As the example shows, the result is true because repeated additions of 0.2 have resulted in a loss of precision.

### Incorrect example

The wrong example is in the page _following_ this excerpt:

> However, when you want to compare a Boolean variable with the literal Boolean value true or false, it is not necessary to do an explicit comparison, because the result of evaluating a Boolean value is that Boolean value. For example, the expressions
>
> `if (booleanValue) {`
>
> and

Presumably the example should be

    if (booleanValue == true) {

However it is:

```
using System;

public class Example
{
   public static void Main()
   {
      bool[] hasServiceCharges = { true, false };
      Decimal subtotal = 120.62m;
      Decimal shippingCharge = 2.50m;
      Decimal serviceCharge = 5.00m;

      foreach (var hasServiceCharge in hasServiceCharges) {
         Decimal total = subtotal + shippingCharge +
                                (hasServiceCharge ? serviceCharge : 0);
         Console.WriteLine("hasServiceCharge = {1}: The total is {0:C2}.",
                           total, hasServiceCharge);
      }
   }
}
// The example displays output like the following:
//       hasServiceCharge = True: The total is $128.12.
//       hasServiceCharge = False: The total is $123.12.
```

gewarren:
Fixes #4368

When I ran the snippet, it didn't match the output. The precision didn't seem to be lost during the additions of .2.

Also cleaned up some headings.

_Ignore whitespace changes in diff._

### DIFF
--- a/xml/System/Boolean.xml
+++ b/xml/System/Boolean.xml
@@ -80,7 +80,7 @@
 
 ## Remarks
 
-A <xref:System.Boolean> instance can have either of two values: `true`, or `false`.
+A <xref:System.Boolean> instance can have either of two values: `true` or `false`.
 
  The <xref:System.Boolean> structure provides methods that support the following tasks:
 


xml/System/Boolean.xml
```diff
@@ -79,7 +79,8 @@
       <format type="text/markdown"><![CDATA[
 
 ## Remarks
- A <xref:System.Boolean> instance can have either of two values: `true`, or `false`.
+
+A <xref:System.Boolean> instance can have either of two values: `true`, or `false`.
```
tdykstra, line 83:
```suggestion
A <xref:System.Boolean> instance can have either of two values: `true` or `false`.
```

PR #4610, opened 2020-08-05
Error in docs Vector<256>.WithUpper(..)

dmitry-valyukov:

### DIFF
--- a/xml/System.Runtime.Intrinsics/Vector256.xml
+++ b/xml/System.Runtime.Intrinsics/Vector256.xml
@@ -2559,7 +2559,7 @@
         <typeparam name="T">The type of the input vector.</typeparam>
         <param name="vector">The vector to get the lower 128-bits from.</param>
         <param name="value">The value of the upper 128-bits as a <see cref="T:System.Runtime.Intrinsics.Vector128`1" />.</param>
-        <summary>Creates a new <see cref="T:System.Runtime.Intrinsics.Vector256`1" /> with the upper 128-bits set to the specified value and the upper 128-bits set to the same value as that in the given vector.</summary>
+        <summary>Creates a new <see cref="T:System.Runtime.Intrinsics.Vector256`1" /> with the upper 128-bits set to the specified value and the lower 128-bits set to the same value as that in the given vector.</summary>
         <returns>A new <see cref="T:System.Runtime.Intrinsics.Vector256`1" /> with the upper 128-bits set to the specified value and the upper 128-bits set to the same value as that in <paramref name="vector" />.</returns>
         <remarks>To be added.</remarks>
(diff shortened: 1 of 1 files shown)


xml/System.Runtime.Intrinsics/Vector256.xml
```diff
@@ -2559,7 +2559,7 @@
         <typeparam name="T">The type of the input vector.</typeparam>
         <param name="vector">The vector to get the lower 128-bits from.</param>
         <param name="value">The value of the upper 128-bits as a <see cref="T:System.Runtime.Intrinsics.Vector128`1" />.</param>
-        <summary>Creates a new <see cref="T:System.Runtime.Intrinsics.Vector256`1" /> with the upper 128-bits set to the specified value and the upper 128-bits set to the same value as that in the given vector.</summary>
+        <summary>Creates a new <see cref="T:System.Runtime.Intrinsics.Vector256`1" /> with the upper 128-bits set to the specified value and the lower 128-bits set to the same value as that in the given vector.</summary>
         <returns>A new <see cref="T:System.Runtime.Intrinsics.Vector256`1" /> with the upper 128-bits set to the specified value and the upper 128-bits set to the same value as that in <paramref name="vector" />.</returns>
```
gewarren, line 2563:
```suggestion
        <returns>A vector with the upper 128-bits set to the specified value and the lower 128-bits set to those from <paramref name="vector" />.</returns>
```

xml/System.Runtime.Intrinsics/Vector256.xml
```diff
@@ -2559,7 +2559,7 @@
         <typeparam name="T">The type of the input vector.</typeparam>
         <param name="vector">The vector to get the lower 128-bits from.</param>
         <param name="value">The value of the upper 128-bits as a <see cref="T:System.Runtime.Intrinsics.Vector128`1" />.</param>
```
gewarren, line 2561:
```suggestion
        <param name="value">The value of the upper 128-bits.</param>
```

xml/System.Runtime.Intrinsics/Vector256.xml
dmitry-valyukov: Looks good, but I'm not sure if I should agree or disagree with any proposed changes to the documentation (as an independent developer). Maybe it will be better to create another PR.

xml/System.Runtime.Intrinsics/Vector256.xml
dmitry-valyukov: The same.

dmitry-valyukov: The same.

PR #4611, opened 2020-08-05
Error in docs Vector<128>.WithUpper(..)

dmitry-valyukov:


xml/System.Runtime.Intrinsics/Vector128.xml
```diff
@@ -2653,7 +2653,7 @@
         <typeparam name="T">The type of the input vector.</typeparam>
         <param name="vector">The vector to get the lower 64-bits from.</param>
         <param name="value">The value of the upper 64-bits as a <see cref="T:System.Runtime.Intrinsics.Vector64`1" />.</param>
-        <summary>Creates a new <see cref="T:System.Runtime.Intrinsics.Vector128`1" /> with the upper 64-bits set to the specified value and the upper 64-bits set to the same value as that in the given vector.</summary>
+        <summary>Creates a new <see cref="T:System.Runtime.Intrinsics.Vector128`1" /> with the upper 64-bits set to the specified value and the lower 64-bits set to the same value as that in the given vector.</summary>
         <returns>A new <see cref="T:System.Runtime.Intrinsics.Vector128`1" /> with the upper 64-bits set to the specified value and the upper 64-bits set to the same value as that in <paramref name="vector" />.</returns>
```
gewarren, line 2657:
```suggestion
        <returns>A new <see cref="T:System.Runtime.Intrinsics.Vector128`1" /> with the upper 64-bits set to the specified value and the lower 64-bits set to those in <paramref name="vector" />.</returns>
```

xml/System.Runtime.Intrinsics/Vector128.xml
```diff
@@ -2653,7 +2653,7 @@
         <typeparam name="T">The type of the input vector.</typeparam>
         <param name="vector">The vector to get the lower 64-bits from.</param>
         <param name="value">The value of the upper 64-bits as a <see cref="T:System.Runtime.Intrinsics.Vector64`1" />.</param>
```
gewarren, line 2655:
```suggestion
        <param name="value">The upper 64-bits.</param>
```

xml/System.Runtime.Intrinsics/Vector128.xml
dmitry-valyukov: Looks good, but I'm not sure if I should agree or disagree with any proposed changes to the documentation (as an independent developer). Maybe it will be better to create another PR.

xml/System.Runtime.Intrinsics/Vector128.xml
dmitry-valyukov: The same.

PR #4613, opened 2020-08-05
Update ScrollViewer.xml

DavidHollman:
Refer to ScrollableHeight and ScrollableWidth properties instead of the more complicated explanation in terms of a subtraction. I checked the reference sources and these properties internally perform that same subtraction.


gewarren:
Ping @dotnet/wpf-developers .

SamBent (approved):

gewarren (approved):

PR #4633, opened 2020-08-08
Indicate the fact that the bufferSize param on OpenStandardxxx methods has no effect

john-h-k:


eiriktsarpalis (approved):
lgtm

xml/System/Console.xml
```diff
@@ -1924,7 +1924,7 @@
         <Parameter Name="bufferSize" Type="System.Int32" Index="0" FrameworkAlternate="net-5.0;netcore-2.0;netcore-2.1;netcore-2.2;netcore-3.0;netcore-3.1;netframework-1.1;netframework-2.0;netframework-3.0;netframework-3.5;netframework-4.0;netframework-4.5;netframework-4.5.1;netframework-4.5.2;netframework-4.6;netframework-4.6.1;netframework-4.6.2;netframework-4.7;netframework-4.7.1;netframework-4.7.2;netframework-4.8;netstandard-2.0;netstandard-2.1;xamarinandroid-7.1;xamarinios-10.8;xamarinmac-3.0" />
       </Parameters>
       <Docs>
-        <param name="bufferSize">The internal stream buffer size.</param>
+        <param name="bufferSize">The internal stream buffer size. On both .NET Core and .NET Framework, this parameter has no effect.</param>
```
eiriktsarpalis, line 1927: @carlossanlop can you provide some feedback on wording here? Is the phrase "On both .NET Core and .NET Framework" redundant perhaps?

carlossanlop: It is redundant, unless it has an effect on .NET 5.0. @john-h-k is this the case? (.NET 5.0 is not considered a ".NET Core", that's why I ask).

carlossanlop: @gewarren do we have new guidance on how to group .NET Core (3.1 and older) with .NET 5.0?

gewarren: The wording we used, which is Beth Massi-approved, is ".NET 5 (and .NET Core) and later versions". But I don't think that's what we're looking for here, since this includes .NET Framework (and possibly not .NET 5).

carlossanlop: So it seems there's no point in indicating the .NET versions at all. Right?

jeffhandley, line 1927:
```suggestion
        <param name="bufferSize">This parameter has no effect, but its value must be greater than or equal to zero.</param>
```

xml/System/Console.xml
```diff
@@ -2043,7 +2043,7 @@ The following example illustrates the use of the `OpenStandardInput` method.
...
```
adamsitnik, line 2046:
```suggestion
        <param name="bufferSize">This parameter has no effect, but its value must be greater than or equal to zero.</param>
```

xml/System/Console.xml
```diff
@@ -2163,7 +2163,7 @@ This method can be used to reacquire the standard output stream after it has bee
...
```
adamsitnik, line 2166:
```suggestion
        <param name="bufferSize">This parameter has no effect, but its value must be greater than or equal to zero.</param>
```

adamsitnik (approved):
LGTM, thank you @john-h-k

PR #4639, opened 2020-08-10
Update Task.Run(Task, CancellationToken) example

garogarabedyan:
## Summary

Modify the example to pass 2 cancellation tokens -- one to cancel the task before starting (by the Task class code) and the other to cancel the task in the middle of computation (by user defined code).

This snippet is used on https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.run?view=netcore-3.1#System_Threading_Tasks_Task_Run_System_Func_System_Threading_Tasks_Task__System_Threading_CancellationToken_

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR_System/system.threading.tasks.task.run/cs/Run7.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR_System/system.threading.tasks.task.run/cs/Run7.cs
@@ -8,7 +8,7 @@ using System.Threading.Tasks;
 public class Example
 {
     private static CancellationTokenSource source = new CancellationTokenSource();
-    private static volatile int completedIterations = 0;
+    private static int completedIterations = 0;
     public static void Main()
     {
         var tasks = new List<Task<int>>();


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=4639) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=4639) before we can accept your contribution.<br/><br/>:x: garogarabedyan [sign now](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=4639)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/dotnet-api-docs?pullRequest=4639) it.</sub>

tarekgh:
The suggested changes looks good to me.

@carlossanlop maybe better to wait when @stephentoub take a look just in case if he has any comment on this code sample.

stephentoub:
I'm not following the concern with the original code sample.  @garogarabedyan, you refer to volatility, but the token isn't being changed; what is your concern?

garogarabedyan:
Hi Stephen,

Although, the code is correct as you pointed out, I feel it is too easy for anyone to quickly modify it into thread unsafe program.

Best,
Garo

garogarabedyan:
@stephentoub Although the non-volatile `token` variable is not modified as you pointed out, I feel the token variable and the integer variable `completedIterations` both being non-volatile encourage users in thread unsafe practices.

stephentoub:
> and the integer variable completedIterations

Making `completedIterations` volatile doesn't help anything; in fact it adds unnecessary cost on weaker memory models.  The only manipulation of it is already being performed with Interlocked, where the volatility on the variable will be ignored, and the subsequent read of the variable won't be affected by making it volatile because the Interlocked right before it serves as a full memory barrier.  I don't think it's a good idea to add volatile in places it's not actually needed; that itself causes confusion and thus has downsides.  Just making such a variable volatile is not a best practice.

> the non-volatile token variable

Your change still reads a non-volatile source field from multiple threads, so I'm not sure what the change is trying to fix there.

garogarabedyan:
@stephentoub 

Regarding `completedIterations`: I agree it is bad to have it volatile and will revert it.

Regarding the non-volatile CancellationTokenSource field: CancellationTokenSource class has a private volatile field which keeps the state of the cancellation (whether cancelled or not) which works the same with the original sample and the proposed changes.

Regarding the overall idea: I feel now the summary is misleading and it should be "Pass 2 new CancellationToken structs -- one to cancel the task before starting (by the Task class code) and the other to cancel the task in the middle of computation (by user defined code) to illustrate the usage of `Task.Run(Action, CancellationToken)`".

Do you think the changes now are good enough?

stephentoub:
> Regarding the non-volatile CancellationTokenSource field: CancellationTokenSource class has a private volatile field which keeps the state of the cancellation

I'm not sure what that has to do with this use.  If the concern is around access to state from multiple threads, before you even get to the cancellation token you're accessing the source.  I'm not saying it should be volatile; it shouldn't.  But rather I'm pointing out that making things volatile needs to be done for a specific reason, which doesn't exist here.

> "Pass 2 new CancellationToken structs -- one to cancel the task before starting (by the Task class code) and the other to cancel the task in the middle of computation (by user defined code) to illustrate the usage of Task.Run(Action, CancellationToken)"

I don't understand.  Both before and after your change there's only one token that's both passed as an argument to Task.Run and that's captured by the Task.Run lambda.

garogarabedyan:
I appreciate your comment, @stephentoub. 

I agree CancellationTokenSource should not be volatile and I have already reverted `completedIterations` to non-volatile, because of your argument of the full memory barrier imposed by the Interlocked.Increment method before reading it's value.

I disagree there is a single CancellationToken struct passed to the lambda and the Task.Run method, because struct is a value type and copied when passed. I must admit, however, that is not making a lot of difference from thread safety perspective with the current CancellationToken definition, because it only has a reference to it's parent CancellationTokenSource object.

I still think the change is valuable as it shows the usage of the 2 CancellationTokens -- one to cancel the task before starting (by the Task class code) and the other to cancel the task in the middle of computation (by user defined code). My proposed changes about volatility were not improving the code and were actually making it worse and that is why I have reverted them.

stephentoub:
They're not two different tokens.  Yes, they're structs, but they represent exactly the same source of cancellation.  If you compare them with either == or Equals, it'll return true.

garogarabedyan:
I simply feel passing the CancellationToken as an argument is a better or best thread safety practice for CancelationToken struct and value types in general, compared to capturing them.

stephentoub:
> compared to capturing them

Your code is still capturing it.

garogarabedyan:
I agree the current revision of CancellationToken and CancellationTokenSource does not need these changes and can capture CancellationToken from lamdas and will revert this PR. I, however, have one question for a future possible scenario: Would you consider always passing CancellationToken struct as function arguments, instead of capturing it, if CancellationToken was a stateful struct? 

CancellationTokenSource.Token produces structs with a single implicitly readonly reference to CancellationTokenSource object in the heap. In a future scenario, however, when CancellationToken struct is extended with fields which keep state, capturing such new stateful struct CancellationToken from lamdas will not be thread safe.

stephentoub:
> however, when CancellationToken struct is extended with fields which keep state, capturing such new stateful struct CancellationToken from lamdas will not be thread safe

That would be a breaking change. Too much code relies on being able to use tokens in this manner, free from tearing and other such concerns.

On top of that, even if tearing were a concern, it's not in the current code based on its structure. There is zero correctness benefit to the proposed change, and the refactoring to use a separate function makes the example harder to read.

garogarabedyan:
I appreciate your answer, @stephentoub !

Unfortunately, I do not know what is tiering at least in the context you are using it and maybe that is why I feel passing a stateful struct as function parameter will avoid thread safety problems, because will be a new copy in the call stack.

IMHO, CancellationTokenSource class and CancellationToken struct should be merged into one class, if feasible, which will allow it to be captured from inside lambdas, I think will be easier for the developers to understand and use.

Again, thanks for your review.

PR #4640, opened 2020-08-11
System.Net.HttpListener.BeginGetContext does not invoke callback asyn…

apenn-msft:
…chronously

There is surprising and unexpected behavior in that the API may call the callback synchronously and if the callback is blocking (or recursive) so too then becomes this API, which is supposed to not block.
Confirmed this is expected via the feedback item:
https://developercommunity.visualstudio.com/content/problem/1142285/systemnethttplistenerbegingetcontext-does-not-invo.html

and making the documentation explicit on what "the operation" entails.

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Net/HttpListener.xml
+++ b/xml/System.Net/HttpListener.xml
@@ -372,6 +372,8 @@ Use the [HttpCfg.exe](/windows/win32/http/httpcfg-exe) tool to add SSL certifica
  The asynchronous operation must be completed by calling the <xref:System.Net.HttpListener.EndGetContext%2A> method. Typically, the method is invoked by the `callback` delegate.  
   
  This method does not block while the operation completes. To get an incoming request and block until the operation completes, call the <xref:System.Net.HttpListener.GetContext%2A> method.  
+ > [!NOTE]
+ > The operation does not include the invocation of the supplied callback; the callback may be invoked synchronously if the operation has already completed.
   
  For detailed information about using the asynchronous programming model, see [Calling Synchronous Methods Asynchronously](/dotnet/standard/asynchronous-programming-patterns/calling-synchronous-methods-asynchronously)  
   


xml/System.Net/HttpListener.xml
```diff
@@ -372,6 +372,8 @@ Use the [HttpCfg.exe](/windows/win32/http/httpcfg-exe) tool to add SSL certifica
  The asynchronous operation must be completed by calling the <xref:System.Net.HttpListener.EndGetContext%2A> method. Typically, the method is invoked by the `callback` delegate.  
   
  This method does not block while the operation completes. To get an incoming request and block until the operation completes, call the <xref:System.Net.HttpListener.GetContext%2A> method.  
+ > [!NOTE]
```
karelz, line 375: This is general APM note - do other APIs using APM have same warning? Should they have it? Should we rather clarify APM docs?

apenn-msft: The general APM concept of an "async callback" states it must be called on another thread:
https://docs.microsoft.com/en-us/dotnet/api/system.asynccallback?view=netcore-3.1
"Use an AsyncCallback delegate to process the results of an asynchronous operation in a separate thread."

I'm thinking there are 3 cases then:
1. Just BeginGetContext deviates from APM spec
2. All Begin* APIs in HttpListener deviate this way
3. All APM components in .Net system libraries behave this way

I hope it's just case 1 and in that case I'd much rather prefer we fix the deviation as a bug than document it, but I also know that HttpListener is not really supported anymore and it would change its behavior in a meaningful way (e.g. may some callers today expect the callback to happen on the same thread under the circumstances where it would happen (the request is already complete).

If we decide not to fix it and it is case 2, then yea the note should be propagated to all HttpListener Begin* methods, or on the main documentation page for HttpListener. 

If it's case 3 however, then it would be a bug between implementation and specification, and yep, we'd have to change the APM specification/documentation (assuming we can't fix all implementations :)). I do hope it's not this case though because even before reading APM docs, I already intuitively expected an "AsyncCallback" to be ... async. After reading the APM docs I can say at least that seems to be the expectation for .Net too.

BillWagner:

carlossanlop:
@karelz @aik-jahoda @scalablecory can you follow up on this?

xml/System.Net/HttpListener.xml
karelz: It is option 3 to my best knowledge and the docs are just not clear. It is not bug in spec.

ManickaP:
IMO, this is more of an issue of conceptual docs about APM and not specific to this API. To support my case, [`IAsyncResult`](https://docs.microsoft.com/en-us/dotnet/api/system.iasyncresult.completedsynchronously?view=net-5.0) has a property saying whether the operation is completed synchronously. Moreover, our docs doesn't state that callback is always called on a separate thread. It states: 
> This method does not block while the operation completes.

Which is true if the result is a available immediately.

I don't think this should be merged. 

However, we could improve the conceptual docs and more vocally specify that the callback might be issued on the calling thread of the `Begin` method.

apenn-msft:
@ManickaP:
> Moreover, our docs doesn't state that callback is always called on a separate threa

see my earlier response here:
> https://docs.microsoft.com/en-us/dotnet/api/system.asynccallback?view=netcore-3.1
"Use an AsyncCallback delegate to process the results of an asynchronous operation in a separate thread."

The docs do state explicitly that an async callback should be called on a separate thread.
One might expect the responsibility to fall on AsyncCallback to ensure this; but realistically it's implemented as a delegate and so ti doesn't have control over method invocation. It's up to the invoker to enforce the async invocation that AsyncCallback documents it has. Or change the documentation.

Personally, I would prefer to change the implementation since "Async" callback should, well... be async!
This could be done by changing AsyncCallback to first-class type that wraps a delegate, but I don't think C# supports callable types? (other than delegates). A way to get around this is by providing a conversion operator from this AsyncCallback type to a delegate, but I'm not sure when conversion operators were added to the language or if they would work in all contexts AsyncCallback is being invoked in today.

Practically speaking, it's probably easier to preserve the existing behavior and change the documentation to match it; adding a caveat that "Async" doesn't always mean async and the callback can be invoked on the calling thread/same thread as the operation.

ManickaP:
> The docs do state explicitly that an async callback should be called on a separate thread.

But that's `AsyncCallback` documentation, not `BeginGetContext`. I'd be fine if the PR was changing that and not adding a note to `BeginGetContext`.

apenn-msft:
Right but the documentation bug exists because of the implementation details: AsyncCallback is just a delegate and has no way to enforce this. It is brittle to encode caller-behavior into the documentation of the callee. Today some callers may always call AsyncCallbacks on a new thread, some may never, and like BeginContext some may do mixed-mode; therefore it is up to the user of AsyncCallback to document whether it is called asynchronously.

In practice it might just happen to be that all of .NET and all of APM is intended to use AsyncCallback the same way BeginContext does, but even then, that may not be true for all C# code. The reality is AsyncCallback has no way to guarantee the behavior it is documenting, so it should not be documenting itself as async. It might be acceptable to note in AsyncCallback that it is up to the caller to determine whether it is indeed "Async", but then we also need documentation in each API that uses it (e.g. BeginContext) to state whether it is called async, sync, or both.

ManickaP:
Yes, `AsyncCallback` cannot guarantee how it is called, that's why it shouldn't state it's called on a different thread.

As I said previously, there are no guarantees anywhere. Callbacks might be invoked synchronously everywhere. The calling code should not depend on where the callbacks are called. If it needs to be execute on a different thread, it must ensure that on it's own (e.g. [ThreadPool.QueueUserWorkItem](https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadpool.queueuserworkitem?view=net-5.0)).

We shouldn't pollute documentation of a single `Begin` method with these details just because that's the one where it caused an issue.

apenn-msft:
> Callbacks might be invoked synchronously everywhere. 
AsyncCallback is quite an unfortunate name we've chosen then. At best, it could be ambiguous whether "Async" refers to the IAsyncResult parameter or it being asynchronously executed (once we remove the part of the documentation that explicitly states the latter)
 
Can we link a follow-up PR/issue to fix the documentation here to not mention "in another thread": 
https://docs.microsoft.com/en-us/dotnet/api/system.asynccallback?view=netcore-3.1
https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/using-an-asynccallback-delegate-to-end-an-asynchronous-operation

there may be other instances in the surrounding APM documentation so it would be good to also conduct a brief review.

carlossanlop:
Ping @ManickaP for follow up.

ManickaP:
We should close this PR and instead fix the conceptual docs mentioned in https://github.com/dotnet/dotnet-api-docs/pull/4640#issuecomment-690732459

apenn-msft:
> We should close this PR and instead fix the conceptual docs mentioned in [#4640 (comment)](https://github.com/dotnet/dotnet-api-docs/pull/4640#issuecomment-690732459)

don't forget to create a new issue and link it before closing one.
I've created one here we can use to track the work in that comment:
https://github.com/dotnet/dotnet-api-docs/issues/5338

jeffhandley:
Thanks, @apenn-msft! Sorry about that.

PR #4651, opened 2020-08-12
Vector improvements

gewarren:
- Remove explicit types from parameter and return type descriptions, since they're added automatically.
- 128-bits to 128 bits.
- Also fixes a typo right at the end (upper bits -> lower bits).


carlossanlop (approved):
LGTM. Thanks!

PR #4655, opened 2020-08-12
Fix up MSBuild API ref

gewarren:
Remarks tags were inserted into summary tags resulting in things like this:

image

Also improves other property summaries and values.

### DIFF
--- a/xml/Microsoft.Build.Framework/SdkResult.xml
+++ b/xml/Microsoft.Build.Framework/SdkResult.xml
@@ -14,13 +14,17 @@
   </Base>
   <Interfaces />
   <Docs>
-    <summary>
-                An abstract interface class to indicate SDK resolver success or failure.
-                <remarks>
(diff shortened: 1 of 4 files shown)


xml/Microsoft.Build.Framework/SdkResult.xml
```diff
@@ -14,13 +14,17 @@
   </Base>
   <Interfaces />
   <Docs>
-    <summary>
-                An abstract interface class to indicate SDK resolver success or failure.
-                <remarks>
```
gewarren, line 19: @joelmartinez It seems like a build warning could/should be generated if `<remarks>` tags are found within `<summary>` or `<returns>` tags?

joelmartinez: @mimisasouvanh this sounds like a great customer suggestion :)

gewarren:
Close/reopen.

gewarren:
Ping @ghogen.

PR #4668, opened 2020-08-13
Automatic port of System.Formats.Asn1 documentation

carlossanlop:
Ported with [DocsPortingTool](https://github.com/carlossanlop/DocsPortingTool). Includes exceptions.

### DIFF
--- a/xml/System.Formats.Asn1/AsnWriter.xml
+++ b/xml/System.Formats.Asn1/AsnWriter.xml
@@ -906,7 +906,7 @@ This method only checks that the tag and length are encoded according to the cur
 ## Remarks
 
 The index of the bit array corresponds to the bit number in the encoded format, which is different than the value produced by <xref:System.Collections.BitArray.CopyTo(System.Array,System.Int32)> with a byte array.
-              For example, the bit array `{ `false`, `true`, `true` }` encodes as `0b0110_0000` with 5 unused bits.
+              For example, the bit array `{ false, true, true }` encodes as `0b0110_0000` with 5 unused bits.
 
           ]]></format>
         </remarks>


xml/System.Formats.Asn1/AsnWriter.xml
```diff
@@ -757,10 +897,23 @@
...
+        <param name="value">The bits to write</param>
+        <param name="tag">The tag to write, or <see langword="null" /> for the default tag (Universal 3).</param>
+        <summary>Write a bit array value as a NamedBitList with a specified tag.</summary>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+The index of the bit array corresponds to the bit number in the encoded format, which is different than the value produced by <xref:System.Collections.BitArray.CopyTo(System.Array,System.Int32)> with a byte array.
+              For example, the bit array `{ `false`, `true`, `true` }` encodes as `0b0110_0000` with 5 unused bits.
```
bartonjs, line 909: Looks like the tool didn't do the best possible job here.

What it wrote (`{ `false`, `true`, `true` }`) doesn't print nicely.

```suggestion
              For example, the bit array `{ false, true, true }` encodes as `0b0110_0000` with 5 unused bits.
```

`{ false, true, true }`  is more what's expected.  (And matches the xml content... is there a late filter that wraps boolean values in backticks?  It might need to learn if it's already in an escaped context?)

carlossanlop: Yes, the tool finds and wraps `true` and `false` with backticks. It's not perfect. This is a rare case where the larger string was already escaped, I think I can get it fixed.

PR #4686, opened 2020-08-16
OpCodes.Call valid on ConstructorInfo

artemus-harper:
## Summary

It is valid to use Call on constructors when implementing the base() or this() constructor call.


GrabYourPitchforks (approved):

gewarren:
@artemus-harper Thank you!

PR #4689, opened 2020-08-17
Automatic port of System.Collections exceptions

carlossanlop:
Ported with [DocsPortingTool](https://github.com/carlossanlop/DocsPortingTool).

### DIFF
--- a/xml/System.Collections.Immutable/ImmutableList`1.xml
+++ b/xml/System.Collections.Immutable/ImmutableList`1.xml
@@ -1882,7 +1882,6 @@
   
  ]]></format>
         </remarks>
-        <exception cref="T:System.NotImplementedException" />
         <exception cref="T:System.NotSupportedException">Always thrown.</exception>
       </Docs>
     </Member>


xml/System.Collections.Immutable/ImmutableList`1.xml
```diff
@@ -1883,6 +1883,7 @@
  ]]></format>
         </remarks>
         <exception cref="T:System.NotImplementedException" />
+        <exception cref="T:System.NotSupportedException">Always thrown.</exception>
```
carlossanlop, line 1886: This one is weird, because the previous line indicates a `NotImplementedException` already being thrown. Can you please verify, @eiriktsarpalis / @layomia ?

eiriktsarpalis: This is what exists in the method implementation https://github.com/dotnet/runtime/blob/dfb3022b59700f33ff8f7034982e5ce642d80688/src/libraries/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableList_1.cs#L940

eiriktsarpalis: I think the tool should have replaced the previous line?

carlossanlop: No, the tool does not replace anything pre-existing in the xml. I'm not sure why `NotImplementedException` is there, so I'll remove it.

carlossanlop: Thanks for checking, I was having trouble finding that file.

carlossanlop, line 1886:
```suggestion
        <exception cref="T:System.NotSupportedException">Always thrown.</exception>
```

PR #4704, opened 2020-08-19
Update Debugger.xml

danmoseley:
Relates to https://github.com/dotnet/runtime/issues/41029

## Summary

Notes that Debugger.Launch() only works on Windows.

### DIFF
--- a/xml/System.Diagnostics/Debugger.xml
+++ b/xml/System.Diagnostics/Debugger.xml
@@ -356,7 +356,8 @@ Console.WriteLine("Hello, world.");
   
 ## Remarks  
  If a debugger is already attached, nothing happens.  
- Debugger launch is only supported on Windows. On Unix, the method will return true without launching a debugger.
+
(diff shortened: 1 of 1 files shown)


xml/System.Diagnostics/Debugger.xml
```diff
@@ -356,6 +356,7 @@ Console.WriteLine("Hello, world.");
   
 ## Remarks  
  If a debugger is already attached, nothing happens.  
+ Debugger launch is only supported on Windows. On Unix, the method will return true without launching a debugger.
```
gewarren, line 359:
```suggestion

Debugger launch is only supported on Windows. On Unix operating systems, the method returns `true` without launching a debugger.
```

danmosemsft: Thanks, applied. BTW in future don't hesitate to apply suggested edits, I'm never going to mind :)

PR #4728, opened 2020-08-26
fill in todo's for GC APIs

Maoni0:
## Summary

Fill in info for some GC APIs' returns.

### DIFF
--- a/xml/System/GC.xml
+++ b/xml/System/GC.xml
@@ -206,7 +206,7 @@
         <param name="length">Specifies the length of the array.</param>
         <param name="pinned">Specifies whether the allocated array must be pinned.</param>
         <summary>Allocates an array.</summary>
-        <returns>To be added.</returns>
+        <returns>An array object.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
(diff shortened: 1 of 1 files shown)


Maoni0:
@carlossanlop

xml/System/GC.xml
```diff
@@ -250,7 +250,7 @@ If pinned is set to `true`, `T` must not be a reference type or a type that cont
         <param name="length">Specifies the length of the array.</param>
         <param name="pinned">Specifies whether the allocated array must be pinned.</param>
         <summary>Allocates an array while skipping zero-initialization, if possible.</summary>
-        <returns>To be added.</returns>
+        <returns>An array object with uninitialized memory except if it contains references or if it's too small for unpinned.</returns>
```
carlossanlop, line 253: > except if it contains references or if it's too small for unpinned.

Does this mean it will throw an exception for those cases? If yes, which exception? Exceptions need to be added as extra xml fields, right below `<remarks>`. I can help with that if necessary.

Maoni0: ahh yes that's a great thing to mention. for those cases it would simply allocated a normal array, ie, just do new T[length]. no exception is thrown.

PR #4732, opened 2020-08-27
Add missing api doc for System.Net.HttpVersion

aik-jahoda:
part of: dotnet/runtime#41446

### DIFF
--- a/xml/System.Net/HttpVersion.xml
+++ b/xml/System.Net/HttpVersion.xml
@@ -207,32 +207,9 @@
         <ReturnType>System.Version</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Defines a <see cref="T:System.Version" /> instance for HTTP/2.</summary>
+        <summary>Defines a <see cref="T:System.Version" /> instance for HTTP 2.0.</summary>
         <remarks>To be added.</remarks>
       </Docs>
-    </Member>
-    <Member MemberName="Version30">
-      <MemberSignature Language="C#" Value="public static readonly Version Version30;" />
-      <MemberSignature Language="ILAsm" Value=".field public static initonly class System.Version Version30" />
-      <MemberSignature Language="DocId" Value="F:System.Net.HttpVersion.Version30" />
-      <MemberSignature Language="VB.NET" Value="Public Shared ReadOnly Version30 As Version " />
-      <MemberSignature Language="C++ CLI" Value="public: static initonly Version ^ Version30;" />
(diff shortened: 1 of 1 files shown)


xml/System.Net/HttpVersion.xml
```diff
@@ -230,7 +230,7 @@
         <ReturnType>System.Version</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Defines a <see cref="T:System.Version" /> instance for HTTP/3.</summary>
```
ManickaP, line 233: `Version30` has been removed: https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.Primitives/ref/System.Net.Primitives.cs#L205-L211
https://github.com/dotnet/runtime/pull/40468/files#diff-03c1aae84de79481c30bdcd38f3ef9bd
cc: @scalablecory

aik-jahoda: I have removed whole member section

xml/System.Net/HttpVersion.xml
```diff
@@ -210,29 +210,6 @@
...
-      </AssemblyInfo>
-      <ReturnValue>
-        <ReturnType>System.Version</ReturnType>
-      </ReturnValue>
-      <Docs>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
-      </Docs>
-    </Member>
+    </Member>
```
ManickaP, line 213: I'm not sure how docs and auto-generating tools will handle this removal.
Maybe it should be left with `To be added.` (which is skipped) until auto-removed. If that's what the tool will do once it sees the property is gone.

ManickaP (approved):
Content-wise LGTM.

xml/System.Net/HttpVersion.xml
aik-jahoda: I believe the tool will recreate the section once it discover the api again. By the way it should also fill comments from `///` in code during first run.

@carlossanlop am I right?

carlossanlop: Don't remove APIs manually.

If this is an API that has been removed from runtime, then we have to wait for the MS Docs build system to catch up and remove it automatically.

Is this API gone in Preview8 or above? Because we haven't updated MS Docs yet with that preview.

carlossanlop:
I'm closing this PR since we should not manually modify the autogenerated xmls.

PR #4760, opened 2020-09-01
Add SslStream missing API documentation

aik-jahoda:
## Summary

SslStream: add missing API documentation.

part of: dotnet/runtime#41446


aik-jahoda:
I don't know what causes the invalid xref so I created an issue: https://github.com/dotnet/dotnet-api-docs/issues/4763

carlossanlop:
The warning is indeed unrelated to your change and you opened an issue for it, so I think we can get this merged since it LGTM.

PR #4787, opened 2020-09-03
Fix double asterisk, add extra type summary

carlossanlop:
@JasonCard @gewarren @BillWagner PTAL

Fixed the issue reported by Jason, and also found a type summary that we can fix at the same time.

### DIFF
--- a/xml/Microsoft.Build.Globbing/MSBuildGlob.xml
+++ b/xml/Microsoft.Build.Globbing/MSBuildGlob.xml
@@ -21,9 +21,9 @@
     <summary>
                 Represents a parsed MSBuild glob.
                 An MSBuild glob is composed of three parts:
-                - fixed directory part: <c>a/b/</c> in <c>a/b/**/*test*/**/*.cs</c>
-                - wildcard directory part: <c>**/*test*/**/</c> in <c>a/b/**/*test*/**/*.cs</c>
-                - file name part: <c>*.cs</c> in <c>a/b/**/*test*/**/*.cs</c>
+                - fixed directory part: <c>a/b/</c> in <c>a/b/\*\*/*test\*/\*\*/\*.cs</c>
+                - wildcard directory part: <c>\*\*/\*test\*/\*\*/</c> in <c>a/b/\*\*/\*test\*/\*\*/\*.cs</c>
+                - file name part: <c>\*.cs</c> in <c>a/b/\*\*/\*test\*/\*\*/\*.cs</c>
             </summary>
     <remarks>To be added.</remarks>
   </Docs>


xml/Microsoft.Build.Globbing/MSBuildGlob.xml
```diff
@@ -21,9 +21,9 @@
     <summary>
                 Represents a parsed MSBuild glob.
                 An MSBuild glob is composed of three parts:
-                - fixed directory part: "a/b/" in "a/b/**/*test*/**/*.cs"
-                - wildcard directory part: "**/*test*/**/" in "a/b/**/*test*/**/*.cs"
-                - file name part: "*.cs" in "a/b/**/*test*/**/*.cs"
+                - fixed directory part: <c>a/b/</c> in <c>a/b/**/*test*/**/*.cs</c>
+                - wildcard directory part: <c>**/*test*/**/</c> in <c>a/b/**/*test*/**/*.cs</c>
+                - file name part: <c>*.cs</c> in <c>a/b/**/*test*/**/*.cs</c>
             </summary>
```
carlossanlop, line 27: This is not markdown, so I don't know why the double star is being interpreted as bold. @joelmartinez do you know if my fix is correct so the star characters show up?

joelmartinez: hmm ... Admittedly, I too was under the impression that summary content was not treated as markdown, however inspecting the code for Ecma2Yaml, I can find this (for overloads), which shows that it's handling the summary in this case as markdown. This means the fix you've added here is required from an authoring perspective.  
https://github.com/docascode/ECMA2Yaml/blob/49078f3b9d9be2c6d034b089ecffd213bd7e5ab1/ECMA2Yaml/ECMA2Yaml/YamlHelpers/YamlHeaderWriter.cs#L17-L20

        [JsonProperty("summary")]
        [MarkdownContent]
        [YamlMember(Alias = "summary")]
        public string Summary { get; set; }

@TianqiZhang, can you verify whether the summary is treated in ecma2yaml as markdown across the board?

carlossanlop: Found an existing example that works well: https://github.com/dotnet/dotnet-api-docs/blob/09f58cd587c249c585def8087f88fa7004f24495/xml/Microsoft.Extensions.FileProviders/PhysicalFileProvider.xml#L370

xml/Microsoft.Build.Globbing/MSBuildGlob.xml
```diff
@@ -21,9 +21,9 @@
     <summary>
                 Represents a parsed MSBuild glob.
                 An MSBuild glob is composed of three parts:
-                - fixed directory part: "a/b/" in "a/b/**/*test*/**/*.cs"
-                - wildcard directory part: "**/*test*/**/" in "a/b/**/*test*/**/*.cs"
-                - file name part: "*.cs" in "a/b/**/*test*/**/*.cs"
+                - fixed directory part: <c>a/b/</c> in <c>a/b/**/*test*/**/*.cs</c>
+                - wildcard directory part: <c>**/*test*/**/</c> in <c>a/b/**/*test*/**/*.cs</c>
+                - file name part: <c>*.cs</c> in <c>a/b/**/*test*/**/*.cs</c>
```
carlossanlop, line 26:
```suggestion
                - fixed directory part: <c>a/b/</c> in <c>a/b/\*\*/*test\*/\*\*/\*.cs</c>
                - wildcard directory part: <c>\*\*/\*test\*/\*\*/</c> in <c>a/b/\*\*/\*test\*/\*\*/\*.cs</c>
                - file name part: <c>\*.cs</c> in <c>a/b/\*\*/\*test\*/\*\*/\*.cs</c>
```

xml/Microsoft.Build.Globbing/MSBuildGlob.xml
carlossanlop: Here's how that one looks: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.fileproviders.physicalfileprovider.watch?view=dotnet-plat-ext-3.1#parameters

JasonCard: I confirmed that example had no issues for loc, so this fix should work.

JasonCard (approved):

PR #4791, opened 2020-09-03
Automatic port of System.Numerics docs

carlossanlop:


xml/System.Numerics/BigInteger.xml
```diff
@@ -1588,9 +1588,17 @@
...
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Gets the number of bits required for shortest two's complement representation of the current instance without the sign bit.</summary>
+        <returns>The minimum non-negative number of bits in two's complement notation without the sign bit.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+This method returns 0 iff the value of current object is equal to <xref:System.Numerics.BigInteger.Zero> or <xref:System.Numerics.BigInteger.MinusOne>. For positive integers the return value is equal to the ordinary binary representation string length.
```
pgovind, line 1598:
```suggestion
This method returns 0 if the value of current object is equal to <xref:System.Numerics.BigInteger.Zero> or <xref:System.Numerics.BigInteger.MinusOne>. For positive integers the return value is equal to the ordinary binary representation string length.
```

pgovind (approved):
LGTM!

PR #4804, opened 2020-09-04
CI update for mdoc 5.8

gewarren:


xml/System/Half.xml
```diff
@@ -784,7 +784,7 @@
       <Docs>
         <param name="s">A string that contains a number to convert.</param>
         <summary>Converts the string representation of a number to its half-precision floating-point number equivalent.</summary>
-        <returns>A half-precision floating-point number equivalent to the numeric value or symbol specified in <paramref name="s"/>.</returns>
+        <returns>A half-precision floating-point number equivalent to the numeric value or symbol specified in <paramref name="s" />.</returns>
```
gewarren, line 787: Is adding the missing space a new feature of mdoc 5.8 @joelmartinez ?

xml/Microsoft.Extensions.Logging/LoggingBuilderExtensions.xml
```diff
@@ -126,25 +126,6 @@
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
-    <Member MemberName="Configure">
```
gewarren, line 129: This member deletion is the only thing that seems "off" to me - the binaries haven't changed as far as I'm aware. @joelmartinez

xml/System/Half.xml
joelmartinez: This is just the built in XML APIs formatting the XML ... hasn't changed in years. But it's a common question from authors who don't expect changes in their written content :)

xml/Microsoft.Extensions.Logging/LoggingBuilderExtensions.xml
joelmartinez: Hmm ... looking at the changes to the framework index files, it doesn't show this member being deleted there, which suggests it was never actually a part of the manifest that mdoc published. Perhaps this is just an artifact from changes to the .net 5 preview? Seems like an anomaly?

gewarren: @carlossanlop Can you help me figure out why this is getting deleted from the docs? It appears to still be in the source code: https://github.com/dotnet/runtime/blob/a9f6fd69b362f0c4f986bd7bf241bf5ef7d06dce/src/libraries/Microsoft.Extensions.Logging/src/LoggingBuilderExtensions.cs#L58.

gewarren: It's still in the framework index file at https://raw.githubusercontent.com/dotnet/dotnet-api-docs/master/xml/FrameworksIndex/dotnet-plat-ext-5.0.xml:

`<Member Id="M:Microsoft.Extensions.Logging.LoggingBuilderExtensions.Configure(Microsoft.Extensions.Logging.ILoggingBuilder,System.Action{Microsoft.Extensions.Logging.LoggerFactoryOptions})" />`

gewarren: Also tagging @safern for comment.

joelmartinez: @gewarren ok ... I've created a bug to track this issue:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/298401 

We'll figure out what's happening and include the fix in the next release. In the meantime, you could work around the issue by manually adding this method back into the XML before merging this branch.

gewarren:
Bill or Joel, could you approve this PR please? I'll add the deleted method back separately.

joelmartinez (approved):
👍🏼

PR #4805, opened 2020-09-04
Update RichTextBox.xml

NetMage:
## Summary

Removed incorrect statements that method highlights found text.


xml/System.Windows.Forms/RichTextBox.xml
```diff
@@ -1109,7 +1109,7 @@
    
   
 ## Examples  
- The following code example searches the entire contents of a <xref:System.Windows.Forms.RichTextBox> for the first instance of a search string passed into the text parameter of the method. If the search string is found in the <xref:System.Windows.Forms.RichTextBox>, the method returns a value of `true` and highlights the search text, otherwise it returns `false`. The example requires that this method is placed in the class of a <xref:System.Windows.Forms.Form> that contains a <xref:System.Windows.Forms.RichTextBox> named `richTextBox1`.  
+ The following code example searches the entire contents of a <xref:System.Windows.Forms.RichTextBox> for the first instance of a search string passed into the text parameter of the method. If the search string is found in the <xref:System.Windows.Forms.RichTextBox>, the method returns a value of `true`; otherwise it returns `false`. The example requires that this method is placed in the class of a <xref:System.Windows.Forms.Form> that contains a <xref:System.Windows.Forms.RichTextBox> named `richTextBox1`.
```
gewarren, line 1112: The change is to remove "and highlights the search text".

merriemcgaw: @RussKie is there a bug in our RichTextBox we should address or do you think it's always been that way? @Tanya-Solyanik too.

merriemcgaw: @Olina-Zhang do you think you can find out if this ever worked as advertised? Do we highlight the selected text and is this a regression from an earlier Framework version? If so we should create an issue to be fixed and not update the docs. (FYI @dreddy-work )

Olina-Zhang: @merriemcgaw, verified that test code, the search text can be chosen and highlighted in .Net frameowrk versions and .Net Core 3.1, Net 5.0. The results have a little difference for search text. Please see following screenshot:
image

NetMage: What test code are you verifying? None of the example code on the page highlights anything (they are all find methods) and my tests in .Net 4.8 and .Net Core 3.1 show that RichTestBox.Find does not highlight.

NetMage: I think I understand - the text isn't what I would call highlighted, rather it is _selected_ which only shows when the RichTextBox has focus. I think that would be the right improvement in the documentation.

Olina-Zhang:
Test code is in:
https://github.com/dotnet/dotnet-api-docs/blob/master/samples/snippets/csharp/VS_Snippets_Winforms/Classic%20RichTextBox%20Example/CS/source.cs.

获取 Outlook for Android<https://aka.ms/ghei36>

RussKie:
I totally missed this. I will review in the next few days.

RussKie (commented):
The wording must use "select" (and derivative forms) and be consistent with wording in https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.richtextbox#examples section. E.g.
image

carlossanlop:
@NetMage can you please follow up on the suggestions?

PR #4842, opened 2020-09-13
Update the conditions on GetEncodings

yecril71pl:
## Summary

Explain how the list of encodings returned by `GetEncodings` depends on the version.

### DIFF
--- a/xml/System.Text/Encoding.xml
+++ b/xml/System.Text/Encoding.xml
@@ -3906,7 +3906,7 @@ In .NET 5.0 and later versions, the code page name `utf-7` is not supported.
 >
 > In .NET Framework and .NET Core, the returned list of supported encodings *does not* include these additional encodings.
 >
-> Therefore, if running on .NET Framework or running on .NET 5.0 with the code pages encoding provider registered via `Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);`, you will get a result similar to the list in the following code examples. If running on .Net Core versions, will get the basic list as in the [remark section](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding?view=netcore-3.1#remarks).
+> Therefore, if running on .NET Framework or .NET 5.0+ with the code pages encoding provider registered via `Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);`, you'll get a result similar to the list in the previous code example. If running on .NET Core versions up to version 3.x, you'll get the basic list.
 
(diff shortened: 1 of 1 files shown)


xml/System.Text/Encoding.xml
```diff
@@ -3905,6 +3905,8 @@ In .NET 5.0 and later versions, the code page name `utf-7` is not supported.
 > In .NET 5.0 and later versions, the list of supported encodings that <xref:System.Text.Encoding.GetEncodings%2A> returns includes any additional encodings made available by any <xref:System.Text.EncodingProvider> implementations that were registered by calling <xref:System.Text.Encoding.RegisterProvider%2A>.
 >
 > In .NET Framework and .NET Core, the returned list of supported encodings *does not* include these additional encodings.
+>
+> Therefore, if running on .NET Framework or running on .NET 5.0 with the code pages encoding provider registered via `Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);`, you will get a result similar to the list in the following code examples. If running on .Net Core versions, will get the basic list as in the [remark section](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding?view=netcore-3.1#remarks).
```
gewarren, line 3909:
```suggestion
> Therefore, if running on .NET Framework or .NET 5.0 with the code pages encoding provider registered via `Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);`, you will get a result similar to the list in the following code examples. If running on .NET Core versions, will get the basic list as in the [Remarks section](#remarks).
```

However, I don't understand why .NET Framework and .NET 5.0 are grouped together. Shouldn't .NET Framework and .NET Core be grouped together?

tarekgh (approved):
LGTM with @gewarren change suggestion.

xml/System.Text/Encoding.xml
jeffhandley: I interpreted it as .NET Framework and .NET 5+ have the same behavior, but .NET Core (up to 3.x) had different behavior. Is that right, @tarekgh?

tarekgh: > I interpreted it as .NET Framework and .NET 5+ have the same behavior, but .NET Core (up to 3.x) had different behavior. Is that right, @tarekgh?

Yes, that is correct.

---
In reply to: [575897830](https://github.com/dotnet/dotnet-api-docs/pull/4842#discussion_r575897830) [](ancestors = 575897830)

tarekgh (approved):

xml/System.Text/Encoding.xml
jeffhandley:
```suggestion
> Therefore, if running on .NET Framework or .NET 5.0+ with the code pages encoding provider registered via `Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);`, you'll get a result similar to the list in the previous code example. If running on .NET Core versions up to version 3.x, you'll get the basic list.
```

jeffhandley:
Thanks for the contribution, @yecril71pl!

PR #4846, opened 2020-09-14
.NET 5.0 RC1 updates

gewarren:
This doesn't seem correct to me. There isn't a single meaningful change in this PR. (The deletion of Configure() is a known bug.) I know there were API changes for RC1, for example, `MinimumOSPlatformAttribute` and `RemovedInOSPlatformAttribute` were renamed.


safern:
I'll take a look

gewarren:
Closing due to user error in the binaries.

PR #4855, opened 2020-09-15
Update TextFieldParser.xml

ericmutta:
Add the word `using` to make the following sentence correct:

> ...while **[USING]** the `ReadFields` method to extract fields of text is similar to splitting the strings.


gewarren (approved):
@ericmutta Thank you!

PR #4870, opened 2020-09-17
HTTP Version Selection

aik-jahoda:
## Summary

Documentation of https://github.com/dotnet/runtime/pull/39201

### DIFF
--- a/xml/System.Net.Http/HttpClient.xml
+++ b/xml/System.Net.Http/HttpClient.xml
@@ -542,7 +542,7 @@ The `DefaultRequestVersion` property can be changed as long as the <xref:System.
       </ReturnValue>
       <Docs>
         <summary>Gets or sets the default version policy for implicitly created requests in convenience methods, for example, <see cref="M:System.Net.Http.HttpClient.GetAsync(System.String)" /> and <see cref="M:System.Net.Http.HttpClient.PostAsync(System.String,System.Net.Http.HttpContent)" />.</summary>
-        <value>To be added.</value>
+        <value>The HttpVersionPolicy used when the HTTP connection is established.</value>
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
--- a/xml/System.Net.Http/HttpRequestMessage.xml
+++ b/xml/System.Net.Http/HttpRequestMessage.xml
@@ -570,7 +570,7 @@ The following table lists the default message version based on the version of .N
       </ReturnValue>
       <Docs>
         <summary>Gets or sets the policy that determines how <see cref="System.Net.Http.HttpRequestMessage.Version" /> is interpreted and how the final HTTP version is negotiated with the server.</summary>
-        <value>To be added.</value>
+        <value>The HttpVersionPolicy used when the HTTP connection is established.</value>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>


xml/System.Net.Http/HttpVersionPolicy.xml
```diff
@@ -13,7 +13,7 @@
     <BaseTypeName>System.Enum</BaseTypeName>
   </Base>
   <Docs>
-    <summary>To be added.</summary>
+    <summary>Determines behavior when selecting and negotiating HTTP version for a request.</summary>
```
gewarren, line 16:
```suggestion
    <summary>Specifies behaviors for selecting and negotiating the HTTP version for a request.</summary>
```

xml/System.Net.Http/HttpVersionPolicy.xml
```diff
@@ -34,7 +34,15 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Uses only the requested version. Throwing <see cref="HttpRequestException" /> if a connection with the exact version cannot be established.</summary>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+Note that this option allows to use prenegotiated clear text connection for the requested version.
```
gewarren, line 42:
```suggestion
This option allows for use of a prenegotiated clear text connection for the requested version.
```

xml/System.Net.Http/HttpVersionPolicy.xml
```diff
@@ -34,7 +34,15 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Uses only the requested version. Throwing <see cref="HttpRequestException" /> if a connection with the exact version cannot be established.</summary>
```
gewarren, line 37:
```suggestion
        <summary>Only use the requested version. Throws an <see cref="HttpRequestException" /> if a connection with the exact version cannot be established.</summary>
```

Question: Which method(s) would throw that exception? We should include that here, or a general description of those methods, because the enum member doesn't throw it.

xml/System.Net.Http/HttpVersionPolicy.xml
```diff
@@ -54,7 +62,17 @@
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Tries to uses highest available version, downgrading only to the requested version, not bellow. Throwing <see cref="HttpRequestException" /> if a connection with higher or equal version cannot be established.</summary>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+If the server supports higher than requested version, either negotiated via ALPN (H2) or advertised via Alt-Svc (H3), as well as secure connection is being requested, the result is the highest available one. Otherwise, downgrades to the <see cref="HttpRequestMessage.Version" />.
```
gewarren, line 70:
```suggestion
If the server supports a higher version than the requested version (either negotiated via ALPN (H2) or advertised via Alt-Svc (H3)) and a secure connection is requested, the result is the highest available version. Otherwise, the version downgrades to <see cref="HttpRequestMessage.Version" />.
```

xml/System.Net.Http/HttpVersionPolicy.xml
```diff
@@ -54,7 +62,17 @@
...
+
+Note that this option allows to use prenegotiated clear text connection for the requested version but not for anything higher.
```
gewarren, line 72:
```suggestion
This option allows use of a prenegotiated clear text connection for the requested version but not for a higher version.
```

xml/System.Net.Http/HttpVersionPolicy.xml
```diff
@@ -54,7 +62,17 @@
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Tries to uses highest available version, downgrading only to the requested version, not bellow. Throwing <see cref="HttpRequestException" /> if a connection with higher or equal version cannot be established.</summary>
```
gewarren, line 65:
```suggestion
        <summary>Use the highest available version, downgrading only to the requested version but not below. Throws an <see cref="HttpRequestException" /> if a connection with higher or equal version cannot be established.</summary>
```

Same comment about throwing the exception.

xml/System.Net.Http/HttpVersionPolicy.xml
```diff
@@ -74,7 +92,16 @@
       </ReturnValue>
       <MemberValue>0</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Default behavior, either uses requested version or downgrades to a lower one.</summary>
```
gewarren, line 95:
```suggestion
        <summary>Use the requested version or downgrade to a lower one. This is the default behavior.</summary>
```

xml/System.Net.Http/HttpVersionPolicy.xml
```diff
@@ -74,7 +92,16 @@
...
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+If the server supports the requested version, either negotiated via ALPN (H2) or advertised via Alt-Svc (H3), as well as a secure connection is being requested, the result is the <see cref="HttpRequestMessage.Version" />. Otherwise, downgrades to HTTP/1.1.
+
+Note that this option does not allow use of prenegotiated clear text connection, e.g. H2C.
```
gewarren, line 102:
```suggestion
This option does not allow use of a prenegotiated clear text connection, for example, H2C.
```

xml/System.Net.Http/HttpVersionPolicy.xml
```diff
@@ -74,7 +92,16 @@
       </ReturnValue>
       <MemberValue>0</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Default behavior, either uses requested version or downgrades to a lower one.</summary>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+If the server supports the requested version, either negotiated via ALPN (H2) or advertised via Alt-Svc (H3), as well as a secure connection is being requested, the result is the <see cref="HttpRequestMessage.Version" />. Otherwise, downgrades to HTTP/1.1.
```
gewarren, line 100:
```suggestion
If the server supports the requested version, either negotiated via ALPN (H2) or advertised via Alt-Svc (H3), and a secure connection is being requested, the result is the <see cref="HttpRequestMessage.Version" />. Otherwise, the version downgrades to HTTP/1.1.
```

xml/System.Net.Http/HttpClient.xml
```diff
@@ -541,9 +541,15 @@ The `DefaultRequestVersion` property can be changed as long as the <xref:System.
         <ReturnType>System.Net.Http.HttpVersionPolicy</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Gets or sets the default value of <see cref="HttpRequestMessage.VersionPolicy" /> for implicitly created requests in convenience methods, e.g.: <see cref="GetAsync(string?)" />, <see cref="PostAsync(string?, HttpContent)" />.</summary>
```
gewarren, line 544:
```suggestion
        <summary>Gets or sets the default version policy for implicitly created requests in convenience methods, for example, <see cref="M:System.Net.Http.HttpClient.GetAsync(System.String)" /> and <see cref="M:System.Net.Http.HttpClient.PostAsync(System.String,System.Net.Http.HttpContent)" />.</summary>
```

xml/System.Net.Http/HttpVersionPolicy.xml
ManickaP: `SocketsHttpHandler` send methods since the property (`HttpRequestMessage.VersionPolicy`) is interpreted there. This also influences the behavior of `HttpClient` if it's used with default handlers. OTOH, it has no effect on `WinHttpHandler` for instance.

ManickaP (approved):
Content-wise LGTM.
Thank you!

gewarren:
Oops!

xml/System.Net.Http/HttpVersionPolicy.xml
aik-jahoda: The HttpClient has a HttpRequestException exception mentioned but in general way. On other side this description is very specific for SocketsHttpHandler and this policy.

I would leave it like it is or delete it from the `HttpVersionPolicy.xml` as the reason why it can be thrown is quite obvious.

@gewarren, what is your opinion there?

xml/System.Net.Http/HttpClient.xml
```diff
@@ -541,9 +541,15 @@ The `DefaultRequestVersion` property can be changed as long as the <xref:System.
         <ReturnType>System.Net.Http.HttpVersionPolicy</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Gets or sets the default version policy for implicitly created requests in convenience methods, for example, <see cref="M:System.Net.Http.HttpClient.GetAsync(System.String)" /> and <see cref="M:System.Net.Http.HttpClient.PostAsync(System.String,System.Net.Http.HttpContent)" />.</summary>
         <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <remarks>
```
pgovind, line 546: I think `<value>` needs to be populated, but @carlossanlop should confirm

xml/System.Net.Http/HttpRequestMessage.xml
```diff
@@ -569,7 +569,7 @@ The following table lists the default message version based on the version of .N
         <ReturnType>System.Net.Http.HttpVersionPolicy</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Gets or sets the policy that determines how <see cref="System.Net.Http.HttpRequestMessage.Version" /> is interpreted and how the final HTTP version is negotiated with the server.</summary>
```
pgovind, line 572: Same here

xml/System.Net.Http/HttpVersionPolicy.xml
```diff
@@ -34,7 +34,15 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Only use the requested version. Throws an <see cref="HttpRequestException" /> if a connection with the exact version cannot be established.</summary>
```
pgovind, line 37: The summary here suggests that a possible exception is missing in the doc? If you add a `/// <exception><exception/>` comment in the src, the DocsPort tool will add the exception here correctly :)

xml/System.Net.Http/HttpVersionPolicy.xml
```diff
@@ -54,7 +62,17 @@
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>To be added.</summary>
+        <summary>Use the highest available version, downgrading only to the requested version but not below. Throws an <see cref="HttpRequestException" /> if a connection with higher or equal version cannot be established.</summary>
```
pgovind, line 65: Same here with exception

xml/System.Net.Http/HttpVersionPolicy.xml
gewarren: I think just remove that part of the sentence. I'll edit my suggestion to reflect that.

xml/System.Net.Http/HttpVersionPolicy.xml
gewarren, line 65:
```suggestion
        <summary>Use the highest available version, downgrading only to the requested version but not below.</summary>
```

xml/System.Net.Http/HttpVersionPolicy.xml
carlossanlop: This is an enum value, it's only mentioning that if the value is consumed, the API that uses it will throw.
So this is fine as it is.

xml/System.Net.Http/HttpVersionPolicy.xml
carlossanlop: Same, it's just an enum value, so nothing to do here.

xml/System.Net.Http/HttpVersionPolicy.xml
carlossanlop: This is an enum value, so the exception should not be mentioned here. The exception should be mentioned in the API that consumes this enum value:

```suggestion
        <summary>Only use the requested version.</summary>
```

carlossanlop (approved):

PR #4886, opened 2020-09-21
Add missing docs for AsnContentException.

bartonjs:
Add missing docs for AsnContentException.


carlossanlop (changes requested):
The changes look good, but we are currently merging everything to the net5-rc2 branch, not master. Can you redirect your PR to that branch, please?

PR #4887, opened 2020-09-21
Automatic port of System.Threading.ExecutionContext.Restore doc

carlossanlop:
Ported with [DocsPortingTool](https://github.com/carlossanlop/DocPortingTool).

### DIFF
--- a/xml/System.Threading/ExecutionContext.xml
+++ b/xml/System.Threading/ExecutionContext.xml
@@ -366,8 +366,7 @@
 
 ## Remarks
 
-To revert to the current execution context; capture it before Restore, and Restore it again.
-            It will not automatically be reverted unlike seealsoContents.
+To revert to the current execution context; capture it before calling Restore, and then call Restore it again.
 
           ]]></format>
         </remarks>


xml/System.Threading/ExecutionContext.xml
```diff
@@ -359,9 +359,18 @@
         <Parameter Name="executionContext" Type="System.Threading.ExecutionContext" Index="0" FrameworkAlternate="net-5.0" />
       </Parameters>
       <Docs>
-        <param name="executionContext">To be added.</param>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="executionContext">The ExecutionContext to set.</param>
+        <summary>Restores a captured execution context to on the current thread.</summary>
```
pgovind, line 363:
```suggestion
        <summary>Restores a captured execution context on to the current thread.</summary>
```

pgovind: I'm just looking at the wording. I have no idea what the method actually does!

pgovind (approved):
Wording looks good. Should get the area owner to check if the description of the methods is right!

xml/System.Threading/ExecutionContext.xml
```diff
@@ -359,9 +359,18 @@
...
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <param name="executionContext">The ExecutionContext to set.</param>
+        <summary>Restores a captured execution context on to the current thread.</summary>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+To revert to the current execution context; capture it before Restore, and Restore it again.
```
carlossanlop, line 369:
```suggestion
To revert to the current execution context; capture it before restoring, and then restore it again.
```

xml/System.Threading/ExecutionContext.xml
```diff
@@ -359,9 +359,18 @@
...
+            It will not automatically be reverted unlike seealsoContents.
```
carlossanlop, line 370:
```suggestion
            It will not automatically be reverted unlike when calling the Contents method.
```

kouvel (approved):
Thanks! There should be an exception case for null, I'll take care of it separately along with updating the XML comments.

PR #4891, opened 2020-09-21
Automatic port of Rune.CompareTo EII

carlossanlop:


xml/System.Text/Rune.xml
```diff
@@ -1694,10 +1694,20 @@ For more information, see <xref:System.Char.IsWhiteSpace%2A?displayProperty=name
         <Parameter Name="obj" Type="System.Object" Index="0" FrameworkAlternate="net-5.0" />
       </Parameters>
       <Docs>
-        <param name="obj">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <param name="other">The object to compare with the current instance.</param>
+        <summary>Compares the current instance to the specified object.</summary>
```
pgovind, line 1698: I wonder if `/// <inheritdoc>` works here

pgovind (approved):

xml/System.Text/Rune.xml
```diff
@@ -1694,10 +1694,20 @@ For more information, see <xref:System.Char.IsWhiteSpace%2A?displayProperty=name
...
+        <returns>A signed integer indicating the position of this instance in the sort order in relation to <paramref name="other" />: <br />- Less than zero: This instance precedes <paramref name="other" />.
```
pgovind, line 1699: Maybe consider re-wording this similar to https://docs.microsoft.com/en-us/dotnet/api/system.single.compareto?view=netcore-3.1#System_Single_CompareTo_System_Object_ ?

xml/System.Text/Rune.xml
carlossanlop: It probably does work, but I found a bug in the tool that made it crash when encountering a special character. So I transcribed it manually.

PR #4892, opened 2020-09-21
Add missing documentation for S.S.Cryptography

bartonjs:
Adds missing documentation for new method overloads which take a ReadOnlySpan where the existing method took `byte[]` or `string`.

In general, the docs were just copied from the existing method. In some cases I felt compelled to fix up the existing documentation to better match (my interpretation of) the docs style guidelines.


carlossanlop:
Can you please fetch the upstream/net5-rc2 branch, check it out locally, then put your Security commits on top of that?

carlossanlop (approved):
LGTM!

PR #4898, opened 2020-09-22
Document Type.IsAssignableTo

carlossanlop:
This API was merged in this PR without triple slash comments: https://github.com/dotnet/runtime/pull/40326

Here is the API proposal: https://github.com/dotnet/runtime/issues/30714

Since this API is doing the exact same as `Type.IsAssignableFrom`, but in the opposite direction, I decided to use the same documentation from that API, but switch mentions of the parameter with the current instance.


pgovind (approved):
LGTM!

PR #4908, opened 2020-09-24
Rc2 doc comments

carlossanlop:
## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Net.Http/ByteArrayContent.xml
+++ b/xml/System.Net.Http/ByteArrayContent.xml
@@ -159,14 +159,7 @@
         <param name="cancellationToken">The cancellation token to cancel the operation.</param>
         <summary>Creates an HTTP content stream for reading. It uses the memory from the <see cref="T:System.Net.Http.ByteArrayContent" /> as a backing store.</summary>
         <returns>The HTTP content stream.</returns>
-        <remarks>
-          <format type="text/markdown"><![CDATA[  
-  
-## Remarks  
-This operation blocks until all of the content stream has been created.  
-  
- ]]></format>
-        </remarks>
+        <remarks>To be added.</remarks>
       </Docs>
     </Member>
     <Member MemberName="CreateContentReadStreamAsync">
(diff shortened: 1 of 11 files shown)


xml/System.Globalization/CompareInfo.xml
```diff
@@ -3097,14 +3107,24 @@ Use <xref:System.Globalization.CompareInfo.GetSortKeyLength(System.ReadOnlySpan{
         <Parameter Name="matchLength" Type="System.Int32" RefType="out" Index="3" FrameworkAlternate="net-5.0" />
       </Parameters>
       <Docs>
-        <param name="source">To be added.</param>
-        <param name="suffix">To be added.</param>
```
gewarren, line 3101: Here's the weird line where the latest CI job showed a `suffix` parameter.

gewarren (approved):

xml/System.Globalization/CompareInfo.xml
carlossanlop, line 3101: I see the `suffix` -> `prefix` change you mentioned, @gewarren. Thanks for fixing it.

gewarren: @joelmartinez We're a bit confused why mdoc didn't realize the `suffix` parameter was renamed to `prefix`, and instead just added `prefix` as an additional parameter. The deletion of `suffix` here is something I did manually.

joelmartinez: @gewarren mdoc will remake the `Parameter` elements because those come directly from the metadata so it remakes them every time ... but in general it's a bit more tentative with removing elements in the `Docs` node, because those can contain user-written content, so in cases like this it's best just to adjust manually as you've done here

gewarren: Would that even have shown up in the rendered page if I hadn't removed it?

joelmartinez: @gewarren no I believe it wouldn't, because the rendering is driven by the Parameter entries ... it also shouldn't affect the intellisense, because it would just kind of be ignored, so even if we leave it it wouldn't be world-ending :)

PR #4920, opened 2020-09-30
correct default info for CertificateRevocationCheckMode

wfurt:
The documentation incorrectly claims default revocation mode for Client and Server Options is online. 
This change puts it in sync with reality. (+one small spelling fix)

### DIFF
--- a/xml/System.Net.Security/SslClientAuthenticationOptions.xml
+++ b/xml/System.Net.Security/SslClientAuthenticationOptions.xml
@@ -137,7 +137,7 @@
       </ReturnValue>
       <Docs>
         <summary>Gets or sets the certificate revocation mode for certificate validation.</summary>
-        <value>One of the values in <see cref="T:System.Security.Cryptography.X509Certificates.X509RevocationMode" />: <see langword="NoCheck" />, <see langword="Online" />, or <see langword="NoCheck" />. The default is <see langword="Online" />.</value>
+        <value>One of the values in <see cref="T:System.Security.Cryptography.X509Certificates.X509RevocationMode" />. The default is <see langword="NoCheck" />.</value>
         <remarks>
           <format type="text/markdown"><![CDATA[  
   


xml/System.Net.Security/SslClientAuthenticationOptions.xml
```diff
@@ -137,14 +137,14 @@
       </ReturnValue>
       <Docs>
         <summary>Gets or sets the certificate revocation mode for certificate validation.</summary>
-        <value>One of the values in <see cref="T:System.Security.Cryptography.X509Certificates.X509RevocationMode" />: <see langword="NoCheck" />, <see langword="Online" />, or <see langword="Offline" />. The default is <see langword="Online" />.</value>
+        <value>One of the values in <see cref="T:System.Security.Cryptography.X509Certificates.X509RevocationMode" />: <see langword="NoCheck" />, <see langword="Online" />, or <see langword="NoCheck" />. The default is <see langword="Online" />.</value>
```
bartonjs, line 140: I feel like you meant to edit the second Online, but you edited the first... making it twice as wrong as it used to be.  I'd, personally, drop enumerating the enumeration:

```suggestion
        <value>One of the values in <see cref="T:System.Security.Cryptography.X509Certificates.X509RevocationMode" />. The default is <see langword="NoCheck" />.</value>
```

(Though NoCheck should probably be cref instead of langword):

```suggestion
        <value>One of the values in <see cref="T:System.Security.Cryptography.X509Certificates.X509RevocationMode" />. The default is <see cref="F:System.Security.Cryptography.X509Certificates.X509RevocationMode.NoCheck" />.</value>
```

wfurt: yes, you are right. I edited the raw xml and misted that as the preview pages were not available at that time.

jeffhandley (approved):

PR #4927, opened 2020-10-01
Automatic port of Microsoft.Extensions.Diagnostic* docs

carlossanlop:
Area owners of Microsoft.Extensions.Diagnostic*: @maryamariyan @ericstj 
Note: The area owners page does not specify the label `area-Extensions-DiagnosticAdapter`.

### DIFF
--- a/xml/Microsoft.Extensions.DiagnosticAdapter.Internal/ProxyBase.xml
+++ b/xml/Microsoft.Extensions.DiagnosticAdapter.Internal/ProxyBase.xml
@@ -112,8 +112,8 @@
       <Parameters />
       <Docs>
         <typeparam name="T">To be added.</typeparam>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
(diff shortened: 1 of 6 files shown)


ericstj:
Actually this is owned by @Pilchie's team and is no longer part of `net5.0`.  The source code resides in https://github.com/dotnet/extensions/tree/release/3.1/src/DiagnosticAdapter

ericstj:
Correction here.  Title is actually wrong.  The only problem reported is from Microsoft.Extensions.Http.Polly.  This appears to be a complete duplicate of https://github.com/dotnet/dotnet-api-docs/pull/4932.

ericstj:
Apologies, I misunderstood here.  I see this is adding docs and I shouldn't be focused on the validation status

ericstj:
So as I mentioned above we didn't even touch this assembly in .NET 5.0, it only builds in 3.1.  There was no change to this component's docs, but some interfaces that it implements got better docs which is what seems to have caused this change.

ericstj (approved):
This looks to just be copying interface docs to undocumented members that implement that interface.

PR #4936, opened 2020-10-02
Minor fix to description: only one returns a value

fazl:
## Summary

Describe your changes here.


mconnew:
The original text was correct. 2 way operations have a reply SOAP message even when the return type is void. The reply message isn't sent until the operation has completed on the server side. If the operation throws an exception, you will get SOAP Fault. A one-way operation only returns a transport level acknowledgement once it's verified the incoming message is correct (all headers understood etc.). For HTTP that's a 200 response with no content body. For NetTcp it's a single byte ACK. If the service throws an exception, the client is never informed.

PR #4940, opened 2020-10-02
Activity.Tags list order remarks

gewarren:
Contributes to dotnet/docs#20776.
Also corrects some other properties.


tarekgh (approved):
Thanks @gewarren, LGTM.

PR #4943, opened 2020-10-03
Fixing typo

leobasilio:
The exception is thrown if `IsValid` returns false, not `Validate`.


gewarren (approved):
@leobasilio Thank you!

PR #4952, opened 2020-10-06
Add missing Polly APIs

BillWagner:
The xrefs to the Polly classes and interfaces are generating build warnings and improper rendering. 

This should fix it.

### DIFF
--- a/_zip/missingapi.yml
+++ b/_zip/missingapi.yml
@@ -33,7 +33,7 @@ references:
   name: IPolicyRegistry<TKey>
   fullname: Polly.Registry.IPolicyRegistry<TKey>
   href: https://github.com/App-vNext/Polly/wiki
-- uid: Polly.Registry.IReadOnlyPolicyRegistry%`1
+- uid: Polly.Registry.IReadOnlyPolicyRegistry`1
   name: IReadOnlyPolicyRegistry<TKey>
   fullname: Polly.Registry.IReadOnlyPolicyRegistry<TKey>
   href: https://github.com/App-vNext/Polly/wiki


BillWagner:
@joelmartinez Can you take a look? I think I put in the correct syntax for the UIDs and such.

Thanks.

_zip/missingapi.yml
```diff
@@ -25,6 +25,22 @@ references:
...
+- uid: Polly.Registry.PolicyRegistry
+  name: PolicyRegistry
+  fullname: Polly.Registry.PolicyRegistry
+  href: https://github.com/App-vNext/Polly/wiki
+- uid: Polly.Registry.IPolicyRegistry%601
+  name: IPolicyRegistry<TKey>
+  fullname: Polly.Registry.IPolicyRegistry<TKey>
+  href: https://github.com/App-vNext/Polly/wiki
+- uid: Polly.Registry.IReadOnlyPolicyRegistry%601
+  name: IReadOnlyPolicyRegistry<TKey>
```
gewarren, line 37: This is identical to line 41 AFAICT. Is it supposed to be ReadOnlyPolicyRegistry<TKey>?

joelmartinez:
@BillWagner I **think** you don't need to querystring encode that backtick in the UIDs. There's no other examples of generic types in that file unfortunately, but taking a look at the docid of [another type](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Action%601.xml#L4), we see that it's not encoded, so that's probably what OPS is trying to match against.

```xml
<TypeSignature Language="DocId" Value="T:System.Action`1" />
```

_zip/missingapi.yml
BillWagner: Thanks for catching. It shouldn't not have been added. There were only 3 types generating warnings.

_zip/missingapi.yml
```diff
@@ -29,15 +29,11 @@ references:
...
+- uid: Polly.Registry.IPolicyRegistry`1
   name: IPolicyRegistry<TKey>
   fullname: Polly.Registry.IPolicyRegistry<TKey>
   href: https://github.com/App-vNext/Polly/wiki
-- uid: Polly.Registry.IReadOnlyPolicyRegistry%601
-  name: IReadOnlyPolicyRegistry<TKey>
-  fullname: Polly.Registry.IReadOnlyPolicyRegistry<TKey>
-  href: https://github.com/App-vNext/Polly/wiki
-- uid: Polly.Registry.IReadOnlyPolicyRegistry%601
+- uid: Polly.Registry.IReadOnlyPolicyRegistry%`1
```
joelmartinez, line 36: Looks like there's still an extra `%` here :)

BillWagner: Doh!

joelmartinez: Looks like that did the trick ... there's two more that are still showing missing xref warnings. If you add them, this should get wrapped up :)

BillWagner:
@gewarren I think this is ready to merge. Can you give it a final review?

gewarren (approved):

PR #4953, opened 2020-10-06
Exception changes for Deserialize<char>

gewarren:
Contributes to https://github.com/dotnet/docs/issues/20815.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserialize?view=net-5.0&branch=pr-en-us-4953#System_Text_Json_JsonSerializer_Deserialize__1_System_String_System_Text_Json_JsonSerializerOptions_).

### DIFF
--- a/xml/System.Text.Json/JsonSerializer.xml
+++ b/xml/System.Text.Json/JsonSerializer.xml
@@ -293,7 +293,7 @@ There is remaining data in the span beyond a single JSON value.</exception>
         <typeparam name="TValue">The target type of the JSON value.</typeparam>
         <param name="json">The JSON text to parse.</param>
         <param name="options">Options to control the behavior during parsing.</param>
-        <summary>Parses the text representing a single JSON value into an instance of the type specified by <typeparamref name="TValue" />.</summary>
+        <summary>Parses the text representing a single JSON value into an instance of the type specified by a generic type parameter.</summary>
         <returns>A <typeparamref name="TValue" /> representation of the JSON value.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[
@@ -307,10 +307,9 @@ For more information, see [How to serialize and deserialize JSON](/dotnet/standa
 ]]></format>
         </remarks>
         <exception cref="T:System.IndexOutOfRangeException">
-          .NET Core 3.0 and earlier versions only: <typeparamref name="TValue" /> is <see cref="T:System.Char" /> and <paramref name="json" /> is an empty string.</exception>
+          .NET Core 3.0 only: <typeparamref name="TValue" /> is <see cref="T:System.Char" /> and <paramref name="json" /> is an empty string.</exception>
         <exception cref="T:System.ArgumentNullException">
-          <paramref name="json" /> is <see langword="null" /> (for example, <code>JsonSerializer.Deserialize&lt;char&gt;((string)null);
-</code>).</exception>
+          <paramref name="json" /> is <see langword="null" />.</exception>
         <exception cref="T:System.Text.Json.JsonException">The JSON is invalid.
 
 -or-


xml/System.Text.Json/JsonSerializer.xml
```diff
@@ -293,7 +293,7 @@ There is remaining data in the span beyond a single JSON value.</exception>
         <typeparam name="TValue">The target type of the JSON value.</typeparam>
         <param name="json">The JSON text to parse.</param>
         <param name="options">Options to control the behavior during parsing.</param>
-        <summary>Parses the text representing a single JSON value into an instance of the type specified by a generic type parameter.</summary>
+        <summary>Parses the text representing a single JSON value into an instance of the type specified by <typeparamref name="TValue" />.</summary>
```
buyaa-n, line 296: All other APIs with generic type parameter `TValue` used the old phrase `a generic type parameter`, i guess they better to stay the same
image

xml/System.Text.Json/JsonSerializer.xml
```diff
@@ -306,8 +306,11 @@ For more information, see [How to serialize and deserialize JSON](/dotnet/standa
 
 ]]></format>
         </remarks>
+        <exception cref="T:System.IndexOutOfRangeException">
+          .NET Core 3.0 and earlier versions only: <typeparamref name="TValue" /> is <see cref="T:System.Char" /> and <paramref name="json" /> is an empty string.</exception>
         <exception cref="T:System.ArgumentNullException">
-          <paramref name="json" /> is <see langword="null" />.</exception>
+          <paramref name="json" /> is <see langword="null" /> (for example, <code>JsonSerializer.Deserialize&lt;char&gt;((string)null);
+</code>).</exception>
```
buyaa-n, line 313: Somehow this row is not rendered correctly, anyway not sure if we need this example here as the exception apply to any type parameter not only  `char`

xml/System.Text.Json/JsonSerializer.xml
```diff
@@ -306,8 +306,11 @@ For more information, see [How to serialize and deserialize JSON](/dotnet/standa
 
 ]]></format>
         </remarks>
+        <exception cref="T:System.IndexOutOfRangeException">
+          .NET Core 3.0 and earlier versions only: <typeparamref name="TValue" /> is <see cref="T:System.Char" /> and <paramref name="json" /> is an empty string.</exception>
```
buyaa-n, line 310: i believe System.Text.Json is introduced in 3.0, no earlier versions
```suggestion
          .NET Core 3.0 only: <typeparamref name="TValue" /> is <see cref="T:System.Char" /> and <paramref name="json" /> is an empty string.</exception>
```

buyaa-n (approved):
LGTM, thanks!

xml/System.Text.Json/JsonSerializer.xml
ahsonkhan: True, it is built-in starting .NET Core 3.0 and above, but the user can grab the NuGet package which targets netstandard2.0, so you can run it on .NET Core 2.0, 2.1, and .NET 4.7.2 (or above).
https://www.nuget.org/packages/System.Text.Json/5.0.0-rc.1.20451.14

gewarren: Thanks, I'll update it.

ahsonkhan: That said, to clarify, if the user grabs the latest version of the NuGet package, they will get the correct/new behavior. So, what you have is good as-is.

If someone uses the older NuGet package, say [4.6.0](https://www.nuget.org/packages/System.Text.Json/4.6.0), that will have the IOOR behavior. That version shipped at the same time as the .NET Core 3.0 release.

Anything newer than that will have the right behavior.

xml/System.Text.Json/JsonSerializer.xml
```diff
@@ -316,7 +318,11 @@ For more information, see [How to serialize and deserialize JSON](/dotnet/standa
 
 -or-
 
-There is remaining data in the string beyond a single JSON value.</exception>
+There is remaining data in the string beyond a single JSON value.
+
+-or-
+
+.NET Core 3.1 and later versions only: <typeparamref name="TValue" /> is <see cref="T:System.Char" /> and <paramref name="json" /> is a null reference (<code>JsonSerializer.Deserialize&lt;char&gt;("null");</code>) or an empty string.</exception>
```
ahsonkhan, line 325: This is unnecessary info to the user, and isn't accurate. Please revert this part of the change.

PR #4957, opened 2020-10-08
Update JsonSerializer exception

gewarren:


xml/System.Text.Json/JsonSerializer.xml
```diff
@@ -307,7 +307,7 @@ For more information, see [How to serialize and deserialize JSON](/dotnet/standa
 ]]></format>
         </remarks>
         <exception cref="T:System.IndexOutOfRangeException">
-          .NET Core 3.0 only: <typeparamref name="TValue" /> is <see cref="T:System.Char" /> and <paramref name="json" /> is an empty string.</exception>
+          .NET Core 3.0 and earlier versions only: <typeparamref name="TValue" /> is <see cref="T:System.Char" /> and <paramref name="json" /> is an empty string.</exception>
```
ahsonkhan, line 310:
```suggestion
          Only .NET Core 3.0 and older stand-alone package version (4.6.0): <typeparamref name="TValue" /> is <see cref="T:System.Char" /> and <paramref name="json" /> is an empty string.</exception>
```

xml/System.Text.Json/JsonSerializer.xml
```diff
@@ -307,7 +307,7 @@ For more information, see [How to serialize and deserialize JSON](/dotnet/standa
 ]]></format>
         </remarks>
         <exception cref="T:System.IndexOutOfRangeException">
```
ahsonkhan, line 309: Looking at the whole docs experience again, I don' think these version specific exceptions should be in this xml file.

We don't need (nor want) folks to see the Deserialize method can throw IOOR, **especially not on 3.1 and 5.0 (and beyond)**
That was a point-in-time bug in the code for a super uncommon edge case that folks shouldn't run into anyway.

This exception isn't something we want users to deal with, especially going forward.

**Please remove this exception all together.**

The docs are already versioned. A reader will ask, why am I seeing exceptions for the version I am not even using:
image

ahsonkhan (changes requested):
Remove the IOOR exception from the docs all together. We shouldn't encode a point-in-time, version specific, buggy behavior into the public facing docs.

ahsonkhan:
Same with the other note:
https://github.com/dotnet/dotnet-api-docs/pull/4953#discussion_r501964734

xml/System.Text.Json/JsonSerializer.xml
gewarren: Unfortunately the exceptions and remarks aren't versioned, only the API signature. So this is the way we document version-specific information. But I will remove it.

ahsonkhan:
I was going to ask, for a clean change, if we could just revert the whole change from https://github.com/dotnet/dotnet-api-docs/pull/4953

```bash
git revert <commit hash from previous PR>
```

But it looks like we don't mandate using `Squash and merge` for the docs repo to make sure there is only a single commit for each PR. @mairaw, @tdykstra - what's the reason for that? Almost all the other dev repos I have worked in, we almost always "Squash and merge", so much so, that we have disabled doing anything else by default.

https://github.blog/2016-04-01-squash-your-commits/#whats-changing

ahsonkhan (approved):

mairaw:
@ahsonkhan yes I agree that squash and merge is the best approach. The reason why we can’t disable the regular merge is because when we merge to live, we can’t squash the commits.

ahsonkhan:
> because when we merge to live, we can’t squash the commits.

And why is that?

mairaw:
Because we need to keep the commits intact when we merge one branch to the other, otherwise the branches don’t match.

ahsonkhan:
If the goal is to keep the master and live branches in-sync, would it be possible to automate that through per-commit "sync" PRs, or even go around PRs and port the changes directly to the branch?

PR #4962, opened 2020-10-08
Update Interlocked.xml

damageboy:
The documentation re. `Interlocked.Read` for 64-bit reads sized reads on 64-bit systems is simply not correct.
There is no known architecture in existence where a 64-bit value that spans across two cache lines can be read without the use of atomic operations.

For example, consult [Intel's Software Development Manual, Volume 3A](https://software.intel.com/content/dam/develop/public/us/en/documents/253668-sdm-vol-3a.pdf) which clearly states on section 8.1.1:

 > Accesses to cacheable memory that are split across cache lines and page boundaries are not guaranteed to be
atomic by the Intel Core 2 Duo, Intel® Atom™, Intel Core Duo, Pentium M, Pentium 4, Intel Xeon, P6 family,
Pentium, and Intel486 processors. The Intel Core 2 Duo, Intel Atom, Intel Core Duo, Pentium M, Pentium 4, Intel
Xeon, and P6 family processors provide bus control signals that permit external memory subsystems to make split
accesses atomic; however, nonaligned data accesses will seriously impact the performance of the processor and
should be avoided

Atomic operations are required for such cases, and those cases (split cache-line access) while uncommon in .NET are not impossible.

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Threading/Interlocked.xml
+++ b/xml/System.Threading/Interlocked.xml
@@ -2221,7 +2221,7 @@ This method wraps a call to [FlushProcessWriteBuffers](/windows/win32/api/proces
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Threading.Interlocked.Read%2A> method is necessary on 64-bit systems when the location refers to a 64-bit value that may span across cachlines. On 32-bit systems, 64-bit read operations are not atomic unless performed using <xref:System.Threading.Interlocked.Read%2A>.  
+ The <xref:System.Threading.Interlocked.Read%2A> method is only necessary on 64-bit systems when the location refers to a 64-bit value that may span across cache lines. On 32-bit systems, 64-bit read operations are not atomic unless performed using <xref:System.Threading.Interlocked.Read%2A>.  
   
  ]]></format>
         </remarks>


xml/System.Threading/Interlocked.xml
```diff
@@ -2221,7 +2221,7 @@ This method wraps a call to [FlushProcessWriteBuffers](/windows/win32/api/proces
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Threading.Interlocked.Read%2A> method is unnecessary on 64-bit systems, because 64-bit read operations are already atomic. On 32-bit systems, 64-bit read operations are not atomic unless performed using <xref:System.Threading.Interlocked.Read%2A>.  
+ The <xref:System.Threading.Interlocked.Read%2A> method is necessary on 64-bit systems when the location refers to a 64-bit value that may span across cachlines. On 32-bit systems, 64-bit read operations are not atomic unless performed using <xref:System.Threading.Interlocked.Read%2A>.
```
roji, line 2224:
```suggestion
 The <xref:System.Threading.Interlocked.Read%2A> method is only necessary on 64-bit systems when the location refers to a 64-bit value that may span across cache lines. On 32-bit systems, 64-bit read operations are not atomic unless performed using <xref:System.Threading.Interlocked.Read%2A>.  
```

gfoidl:
> while uncommon in .NET are not impossible.

E.g. with such a struct?
```c#
[StructLayout(LayoutKind.Explicit)]
public struct Bar
{
    [FieldOffset(0)]
    public byte A;

    [FieldOffset(1)]
    public long B;

    [FieldOffset(9)]
    public byte C;

    [FieldOffset(10)]
    public long D;
}
```
`B` and `D` aren't aligned to a 8 byte boundary anymore, the size of the struct is 24 bytes so it may cross cache lines.

xml/System.Threading/Interlocked.xml
gfoidl: Maybe there should also be a sentence about when the 64-bit value may span across cache line.
Or the other way around: Fields layed out by the CLI are aligned to 8 byte-boundaries, so this won't happen. Only structs with explicit layout (or there any other cases expect _unsafe_ code?) fall under this.
Right now I don't have any proper wording in mind for this.

Let's think of a ordinary C# developer which has first contacts with Interlocked, so there might be more confusion after reading the docs than before.
This topic isn't an easy one and the docs can't cover all aspect, but they should provide good enough guidance to get things done properly.

damageboy:
> > while uncommon in .NET are not impossible.
> 
> E.g. with such a struct?
> 
> ```cs
> [StructLayout(LayoutKind.Explicit)]
> public struct Bar
> {
>     [FieldOffset(0)]
>     public byte A;
> 
>     [FieldOffset(1)]
>     public long B;
> 
>     [FieldOffset(9)]
>     public byte C;
> 
>     [FieldOffset(10)]
>     public long D;
> }
> ```
> 
> `B` and `D` aren't aligned to a 8 byte boundary anymore, the size of the struct is 24 bytes so it may cross cache lines.

Yeap. This used to be the more common way of getting to this sort of situation, but with the advent of .NET Span you have new ways of getting to this sort of situation.

kouvel:
I believe interlocked operations on 64-bit memory locations have a requirement for the memory location to be aligned to 8 bytes, so they cannot cross a cache line boundary. I don't think there is a way to do atomic operations on unaligned 64-bit memory locations.

damageboy:
> I believe interlocked operations on 64-bit memory locations have a requirement for the memory location to be aligned to 8 bytes, so they cannot cross a cache line boundary. I don't think there is a way to do atomic operations on unaligned 64-bit memory locations.

None of the above is true for Intel / x86 architectures:

1. 8 byte and cache-line alignment are unrelated (although I do understand you intention that some non-aligned 8-byte addresses will also be un-aligned to a 64 bytes cache line)
2. Cross-cache-line atomic are allowed, albeit at a significant performance drop
3. Intra cache-line misaligned atomics are allowed and carry no performance degradation.

Reference Intel docs / text explaining (1)-(3) was linked in the original comment at the top of this issue.

kouvel:
True, but that doesn't seem to be the case for ARM architectures, where load/store-exclusive instructions seem to require alignment in many if not all cases, with perhaps a stricter requirement on the size of the locked region (within which unaligned accesses may be allowed) than the size of a cache line. The interlocked read prevents tearing on 32-bit systems (similarly to `Volatile.Read(ref Int64)`, which would also have the same alignment constraints). The operation offers stronger memory ordering than both a basic aligned read and `Volatile.Read(ref Int64)` even on 64-bit systems. Instead of suggesting when the API is not necessary I think it would be more useful to say something like:

- Some systems may require the memory location to be aligned to 8 bytes, so for compatibility the method should not be used on references that may not be aligned as such.

Maybe some more guidance would be useful, suggestions?

damageboy:
That's something different then... :+1: 

All I'm saying is that this PR, specifically, corrects a false statement:
> ...method is unnecessary on 64-bit systems, because 64-bit

Into a correct one:
> ...method is only necessary on 64-bit systems when the location refers...

I think that the issue of alignment is more of a conceptual issue that is probably best addressed in general in some conceptual-level documentation of the `Interlocked` class...


Maybe something such as a "See Also" link from this page pointing to a dedicated page that discusses alignment, can even (gasp!) link to the actual relevant documentation per supported architecture in .NET (e.g. x86 + arm) and give the proper context...

Makes sense?

kouvel:
Yea maybe the general guidance should be in the `Interlocked` class docs with links to it from method docs and maybe in `Volatile` class docs as well. For the purpose of correcting the incorrect info, that sounds fine to me, but I suggest changing `method is only necessary ... when` to `method is necessary ... when`, as the `only` seems to imply that that's the only case where it may be necessary.

jeffhandley:
Hi @damageboy; I think there's [a piece of feedback](https://github.com/dotnet/dotnet-api-docs/pull/4962#issuecomment-741918320) left to address here. Would you like to update the PR for approval?

Thanks!

PR #4971, opened 2020-10-12
Update List`1.xml

keenam:
## Summary

Updated big O notation for List constructor with a capacity parameter. The O notation is in sync with the default constructor. Setting the capacity doesn't copy items hence there's no operation involved unlike the other List constructor with items passed in.


layomia (approved):

layomia:
Thanks!

PR #4642, opened 2020-08-11
Update System.Net.WebHeaderCollection docs

CarnaViire:
## Summary

Explained behavior of conditional max header length check in System.Net.WebHeaderCollection for different .NET versions.

The check is removed from .NET 5 here https://github.com/dotnet/runtime/pull/40529

Fixes dotnet/runtime#1486

PR #4972, opened 2020-10-12
Fix WebHeaderCollection header restrictions docs

CarnaViire:
Fix WebHeaderCollection header restrictions docs according to its implementation.

Header restrictions are not checked since WebHeaderCollection was transferred to .NET Core.

### DIFF
--- a/xml/System.Net/WebHeaderCollection.xml
+++ b/xml/System.Net/WebHeaderCollection.xml
@@ -84,7 +84,7 @@
 ## Remarks  
  The <xref:System.Net.WebHeaderCollection> class is generally accessed through <xref:System.Net.WebRequest.Headers%2A?displayProperty=nameWithType> or <xref:System.Net.WebResponse.Headers%2A?displayProperty=nameWithType>.
 
- On .NET Framework, some common headers are considered restricted and are either exposed directly by the API (such as `Content-Type`) or protected by the system and cannot be changed. This does *not* apply to .NET Core, .NET 5.0 and later versions - none of the headers are restricted there.
+ On .NET Framework, some common headers are considered restricted and are either exposed directly by the API (such as `Content-Type`) or protected by the system and cannot be changed. This does *not* apply to .NET Core and .NET 5+ - none of the headers are restricted there.
 
  The restricted headers are:  
   


xml/System.Net/WebHeaderCollection.xml
```diff
@@ -82,7 +82,7 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Net.WebHeaderCollection> class is generally accessed through <xref:System.Net.WebRequest.Headers%2A?displayProperty=nameWithType> or <xref:System.Net.WebResponse.Headers%2A?displayProperty=nameWithType>. Some common headers are considered restricted and are either exposed directly by the API (such as `Content-Type`) or protected by the system and cannot be changed.  
+ The <xref:System.Net.WebHeaderCollection> class is generally accessed through <xref:System.Net.WebRequest.Headers%2A?displayProperty=nameWithType> or <xref:System.Net.WebResponse.Headers%2A?displayProperty=nameWithType>. On .NET Framework, some common headers are considered restricted and are either exposed directly by the API (such as `Content-Type`) or protected by the system and cannot be changed. This does *not* apply to .NET Core versions through version 3.1, .NET 5.0 and later versions.
```
gewarren, line 85:
```suggestion
 The <xref:System.Net.WebHeaderCollection> class is generally accessed through <xref:System.Net.WebRequest.Headers%2A?displayProperty=nameWithType> or <xref:System.Net.WebResponse.Headers%2A?displayProperty=nameWithType>.
 
 On .NET Framework, some common headers are considered restricted and are either exposed directly by the API (such as `Content-Type`) or protected by the system and cannot be changed. This does *not* apply to .NET Core and .NET 5.0 and later versions.
```

gewarren (approved):

xml/System.Net/WebHeaderCollection.xml
```diff
@@ -82,8 +82,10 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Net.WebHeaderCollection> class is generally accessed through <xref:System.Net.WebRequest.Headers%2A?displayProperty=nameWithType> or <xref:System.Net.WebResponse.Headers%2A?displayProperty=nameWithType>. Some common headers are considered restricted and are either exposed directly by the API (such as `Content-Type`) or protected by the system and cannot be changed.  
-  
+ The <xref:System.Net.WebHeaderCollection> class is generally accessed through <xref:System.Net.WebRequest.Headers%2A?displayProperty=nameWithType> or <xref:System.Net.WebResponse.Headers%2A?displayProperty=nameWithType>.
+
+ On .NET Framework, some common headers are considered restricted and are either exposed directly by the API (such as `Content-Type`) or protected by the system and cannot be changed. This does *not* apply to .NET Core versions through version 3.1, .NET 5.0 and later versions.
```
karelz, line 87:
```suggestion
 On .NET Framework, some common headers are considered restricted and are either exposed directly by the API (such as `Content-Type`) or protected by the system and cannot be changed. This does *not* apply to .NET Core - none of the headers are restricted there.
```

karelz: I'd leave out the .NET Core versions IMO -- it is missing from the time the APIs were ported to .NET Core -- whichever the version was.

karelz (changes requested):

xml/System.Net/WebHeaderCollection.xml
CarnaViire: I followed the pattern from when I described max header length check behavior change here https://github.com/dotnet/dotnet-api-docs/pull/4642 but I see now that current situation is a bit different (for max length check, it also changed since 5.0, that's why .NET Core version was explicitly specified) -- but I'd still leave ".NET Core, .NET 5.0 and later versions" to avoid confusion

karelz (approved):
Feels weird that .NET has version (5.0) while .NET Core does not (maybe we should leave out both versions?), but if @gewarren agrees that it follows established patterns, I am fine too.

gewarren:
> Feels weird that .NET has version (5.0) while .NET Core does not (maybe we should leave out both versions?), but if @gewarren agrees that it follows established patterns, I am fine too.

It's tricky because .NET 5 is technically .NET Core. We could change it to "This does *not* apply to .NET Core and .NET 5+".

gewarren (approved):

PR #4982, opened 2020-10-14
Remove comment from DataTable.Rows returning null

vcsjones:
## Summary

`DataTable.Rows` contained a remark about returning null if there are no rows. Best of my knowledge, this property will never return null on .NET Core / 5 or in .NET Framework.


David-Engel:
It appears that way to me, too. @roji would be a good authoritative reviewer for this, though.

roji (approved):
LGTM, DataTable.Rows doesn't seem like it can ever return null, and has been annotated accordingly for C# NRTs in .NET 5.0.

PR #4989, opened 2020-10-14
Update .ghal.rules.json to remove area labels for issues

adegeo:


carlossanlop (approved):
Thank you for making this change, @adegeo . Since the 3 involved labels got renamed (NCL, WinForms, Security), this config file would not be working correctly anymore, so it makes sense to remove that code.

LGTM.

maryamariyan (approved):

PR #4992, opened 2020-10-15
CBOR and platform check API updates

gewarren:
Spell out CBOR acronym on first use.
Update platform check method descriptions.

(Hide whitespace changes in diff.)

### DIFF
--- a/xml/System.Formats.Cbor/CborContentException.xml
+++ b/xml/System.Formats.Cbor/CborContentException.xml
@@ -20,7 +20,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>The exception that is thrown when Concise Binary Object Representation (CBOR) content is invalid.</summary>
+    <summary>The exception that's thrown when CBOR data is invalid.</summary>
     <remarks>To be added.</remarks>
   </Docs>
   <Members>
--- a/xml/ns-System.Formats.Cbor.xml
+++ b/xml/ns-System.Formats.Cbor.xml
@@ -1,6 +1,6 @@
 <Namespace Name="System.Formats.Cbor">
   <Docs>
-    <summary>To be added.</summary>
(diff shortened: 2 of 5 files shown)


buyaa-n (approved):
OperatingSystem type-related changes looks good, thanks

xml/System.Formats.Cbor/CborContentException.xml
```diff
@@ -20,7 +20,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>To be added.</summary>
+    <summary>The exception that is thrown when Concise Binary Object Representation (CBOR) content is invalid.</summary>
```
eiriktsarpalis, line 23: Do we need to do this for every type? It feels redundant since the namespace also contains the acronym. I would suggest expanding it only for the CborWriter and CborReader classes, since they are the entrypoints to this particular API.

eiriktsarpalis (approved):
Other than one comment, CBOR changes LGTM

xml/ns-System.Formats.Cbor.xml
```diff
@@ -1,6 +1,6 @@
 <Namespace Name="System.Formats.Cbor">
   <Docs>
-    <summary>To be added.</summary>
+    <summary>Contains types used in reading and writing data in the Concise Binary Object Representation (CBOR) format.</summary>
```
gewarren, line 3: @eiriktsarpalis Thanks for reviewing. I removed the spelled out acronym from all but reader and writer, but I also added a namespace summary here. Does this look okay?

gewarren: Here's the preview: https://review.docs.microsoft.com/en-us/dotnet/api/system.formats.cbor?view=dotnet-plat-ext-5.0&branch=pr-en-us-4992

eiriktsarpalis: That's perfect, thanks!

PR #4939, opened 2020-10-02
Updating MSBuild reference

v-alje:
Update the MSBuild reference from the dlls and XML in their NuGet packages.

CEAPEX PROJECT 314046 MSBuild 16.7- .NET

PR #4994, opened 2020-10-15
MSBuild: remove unresolved refs

ghogen:
## Summary

Describe your changes here.


ghogen:
I didn't resolve the CircularDependencyException reference, because that appears to be a substantive record of an exception that could be thrown, but it's in Microsoft.Build.Backend, which is not published.  So, asked @rainersigwald about how to resolve this.

ghogen:
#sign-off

ghogen:
@v-alje Can you merge this? I resolved the question about CircularDependencyException, so we're good to go with this change to fix the warnings in the other PR: https://github.com/dotnet/dotnet-api-docs/pull/4939

PR #4998, opened 2020-10-15
Fix Stream.CanSeek remarks

Sergio0694:
## Summary

The `Stream.CanSeek` remarks incorrectly state that if `CanSeek` is `false`, calls to `Length`, `SetLength`, `Position` and `Seek` will always throw a `NotSupportedException`. This is not actually always the case, see https://github.com/dotnet/runtime/issues/43448.

Follow up from https://github.com/dotnet/runtime/issues/43448#issuecomment-709426395.


gewarren (approved):
@Sergio0694 Thank you!

PR #5007, opened 2020-10-17
Improve JSON descriptions

gewarren:
Contributes to dotnet/docs#21067.

### DIFF
--- a/xml/System.Text.Json/JsonSerializerDefaults.xml
+++ b/xml/System.Text.Json/JsonSerializerDefaults.xml
@@ -72,7 +72,7 @@ For information about the default property values that are applied, see [JsonSer
 
 This member implies that:
 
--Property names are treated as case-insensitive.
+- Property names are treated as case-insensitive.
 - "camelCase" name formatting should be employed.
 - Quoted numbers (JSON strings for number properties) are allowed.
 
--- a/xml/System.Net.Http.Json/HttpClientJsonExtensions.xml
+++ b/xml/System.Net.Http.Json/HttpClientJsonExtensions.xml
@@ -343,7 +343,7 @@ This method uses <xref:System.Text.Json.JsonSerializerDefaults.Web?displayProper
 
 ## Remarks
 
-This method uses <xref:System.Text.Json.JsonSerializerDefaults.Web?displayProperty=nameWithType> options for serialization, whereas <xref:System.Text.Json.JsonSerializer> deserialization methods do not, by default.
+This method uses <xref:System.Text.Json.JsonSerializerDefaults.Web?displayProperty=nameWithType> options for serialization, whereas <xref:System.Text.Json.JsonSerializer> serialization methods do not, by default.
 
           ]]></format>
         </remarks>
@@ -386,7 +386,7 @@ This method uses <xref:System.Text.Json.JsonSerializerDefaults.Web?displayProper
 
 ## Remarks
 
-This method uses <xref:System.Text.Json.JsonSerializerDefaults.Web?displayProperty=nameWithType> options for serialization, whereas <xref:System.Text.Json.JsonSerializer> deserialization methods do not, by default.
+This method uses <xref:System.Text.Json.JsonSerializerDefaults.Web?displayProperty=nameWithType> options for serialization, whereas <xref:System.Text.Json.JsonSerializer> serialization methods do not, by default.
 
           ]]></format>
         </remarks>
@@ -501,7 +501,7 @@ This method uses <xref:System.Text.Json.JsonSerializerDefaults.Web?displayProper
 
 ## Remarks
 
-This method uses <xref:System.Text.Json.JsonSerializerDefaults.Web?displayProperty=nameWithType> options for serialization, whereas <xref:System.Text.Json.JsonSerializer> deserialization methods do not, by default.
+This method uses <xref:System.Text.Json.JsonSerializerDefaults.Web?displayProperty=nameWithType> options for serialization, whereas <xref:System.Text.Json.JsonSerializer> serialization methods do not, by default.
 
           ]]></format>
         </remarks>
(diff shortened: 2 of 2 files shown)


xml/System.Text.Json/JsonSerializerDefaults.xml
```diff
@@ -63,13 +66,17 @@ This option implies that property names are treated as case-sensitive and that "
...
+        <summary>Option values appropriate to Web-based scenarios.</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
-This option implies that property names are treated as case-insensitive and that "camelCase" name formatting should be employed.
+This member implies that:
+
+-Pproperty names are treated as case-insensitive.
```
layomia, line 77:
```suggestion
-Property names are treated as case-insensitive.
```

xml/System.Text.Json/JsonSerializerDefaults.xml
```diff
@@ -34,13 +34,16 @@
...
         <remarks>
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
-This option implies that property names are treated as case-sensitive and that "PascalCase" name formatting should be employed.
+This member implies that:
+
+- Property names are treated as case-sensitive.
+- "PascalCase" name formatting should be employed.
```
layomia, line 46: Not sure that we have to mention this. If we do, what about all the other options e.g. `ReadCommentHandling`, `DictionaryKeyPolicy` etc.

If the intention is to contrast with `Web` options, maybe we should add that quoted numbers (JSON strings) are not allowed.

layomia (approved):

xml/System.Text.Json/JsonSerializerDefaults.xml
gewarren: I was mainly just reformatting the existing text. But it seems like it would be helpful to list out all the options it prescribes?

gewarren: @tdykstra What do you think?

layomia: I think we should link to here: https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions?view=net-5.0#properties since this ideally should list what the `default` value for each property is.

tdykstra: Agree with @layomia proposal, in which case explicitly mentioning property naming defaults isn't necessary here.

xml/System.Text.Json/JsonSerializerDefaults.xml
```diff
@@ -63,13 +66,17 @@ This option implies that property names are treated as case-sensitive and that "
...
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
-This option implies that property names are treated as case-insensitive and that "camelCase" name formatting should be employed.
+This member implies that:
+
+-Pproperty names are treated as case-insensitive.
+- "camelCase" name formatting should be employed.
+- Quoted numbers (JSON strings) are allowed.
```
tdykstra, line 79:
```suggestion
- Quoted numbers (JSON strings for number properties) are allowed.
```

xml/System.Text.Json/JsonSerializerDefaults.xml
gewarren, line 46:
```suggestion
```

xml/System.Text.Json/JsonSerializerDefaults.xml
```diff
@@ -34,13 +34,16 @@
...
+        <summary>General-purpose option values. These are the same settings that are applied if a <see cref="T:System.Text.Json.JsonSerializerDefaults" /> member isn't specified.</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
-This option implies that property names are treated as case-sensitive and that "PascalCase" name formatting should be employed.
+This member implies that:
+
+- Property names are treated as case-sensitive.
```
gewarren, line 45:
```suggestion
```

xml/System.Text.Json/JsonSerializerDefaults.xml
```diff
@@ -34,13 +34,16 @@
...
       <Docs>
-        <summary>Specifies that general-purpose values should be used. These are the same settings applied if a <see cref="T:System.Text.Json.JsonSerializerDefaults" /> isn't specified.</summary>
+        <summary>General-purpose option values. These are the same settings that are applied if a <see cref="T:System.Text.Json.JsonSerializerDefaults" /> member isn't specified.</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
-This option implies that property names are treated as case-sensitive and that "PascalCase" name formatting should be employed.
+This member implies that:
```
gewarren, line 43:
```suggestion
For information about the default property values that are applied, see [JsonSerializerOptions properties](/dotnet/api/system.text.json.jsonserializeroptions?view=net-5.0#properties).
```

xml/System.Text.Json/JsonSerializerDefaults.xml
```diff
@@ -63,13 +64,17 @@ This option implies that property names are treated as case-sensitive and that "
...
+        <summary>Option values appropriate to Web-based scenarios.</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
-This option implies that property names are treated as case-insensitive and that "camelCase" name formatting should be employed.
+This member implies that:
+
+-Property names are treated as case-insensitive.
```
tdykstra, line 75:
```suggestion
- Property names are treated as case-insensitive.
```

xml/System.Text.Json/JsonSerializerDefaults.xml
gewarren: Interestingly, remarks for enum members seem to be bogus (ignored). I'll double check that that behavior is by design.

https://review.docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializerdefaults?view=net-5.0&branch=pr-en-us-5007

PR #5022, opened 2020-10-21
Change XML Encryption 1.1 to XML Encryption 1.0

dhcgn:
## Summary

This part of the documentation is wrong.

> The .NET Framework XML encryption classes implement the World Wide Web Consortium (W3C) XML Encryption Syntax and Processing Version 1.1 specification.

The identifier http://www.w3.org/2009/ from XML-Enc 1.1 is not present at all in the code base. Which indicates the absence of any implementation of XML Encryption 1.1.

See discussion: https://github.com/dotnet/runtime/issues/43436#issuecomment-713862438


gewarren (approved):
@dhcgn Thank you!

PR #5023, opened 2020-10-21
Add Windows GDI+ SKUs support note

safern:
Fixes: https://github.com/dotnet/runtime/issues/42499

### DIFF
--- a/includes/drawing.md
+++ b/includes/drawing.md
@@ -1,5 +1,3 @@
-> [!WARNING]
-> The `System.Drawing` namespace is not recommended for new development, due to not being supported within a Windows or ASP.NET service and it is not cross-platform. [ImageSharp](https://github.com/SixLabors/ImageSharp) and [SkiaSharp](https://github.com/mono/SkiaSharp) are recommended as alternatives.
-
 > [!CAUTION]
-> On Windows `System.Drawing` depends on GDI+ native library, which is shipped as part of the OS. Some Windows SKUs like Windows Server Core or Windows Nano, don't include this native library as part of the OS, so exceptions will be thrown at runtime as we can't load this library.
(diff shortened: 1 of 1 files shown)


includes/drawing.md
```diff
@@ -1,2 +1,5 @@
 > [!WARNING]
 > The `System.Drawing` namespace is not recommended for new development, due to not being supported within a Windows or ASP.NET service and it is not cross-platform. [ImageSharp](https://github.com/SixLabors/ImageSharp) and [SkiaSharp](https://github.com/mono/SkiaSharp) are recommended as alternatives.
+
+> [!CAUTION]
+> On Windows `System.Drawing` depends on GDI+ native library, which is shipped as part of the OS. Some Windows SKUs like Windows Server Core or Windows Nano, don't include this native library as part of the OS, so exceptions will be thrown at runtime as we can't load this library.
```
gewarren, line 5:
```suggestion
> On Windows, `System.Drawing` depends on the GDI+ native library, which is shipped as part of the OS. Some Windows SKUs, like Windows Server Core or Windows Nano, don't include this native library as part of the OS. Exceptions will be thrown at run time because the library can't be loaded.
```

Also, I think it's best to combine the Caution block and the Warning block into a single block with two bullets.

gewarren (approved):

includes/drawing.md
safern: And keep it just as !WARNING ?

gewarren: Actually I think I would make it a caution block. From the contributor guide:

```
> [!CAUTION]
> Negative potential consequences of an action.

> [!WARNING]
> Dangerous certain consequences of an action.
```

Neither of the bullets seems "dangerous".

safern: Sounds good! Thanks.

gewarren:
Looks good! https://review.docs.microsoft.com/en-us/dotnet/api/system.drawing.drawing2d?view=dotnet-plat-ext-2.1&branch=pr-en-us-5023#remarks

PR #5025, opened 2020-10-22
Update UnmanagedCallersOnlyAttribute.xml

AaronRobinsonMSFT:
## Summary

Fixed formatting and added comment clarifying default platform calling convention.

/cc @333fred @gewarren @jkoritzinsky @elinor-fung

### DIFF
--- a/xml/System.Runtime.InteropServices/UnmanagedCallersOnlyAttribute.xml
+++ b/xml/System.Runtime.InteropServices/UnmanagedCallersOnlyAttribute.xml
@@ -27,10 +27,9 @@
 ## Remarks
 
 Methods marked with this attribute have the following restrictions:
-              * Method must be marked `static`.
-              * Must not be called from managed code.
-              * Must only have [blittable](https://docs.microsoft.com/dotnet/framework/interop/blittable-and-non-blittable-types) arguments.
-
+* Method must be marked `static`.
+* Must not be called from managed code.
+* Must only have [blittable](https://docs.microsoft.com/dotnet/framework/interop/blittable-and-non-blittable-types) arguments.
           ]]></format>
     </remarks>
   </Docs>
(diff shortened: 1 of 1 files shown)


xml/System.Runtime.InteropServices/UnmanagedCallersOnlyAttribute.xml
```diff
@@ -75,7 +75,10 @@ Methods marked with this attribute have the following restrictions:
 ## Remarks
 
 Supplied types must be from the official <xref:System.Runtime.CompilerServices> namespace and be of the form `CallConvXXX`.
-        
+
+The "default platform" calling convention is well defined for all supported platforms except for x86 based platforms. The x86 defaults are as follows:
```
333fred, line 79: Personally, I don't know what "well defined" means. Who defines it?

AaronRobinsonMSFT: The platform itself. There is no ambiguity about the calling convention. That is the idea I am trying to convey. Suggestions?

gewarren, line 79:
```suggestion
The "default platform" calling convention is well defined for all supported platforms except x86-based platforms. The x86 defaults are as follows:
```

gewarren (approved):

xml/System.Runtime.InteropServices/UnmanagedCallersOnlyAttribute.xml
333fred: I'd just say what it is. Yes, most people won't need to care, but being vague about what the platform defines it to be doesn't help me figure out what's going on with a stack corruption bug.

AaronRobinsonMSFT: I am not keen on defining the convention beyond deferring to the platform itself. In most cases there aren't specific names other than "Windows ARM64 calling convention" or something equally vague. This leads to pointing to specs that aren't interesting. What I am trying to convey here is x86 is the only case where this is even a consideration. I can clarify that on non-x86 platforms we follow the "platform defined calling convention"?

333fred: The big thing that bugs me about this is that doesn't actually tell me anything. Where would I look to find what that convention is?

AaronRobinsonMSFT: It depends unfortunately. Some of this is convention and honestly "just known" by users who are working on their respective platform. For ARM64 that official specification can be referenced, but practically that doesn't always work so one would typically search for "Windows ARM64 calling convention" or "mac ARM64 calling convention" or "Linux ARM64 calling convention". In the same way one would search for "Windows x64 calling convention". This is a huge issue and not something we can really document here, which is why the [`CallingConvention.WinApi`](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.callingconvention) is defined as "default platform calling convention".

What we are discussing here is one of the fundamental issues that VM based languages are trying to solve - a well specified platform on which to write software. Native code historically requires each platform for have it defined. So I guess that is the answer - "consult your target platform"?

333fred (approved):

elinor-fung (approved):

PR #5028, opened 2020-10-23
HttpClient: remove redundant and outdated runtime config switch information

antonfirsov:
Removes redundant and outdated (!) information about runtime configuration options for `HttpClient`, referencing the options docs page instead.

Related PR updating the options docs: dotnet/docs#21197

### DIFF
--- a/xml/System.Net.Http/HttpClient.xml
+++ b/xml/System.Net.Http/HttpClient.xml
@@ -117,6 +117,8 @@ Starting with .NET Core 2.1, the <xref:System.Net.Http.SocketsHttpHandler?displa
 - The elimination of platform dependencies, which simplifies deployment and servicing. For example, `libcurl` is no longer a dependency on .NET Core for macOS and .NET Core for Linux.
 - Consistent behavior across all .NET platforms.
 
+If this change is undesirable, on Windows you can still use <xref:System.Net.Http.WinHttpHandler> by referencing it's [NuGet package](https://www.nuget.org/packages/System.Net.Http.WinHttpHandler/) and passing it to [`HttpClient`'s constructor`](xref:System.Net.Http.HttpClient.%23ctor(System.Net.Http.HttpMessageHandler)) manually.
+
 ### Configure behavior using run-time configuration options
 
 Certain aspects of <xref:System.Net.Http.HttpClient>'s behavior are customizable through [Run-time configuration options](/dotnet/core/run-time-config/networking). However, the behavior of these switches differs through .NET versions. For example, in .NET Core 2.1 - 3.1, you can configure whether <xref:System.Net.Http.SocketsHttpHandler> is used by default, but that option is no longer available starting in .NET 5.0.


xml/System.Net.Http/HttpClient.xml
```diff
@@ -112,37 +112,14 @@ On each platform, <xref:System.Net.Http.HttpClient> tries to use the best availa
...
 - A significant performance improvement when compared with the previous implementation.
-
 - The elimination of platform dependencies, which simplifies deployment and servicing. For example, `libcurl` is no longer a dependency on .NET Core for macOS and .NET Core for Linux.
-
 - Consistent behavior across all .NET platforms.
 
-If this change is undesirable, you can configure your application to use the older <xref:System.Net.Http.HttpClientHandler?displayProperty=nameWithType> instead in a number of ways:
-
-- By calling the <xref:System.AppContext.SetSwitch%2A?displayProperty=nameWithType> method as follows:
+### Configuring behavior using Runtime Configuration options
```
gewarren, line 120:
```suggestion
### Configure behavior using run-time configuration options
```

xml/System.Net.Http/HttpClient.xml
```diff
@@ -112,37 +112,14 @@ On each platform, <xref:System.Net.Http.HttpClient> tries to use the best availa
...
-  ```json
-  "runtimeOptions": {
-    "configProperties": {
-        "System.Net.Http.UseSocketsHttpHandler": false
-    }
-  }
-  ```
-  
-- By defining an environment variable named `DOTNET_SYSTEM_NET_HTTP_USESOCKETSHTTPHANDLER` and setting it to either `false` or 0.
+Certain aspects of <xref:System.Net.Http.HttpClient>'s behavior are customizable through [Run-time configuration options for networking](https://docs.microsoft.com/en-us/dotnet/core/run-time-config/networking), although the behavior of these switches differs through .NET versions. For example, in versions 2.1-3.1 it is possible to configure whether <xref:System.Net.Http.SocketsHttpHandler> is used by default, but that option is no longer available since .NET 5.
```
gewarren, line 122:
```suggestion
Certain aspects of <xref:System.Net.Http.HttpClient>'s behavior are customizable through [Run-time configuration options](https://docs.microsoft.com/en-us/dotnet/core/run-time-config/networking). However, the behavior of these switches differs through .NET versions. For example, in .NET Core 2.1 - 3.1, you can configure whether <xref:System.Net.Http.SocketsHttpHandler> is used by default, but that option is no longer available starting in .NET 5.0.
```

gewarren (approved):
Thanks for updating this!

xml/System.Net.Http/HttpClient.xml
```diff
@@ -112,37 +112,14 @@ On each platform, <xref:System.Net.Http.HttpClient> tries to use the best availa
...
-  ```json
-  "runtimeOptions": {
-    "configProperties": {
-        "System.Net.Http.UseSocketsHttpHandler": false
-    }
-  }
-  ```
-  
-- By defining an environment variable named `DOTNET_SYSTEM_NET_HTTP_USESOCKETSHTTPHANDLER` and setting it to either `false` or 0.
+Certain aspects of <xref:System.Net.Http.HttpClient>'s behavior are customizable through [Run-time configuration options](https://docs.microsoft.com/en-us/dotnet/core/run-time-config/networking). However, the behavior of these switches differs through .NET versions. For example, in .NET Core 2.1 - 3.1, you can configure whether <xref:System.Net.Http.SocketsHttpHandler> is used by default, but that option is no longer available starting in .NET 5.0.
```
antonfirsov, line 122: @gewarren is this the right way to reference conceptual docs?
```
[Run-time configuration options](https://docs.microsoft.com/en-us/dotnet/core/run-time-config/networking)
```

gewarren: No, it should be `[Run-time configuration options](/dotnet/core/run-time-config/networking)`. Thanks for checking.

xml/System.Net.Http/HttpClient.xml
```diff
@@ -112,37 +112,14 @@ On each platform, <xref:System.Net.Http.HttpClient> tries to use the best availa
...
-  ```json
-  "runtimeOptions": {
-    "configProperties": {
-        "System.Net.Http.UseSocketsHttpHandler": false
-    }
-  }
-  ```
-  
-- By defining an environment variable named `DOTNET_SYSTEM_NET_HTTP_USESOCKETSHTTPHANDLER` and setting it to either `false` or 0.
+Certain aspects of <xref:System.Net.Http.HttpClient>'s behavior are customizable through [Run-time configuration options](/dotnet/core/run-time-config/networking). However, the behavior of these switches differs through .NET versions. For example, in .NET Core 2.1 - 3.1, you can configure whether <xref:System.Net.Http.SocketsHttpHandler> is used by default, but that option is no longer available starting in .NET 5.0.
```
ManickaP, line 122: Should the `UseSocketsHttpHandler` switch be also marked as available only till .NET Core 3.1 in the referenced docs https://docs.microsoft.com/en-us/dotnet/core/run-time-config/networking?

And would it make sense to mention that using `WinHttpHandler` is still possible via OOB nuget package (manually setting it to `HttpClient` ctor)?

antonfirsov: > Should the `UseSocketsHttpHandler` switch be also marked as available only till .NET Core 3.1 in the referenced docs https://docs.microsoft.com/en-us/dotnet/core/run-time-config/networking?

I'm not a big fan of documenting the same thing in redundant ways, so in this PR I'm replacing the detailed docs of `UseSocketsHttpHandler` in  `HttpClient`'s remarks section by a reference to the runtime switches page, which is being updated in dotnet/docs#21197. Does this answer your question, or did you mean something different?


> And would it make sense to mention that using `WinHttpHandler` is still possible via OOB nuget package (manually setting it to `HttpClient` ctor)?

Addressed in https://github.com/dotnet/dotnet-api-docs/pull/5028/commits/e2285e6c11ca8b073480dbbac69228d75037e19d

ManickaP: No, that's perfect! I wasn't aware of the other PR and the existing note. Thank you!

ManickaP (approved):

PR #5063, opened 2020-10-30
Fix confusing sentence "without assuming architecture dependent alignment of the addresses" (#5342)

### LINKED ISSUE #5342: Confusing sentence in documentation of System.Runtime.CompilerServices.Unsafe.ReadUnaligned
[The documentation](https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.unsafe.readunaligned?view=netcore-3.1) says that

> Reads a value of type T from the given location without assuming architecture dependent alignment of the addresses.

What means "without assuming architecture dependent alignment of the addresses" ?
This sentence is really confusing.
When the first time I see this documentation, I thought that 
"If we forget to assuming architecture dependent alignment of the addresses while read or write from unaligned addresses, we will fail it. So, this method requires alignment."
IMHO, 
"Reads a value of type T correctly from given location even if while we use the platform which requires alignment and given location is unaligned."
is better sentence.

`Unsafe.WriteUnaligned` also has same problem.

RamType0:
Fixes #5342

### DIFF
--- a/xml/System.Runtime.CompilerServices/Unsafe.xml
+++ b/xml/System.Runtime.CompilerServices/Unsafe.xml
@@ -594,7 +594,7 @@
         <param name="destination">The destination address to copy to.</param>
         <param name="source">The source address to copy from.</param>
         <param name="byteCount">The number of bytes to copy.</param>
-        <summary>Copies bytes from the source address to the destination address without assuming architecture dependent alignment of the addresses.</summary>
+        <summary>Copies bytes from the source address to the destination address. Unlike <xref:System.Runtime.CompilerServices.Unsafe.CopyBlock>, this will succeed even while we use the platform which requires alignment and any of the given location is unaligned.</summary>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
@@ -626,7 +626,7 @@
         <param name="destination">The destination address to copy to.</param>
         <param name="source">The source address to copy from.</param>
         <param name="byteCount">The number of bytes to copy.</param>
-        <summary>Copies bytes from the source address to the destination address without assuming architecture dependent alignment of the addresses.</summary>
+        <summary>Copies bytes from the source address to the destination address. Unlike <xref:System.Runtime.CompilerServices.Unsafe.CopyBlock>, this will succeed even while we use the platform which requires alignment and any of the given location is unaligned.</summary>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
@@ -726,7 +726,7 @@
         <param name="startAddress">The address of the start of the memory block to initialize.</param>
         <param name="value">The value to initialize the block to.</param>
         <param name="byteCount">The number of bytes to initialize.</param>
-        <summary>Initializes a block of memory at the given location with a given initial value without assuming architecture dependent alignment of the address.</summary>
+        <summary>Initializes a block of memory at the given location with a given initial value. Unlike <xref:System.Runtime.CompilerServices.Unsafe.InitBlock>, this will succeed even while we use the platform which requires alignment and the given location is unaligned.</summary>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
@@ -758,7 +758,7 @@
         <param name="startAddress">The address of the start of the memory block to initialize.</param>
         <param name="value">The value to initialize the block to.</param>
         <param name="byteCount">The number of bytes to initialize.</param>
-        <summary>Initializes a block of memory at the given location with a given initial value without assuming architecture dependent alignment of the address.</summary>
+        <summary>Initializes a block of memory at the given location with a given initial value. Unlike <xref:System.Runtime.CompilerServices.Unsafe.InitBlock>, this will succeed even while we use the platform which requires alignment and the given location is unaligned.</summary>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
@@ -955,7 +955,7 @@
       <Docs>
         <typeparam name="T">The type to read.</typeparam>
         <param name="source">The location to read from.</param>
-        <summary>Reads a value of type <typeparamref name="T" /> from the given location without assuming architecture dependent alignment of the addresses.</summary>
+        <summary>Reads a value of type <typeparamref name="T" /> from the given location. Unlike <xref:System.Runtime.CompilerServices.Unsafe.Read>, this will succeed even while we use the platform which requires alignment and the given location is unaligned for values of type <typeparamref name="T" />.</summary>
         <returns>An object of type <typeparamref name="T" /> read from the given location.</returns>
         <remarks>To be added.</remarks>
       </Docs>
@@ -989,7 +989,7 @@
       <Docs>
         <typeparam name="T">The type to read.</typeparam>
         <param name="source">The location to read from.</param>
-        <summary>Reads a value of type <typeparamref name="T" /> from the given location without assuming architecture dependent alignment of the addresses.</summary>
+        <summary>Reads a value of type <typeparamref name="T" /> from the given location. Unlike <xref:System.Runtime.CompilerServices.Unsafe.Read>, this will succeed even while we use the platform which requires alignment and the given location is unaligned for values of type <typeparamref name="T" />.</summary>
         <returns>An object of type <typeparamref name="T" /> read from the given location.</returns>
         <remarks>To be added.</remarks>
       </Docs>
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=5063) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=5063) before we can accept your contribution.<br/><br/>:x: RamType0 [sign now](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=5063)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/dotnet-api-docs?pullRequest=5063) it.</sub>

GrabYourPitchforks:
I don't think the suggested corrections add clarity. If we wanted to clarify matters, I think this phrasing flows better:

> Writes a value of type _T_ to the given location. This write will succeed even if the destination address is not properly aligned for values of type _T_.

RamType0:
> I don't think the suggested corrections add clarity. If we wanted to clarify matters, I think this phrasing flows better:
> 
> > Writes a value of type _T_ to the given location. This write will succeed even if the destination address is not properly aligned for values of type _T_.

Do you mean that "succeed" is more clarity than "correctly" or "not properly aligned" is clarity than "unaligned" ?
I don't think that these are better, but I don't care about replacing these sentence.

Adding "for values of type T" sounds good. 

I care about preserving original sentence's replacabillity.
Any of "unaligned" variant's summary was created by just adding "without assuming architecture dependent addresses" to "aligned" method's summary.

In my correction, any of "unaligned" variant's summary was created by just adding "correctly even while we use the platform which requires alignment and (any of) the given location is unaligned" to "aligned" method's summary.



At least, any of new correction(mine or yours) is far much clarity than current document.

danmoseley:
@RamType0 I believe we should use @GrabYourPitchforks suggestion. The change as is I find more confusing.

RamType0:
> @RamType0 I believe we should use @GrabYourPitchforks suggestion. The change as is I find more confusing.

I think his phrase is OK, but I don't know which phrase of mine making it confusing.
I should know what making it confusing for fixing it correctly.

RamType0:
Do you think my phrase is more confusing than current?

IMHO, @GrabYourPitchforks's phrase is greater than current, but it is not greater than mine, and not lesser than mine.
And both mine and @GrabYourPitchforks's phrase is far much greater than current.

Is there any non-native English speaker in reviewer?
I'm not a native English speaker, but the document should be easy to understand for both native English speakers and non-native English speakers, ideally.

GrabYourPitchforks:
> I should know what making it confusing for fixing it correctly.

Sure, let's break it down. I've added __\[ \]__ brackets around your text to help group constructs.

__\[Copies bytes from the source address to the destination address correctly\] \[even while we use the platform which requires alignment\] \[and any of the given location is unaligned\].__

The sentence communicates three separate ideas, as bracketed above. For improved readability, it's preferred to separate out the most important idea. Additional sentences can communicate extra information.

This guideline suggests that the first idea should be its own sentence: "Copies bytes from the source address to the destination address correctly." However, it's awkward to describe a method as performing some operation "correctly." Most developers would expect .NET APIs to behave correctly. Let's keep that in mind for now and move on to the next part.

The next two ideas communicate that: (a) some platforms require alignment; and (b) there may be memory locations which are unaligned. These statements feel redundant. If a value is at an unaligned memory location, then by definition, the value's location violates the platform's normal alignment assumptions. This implies that we can eliminate the middle idea ("even while...").

Now we come back to the "correctly" aspect from earlier. The important part of the "correctly" statement isn't that the method behaves correctly. (Developers would naturally expect this.) Instead, the important part is that the method behaves correctly _even when the values are not aligned_. That's what makes these methods unique compared to standard read / write / copy methods. This implies that the "correctly" word should be joined to the idea that some memory locations are unaligned. Since the first sentence of the text already communicates the core idea that the method reads / writes / copies bytes, we should move the "correctly" word to the second sentence.

With these changes, we end up with (paraphrased):

> Copies bytes from the source address to the destination address. Works correctly if any of the given location is unaligned.

The rest is technical cleanup (using more accurate terminology), tailoring the wording for the target audience, and grammatical corrections.

Hope this helps explain things!

RamType0:
@GrabYourPitchforks Thanks for great description!

BTW, I heard it is ensured that primitive types with size lesser than 4bytes or pointer types are always read or written atomically.
Is this also applied for Unsafe.ReadUnaligned and Unsafe.WriteUnaligned?(I don't think so)
It should be also documented.

GrabYourPitchforks:
> BTW, I heard it is ensured that primitive types with size lesser than 4bytes or pointer types are always read or written atomically. Is this also applied for Unsafe.ReadUnaligned and Unsafe.WriteUnaligned?(I don't think so)

This depends on the processor. For x86 / x64 processors, if the integral value is appropriately aligned for the target address, the read / write is guaranteed atomic.

> It should be also documented.

This is not necessary. The APIs discussed in this issue deal only with alignment concerns, not atomicity. The target developer for this API is somebody who needs low-level access to memory. It should be assumed such a person is intimately familiar with memory ordering models, platform alignment requirements, and similar concepts. We don't need to describe every guarantee or non-guarantee provided by these APIs beyond those in the API's immediate purview.

For an example of API documentation which _does_ describe atomicity guarantees, see [`System.Threading.Interlocked`](https://docs.microsoft.com/dotnet/api/system.threading.interlocked).

RamType0:
I've fixed some sentences with following your instructions.

But I've felt some other missing description in document.

- What is unaligned address for `Unsafe.InitBlock` or `Unsafe.CopyBlock`?
IMHO, pointer is always aligned for values for type `byte`.
[I also referenced OpCodes.Cpblk](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.cpblk?view=net-5.0), but it just says "the natural size of the machine". I didn't understand it.(Is it means the pointer size?)

- Is it not necessary that `Unsafe.CopyBlock` doesn't support overlapping copy?
AFIK,Unsafe.CopyBlock is just a wrapper of `cpblk` IL instruction, and it doesn't supports overlapping copy(Some regions of source and destination is overlapping).
But it is not documented.

RamType0:
Sorry for my bad knowledge, how can I create a link in summary?

>Try searching the docs.microsoft.com contributor guides

It is forbidden
>Post your question in the Docs support channel

Error occurred while starting Teams web app with following information.

Request Id: 2451e15b-3537-425d-af87-c122e9ba5701
Correlation Id: 3b93a12f-118a-498b-ad08-2098aaaf6118
Timestamp: 2021-03-27T15:56:52Z
Message: AADSTS90002: Tenant '72f98bf-86f1-41af-91ab-2d7cd011db47' not found. This may happen if there are no active subscriptions for the tenant. Check to make sure you have the correct tenant ID. Check with your subscription administrator.

danmoseley:
@gewarren ?

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -594,7 +594,7 @@
         <param name="destination">The destination address to copy to.</param>
         <param name="source">The source address to copy from.</param>
         <param name="byteCount">The number of bytes to copy.</param>
-        <summary>Copies bytes from the source address to the destination address without assuming architecture dependent alignment of the addresses.</summary>
+        <summary>Copies bytes from the source address to the destination address. Unlike <xref:System.Runtime.CompilerServices.Unsafe.CopyBlock>, this will succeed even while we use the platform which requires alignment and any of the given location is unaligned.</summary>
```
gewarren, line 597:
```suggestion
        <summary>Copies bytes from the source address to the destination address. Unlike <see cref="Overload:System.Runtime.CompilerServices.Unsafe.CopyBlock">, this write will succeed on platforms that require alignment even if the destination address is unaligned.</summary>
```

gewarren:
> Sorry for my bad knowledge, how can I create a link in summary?

The `xref` style links are for markdown sections only. For XML content, like in the summary, use `<see cref>` links. I left a suggestion that you can follow for the other summaries.

Also, apologies about those internal help links. I've filed a suggestion that they also include a link to the [public contributor's guide](https://docs.microsoft.com/en-us/contribute/) in those PR comments. (Although that guide doesn't seem to include any API ref-specific guidance.)

GrabYourPitchforks:
> Unsafe.CopyBlock is just a wrapper of cpblk IL instruction, and it doesn't supports overlapping copy(Some regions of source and destination is overlapping).
But it is not documented.

This is clearly documented. From ECMA-335, III.3.30: "The behavior of cpblk is unspecified if the source and destination areas overlap."

We don't need to mirror the entirety of the ECMA text into our docs here. These APIs are meant for developers who need to emit specific IL instructions into their apps. It's reasonable to expect callers to have read the appropriate specs.

Honestly I think we're starting to go down a rabbit hole here. In an ideal world we'd have all docs easily approachable and understandable by a wide developer audience, but _these are not APIs for a broad audience_. There is intentionally some amount of gatekeeping in this API surface. We hide these APIs (and name them "unsafe"!) for good reason.

xml/System.Runtime.CompilerServices/Unsafe.xml
RamType0: I used the sentence "any of the given location" because it also takes source address and it could be also unaligned.
What do you think about it?

Any other suggestion seems to be really good! Thanks!

RamType0:
@GrabYourPitchforks How about adding link to `OpCodes.Cpblk` at "See also"?

gewarren (dismissed):
Thank you @RamType0!

danmoseley:
@copilot resolve the merge conflicts in this pull request

PR #5066, opened 2020-10-31
FileStream: Position may not be restored if SetLength fails

jozkee:
## Summary

Add paragraph to FileStream.SetLength explaining how the method works and that Position property may not be restored if the method fails. On Windows this may happen when there is not enough space on disk for the provided length.

### DIFF
--- a/xml/System.IO/FileStream.xml
+++ b/xml/System.IO/FileStream.xml
@@ -3727,6 +3727,8 @@ The following example shows how to read from a file asynchronously.
  If the given value is less than the current length of the stream, the stream is truncated. In this scenario, if the current position is greater than the new length, the current position is moved to the last byte of the stream. If the given value is larger than the current length of the stream, the stream is expanded, and the current position remains the same. If the stream is expanded, the contents of the stream between the old and the new length are undefined.  
   
  A stream must support both writing and seeking for `SetLength` to work.  
+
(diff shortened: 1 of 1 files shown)


carlossanlop:
Context - This PR: https://github.com/dotnet/runtime/pull/43863

xml/System.IO/FileStream.xml
```diff
@@ -3727,6 +3727,8 @@ The following example shows how to read from a file asynchronously.
  If the given value is less than the current length of the stream, the stream is truncated. In this scenario, if the current position is greater than the new length, the current position is moved to the last byte of the stream. If the given value is larger than the current length of the stream, the stream is expanded, and the current position remains the same. If the stream is expanded, the contents of the stream between the old and the new length are undefined.  
   
  A stream must support both writing and seeking for `SetLength` to work.  
+
+ This method sets <xref:System.IO.Stream.Position> to `value` and proceeds to set <xref:System.IO.Stream.Length> to `value`, then restores <xref:System.IO.Stream.Position> to its original value. If this method fails, <xref:System.IO.Stream.Position> may not be restored.
```
jkotas, line 3731:
```suggestion
If the operation fails, the current position is undefined.
```
This is describes the current implementation of this method on Windows. The implementation may change in future; or it may be different between OSes. We should avoid describing implementation details in the documentation where possible. The documentation should describe externally observable behavior or invariants instead.

Jozkee:
@jkotas I will close this PR since https://github.com/dotnet/runtime/pull/44170 will fix this error, let me know if you think otherwise.

PR #5070, opened 2020-11-01
Update SendKeys.xml

nicovv44:
Add space as the or seems to miss one


carlossanlop:
@nicovv44 did you expect to see two spaces? Because I currently see one space on the left side of the diff.

nicovv44:
The problem seems now fixed.

I was seeing no space in Google chrome...

PR #5071, opened 2020-11-01
Document CultureInfo.CurrentCulture default on Unix-likes

### LINKED ISSUE #4254: CultureInfo.CurrentCulture document default on Unix-likes
The procedure described in [How a Thread's Culture Is Determined](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.currentculture#how-a-threads-culture-is-determined) section of the documentation for [`CultureInfo.CurrentCulture`](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.currentculture) currently ends with

> * By calling the Windows `GetUserDefaultLocaleName` function.

Which clearly does not apply to .NET Core on non-Windows systems.  After a bit of digging I've found the following:

* [`CultureInfo.CurrentCulture.get`](https://github.com/dotnet/runtime/blob/v5.0.0-preview.3.20214.6/src/libraries/System.Private.CoreLib/src/System/Globalization/CultureInfo.cs#L388) calls
* [`CultureInfo.GetUserDefaultCulture()`](https://github.com/dotnet/runtime/blob/v5.0.0-preview.3.20214.6/src/libraries/System.Private.CoreLib/src/System/Globalization/CultureInfo.Unix.cs#L11) calls
* [`GlobalizationNative_GetDefaultLocaleName`](https://github.com/dotnet/runtime/blob/1db45ed953745c0e270d09df99828cb58772f4ad/src/libraries/Common/src/Interop/Interop.Locale.cs#L21) defined as
* [`GlobalizationNative_GetDefaultLocaleName`](https://github.com/dotnet/runtime/blob/v5.0.0-preview.3.20214.6/src/libraries/Native/Unix/System.Globalization.Native/pal_locale.c#L208) calls
* [`DetectDefaultLocaleName`](https://github.com/dotnet/runtime/blob/v5.0.0-preview.3.20214.6/src/libraries/Native/Unix/System.Globalization.Native/pal_locale.c#L131) calls
* [`uloc_getDefault`](https://github.com/unicode-org/icu/blob/release-67-1/icu4c/source/common/uloc.cpp#L2103) (in [ICU](https://icu-project.org)) calls
* [`locale_get_default`](https://github.com/unicode-org/icu/blob/release-67-1/icu4c/source/common/locid.cpp#L240) calls
* [`Locale::getDefault`](https://github.com/unicode-org/icu/blob/release-67-1/icu4c/source/common/locid.cpp#L926) calls
* [`locale_set_default_internal`](https://github.com/unicode-org/icu/blob/release-67-1/icu4c/source/common/locid.cpp#L172) calls
* [`uprv_getDefaultLocaleID`](https://github.com/unicode-org/icu/blob/release-67-1/icu4c/source/common/putil.cpp#L1615) calls
* [`uprv_getPOSIXIDForDefaultLocale`](https://github.com/unicode-org/icu/blob/release-67-1/icu4c/source/common/putil.cpp#L1589) calls
* [`uprv_getPOSIXIDForCategory(LC_MESSAGES)`](https://github.com/unicode-org/icu/blob/release-67-1/icu4c/source/common/putil.cpp#L1526) calls
* [`setlocale(3)`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/setlocale.html) with `category` `LC_MESSAGES`, which is specified by XBD [Locale](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap07.html#tag_07) and [Internationalization Variables](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_02) to use the first valid value from `$LC_ALL`, `$LC_MESSAGES`, or `$LANG`.

Assuming ICU is an implementation detail and that developers are more likely to be familiar with `setlocale(3)` than `uloc_getDefault()`, I'd suggest changing the last bullet point to something like:

* By calling the Windows `GetUserDefaultLocaleName` function, or POSIX `setlocale(3)` with `LC_MESSAGES` on Unix-like systems.

With appropriate caveats if this behavior is not guaranteed.

Thanks for considering,
Kevin

kevinoid:
## Summary

The last step of the "How a Thread's Culture Is Determined" section of `CultureInfo.CurrentCulture` currently documents calling the Windows `GetUserDefaultLocaleName` function.  On Unix-like systems [ICU] `uloc_getDefault` is called, which eventually calls POSIX `setlocale(3)` with category `LC_MESSAGES` to get the default, as described in #4254.

This PR documents the use of `setlocale` on Unix-like systems to help developers understand the default on these systems.

Thanks for considering,
Kevin

Fixes #4254

[ICU]: https://icu-project.org/

P.S. If you'd prefer that ICU `uloc_getDefault` is mentioned (exclusively or in addition to `setlocale`) or `setlocale` is linked to [its spec. in POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/setlocale.html), [CRT](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale), or elsewhere, let me know.

### DIFF
--- a/xml/System.Globalization/CultureInfo.xml
+++ b/xml/System.Globalization/CultureInfo.xml
@@ -1208,7 +1208,7 @@ If `name` is <xref:System.String.Empty?displayProperty=nameWithType>, the constr
   
      For more information, see the "Culture and task-based asynchronous operations" section in the <xref:System.Globalization.CultureInfo> topic.  
   
--   By calling the Windows `GetUserDefaultLocaleName` function.  
+-   By calling the `GetUserDefaultLocaleName` function on Windows or the `uloc_getDefault` function from [ICU](https://icu-project.org/), which currently calls the POSIX `setlocale` function with category `LC_MESSAGES`, on Unix-like systems.  
   
  Note that if you set a specific culture that is different from the system-installed culture or the user's preferred culture, and your application starts multiple threads, the current culture of those threads will be the culture that is returned by the `GetUserDefaultLocaleName` function, unless you assign a culture to the <xref:System.Globalization.CultureInfo.DefaultThreadCurrentCulture%2A> property in the application domain in which the thread is executing.  
   


xml/System.Globalization/CultureInfo.xml
```diff
@@ -1208,7 +1208,7 @@ If `name` is <xref:System.String.Empty?displayProperty=nameWithType>, the constr
   
      For more information, see the "Culture and task-based asynchronous operations" section in the <xref:System.Globalization.CultureInfo> topic.  
   
--   By calling the Windows `GetUserDefaultLocaleName` function.  
+-   By calling the `GetUserDefaultLocaleName` function on Windows or the `setlocale` function with category `LC_MESSAGES` on Unix-like systems.
```
tarekgh, line 1211: on Linux we call ICU API uloc_getDefault 

https://github.com/dotnet/runtime/blob/f1de95ede6792c5a3106ea26f2faef42ad29820b/src/libraries/Native/Unix/System.Globalization.Native/pal_locale.c#L132

the setLocale used in old versions but not used anymore starting from 5.0.

kevinoid: > on Linux we call ICU API uloc_getDefault

Indeed.  And `uloc_getDefault` calls `setlocale` (by the call sequence I documented in #4254.  Are you arguing we should document `uloc_getDefault` exclusively?

tarekgh: Yes would be good to document the actual call we use because if the ICU decided to change the way picking up the default locale that will make our doc invalid. we don't have to document the ICU API ourselves. make sense?

kevinoid: Sure, assuming you want ICU to be part of the documented API, rather than an implementation detail.  Also, I suspect many more developers are familiar with how `setlocale` works than `uloc_getDefault`.  Would it be acceptable to note that `uloc_getDefault` currently uses `setlocale`?

tarekgh: > Would it be acceptable to note that uloc_getDefault currently uses setlocale?

Sure, I don't mind to phrase it this way.

kevinoid: Great!  I've updated the PR to document `uloc_getDefault` and mention `setlocale`.  Let me know what you think.

xml/System.Globalization/CultureInfo.xml
```diff
@@ -1208,7 +1208,7 @@ If `name` is <xref:System.String.Empty?displayProperty=nameWithType>, the constr
   
      For more information, see the "Culture and task-based asynchronous operations" section in the <xref:System.Globalization.CultureInfo> topic.  
   
--   By calling the Windows `GetUserDefaultLocaleName` function.  
+-   By calling the `GetUserDefaultLocaleName` function on Windows or the `uloc_getDefault` function from [ICU](https://icu-project.org/), which currently calls the POSIX `setlocale` function with category `LC_MESSAGES`, on Unix-like systems.
```
tarekgh, line 1211: @carlossanlop could you please double check if we use the md hyperlinks styles in our docs as it is used here in the ICU link?

tarekgh (approved):

xml/System.Globalization/CultureInfo.xml
carlossanlop: Yes, the way it was used in this change is correct. You can also confirm that it looks good if you go to the preview URL: https://review.docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.currentculture?view=dotnet-uwp-10.0&branch=pr-en-us-5071#how-a-threads-culture-is-determined

Scroll a bit down after the code snippet.

PR #5074, opened 2020-11-02
Fix documentation for Process.WaitForExit

eiriktsarpalis:
Fixes https://github.com/dotnet/runtime/issues/44143.


adamsitnik (approved):
LGTM, thank you @eiriktsarpalis 

FWIW I've verified the magic numbers: https://github.com/dotnet/runtime/blob/6072e4d3a7a2a1493f514cdf4be75a3d56580e84/src/libraries/System.Private.CoreLib/src/System/Threading/Timeout.cs#L13

@eiriktsarpalis should we add a `.WaitForExit(Timespan)` overload to avoid problems with the magic integer values?

PR #5084, opened 2020-11-04
Add hard newlines to fix formatting

gewarren:


xml/System/String.xml
```diff
@@ -5301,14 +5301,14 @@ Invalid sequences are represented in the enumeration by <xref:System.Text.Rune.R
...
 
- [Get started with the String.Format method](#Starting)
- [Which method do I call?](#FTaskList)
- [The Format method in brief](#Format_Brief)
- [The Format item](#FormatItem)
- [How arguments are formatted](#HowFormatted)
- [Format items that have the same index](#SameIndex)
- [Formatting and culture](#Format_Culture)
- [Custom formatting operations](#Format_Custom)
+ [Get started with the String.Format method](#Starting)\
```
BillWagner, line 5304: In the conceptual docs, we've standardized on lists for these links. I'll approve, and let you consider which you prefer.

xml/System/String.xml
```diff
@@ -5410,15 +5410,15 @@ Invalid sequences are represented in the enumeration by <xref:System.Text.Rune.R
 
  A format item has the following elements:
 
- *index*
+ *index*\
```
BillWagner, line 5413: Did you want italic here, or bold? (3 occurrences)

BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

xml/System/String.xml
gewarren: I'm not sure, but since I didn't introduce it, I'm going to leave it as-is for now. There are a lot of improvements we can make to this file, like moving out large swaths of text into conceptual docs.

gewarren:
Thanks Bill!

PR #5086, opened 2020-11-04
Update Span`1.xml

marafiq:
typo and may be better sentence

## Summary

Describe your changes here.

### DIFF
--- a/xml/System/Span`1.xml
+++ b/xml/System/Span`1.xml
@@ -980,7 +980,7 @@ Two <xref:System.Span%601> objects are equal if they have different lengths or i
         <remarks>
           <format type="text/markdown"><![CDATA[  
 
-This method performs a heap allocation and therefore should be avoided if possible. It is sometimes necessary when APIs that support <xref:System.Span%601> objects are not available. And those APIs work with arrays instead.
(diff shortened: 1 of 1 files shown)


xml/System/Span`1.xml
```diff
@@ -980,7 +980,7 @@ Two <xref:System.Span%601> objects are equal if they have different lengths or i
         <remarks>
           <format type="text/markdown"><![CDATA[  
 
-This method performs a heap allocation and therefore should be avoided if possible. It it sometimes necessary when APIs that support <xref:System.Span%601> objects are not available but APIs that work with arrays are.
+This method performs a heap allocation and therefore should be avoided if possible. It is sometimes necessary when APIs that support <xref:System.Span%601> objects are not available. And those APIs work with arrays instead.
```
carlossanlop, line 983: May I suggest an alternative wording to make the message a bit clearer?:

```suggestion
This method performs a heap allocation and therefore should be avoided if possible. Heap allocations are expected in APIs that work with arrays. Using such APIs is unavoidable if an alternative API overload that takes a <xref:System.Span%601> does not exist.
```

carlossanlop (approved):
LGTM. Thanks!

PR #5095, opened 2020-11-06
Remove "gets or sets" from fields

gewarren:
See https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Field for field guidance.


buyaa-n (approved):
Seems there is one field left, https://github.com/dotnet/dotnet-api-docs/blob/c865806a2681d6fb80f8091c1509dd3783b4b584/xml/System.Xml.Serialization/XmlDeserializationEvents.xml#L61

Overall looks good

PR #5108, opened 2020-11-10
Correct Rfc2898DeriveBytes example encryption class reference

dhivarson:
The example is using the Aes class for encryption not the referenced TripleDES class.

## Summary

Corrected the encryption class reference from TripleDES to Aes in the examples section of the Rfc2898DeriveBytes class.


gewarren (approved):
@dhivarson Thank you!

PR #5121, opened 2020-11-14
Update XsltArgumentList.xml

ericamick:
As written, the documentation currently suggests only one parameter per namespace is allowed. It should be one parameter with a given name per namespace.

## Summary

Make clear that a duplicate name in the same namespace throws the exception.

### DIFF
--- a/xml/System.Xml.Xsl/XsltArgumentList.xml
+++ b/xml/System.Xml.Xsl/XsltArgumentList.xml
@@ -258,7 +258,7 @@
   
  The <paramref name="name" /> is not a valid name according to the W3C XML specification.  
   
- The <paramref name="namespaceUri" /> already has a parameter associated with it.</exception>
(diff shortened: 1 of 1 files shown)


xml/System.Xml.Xsl/XsltArgumentList.xml
```diff
@@ -258,7 +258,7 @@
   
  The <paramref name="name" /> is not a valid name according to the W3C XML specification.  
   
- The <paramref name="namespaceUri" /> already has a parameter associated with it.</exception>
+ The <paramref name="namespaceUri" /> already has a parameter called <paramref name="name" /> associated with it.</exception>
```
buyaa-n, line 261: I think it meant it already has a parameter associated with the given `name`
```suggestion
 The <paramref name="namespaceUri" /> already has a parameter associated with the <paramref name="name" />.</exception>
```

PR #5128, opened 2020-11-18
Add 2 OOB .NET extension assemblies

gewarren:
Adds System.Memory.Data.dll and System.ServiceModel.Federation.dll APIs to .NET Core 2.0+ extensions.

### DIFF
--- a/xml/FrameworksIndex/dotnet-plat-ext-2.0.xml
+++ b/xml/FrameworksIndex/dotnet-plat-ext-2.0.xml
@@ -41,6 +41,7 @@
     <Assembly Name="Microsoft.Extensions.Options" Version="2.0.0.0" />
     <Assembly Name="Microsoft.Extensions.Primitives" Version="2.0.0.0" />
     <Assembly Name="Microsoft.Extensions.WebEncoders" Version="2.0.0.0" />
+    <Assembly Name="System.Memory.Data" Version="1.0.1.0" />
     <Assembly Name="System.ServiceModel.Federation" Version="4.8.0.0" />
   </Assemblies>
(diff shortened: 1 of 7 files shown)


safern (approved):

gewarren:
This is what the ["Applies to" section](https://review.docs.microsoft.com/en-us/dotnet/api/system.binarydata?view=dotnet-plat-ext-2.0&branch=pr-en-us-5128#moniker-applies-to) looks like for BinaryData. It doesn't mention .NET Standard. Is this okay?

image

safern:
> It doesn't mention .NET Standard. Is this okay?

I think this is how it shows for others, right? Or do we have a .NET Standard Platform Extensions moniker as well?

jeffhandley:
I noticed "To be added" throughout the docs. Does this step add the type/member definitions for the docs to get inserted to after a subsequent step?

gewarren:
> I think this is how it shows for others, right? Or do we have a .NET Standard Platform Extensions moniker as well?

We don't currently have a .NET Standard Platform Extensions moniker. Maira suggested we ask @terrajobst about whether we should have such a moniker.

> I noticed "To be added" throughout the docs. Does this step add the type/member definitions for the docs to get inserted to after a subsequent step?

Yes, since we didn't import doc comments for this PR, Carlos will bring them over in a subsequent step. This was intentional so as not to overwrite docs for other APIs.

gewarren:
I'll go ahead and merge this. If we decide to add a .NET Standard Platform Extensions moniker, we can add this to it later.

terrajobst:
Makes sense. Thanks!

PR #5130, opened 2020-11-18
Update FlagsAttribute.xml

mikedamay:
The only effect of the `[Flags] attribute is to modify the behavior of calls to `ToString()`.  Although it is difficult to think of direct practical negative consequences of implying that `FlagsAttribute` is necessary to use an enumeration as a bit field in the way that the article suggests, a small unnecessary cognitive load is imposed when official documentation does not quite tally with one's experience.  Precision is worth it.

## Summary

Describe your changes here.

### DIFF
--- a/xml/System/FlagsAttribute.xml
+++ b/xml/System/FlagsAttribute.xml
@@ -52,7 +52,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>Indicates that an enumeration can be treated as a bit field; that is, a set of flags.</summary>
+    <summary>Ensures that ToString() will return well formatted combinations of enum values where an enumeration is treated as a bit field.</summary>
     <remarks>
       <format type="text/markdown"><![CDATA[  
   


xml/System/FlagsAttribute.xml
```diff
@@ -52,7 +52,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>Indicates that an enumeration can be treated as a bit field; that is, a set of flags.</summary>
+    <summary>Ensures that ToString() will return well formatted combinations of enum values where an enumeration is treated as a bit field.</summary>
```
tannergooding, line 55: I don't think this is a good change. While it's (as far as I remember) true that `System.Enum` is largely only impacted on `ToString`, this flag is intended to be general metadata that other things can and may depend on to understand that `Flag1 | Flag2` is "valid", and not just unknown or undefined behavior.

tannergooding: For example, `System.ComponentModel.Annotations` checks for this flag to customize its own behavior on repporting whether a value is valid.

gewarren:
Based on Tanner's feedback, I'm going to close this PR.

PR #5136, opened 2020-11-21
Match labels to trackbars according to RGB order

TarenLewis:
## Summary

Matched Label2 with Green value, and Label3 with Blue value in accordance with RGB order (1,2,3), and also to match the trackbar numbers (1,2,3). Improves readability and lowers risk of confusion.

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_Winforms/Trackbar/CS/form1.cs
+++ b/samples/snippets/csharp/VS_Snippets_Winforms/Trackbar/CS/form1.cs
@@ -108,14 +108,14 @@ namespace TrackBar
             // 
             // label2
             // 
-            this.label2.Location = new System.Drawing.Point(8, 16);
+            this.label2.Location = new System.Drawing.Point(600, 16);
             this.label2.Name = "label2";
             this.label2.Size = new System.Drawing.Size(120, 16);
             this.label2.TabIndex = 5;
             // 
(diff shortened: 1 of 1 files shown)


gewarren:
@TarenLewis I don't actually see that the code you modified is used in the documentation anywhere, only snippet4 (lines 64-93). Were you just browsing the code when you noticed this?

TarenLewis:
@gewarren Yes, I noticed this "issue" in the documentation located here:
https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.trackbar.scroll?view=netcore-3.1
It seems counterintuitive to me that the naming conventions would be:

label1 --> trackbar1
label3 --> trackbar2
label2 --> trackbar3 

Thats why I swapped the coordinates, and renamed the labels accordingly to fit RGB order and "better" naming conventions. The code seemed confusing to me when I first looked at it, that's why I did this.

gewarren:
I see it now, thanks. One problem is that the C++ and VB snippets still have the previous order of the labels. Could you update those too?

TarenLewis:
Yes, I will take care of those fixes as well.

gewarren:
@TarenLewis Are you still working on the VB/C++ snippets? Could you add them to this PR?

gewarren:
Closing in favor of #9997.

PR #5141, opened 2020-11-24
Add an observation about delegates closed over their first argument.

teo-tsirpanis:
## Summary

I discovered that they cannot be created when it is a boxed value type, failing with a signature mismatch.


xml/System/Delegate.xml
```diff
@@ -96,7 +96,7 @@
...
   
- When a delegate represents a static method, the delegate stores a reference to the method's entry point. When a delegate represents a static method closed over its first argument, the delegate stores a reference to the method's entry point and a reference to a target object assignable to the type of the method's first argument. When the delegate is invoked, the first argument of the static method receives the target object.  
+ When a delegate represents a static method, the delegate stores a reference to the method's entry point. When a delegate represents a static method closed over its first argument, the delegate stores a reference to the method's entry point and a reference to a target object assignable to the type of the method's first argument. When the delegate is invoked, the first argument of the static method receives the target object. This first argument must be a reference type; not a boxed value type.
```
tannergooding, line 99: Can you provide an example of the code that fails that you were expecting to work?

tannergooding: For example, the following program works as expected and so it's unclear to me what this new text is attempting to cover:

```csharp
using System;
using System.Numerics;

static Func<float> M1(Vector4 x)
{
    Func<float> a = x.Length;
    return a;
}

static Func<float> M2(Vector4 x)
{
    var methodInfo = typeof(Vector4).GetMethod("Length");
    Func<float> a = (Func<float>)Delegate.CreateDelegate(typeof(Func<float>), x, methodInfo);
    return a;
}

Vector4 x = default;
var a1 = M1(x);
var a2 = M2(x);
Console.WriteLine(a1());
Console.WriteLine(a2());
```

teo-tsirpanis: It wasn't immediately clear from the issue's title, but the issue covers _static_ methods whose first argument is a struct (I think that's what "closed over their first argument" means). Your example failed to reproduce it because `Vector4.Length` is an _instance_ method.

Here is some code that fails. It throws an exception in `StructStatic` and does not in `ClassStatic` and `StructInstance`:

``` csharp
using System;
using System.Numerics;

static class TestClass {
	public record Vector4Ref(Vector4 Value);

	public static float GetLength(Vector4 x) => x.Length();
	public static float GetLengthRef(Vector4Ref x) => x.Value.Length();

	static Func<float> ClassInstance(Vector4Ref x)
	{
		var methodInfo = typeof(TestClass).GetMethod(nameof(GetLengthRef));
		return (Func<float>)Delegate.CreateDelegate(typeof(Func<float>), x, methodInfo);
	}

	static Func<float> StructInstance(Vector4 x)
	{
	    var methodInfo = typeof(Vector4).GetMethod(nameof(Vector4.Length));
	    Func<float> a = (Func<float>)Delegate.CreateDelegate(typeof(Func<float>), x, methodInfo);
	    return a;
	}
	
	static Func<float> StructStatic(Vector4 x)
	{
		var methodInfo = typeof(TestClass).GetMethod(nameof(GetLength));
		return (Func<float>)Delegate.CreateDelegate(typeof(Func<float>), x, methodInfo);
	}

	static void Main()
	{
		Vector4 x = default;
		Vector4Ref xRef = new(x);
		Console.WriteLine(ClassInstance(xRef)());
		Console.WriteLine(StructInstance(x)());
		Console.WriteLine(StructStatic(x)());
	}
}
```

Unless this is a runtime bug.

teo-tsirpanis: I even tried to make `GetLength` accept a `ref Vector4`, just in case it worked (because `this` in struct instance methods is a byref and who knows what would happen), but it still threw with the same error.

By the way this issue is mostly of theoretical nature and does not affect any of my code.

tannergooding: Looks like it comes down to https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/comdelegate.cpp#L2728.

@jkotas, do you know who the right person to review this documentation change is and potentially why this limitation exists?

jkotas, line 99: > not a boxed value type

I think you should drop this part. Boxed value type acts like a type reference. It actually works fine 
```suggestion
 When a delegate represents a static method, the delegate stores a reference to the method's entry point. When a delegate represents a static method closed over its first argument, the delegate stores a reference to the method's entry point and a reference to a target object assignable to the type of the method's first argument. When the delegate is invoked, the first argument of the static method receives the target object. This first argument must be a reference type.
```
boxed value type acts like a reference type. It actually works fine with boxed value types. E.g. this is using boxed value type and it works just fine:
```
using System;
using System.Numerics;

static class TestClass
{
	public static float GetLength(object x) => ((Vector4)x).Length();
	
	static Func<float> StructStatic(Vector4 x)
	{
		var methodInfo = typeof(TestClass).GetMethod(nameof(GetLength));
		return (Func<float>)Delegate.CreateDelegate(typeof(Func<float>), x, methodInfo);
	}

	static void Main()
	{
		Vector4 x = default;
		Console.WriteLine(StructStatic(x)());
	}
}
```
>  potentially why this limitation exists?

Too complicated to handle this corner case in the runtime. Not worth it.
https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/comdelegate.cpp#L2716

teo-tsirpanis: Done @jkotas.

jkotas (approved):

tannergooding (approved):

PR #5142, opened 2020-11-25
Minor spacing fix

peteraritchie:
## Summary

Added space after `**Note**` so that it does not run into `If`


mairaw (approved):
Thanks @peteraritchie!

PR #5151, opened 2020-12-03
Adding a little more detail on the descriptions

philipwolfe:
At first I didn't like that this referenced DataSet because it is irrelevant for parameter direction.  Then I decided to add some more description.

## Summary
Adding more detail.

### DIFF
--- a/xml/System.Data/ParameterDirection.xml
+++ b/xml/System.Data/ParameterDirection.xml
@@ -115,7 +115,7 @@
       </ReturnValue>
       <MemberValue>3</MemberValue>
       <Docs>
-        <summary>The parameter is capable of both input and output.  See the descriptions of input and output parameters.</summary>
+        <summary>The parameter is capable of both input and output. See the descriptions of input and output parameters.</summary>
       </Docs>
     </Member>
     <Member MemberName="Output">


xml/System.Data/ParameterDirection.xml
```diff
@@ -152,7 +152,7 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>The parameter is an output parameter.</summary>
+        <summary>The parameter is an output parameter.  This is useful when a stored procedure returns data rows and it is necessary to return additional values such as a sum total or number of errors.  An output parameter is different from a return value.  A stored procedure can use an output parameter and a return value to return different data.</summary>
```
roji, line 155: It's important to remember that these types are database-agnostic, and so we need to be careful not to include any database-specific information; in many databases, there's no distinction between output parameters and return values (indeed, the concepts themselves don't exist in many cases). So it's better not to go into too much detail. Output values are also used in various scenarios where the sproc doesn't return data rows at all.

```suggestion
        <summary>The parameter is an output parameter, which can be set by a stored procedure or function to return a value. This value is separate from and can be returned in addition to any data rows.</summary>
```

xml/System.Data/ParameterDirection.xml
```diff
@@ -189,7 +189,7 @@
       </ReturnValue>
       <MemberValue>6</MemberValue>
       <Docs>
-        <summary>The parameter represents a return value from an operation such as a stored procedure, built-in function, or user-defined function.</summary>
+        <summary>The parameter represents a return value from an operation such as a stored procedure, built-in function, or user-defined function.  Specifying a named parameter enables the <see cref="T:System.Data.IDbCommand" /> to locate and read the return value, even when data rows are also returned.</summary>
```
roji, line 192: * How is the parameter being named specifically relevant here? Also, named and non-named parameter are once again a database-specific detail (some databases don't support named parameters at all).
* We generally discourage use of the ADO.NET interfaces (e.g. `IDbCommand`), which have been largely replaced by the abstract base classes (`DbCommand`). This is because the interfaces could not be evolved with new APIs (e.g. async execution) without breaking existing implementation.

In general, the existing text here seems sufficient to me.

roji (changes requested):
Thanks for this doc change proposal, see comments below.

PR #5154, opened 2020-12-03
NOISSUE - Update the documentation relating to the setting of the DF bit

jmcconnell26:
in the IP header

## Summary

Updating the documentation to make the behaviour of the DontFragment value of the Socket clear.

### DIFF
--- a/xml/System.Net.Sockets/Socket.xml
+++ b/xml/System.Net.Sockets/Socket.xml
@@ -4154,9 +4154,9 @@ This method populates the <xref:System.Net.Sockets.Socket> instance with data ga
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets or sets a <see cref="T:System.Boolean" /> value that specifies whether the <see cref="T:System.Net.Sockets.Socket" /> sets the DF bit in the IP header, that allows Internet Protocol (IP) datagrams to be fragmented.</summary>
+        <summary>Gets or sets a value that specifies whether the <see cref="T:System.Net.Sockets.Socket" /> allows Internet Protocol (IP) datagrams to be fragmented.</summary>
         <value>
-          <see langword="true" /> if the <see cref="T:System.Net.Sockets.Socket" /> should set the DF bit in the IP Header, to forbid datagram fragmentation; otherwise, <see langword="false" />. The default is <see langword="true" />.</value>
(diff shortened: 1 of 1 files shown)


xml/System.Net.Sockets/Socket.xml
```diff
@@ -4154,9 +4154,9 @@ This method populates the <xref:System.Net.Sockets.Socket> instance with data ga
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets or sets a <see cref="T:System.Boolean" /> value that specifies whether the <see cref="T:System.Net.Sockets.Socket" /> allows Internet Protocol (IP) datagrams to be fragmented.</summary>
+        <summary>Gets or sets a <see cref="T:System.Boolean" /> value that specifies whether the <see cref="T:System.Net.Sockets.Socket" /> sets the DF bit in the IP header, that allows Internet Protocol (IP) datagrams to be fragmented.</summary>
```
gewarren, line 4157:
```suggestion
        <summary>Gets or sets a value that specifies whether the <see cref="T:System.Net.Sockets.Socket" /> allows Internet Protocol (IP) datagrams to be fragmented.</summary>
```

xml/System.Net.Sockets/Socket.xml
```diff
@@ -4154,9 +4154,9 @@ This method populates the <xref:System.Net.Sockets.Socket> instance with data ga
...
         <value>
-          <see langword="true" /> if the <see cref="T:System.Net.Sockets.Socket" /> allows datagram fragmentation; otherwise, <see langword="false" />. The default is <see langword="true" />.</value>
+          <see langword="true" /> if the <see cref="T:System.Net.Sockets.Socket" /> should set the DF bit in the IP Header, to forbid datagram fragmentation; otherwise, <see langword="false" />. The default is <see langword="true" />.</value>
```
gewarren, line 4159:
```suggestion
          <see langword="true" /> if the <see cref="T:System.Net.Sockets.Socket" /> doesn't allow datagram fragmentation; otherwise, <see langword="false" />. The default is <see langword="true" />.</value>
```

gewarren (commented):
@jmcconnell26 We don't normally include implementation details in the API docs. What's the reason for adding info about setting the DF bit?

jmcconnell26:
> @jmcconnell26 We don't normally include implementation details in the API docs. What's the reason for adding info about setting the DF bit?

Hi @gewarren, I had added the mention of the DF bit simply because it made it easier to understand. 
I'm happy to drop this if it goes against the guidelines!
I've pushed your recommended changes, if that looks good to you?

Thanks,
Josh

gewarren (approved):
Looks good, thanks!

PR #5160, opened 2020-12-08
Compare CompareTo return with zero in examples

chwarr:
In the examples for `IComparable<T>.CompareTo`, the comparison operator
implementations are checking that `Double.CompareTo` returns exactly -1
or 1. The contract for `IComparable<T>.CompareTo` is that the return
value is negative, zero, or positive. A valid implementation of
`IComparable<T>.CompareTo` can return 87 or -2844587, for example.

Compare to zero instead of checking for equality with -1 and 1.

---

The managed C++ example returns true for the following code when it
should return false:

    Temperature^ zero = gcnew Temperature(0);
    return zero > zero;

Fix the C++ `operator>` implementation.

### DIFF
--- a/samples/snippets/cpp/VS_Snippets_CLR/IComparable`1 Example/CPP/source.cpp
+++ b/samples/snippets/cpp/VS_Snippets_CLR/IComparable`1 Example/CPP/source.cpp
@@ -27,19 +27,19 @@ public:
        // Define the is greater than operator.
     bool operator>=  (Temperature^ other)
     {
-       return CompareTo(other) == 1;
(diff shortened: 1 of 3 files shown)


gewarren:
Preview URL: https://review.docs.microsoft.com/en-us/dotnet/api/system.icomparable-1.compareto?view=net-5.0&branch=pr-en-us-5160#System_IComparable_1_CompareTo__0_

@jeffhandley Are you the right person to review this? It looks good to me.

jeffhandley (approved):
Thanks for submitting these changes, @chwarr!

PR #5162, opened 2020-12-10
Add default timeout mention.

NN---:
## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Net.NetworkInformation/Ping.xml
+++ b/xml/System.Net.NetworkInformation/Ping.xml
@@ -64,7 +64,9 @@
  Network topology can determine whether <xref:System.Net.NetworkInformation.Ping> can successfully contact a remote host. The presence and configuration of proxies, network address translation (NAT) equipment, or firewalls can prevent <xref:System.Net.NetworkInformation.Ping> from succeeding. A successful <xref:System.Net.NetworkInformation.Ping> indicates only that the remote host can be reached on the network; the presence of higher level services (such as a Web server) on the remote host is not guaranteed.  
   
  This class provides functionality similar to the Ping.exe command line tool. The <xref:System.Net.NetworkInformation.Ping.Send%2A> and <xref:System.Net.NetworkInformation.Ping.SendAsync%2A> methods send an Internet Control Message Protocol (ICMP) echo request message to a remote computer and waits for an ICMP echo reply message from that computer. For a detailed description of ICMP messages, see RFC 792, available at [https://www.ietf.org](https://www.ietf.org/).  
-  
+ 
+ The default timeout when timeout is not specified is 5 seconds similar to Ping.exe command line tool timeout.
+
  The following types are used with the <xref:System.Net.NetworkInformation.Ping> class and are described in detail below.  
   
 |**Type name**|**Description**|  


xml/System.Net.NetworkInformation/Ping.xml
```diff
@@ -64,7 +64,9 @@
  Network topology can determine whether <xref:System.Net.NetworkInformation.Ping> can successfully contact a remote host. The presence and configuration of proxies, network address translation (NAT) equipment, or firewalls can prevent <xref:System.Net.NetworkInformation.Ping> from succeeding. A successful <xref:System.Net.NetworkInformation.Ping> indicates only that the remote host can be reached on the network; the presence of higher level services (such as a Web server) on the remote host is not guaranteed.  
   
  This class provides functionality similar to the Ping.exe command line tool. The <xref:System.Net.NetworkInformation.Ping.Send%2A> and <xref:System.Net.NetworkInformation.Ping.SendAsync%2A> methods send an Internet Control Message Protocol (ICMP) echo request message to a remote computer and waits for an ICMP echo reply message from that computer. For a detailed description of ICMP messages, see RFC 792, available at [https://www.ietf.org](https://www.ietf.org/).  
-  
+ 
+ The default timeout when timeout is not specified is 5 seconds similar to Ping.exe command line tool timeout.
```
ManickaP, line 68: Could you please link a source of this information? Is it also true on *nixes?

ManickaP (commented):
What's driving this addition? Is there any opened issue to which is this change related?

NN---:
It is directly from the .NET code.

https://github.com/dotnet/runtime/blob/master/src/libraries/System.Net.Ping/src/System/Net/NetworkInformation/Ping.cs#L15

xml/System.Net.NetworkInformation/Ping.xml
wfurt: I see no reason to add reference to ping.exe for public API doc. stating the default should be sufficient IMHO.

ManickaP:
This is not true though:
- Ping.exe says 4000 ms = 4 seconds: https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/ping#parameters ==> so it is not like ping.exe.
- The timeout also depends on whether you can send ICMP over raw sockets or not. For example following code will timeout in 3 seconds on my Linux box:
```C#
var p = new Ping();
var sw = Stopwatch.StartNew();
var x = p.Send(new IPAddress(new byte[]{192, 168, 0, 42}));
Console.WriteLine(sw.Elapsed);
Console.WriteLine(x.Status);
```
However, it will timeout in 5 seconds if ran under `sudo`, because that will give it permission to send ICMP directly.

@wfurt could you please chime in here?

Also I digged up the original issue in runtime repo: https://github.com/dotnet/runtime/issues/45888 It would be nice to include it in the description with a small text explaining why is this change happening next time.

xml/System.Net.NetworkInformation/Ping.xml
ManickaP: I didn't mean link it here, I was curious where the 5 seconds came from.

PR #5164, opened 2020-12-11
InterlockedAdd -- clarify atomic operation result value

glenn-slayden:
## Summary

The word "new" describing the return value seemed odd to me, since the whole point of `Interlocked` is that such values may not still be prevailing. I couldn't settle on a wording that clarifies that the entire notion of "new value stored" is a poor choice here, since lock-free operations are necessarily and inherently only retrospectively (past-looking) correct. Perhaps *...value that was stored at the time the operation was applied.*?

So I was also trying to figure out a concise way to mention that, if you need to know the stored value *to which this operation was atomically applied*, the only way to recover that is by subtracting the value you supplied from the return value, that is, by reversing your operation, after the fact.

### DIFF
--- a/xml/System.Threading/Interlocked.xml
+++ b/xml/System.Threading/Interlocked.xml
@@ -143,7 +143,7 @@
         <param name="location1">A variable containing the first value to be added. The sum of the two values is stored in <paramref name="location1" />.</param>
         <param name="value">The value to be added to the integer at <paramref name="location1" />.</param>
         <summary>Adds two 32-bit integers and replaces the first integer with the sum, as an atomic operation.</summary>
-        <returns>The new value stored at <paramref name="location1" />.</returns>
+        <returns>The new value that was stored at <paramref name="location1" /> exclusively as a direct result of this operation.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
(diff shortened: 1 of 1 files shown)


carlossanlop:
Ping @kouvel  can you please take a look?

kouvel:
Makes sense to me, though "exclusively" seems a bit confusing.

How about a simpler alternative: "The new value that was stored at ... exclusively as a direct result of this operation." -> "The new value that was stored at ... by this operation."

glenn-slayden:
> I couldn't figure out the best area label to add to this PR. If you have write-permissions please help me learn by adding exactly one area label.

I assume I don't have write-permissions.

glenn-slayden:
> Makes sense to me, though "exclusively" seems a bit confusing.
> 
> How about a simpler alternative: "The new value that was stored at ... exclusively as a direct result of this operation." -> "The new value that was stored at ... by this operation."

That sounds good to me; it's maintains the improvment.

glenn-slayden:
Any progress on this? I'm hoping to delete my repo fork which is only in support of this pr.

kouvel:
Seems like I don't have permissions to add a commit with the update above to your branch. I have opened https://github.com/dotnet/dotnet-api-docs/pull/6869 in my fork with the update. Thanks!

PR #5171, opened 2020-12-14
FieldInfo - format output as code

MSDN-WhiteKnight:
Format console output as code to prevent localization


gewarren (approved):
@MSDN-WhiteKnight Thanks, the preview looks good.

PR #5194, opened 2020-12-24
Fix misplaced content string in Window1.xaml

Scrivener07:
## Summary

I moved the grid view column header string to a `Content` parameter.


SamBent:
This change isn't needed.  The lines 
```
<GridViewColumnHeader>Last Name</GridViewColumnHeader>
```
and
```
<GridViewColumnHeader Content="Last Name"></GridViewColumnHeader>
```
do the same thing.  The first form is probably more common in practice - no reason not to use it in the snippet.

For more detail, see  [ContentPropertyAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.windows.markup.contentpropertyattribute?view=net-5.0).

gewarren:
Closing per Sam's comment.

PR #5199, opened 2021-01-05
Add a sample for N0 specifier

KirillOsenkov:
N0 specifier is commonly needed, but not mentioned anywhere. I had to dig through StackOverflow to find it.


gewarren (approved):

tannergooding (approved):
LGTM.

However, perhaps it would be good for the samples to have a separate section for `*0` and `*#` for all specifiers.

That is `F0` can also be interesting, as can `F1` or `F3` to differentiate what is "default" vs "explicitly zero" vs "explicitly some other number"

gewarren:
Should we merge this as-is or wait for the addition of a separate section for *0 and *# for all specifiers?

tannergooding:
> Should we merge this as-is or wait for the addition of a separate section for *0 and *# for all specifiers?

I don't have a preference. The current change is additive and unblocks the scenario Kirill cares about.

I believe my proposal to extend to covering explicit `*0` and `*#` would also be additive and is mostly nice to have.

KirillOsenkov:
Let’s merge this as is and Tanner can send a separate PR for his suggestion ;)

pgovind (approved):
LGTM. I think we can merge this and have Tanner's suggestion in a follow up PR too

PR #5201, opened 2021-01-05
Add remark that StringSplitOptions.TrimEntries is only available in .NET 5

meum:


tannergooding:
@gewarren, is this something that could be better represented by the docs page?

In particular, this does only show on the .NET 5 page (https://docs.microsoft.com/en-us/dotnet/api/system.stringsplitoptions?view=net-5.0) and not the .NET Core 3.1 page (https://docs.microsoft.com/en-us/dotnet/api/system.stringsplitoptions?view=netcore-3.1).
However, most other members (like methods and constructors struct/class fields) have an `Applies to` section, while `enum` members in particular do not.

gewarren:
@tannergooding We don't show remarks for enum fields either, mainly because there's not really a place to put them. Currently they are only displayed in the table; they don't have individual pages.

For this note/remark, it might be better to include it as part of the summary for the `TrimEntries` field so it's more "associated".

tannergooding:
> For this note/remark, it might be better to include it as part of the summary for the TrimEntries field so it's more "associated"

I think this sounds fine. My only concern is that it will show up in IntelliSense as well and is a (I believe) new convention.

> We don't show remarks for enum fields either, mainly because there's not really a place to put them

Do you have any concern with me opening an issue suggesting we also have a doc page per enum field, similarly to what we do for struct/class fields?
It seems unfortunate that there is no where to put this information when it does exist.

gewarren:
> Do you have any concern with me opening an issue suggesting we also have a doc page per enum field, similarly to what we do for struct/class fields?

I don't have a concern with that. You can open a customer suggestion here and also link it to this one: https://ceapex.visualstudio.com/Engineering/_workitems/edit/325763.

carlossanlop:
@gewarren I see the internal issue is closed, and it is linked to another issue with the commit that fixed it.
Is there a process we need to follow to get enum member remarks to show up in dotnet-api-docs xml files? I am unfamiliar with the docs dev process.

gewarren:
> @gewarren I see the internal issue is closed, and it is linked to another issue with the commit that fixed it.
> Is there a process we need to follow to get enum member remarks to show up in dotnet-api-docs xml files? I am unfamiliar with the docs dev process.

@carlossanlop That internal issue was fixed; I just used it here as an example that you could "Create a copy of work item" from, since it's quite similar to what you want to request. The process is simply that - to create a customer suggestion. I usually always click "Follow" after I create it so I can track any updates. Do you want me to create it instead?

xml/System/StringSplitOptions.xml
```diff
@@ -55,6 +55,8 @@ The <xref:System.String.Split%2A?displayProperty=nameWithType> method returns an
 
 Specify the `None` field to invoke the default behavior of the <xref:System.String.Split%2A?displayProperty=nameWithType> method, which is not to trim white-space characters and to include empty substrings.
 
+The `TrimEntries` field is only available in .NET 5.
```
gewarren, line 58:
```suggestion
The `TrimEntries` field is only available in .NET 5 and later versions.
```

PR #5241, opened 2021-01-15
Fix WsHttpBinding.Security.Transport docs

florianknecht:
The default value of WsHttpBinding.Security.Transport.ClientCredentialType is actually HttpClientCredentialType.Windows, when no explicit HttpTransportSecurity object is provided.
This is what I found when migrating an exisiting BasicHttpBinding to WSHttpBinding - my client would throw errors containing HTTP 401 Negotiate errors.

If I create a new HttpTransportSecurity object, its default value is indeed HttpClientCredentialType.None.

Can you verify this behaviour?


mconnew (approved):
The submitted doc change correctly reflects the code behavior

gewarren:
@florianknecht Thank you!

PR #5247, opened 2021-01-16
Updating FileSystem.AccessControl examples

carlossanlop:
- Make sure the samples are using the latest .NET 5.0 features
- Move them to sample files instead of having them directly written in the remarks

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR/IO.FileSystem.AccessControl/DirectoryInfoCreateDirectory.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR/IO.FileSystem.AccessControl/DirectoryInfoCreateDirectory.cs
@@ -14,7 +14,9 @@ namespace MyNamespace
         {
             // Create the file security object
 
-            SecurityIdentifier identity = new SecurityIdentifier(WellKnownSidType.BuiltinUsersSid, null);
+            SecurityIdentifier identity = new SecurityIdentifier(
+                WellKnownSidType.BuiltinUsersSid, // This maps to "Everyone" user group in Windows
+                null); // null is OK for this particular user group. For others, a non-empty value might be required
             FileSystemAccessRule accessRule = new FileSystemAccessRule(identity, FileSystemRights.FullControl, AccessControlType.Allow);
 
             DirectorySecurity expectedSecurity = new DirectorySecurity();
--- a/samples/snippets/csharp/VS_Snippets_CLR/IO.FileSystem.AccessControl/FileInfoCreate.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR/IO.FileSystem.AccessControl/FileInfoCreate.cs
@@ -45,24 +45,7 @@ namespace MyNamespace
                 byte[] writeBuffer = new UTF8Encoding(encoderShouldEmitUTF8Identifier: true).GetBytes(text);
 
                 stream.Write(writeBuffer, 0, writeBuffer.Length);
-            } // Dispose flushes
-
-            // Reopen the file in new stream and read the contents to confirm they are correct
-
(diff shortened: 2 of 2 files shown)


gewarren (approved):
LGTM - [preview link](https://review.docs.microsoft.com/en-us/dotnet/api/system.io.filesystemaclextensions.create?view=net-5.0&branch=pr-en-us-5247#System_IO_FileSystemAclExtensions_Create_System_IO_DirectoryInfo_System_Security_AccessControl_DirectorySecurity_).

samples/snippets/csharp/VS_Snippets_CLR/IO.FileSystem.AccessControl/DirectoryInfoCreateDirectory.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class MyClassCS
+    {
+        // Attribute to address CA1416 warning:
+        // System.IO.FileSystem.AccessControl APIs are  only available on Windows
+        [SupportedOSPlatform("windows")]
+        static void Main()
+        {
+            // Create the file security object
+
+            SecurityIdentifier identity = new SecurityIdentifier(WellKnownSidType.BuiltinUsersSid, null);
```
adamsitnik, line 17: I would add a comment explaining what `BuiltinUsersSid` is and why the 2nd parameter can be null. 

```suggestion
            SecurityIdentifier identity = new SecurityIdentifier(
                WellKnownSidType.BuiltinUsersSid, // this maps to "Everyone" user group in Windows
                null); // null is OK for this particualr user group, for others a non empty value might be required
```

samples/snippets/csharp/VS_Snippets_CLR/IO.FileSystem.AccessControl/FileInfoCreate.cs
```diff
@@ -0,0 +1,68 @@
...
+                {
+                    Console.WriteLine(encoding.GetString(readBuffer));
+
+                    /*
+                        Output:
+                            Hello world!
+                    */
+                }
+            }
+        }
```
adamsitnik, line 66: the description of this example says:

> The following code example creates a new text file inside the user's temporary folder, explicitly specifying all security attributes

we should rather focus on the security aspect here (passing `FileSecurity` to the right method) and just create the file, without writing and reading the "Hello World". It just adds a lot of code to the example and makes it more complex and harder to understand. Just keep it simple.

samples/snippets/csharp/VS_Snippets_CLR/IO.FileSystem.AccessControl/DirectoryInfoCreateDirectory.cs
carlossanlop: Sounds good. I'll add it. I adjusted a small typo.

samples/snippets/csharp/VS_Snippets_CLR/IO.FileSystem.AccessControl/FileInfoCreate.cs
carlossanlop: I see what you mean. I wanted to show that the file was still readable after altering the file security. But sounds more like something we would do in a unit test, not in an example.

samples/snippets/csharp/VS_Snippets_CLR/IO.FileSystem.AccessControl/FileInfoCreate.cs
```diff
@@ -0,0 +1,68 @@
...
+                while (stream.Read(readBuffer, 0, readBuffer.Length) > 0)
+                {
+                    Console.WriteLine(encoding.GetString(readBuffer));
+
+                    /*
+                        Output:
+                            Hello world!
+                    */
+                }
+            }
```
carlossanlop, line 65:
```suggestion
            } // Dispose flushes
```

Issue #5253, opened 2021-01-18
Obsoletion text in topic stopped not being rendered for .NET 5.0+


danmosemsft:
Eg., for [Process.NonpagedSystemMemorySize Property](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.nonpagedsystemmemorysize?view=net-5.0) there is no indication it is obsolete (and has been for 10-15 years)

But if I change to the older versions I do see it: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.nonpagedsystemmemorysize?view=netcore-2.2. The cause is that `FrameworkAlternate` attribute does not include `net-5.0`. If I edit the HTML to add that, it appears rendered.

Does a script need to be written to add 'net-5.0' to all such tags in all topics? 

@gewarren

danmosemsft:
https://github.com/dotnet/runtime/pull/47137/files

gewarren:
@danmosemsft This is a [bug in "mdoc"](https://ceapex.visualstudio.com/Engineering/_workitems/edit/322199) - the tool that generates the API docs. Hopefully it will be fixed and released in the next week, and we'll update the docs as soon as the fix is available.

Related to #5181.

danmosemsft:
Oh missed the dupe thanks

PR #5254, opened 2021-01-18
Clarify Process.XX obsoletion

danmoseley:
Match https://github.com/dotnet/runtime/pull/47137 

Note it doesn't show up right now because of bug https://github.com/dotnet/dotnet-api-docs/issues/5253


danmosemsft:
I already updated the code to use the same string in the attributes.

gewarren (approved):
@danmosemsft This XML is auto-generated so it will be overwritten next time we refresh the .NET API ref docs. But for now, I think it's okay to merge this. Presumably it will be overwritten with the exact same text.

danmosemsft:
Oh, I didn't know that.

PR #5256, opened 2021-01-19
Removing unnecessary class creation

Israfil22:
## Summary

Alice's class code contatin unnecessary operation - repeated generating CngKey class on 21 line.
20 line already contain needed CngKey, that should need to be used on next line.

## Changes

Class creation on 21 line was replaced with key on 20 line.


eiriktsarpalis:
cc @bartonjs @GrabYourPitchforks

bartonjs (approved):

PR #5272, opened 2021-01-22
Fix typo in HttpMessageHandler.xml

ericmutta:
The sentence:

> There are various HTTP message handles that can be used.

Should read:

> There are various HTTP message handlers that can be used.


gewarren (approved):
@ericmutta Thanks!

PR #5273, opened 2021-01-22
There's no "key" argument in Queue.TryDequeue

R2D221:
## Summary

The documentation says TryDequeue can throw an ArgumentNullException if `key` is null, but there's no `key` argument in the method signature to begin with.


eiriktsarpalis (approved):
Seems reasonable, thanks!

PR #5286, opened 2021-01-27
Update Image.xml

danmoseley:
Fix https://github.com/dotnet/runtime/issues/46405

### DIFF
--- a/xml/System.Drawing/Image.xml
+++ b/xml/System.Drawing/Image.xml
@@ -723,6 +723,7 @@
  -or-  
   
  <paramref name="stream" /> is <see langword="null" />.</exception>
+        <exception cref="T:System.OutOfMemoryException">The stream does not have a valid image format.</exception>
       </Docs>
     </Member>
     <Member MemberName="FromStream">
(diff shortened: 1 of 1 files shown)


xml/System.Drawing/Image.xml
```diff
@@ -838,6 +840,7 @@
  ]]></format>
         </remarks>
         <exception cref="T:System.ArgumentException">The stream does not have a valid image format.</exception>
+        <exception cref="T:System.OutOfMemoryException">The stream does not have a valid image format.</exception>
```
gewarren, line 843: What's the difference between ArgumentException and OOMException?

danmosemsft: I believe it depends on the way the format is invalid. Eg., at a guess maybe if the file claims its huge and that would require us to allocate too much memory it might be OOME, if it's got an unexpected header maybe it's AE. I don't know exactly, and it may well vary between Unix and Windows - I don't know that it's by design.

danmosemsft: We would probably be comfortable with a change that made it throw AE consistently -- I don't think that would be too breaking -- but we don't have any immediate plans to do it, hence this PR.

safern (approved):

PR #5300, opened 2021-02-02
mdoc 5.8.1 preview II

gewarren:


xml/System.Diagnostics/Process.xml
```diff
@@ -2662,6 +2663,10 @@ The calling process is a member of the associated process' descendant tree.</exc
         <AssemblyVersion>2.1.0.0</AssemblyVersion>
       </AssemblyInfo>
       <Attributes>
+        <Attribute FrameworkAlternate="net-5.0">
+          <AttributeName Language="C#">[set: System.Runtime.Versioning.SupportedOSPlatform("windows")]</AttributeName>
```
gewarren, line 2667: macos and freebsd should be listed here too - see https://github.com/dotnet/runtime/blob/ad407d2be23dce89caf1e4ee5fe30c93b9a6a010/src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.cs#L268.

PR #5310, opened 2021-02-04
Fix complete view of the MMF description

KubaSzostak:
> To create a complete view of the memory-mapped file, specify 0 (zero) for the size parameter. If you do this, the size of the view might be **smaller** or larger than the size of the source file on disk. This is because views are provided in units of system pages, and the size of the view is **rounded up to the next** system page size.

Above statement has a small mistake. The view can't be smaller. It is indeed **rounded up** to the page size.

## Summary

Describe your changes here.


eiriktsarpalis:
cc @adamsitnik @carlossanlop @jozkee

adamsitnik:
Hi @kubaszostak 

Big thanks for your contribution! There was another PR opened to address this issue (#4297) and it got already merged, so I am afraid that I have to close this one.

Thanks,
Adam

PR #5313, opened 2021-02-04
Fix another typo in Task.xml

ericmutta:
The sentence:

> ...each of which sleeps for an interval determine by a random number generator. 

Should read:

> ...each of which sleeps for an interval determined by a random number generator.


adamsitnik (approved):
LGTM, thank you @ericmutta

PR #5320, opened 2021-02-10
Bring in changes from new mdoc 5.8.1

gewarren:
- Improved nullable types handling
- Improved obsolete attributes handling

PR #5322, opened 2021-02-10
Don't filter new platform attributes

gewarren:
We can't verify that any of these display properly until #5320 is merged (which brings in these attribute decorations for .NET 5).


buyaa-n (approved):
Thanks!

PR #5323, opened 2021-02-11
Fix typo in Span<T>.xml

ericmutta:
The following sentence:

> ...methods of the `Span<T>` **class**...

Uses the word `class` to refer to the `Span<T>` type which is in fact a (ref) `struct`. I have changed it to read as follows:

> ...methods of the `Span<T>` **type**...


ericmutta:
Many thanks for following up @adamsitnik :+1:

PR #5325, opened 2021-02-11
Update StructLayoutAttribute.xml

### LINKED ISSUE #5318: Incorrect statement in Remarks section of StructLayoutAttribute.Pack documentation
The Remarks section of the `StructLayoutAttribute.Pack` documentation contains the following statement regarding the field:

> It affects both `LayoutKind.Sequential` and `LayoutKind.Explicit`.

https://github.com/dotnet/dotnet-api-docs/blob/733b2cecdd45ab3ed5efee2f001d9b07730f06e1/xml/System.Runtime.InteropServices/StructLayoutAttribute.xml#L265

According to a comment in an issue on the dotnet/runtime repo, `StructLayoutAttribute.Pack` does not apply to `LayoutKind.Explicit`:

> Pack should not have any affect in combination with `LayoutKind.Explicit`.

https://github.com/dotnet/runtime/issues/47907#issuecomment-774719848

I recommend that the statement in the Remarks section be updated to reflect that the `StructLayoutAttribute.Pack` field only affects `LayoutKind.Sequential`.

WizardBrony:
Fix remark about what `LayoutKind`s `StructLayoutAttribute.Pack` affects (only `Sequential`)

## Summary

Removed mention of `StructLayoutAttribute.Pack` affecting `LayoutKind.Explicit`.

Fixes #5318

### DIFF
--- a/xml/System.Runtime.InteropServices/StructLayoutAttribute.xml
+++ b/xml/System.Runtime.InteropServices/StructLayoutAttribute.xml
@@ -262,7 +262,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Runtime.InteropServices.StructLayoutAttribute.Pack> field controls the alignment of a type's fields in memory.  It affects both <xref:System.Runtime.InteropServices.LayoutKind.Sequential?displayProperty=nameWithType> and <xref:System.Runtime.InteropServices.LayoutKind.Explicit?displayProperty=nameWithType>. By default, the value is 0, indicating the default packing size for the current platform. The value of <xref:System.Runtime.InteropServices.StructLayoutAttribute.Pack> must be 0, 1, 2, 4, 8, 16, 32, 64, or 128:  
+ The <xref:System.Runtime.InteropServices.StructLayoutAttribute.Pack> field controls the alignment of a type's fields in memory.  It affects <xref:System.Runtime.InteropServices.LayoutKind.Sequential?displayProperty=nameWithType>. By default, the value is 0, indicating the default packing size for the current platform. The value of <xref:System.Runtime.InteropServices.StructLayoutAttribute.Pack> must be 0, 1, 2, 4, 8, 16, 32, 64, or 128:  
   
  The fields of a type instance are aligned by using the following rules:  
   


xml/System.Runtime.InteropServices/StructLayoutAttribute.xml
```diff
@@ -262,7 +262,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Runtime.InteropServices.StructLayoutAttribute.Pack> field controls the alignment of a type's fields in memory.  It affects both <xref:System.Runtime.InteropServices.LayoutKind.Sequential?displayProperty=nameWithType> and <xref:System.Runtime.InteropServices.LayoutKind.Explicit?displayProperty=nameWithType>. By default, the value is 0, indicating the default packing size for the current platform. The value of <xref:System.Runtime.InteropServices.StructLayoutAttribute.Pack> must be 0, 1, 2, 4, 8, 16, 32, 64, or 128:
```
AaronRobinsonMSFT, line 265: Unsure how to clarify this but it does impact both at least at the metadata layer - [sharp lab](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxASwDYB8ACAmARgFgAoHABgAIcCA6AJQFcA7GNMOWgSVcQHsADgGVEANzQBjOFADcpHAGZqeSgGFKAb1KUdlANpCYCRhJgAZCAE8+jGAAoL12wGk0zACa0AogA8BGSTQYJEoABQgJAGtKAF5KAA4ASgBdbV1FSlhjU0pfATSdLRJdEv0AMTQ4DHcAeQAzOqg4ewAWFILSjLcYSgBBOWLdAF8OgyMTcysbe0dp1w9aEQBHRjhWNAgMEPCo2Mo21MGdDKyJymWOotL0pW6+gZKRkiGgA). Regarding how the runtime interprets that metadata, I believe the statement is correct. I consider the documentation to describe the contract between API and metadata not runtime behavior per se. Clarifying the distinction seems appropriate but removing any mention of `Explicit` isn't correct in my opinion.

AaronRobinsonMSFT (changes requested):

xml/System.Runtime.InteropServices/StructLayoutAttribute.xml
WizardBrony: @AaronRobinsonMSFT

The only reason it impacts both is because of a legacy bug in the C# compiler:

https://github.com/dotnet/roslyn/issues/51151

The runtime just happens to handle this scenario correctly by ignoring `.pack`. My personal opinion is that the documentation should describe actual behavior. But if it is desired to leave mention of `Explicit` because it affects the metadata, I would at least recommend mentioning that it is only as a result of a bug that can't be fixed without a breaking change, especially since this artifact violates ECMA-335.

AaronRobinsonMSFT: From the referenced issue:

> According to the ECMA CLI Spec:

> > The .pack directive shall not be supplied for any type with explicit layout control.

The spec is sufficient for me.

>My personal opinion is that the documentation should describe actual behavior.

I disagree strongly. The spec defines the behavior not the implement because this is an API definition. In this case the spec clearly defines this so I agree it should be documented as such. Thank you for the spec reference.

AaronRobinsonMSFT (approved):

xml/System.Runtime.InteropServices/StructLayoutAttribute.xml
WizardBrony: > The spec defines the behavior not the implement because this is an API definition. In this case the spec clearly defines this so I agree it should be documented as such.

That's fair. Thanks for the discussion.

PR #5334, opened 2021-02-13
Add paramref in returns-comment.

ulfakerstedt:
Add paramref to the referenced type parameter TSource in the overload without equality comparer of ToHashSet<TSource>'s returns-comment.
(The overload with the equality comparer already has the type parameter enclosed in a paramref tag.)
(Perhaps it should be a typeparamref? However paramref seems to be used everywhere else.)


eiriktsarpalis (approved):

gewarren:
Thank you @ulfakerstedt

PR #5335, opened 2021-02-15
null cannot be boxed

sgurukrupa:
The use of the term boxing is not appropriate in case of 'null'.

## Summary

Describe your changes here.


BillWagner (approved):
I like this change @sgurukrupa It's more correct.

I'll :shipit: now.

PR #5350, opened 2021-02-18
Fix .NET 5.0 C# bool.TryParse signature.

ILMTitan:
In `bool.TryParse`, `string value` was correctly updated to `string? value`, but `out bool result` was incorrectly updated to `out bool? result`.

## Summary

Remove improper nullable modifier.

See issue #5351


gewarren (approved):
@ILMTitan Thanks for fixing this. This XML is autogenerated so it will get overwritten in the next release (but hopefully correctly). But for now, I'll merge your fix.

PR #5359, opened 2021-02-22
Fix default value for DllImportAttribute.PreserveSig

kant2002:
## Summary

See for proof.
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/DllImportAttribute.cs#L22

Another information which hint that likely this documentation reference is referred to Metadata PreserveSig value.
https://github.com/dotnet/runtime/blob/e6453ecf2ef5ca92d52f8b3bfa364bc832c3a4e8/src/mono/System.Private.CoreLib/src/System/Reflection/Emit/MethodBuilder.Mono.cs#L433-L434

### DIFF
--- a/xml/System.Runtime.InteropServices/DllImportAttribute.xml
+++ b/xml/System.Runtime.InteropServices/DllImportAttribute.xml
@@ -488,7 +488,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- Set the <xref:System.Runtime.InteropServices.DllImportAttribute.PreserveSig> field to `true` to translate unmanaged signatures with `HRESULT` values directly; set it to `false` to automatically convert `HRESULT` return values to exceptions. By default, the <xref:System.Runtime.InteropServices.DllImportAttribute.PreserveSig> field is `true`.  
+ Set the <xref:System.Runtime.InteropServices.DllImportAttribute.PreserveSig> field to `true` to translate unmanaged signatures with `HRESULT` values directly; set it to `false` to automatically convert `HRESULT` return values to exceptions. By default, the <xref:System.Runtime.InteropServices.DllImportAttribute.PreserveSig> field is `false`.  
   
  When `true`, the managed method signature returns an integer value that contains the `HRESULT` value.  In this case, you must manually inspect the return value and respond accordingly in your application.  
   


xml/System.Runtime.InteropServices/DllImportAttribute.xml
```diff
@@ -488,7 +488,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- Set the <xref:System.Runtime.InteropServices.DllImportAttribute.PreserveSig> field to `true` to translate unmanaged signatures with `HRESULT` values directly; set it to `false` to automatically convert `HRESULT` return values to exceptions. By default, the <xref:System.Runtime.InteropServices.DllImportAttribute.PreserveSig> field is `true`.  
+ Set the <xref:System.Runtime.InteropServices.DllImportAttribute.PreserveSig> field to `true` to translate unmanaged signatures with `HRESULT` values directly; set it to `false` to automatically convert `HRESULT` return values to exceptions. By default, the <xref:System.Runtime.InteropServices.DllImportAttribute.PreserveSig> field is `false`.
```
jkotas, line 491: I do not think that this is change is right.

This field is `true` by default when one is using this attribute. For example, if you write `[DllImport("my.dll")]`, it will result into `PreserveSig` set to true.

jkotas:
> See for proof.
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/DllImportAttribute.cs#L22

DllImportAttribute is one of the special pseudo-custom attributes. It is special cased by the compilers and it has special encoding in the metadata. This special casing sets it to `true` by default.

jkotas:
> Another information which hint that likely this documentation reference is referred to Metadata PreserveSig value.
https://github.com/dotnet/runtime/blob/e6453ecf2ef5ca92d52f8b3bfa364bc832c3a4e8/src/mono/System.Private.CoreLib/src/System/Reflection/Emit/MethodBuilder.Mono.cs#L433-L434

Reflection.Emit mimics the special handling of this attribute that is done by the compilers. This comment is talking about exactly that. The value used by default (e.g. what you get from `[DllImport("my.dll")]`) is true vs. the default value (e.g. what you get from `new DllImportAttribute(...)`) is false. I agree that this is a discrepancy, but I do not think the documentation needs to be highlighting it. `new DllImportAttribute(...)` is extremely rare.

PR #5365, opened 2021-02-23
Update ZipFile.xml

PalashBansal:
I am using 4.0.0.0 and ZipFile is in System.IO.Compression.FileSystem.dll
image

### DIFF
--- a/xml/System.IO.Compression/ZipFile.xml
+++ b/xml/System.IO.Compression/ZipFile.xml
@@ -7,7 +7,6 @@
   <TypeSignature Language="F#" Value="type ZipFile = class" />
   <AssemblyInfo>
     <AssemblyName>System.IO.Compression.ZipFile</AssemblyName>
-    <AssemblyVersion>4.0.0.0</AssemblyVersion>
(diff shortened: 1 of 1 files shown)


MSDN-WhiteKnight:
First, the XML metadata is automatically generated by mdoc, and should not be edited manually, because changes will be overwritten by CI (the only manually-editable content is between here is between the `<Docs>` tags). Second, are you sure this is correct change to remove that tag? What issue it solves? Currently, the docs site correctly shows that ZipFile lives in System.IO.Compression.FileSystem.dll for .NET Framework and in System.IO.Compression.ZipFile.dll for .NET 5. What .NET version this "4.0.0.0" is corresponding to? In NET 4.0 ZipFile class is not available at all.

PalashBansal:
@MSDN-WhiteKnight Right! I am using 4.7.1 .net framework as can be seen from the screenshot. Switched to this version in docs and it shows correct assembly i.e. FileSystem.dll
Closing this PR.
PS: Still wondering why assembly version 4.0.0.0 is appearing twice for both FileSystem.dll and Zipfile.dll in ZipFile.xml

PR #5368, opened 2021-02-24
Fix duplicated 'for' in MemoryExtensions.xml

jo-ninja:
## Summary

Fixes duplicated "for" in this sentence:

> Provides extension methods for for the memory- and span-related types...


adamsitnik (approved):
LGTM, thank you @jo-ninja

PR #5377, opened 2021-02-25
Update codeowners file to match runtime repo

gewarren:
Updating based on https://github.com/dotnet/runtime/blob/master/docs/area-owners.md.
Also replaces `**/**` syntax that seems to be invalid.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -10,7 +10,6 @@
 
 *     @dotnet/docs
 
-
 # Order is important. The last matching pattern has the most precedence.
 # The folders are ordered as follows:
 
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -19,80 +18,105 @@
...
-/xml/System.Threading**/** @kouvel
-/xml/System.Threading.Channels**/** @tarekgh @stephentoub
-/xml/System.Threading.Tasks**/** @tarekgh @stephentoub
-/xml/System.Transactions**/** @dasetser @HongGit
-/xml/System.Web**/** @scottaddie
-/xml/System.Windows**/** @dotnet/wpf-developers
-/xml/System.Windows.Forms**/** @RussKie @merriemcgaw
-/xml/System.Xaml**/** @dotnet/wpf-developers
-/xml/System.Xml**/** @buyaa-n @krwq
+/xml/Microsoft.CSharp*/ @cston @333fred
```
carlossanlop, line 21: Does the suffix `*/` also include any files inside that folder, as opposed to using `*/*`?

And I know we don't have any folders inside `xml/Namespace/`, only xmls, so it makes sense to not use `**`.

.github/CODEOWNERS
```diff
@@ -19,80 +18,105 @@
...
+/xml/Microsoft.Extensions*/ @maryamariyan @michaelgsharp @safern @tarekgh
```
eiriktsarpalis, line 22: this would need to be updated every time area-owners.md is changed. Any chance we could consolidate into one document? (e.g. by adding a column linking to the relevant docs folders in area-owners.md)

cc @carlossanlop @jeffhandley

.github/CODEOWNERS
gewarren: Yes, I believe `*/` includes files inside that folder (including in subdirectories, according to [this](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax)).

We think it's invalid because it's not currently functional in requesting PR reviews. And the link above states:

"If any line in your CODEOWNERS file contains invalid syntax, the file will not be detected and will not be used to request reviews."

.github/CODEOWNERS
gewarren: I will leave it up to you and Carlos to decide how reviewers are tagged. If we decide to rely on fabricbot, I think we should just delete this file.

carlossanlop (approved):

gewarren:
Looks like it's working correctly now. Yippee!

https://github.com/dotnet/dotnet-api-docs/pull/5390

image

PR #5382, opened 2021-02-27
Added clarifying examples in Path.xml #3579

### LINKED ISSUE #3579: System.IO.Path documentation claims that some platforms limit file extensions to three characters
"The format of a file name extension is platform-dependent; for example, some systems limit extensions to three characters, and others do not."

https://github.com/dotnet/dotnet-api-docs/blob/3a6c7ff2a7723d194b2d3109a47f42d3bf2bbfed/xml/System.IO/Path.xml#L44

The last major desktop platform that limited file extensions to three characters was Windows 3.x. .NET, neither Framework nor Core, has never supported it. I'm not aware of any other platforms .NET has ever supported with a three-character limit either.

As such, I suggest removing the remark about three-character extensions.

shdvr:
Added examples to clarify when a file extension might be limited to three characters (issue #3579).

## Summary

Proposed wording change to improve clarity based on the discussion from the issue.

Fixes #3579


adamsitnik (approved):
LGTM, thank you @shdvr !

PR #5400, opened 2021-03-09
HttpCapabilitiesBase code example update

OwainWilliams:
The previous example created an error:
Cannot implicitly convert type 'System.Web.HttpBrowserCapabilitiesBase' to 'System.Web.HttpBrowserCapabilities'

The change to the example fixes that issue. 

I have tested the code and it now works as expected. I hope that my understanding of the issue is correct and that this is the correct change to make.


BillWagner (approved):
Thanks for spotting this and submitting the fix @OwainWilliams We appreciate it.

Thanks for making your first contribution to docs! 🎆

PR #5409, opened 2021-03-11
Test PR - .NET 6 APIs

gewarren:
Test PR that includes doc XML files for two assemblies (System.Numerics.Vectors.xml and System.IO.Compression.Brotli.xml).

PR #5410, opened 2021-03-11
[DO NOT MERGE] Compare brotli and numerics (backported) and http (not backported)

carlossanlop:
[This ](https://github.com/dotnet/dotnet-api-docs/pull/5409) was the original PR, but it was too large to view the diff.

I created this PR with `main` as the baseline, and only cherry-picked files from System.IO.Compression and System.Numerics and System.Net.Http.

There are only 2 assemblies that contained intellisense.xml files: Compression.Brotli and Numerics.Vectors.

Everything else from Compression, Numerics, and all of System.Net.Http, was also added to this PR as control, to confirm that the docs remained _untouched_.

### DIFF
--- a/xml/System.IO.Compression/BrotliDecoder.xml
+++ b/xml/System.IO.Compression/BrotliDecoder.xml
@@ -1,15 +1,16 @@
 <Type Name="BrotliDecoder" FullName="System.IO.Compression.BrotliDecoder">
+  <TypeSignature Language="C++ CLI" Value="public value class BrotliDecoder : IDisposable" />
   <TypeSignature Language="C#" Value="public struct BrotliDecoder : IDisposable" />
   <TypeSignature Language="ILAsm" Value=".class public sequential ansi sealed beforefieldinit BrotliDecoder extends System.ValueType implements class System.IDisposable" />
   <TypeSignature Language="DocId" Value="T:System.IO.Compression.BrotliDecoder" />
   <TypeSignature Language="VB.NET" Value="Public Structure BrotliDecoder&#xA;Implements IDisposable" />
-  <TypeSignature Language="C++ CLI" Value="public value class BrotliDecoder : IDisposable" />
   <TypeSignature Language="F#" Value="type BrotliDecoder = struct&#xA;    interface IDisposable" />
   <AssemblyInfo>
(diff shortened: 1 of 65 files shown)


xml/System.IO.Compression/BrotliDecoder.xml
```diff
@@ -143,15 +140,7 @@ The return value can be as follows:
         <summary>Attempts to decompress data that was compressed with the Brotli algorithm.</summary>
         <returns>
           <see langword="true" /> on success; <see langword="false" /> otherwise.</returns>
-        <remarks>
```
carlossanlop, line 146: I'm curious to see remarks like this: They were fully transformed to xml, but the newlines are probably not going to be respected in the preview. Which means I will have to modify DocsPortingTool so that it wraps every line with `<para></para>` (if the tool detects at least two paragraphs).

gewarren:
Close/reopen for build.

xml/System.IO.Compression/BrotliDecoder.xml
gewarren: Yep, we will need to add the `<para>` tags: https://review.docs.microsoft.com/en-us/dotnet/api/system.io.compression.brotlidecoder.decompress?view=net-5.0&branch=pr-en-us-5410#remarks

gewarren:
Yay, no build warnings!

gewarren:
Any idea why ZLibStream didn't show up in the preview? https://github.com/carlossanlop/dotnet-api-docs/blob/testxmls/xml/System.IO.Compression/ZLibStream.xml

xml/System.IO.Compression/BrotliDecoder.xml
carlossanlop: *sigh* ... This will make our triple slash comments less readable. If that's the only choice we have, then I'll work on fixing the tool to address this.

carlossanlop:
> Any idea why ZLibStream didn't show up in the preview? https://github.com/carlossanlop/dotnet-api-docs/blob/testxmls/xml/System.IO.Compression/ZLibStream.xml

Two reasons:

- It's a brand new type, introduced in 6.0.
- It belongs to the assembly `System.IO.Compression`. We only included the intellisense.xml file for the `System.IO.Compression.Brotli` assembly.

PR #5417, opened 2021-03-11
Deleted the code to call RemoveFirst() before trying to find the last occurrence of word "the".

lavychava:
This example removes the first node before trying to find the last occurrence of word "the".  This makes the list have only one "the" in the example.

## Summary

Describe your changes here.

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR/Generic.LinkedList/cs/source.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR/Generic.LinkedList/cs/source.cs
@@ -44,7 +44,6 @@ public class Example
 
         // <Snippet12>
         // Indicate the last occurence of 'the'.
-        sentence.RemoveFirst();
         LinkedListNode<string> current = sentence.FindLast("the");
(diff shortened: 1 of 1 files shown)


antonfirsov:
@lavychava have you tried running the sample code? It produces the output as expected and listed in the end of the sample in comments.

In `Test 4`, the word `yesterday` is being moved to the beginning of the list:

https://github.com/dotnet/dotnet-api-docs/blob/ea6728e2ab726c9b694c0580694a83e78f26b715/samples/snippets/csharp/VS_Snippets_CLR/Generic.LinkedList/cs/source.cs#L37-L43

In `Test 5` we then start by removing that word, because it's no longer relevant for the rest of the sample:

https://github.com/dotnet/dotnet-api-docs/blob/ea6728e2ab726c9b694c0580694a83e78f26b715/samples/snippets/csharp/VS_Snippets_CLR/Generic.LinkedList/cs/source.cs#L204-L208

If anything, I would move `sentence.RemoveLast();` so it's less confusing.

PR #5419, opened 2021-03-12
Remove reference to paths ending with backslashes, make more platform-agnostic

Hex052:
## Summary

Change reference to backslash to `Path.DirectorySeparatorChar`, since Linux and macOS don't use backslashes as path separators.

I don't think I made this change quite correctly, but if I haven't I'd appreciate feedback on how to do it correctly!

### DIFF
--- a/xml/System.IO/Path.xml
+++ b/xml/System.IO/Path.xml
@@ -1911,7 +1911,7 @@ Paths are resolved by calling the <xref:System.IO.Path.GetFullPath%2A> method be
       <Parameters />
       <Docs>
         <summary>Returns the path of the current user's temporary folder.</summary>
-        <returns>The path to the temporary folder, ending with a `DirectorySeparatorChar`.</returns>
+        <returns>The path to the temporary folder, ending with a  <see cref="DirectorySeparatorChar"/>.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[  
   


Hex052:
So it looks like I maybe should've put it more like this? I've not edited this docs before.

`<see cref="M:System.IO.Path.DirectorySeparatorChar" />`

xml/System.IO/Path.xml
```diff
@@ -1911,7 +1911,7 @@ Paths are resolved by calling the <xref:System.IO.Path.GetFullPath%2A> method be
       <Parameters />
       <Docs>
         <summary>Returns the path of the current user's temporary folder.</summary>
-        <returns>The path to the temporary folder, ending with a backslash.</returns>
+        <returns>The path to the temporary folder, ending with a `DirectorySeparatorChar`.</returns>
```
antonfirsov, line 1914: You can add a semantic reference to the `DirectorySeparatorChar` field:

```suggestion
        <returns>The path to the temporary folder, ending with a  <see cref="DirectorySeparatorChar"/>.</returns>
```

antonfirsov (approved):
Looks good, with one small suggestion. Thanks for spotting!

xml/System.IO/Path.xml
Hex052: Ah, that's how you do it, thank you!

adamsitnik (approved):
LGTM, thank you @Hex052 !

PR #5435, opened 2021-03-14
Fix copy paste mistake

Suchiman:
Fix copy paste mistake in `RequestHeaderEncodingSelector` (from `ResponseHeaderEncodingSelector`)

### DIFF
--- a/xml/System.Net.Http/SocketsHttpHandler.xml
+++ b/xml/System.Net.Http/SocketsHttpHandler.xml
@@ -724,8 +724,8 @@ For example, if the value is 64, then 65,536 bytes are allowed for the maximum r
         <ReturnType>System.Net.Http.HeaderEncodingSelector&lt;System.Net.Http.HttpRequestMessage&gt;</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets or sets a callback that encodes request headers values.</summary>
-        <value>The header encoding selector callback to encode the value for the specified request header name, or <see langword="null" /> to indicate the default behavior.</value>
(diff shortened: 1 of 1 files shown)


xml/System.Net.Http/SocketsHttpHandler.xml
```diff
@@ -724,8 +724,8 @@ For example, if the value is 64, then 65,536 bytes are allowed for the maximum r
         <ReturnType>System.Net.Http.HeaderEncodingSelector&lt;System.Net.Http.HttpRequestMessage&gt;</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets or sets a callback that decodes request headers values.</summary>
-        <value>The header encoding selector callback to decode the value for the specified response header name, or <see langword="null" /> to indicate the default behavior.</value>
+        <summary>Gets or sets a callback that encodes request headers values.</summary>
```
antonfirsov, line 727: The whole phrasing seems to be very inaccurate. I would rather write:
```suggestion
        <summary>Gets or sets a callback that selects the <see cref="System.Text.Encoding"/> to encode header values.</summary>
```

antonfirsov (commented):
I think we should improve the phrasing even more. /CC @MihaZupan

xml/System.Net.Http/SocketsHttpHandler.xml
antonfirsov: This text has to be also fixed in `ResponseHeaderEncodingSelector` .

scalablecory (approved):
Thanks!

PR #5439, opened 2021-03-16
Replace reference to \r\n with System.Environment.NewLine

Hex052:
## Summary

Remove a reference to `"\r\n"` explicitly, and replace it with `System.Environment.NewLine`


adamsitnik (approved):
LGTM, thank you @Hex052

PR #5462, opened 2021-03-24
CI update - System folder

gewarren:
Updates signatures.


carlossanlop (approved):
It's just moving C++ above F#, right? What was the motivation for the PR? Were we seeing warnings due to wrong order?

gewarren:
> It's just moving C++ above F#, right? What was the motivation for the PR? Were we seeing warnings due to wrong order?

There's a new version of mdoc that gives names to unnamed parameters, but it was impossible to see those changes because of these signature position moves. So this group of PRs is to set the baseline so we can see the meaningful differences in the new version of mdoc. It's unfortunate, and I'm not sure why there is this change in position of the signatures. If you are inclined, could you approve the other 6 PRs? (I broke the PRs up because every single API is affected - 16K files.)

PR #5470, opened 2021-03-24
IgnoreNullValues isn't for deserialization

tdykstra:
Fixes https://github.com/dotnet/docs/issues/23450

### DIFF
--- a/xml/System.Text.Json/JsonSerializerOptions.xml
+++ b/xml/System.Text.Json/JsonSerializerOptions.xml
@@ -364,6 +364,8 @@ For more information, see [How to write custom converters for JSON serialization
 
 ## Remarks
 
+This setting doesn't apply to deserialization.
+
 For more information, see [How to ignore properties](/dotnet/standard/serialization/system-text-json-ignore-properties).
 
  ]]></format>


gewarren (approved):
Should you also add a remark in the remarks section that it does not apply to deserialization?

tdykstra:
@gewarren Thanks for the review!

layomia:
Actually, `IgnoreNullValues` _is_ honored when deserializing. We did this for compat with `Newtonsoft.Json` ([NullValueHandling](https://www.newtonsoft.com/json/help/html/NullValueHandlingIgnore.htm)) in v1 (.NET Core 3.x), but ended up obsoleting it after introducing `DefaultIgnoreCondition` in .NET 5.

`DefaultIgnoreCondition` (currently) only applies to serialization. When designing the new option, we couldn't think of a good reason to ignore `null` values in JSON payloads when reading. Most `Newtonsoft.Json` users don't even know that `NullValueHandling` is honored when deserializing. We decided not to carry over the "mistake" we made with `IgnoreNullValues`.

layomia:
cc @tdykstra - https://github.com/dotnet/dotnet-api-docs/pull/5470#issuecomment-806306430.

PR #5474, opened 2021-03-25
Fix 'returns' documentation for method bool System.Type.IsAssignableTo(Type)

hamarb123:
Fixes https://github.com/dotnet/runtime/issues/50221.

(Copy of issue at runtime):
On https://docs.microsoft.com/en-us/dotnet/api/system.type.isassignableto?view=net-5.0 it says (in the returns section) the following:

`false` if none of these conditions are true, or if `targetType` or `this` is `null`.

This is a new API to .NET 5, and the documentation is incorrect as it will not return false if the `this` object is null, it will throw a `System.NullReferenceException` since it's an instance method, not an extension method.
I've tested this with .NET 6.0.100-preview.2.21158.2 and it should have the same result on other versions.

Code to reproduce:
```cs
((Type)null).IsAssignableTo(null)
```

(Additional information):
No other method specifies `System.NullReferenceException` as an exception it should throw since it's essentially caused by a failure to invoke the method, not an exception from the method.

### DIFF
--- a/xml/System/Type.xml
+++ b/xml/System/Type.xml
@@ -11299,7 +11299,7 @@ GetType(Array).IsAssignableFrom(type)
   
 -   The current type represents a value type, and <paramref name="targetType" /> represents <c>Nullable&lt;c&gt;</c> (<c>Nullable(Of c)</c> in Visual Basic).  
   
- <see langword="false" /> if none of these conditions are true, or if <paramref name="targetType" /> or <see langword="this" /> is <see langword="null" />.</returns>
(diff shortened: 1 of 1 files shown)


hamarb123:
Please review and merge @gewarren.
Thanks!

gewarren:
@mangod9 Is this correct? It looks fine from a formatting standpoint.

mangod9 (approved):

gewarren (approved):

gewarren:
@hamarb123 Thanks for updating this!

PR #5488, opened 2021-03-30
Remove useless line in sample code for EventSource

sywhang:
## Summary

There is a useless line of code in the sample code for EventSource. Removing this to avoid causing any confusion.

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR/etwtracesmall/cs/program.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR/etwtracesmall/cs/program.cs
@@ -20,7 +20,6 @@ namespace Demo1
     {
         static void Main(string[] args)
         {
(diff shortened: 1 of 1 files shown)


samples/snippets/csharp/VS_Snippets_CLR/etwtracesmall/cs/program.cs
```diff
@@ -21,7 +21,6 @@ class Program
         static void Main(string[] args)
         {
             string name = MyCompanyEventSource.GetName(typeof(MyCompanyEventSource));
```
Jozkee, line 23: Isn't this line also useless? VB snippet doesn't have it either.

gewarren (approved):

GrabYourPitchforks (approved):

PR #6604, opened 2021-04-04
Update SortDescription.xml

Omcsesz:
## Summary

I added that if someone uses empty string as the value for parameter propertyName.

### DIFF
--- a/xml/System.ComponentModel/SortDescription.xml
+++ b/xml/System.ComponentModel/SortDescription.xml
@@ -39,7 +39,7 @@
         <Parameter Name="direction" Type="System.ComponentModel.ListSortDirection" />
       </Parameters>
       <Docs>
-        <param name="propertyName">The name of the property to sort the list by. Should be set to "" if the collection to be sorted contains strings, integers, etc..</param>
+        <param name="propertyName">The name of the property to sort the list by. If the collection contains strings or value types, set this parameter to the empty string (`""`).</param>
         <param name="direction">The sort order.</param>
         <summary>Initializes a new instance of the <see cref="T:System.ComponentModel.SortDescription" /> structure.</summary>
         <remarks>To be added.</remarks>


xml/System.ComponentModel/SortDescription.xml
```diff
@@ -39,7 +39,7 @@
         <Parameter Name="direction" Type="System.ComponentModel.ListSortDirection" />
       </Parameters>
       <Docs>
-        <param name="propertyName">The name of the property to sort the list by.</param>
+        <param name="propertyName">The name of the property to sort the list by. Should be set to "" if the collection to be sorted contains strings, integers, etc..</param>
```
tarekgh, line 42: > Should be set to "" if the collection to be sorted contains strings, integers, etc.

Should be set to empty string if the collection to be sorted contains types like string or numeric types.

tarekgh (approved):
I am not the best at rewording it but @gewarren may help in better rewording :-)

xml/System.ComponentModel/SortDescription.xml
Omcsesz: Agreed.

gewarren:
```suggestion
        <param name="propertyName">The name of the property to sort the list by. If the collection contains strings or value types, set this parameter to the empty string (`""`).</param>
```

maryamariyan (approved):
@gewarren is it OK to include `""` in the docs as provided in this PR?

gewarren:
> @gewarren is it OK to include `""` in the docs as provided in this PR?

I had left a suggestion to code fence the quotes (and also to reword it slightly) but the conversation was somehow resolved.

https://github.com/dotnet/dotnet-api-docs/pull/6604/files#r1002113425

tarekgh:
Thanks @gewarren, I have applied your suggestion and we can merge after building.

PR #6606, opened 2021-04-05
Fix typo in HttpClientHandler.xml

ericmutta:
The sentence:

> Gets the collection of security certificates that are associated requests to the server.

Is missing the word "with" and should read:

> Gets the collection of security certificates that are associated **with** requests to the server.


karelz (approved):

karelz:
Thanks for the fix @ericmutta!

ericmutta:
Thanks for following up @karelz :+1:

PR #6636, opened 2021-04-13
Link to System.Drawing.Common official MS pkg

alex-jitbit:
Also, is this caution even relevant, now that we have System.Drawing.Common?

## Summary

Link to System.Drawing.Common official Microsoft package


gewarren (approved):
@alex-jitbit Thank you for adding the link. I think it's still relevant.

PR #6642, opened 2021-04-14
Update JsonDocument.xml

watfordgnf:
Removes the incorrect "character" modifier for "value". `JsonDocument.Parse(ReadOnlyMemory<char>...` reads a single JSON _value_:
```
        /// <summary>
        ///   Parse text representing a single JSON value into a JsonDocument.
        /// </summary>
```
https://github.com/dotnet/runtime/blob/209a9b1c1d3022f9178d74be3e0850f41f3f26f5/src/libraries/System.Text.Json/src/System/Text/Json/Document/JsonDocument.Parse.cs#L189-L208


eiriktsarpalis (approved):
LGTM, thanks!

PR #6644, opened 2021-04-14
Fix BitConverter issues

### LINKED ISSUE #4327: BitConverter issues
`public static ulong ToUInt64(ReadOnlySpan<byte> value);` is missing the following exception

```
        // Exceptions:
        //   T:System.ArgumentOutOfRangeException:
        //     The length of value is less than the length of a System.Int64 value.
```

...but I'm not sure if that is intended or whether `System.Int64` should be a hardcoded `8`. Different methods have alternating styling. The implementations all compare `sizeof(T)` and don't use constants.

`public static bool ToBoolean(ReadOnlySpan<byte> value)` has an incorrect exception.
```
        // Exceptions:
        //   T:System.ArgumentOutOfRangeException:
        //     The length of value is less than 0.
```
The 0 should be a 1.
It should read `The length of value is less than the length of a System.Byte value.` or something more explicit like "The length of the value is zero."

`public static ushort ToUInt16(ReadOnlySpan<byte> value)` has a typo in the exception signature.

```
        // Exceptions:
        //   T:System.ArgumentOutOfRangeException:
        //     The length of value is less than the 2.
```

Note, I just did a cursory glance so there may be other issues. I can submit a PR if you specify which styling you'd prefer (sizeof(T) vs constant)

gewarren:
- Add new docs for half-related APIs
- Fix #4327

### DIFF
--- a/xml/System/BitConverter.xml
+++ b/xml/System/BitConverter.xml
@@ -1518,10 +1518,10 @@
         <Parameter Name="value" Type="System.ReadOnlySpan&lt;System.Byte&gt;" Index="0" FrameworkAlternate="net-6.0" />
       </Parameters>
       <Docs>
-        <param name="value">To be added.</param>
         <param name="value">A read-only span containing the bytes to convert.</param>
         <summary>Converts a read-only byte span into a half-precision floating-point value.</summary>
         <returns>A half-precision floating-point value that represents the converted bytes.</returns>
+        <remarks>To be added.</remarks>
       </Docs>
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: or wait for the product team to review.

tannergooding (approved):

gewarren:
Thanks for the reviews Bill and Tanner.

PR #6647, opened 2021-04-15
Minor change at Line 18 to align with the output.

Trident30:
## Summary

Aligns the code with the output displayed.

The code at Line 18 doesn't match with the output. Line should be as below:
Console.WriteLine("{0} thread ID: {1}",
                        s, Thread.CurrentThread.ManagedThreadId);


BillWagner (approved):
Thanks for making this change @Trident30 

I've reviewed the change and I'll :shipit: now.

Thanks again.

PR #6648, opened 2021-04-15
Update ApplicationException remarks

ChristopherHaws:
## Summary

The current docs for `ApplicationException` state that it "Serves as the base class for application-defined exceptions" however the remarks section states "You should derive custom exceptions from the `Exception` class rather than the `ApplicationException` class." which are contradictory. I believe that the intention for the remarks statement is that libraries should not throw/catch `ApplicationException` since this type should only be used in application code directly.

### DIFF
--- a/xml/System/ApplicationException.xml
+++ b/xml/System/ApplicationException.xml
@@ -53,7 +53,7 @@
 ## Remarks  
   
 > [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


BillWagner:
Thanks for adding this @ChristopherHaws 

I am going to recommend closing this, even though I agree there is some work to do here.

The history is that `ApplicationException` looked like a good idea when .NET was first being developed. In practice, this didn't turn out to be true. Our recommendation now is to use one of the already defined exception types where appropriate, and define your own exception classes when needed. Those exception classes should be derived from `System.Exception`. 

I'll close this, but I'd welcome a new PR that reflects that history.

PR #6652, opened 2021-04-19
Fix code snippet in `Math.Round` for `RoundApproximate`

OronDF343:
## Summary

See https://docs.microsoft.com/en-us/dotnet/api/system.math.round?view=net-5.0#rounding-and-precision

The `RoundApproximate` example snippet is broken due to too much indentation (which caused it to be recognized as an inline code block rather than a snippet), I have fixed it so it should match the other examples.


BillWagner (approved):
Thank you @OronDF343 
This change looks great. I'll :shipit: now.

You should see the changes on the live site in the next day or two.  Thanks for contributing to docs! 🎆

PR #6656, opened 2021-04-21
Update StrongNameKeyPair docs

### LINKED ISSUE #4896: It's not stated in its documentation page that System.Reflection.StrongNameKeyPair.PublicKey throws on .NET Core/.NET 5+
[As seen in .NET's source code](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Reflection/StrongNameKeyPair.cs), the `System.Reflection.StrongNameKeyPair.PublicKey` property throws an exception when any of its members are used. However, [the property's documentation](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.strongnamekeypair.publickey?view=netcore-3.1) does not say anything about this, leaving us only with [the APIs that always throw](https://docs.microsoft.com/en-us/dotnet/core/compatibility/unsupported-apis#systemreflection) to let us know.

MSDN-WhiteKnight:
Add PlatformNotSupportedException for APIs not supported in .NET Core

Fixes #4896

### DIFF
--- a/xml/System.Reflection/StrongNameKeyPair.xml
+++ b/xml/System.Reflection/StrongNameKeyPair.xml
@@ -231,6 +231,7 @@
         <exception cref="T:System.ArgumentNullException">
           <paramref name="keyPairContainer" /> is <see langword="null" />.</exception>
         <exception cref="T:System.Security.SecurityException">The caller does not have the required permission.</exception>
+        <exception cref="T:System.PlatformNotSupportedException">.NET 5+ and .NET Core only: This member is not supported.</exception>
         <altmember cref="T:System.Array" />
       </Docs>
     </Member>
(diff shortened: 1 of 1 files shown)


buyaa-n:
We have obsoleted this type in .net core recently https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Reflection/StrongNameKeyPair.cs#L9, now all constructors and APIs will throw. Not sure how this PR should reflect that CC @gewarren

gewarren:
> We have obsoleted this type in .net core recently https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Reflection/StrongNameKeyPair.cs#L9, now all constructors and APIs will throw. Not sure how this PR should reflect that CC @gewarren

I've created https://github.com/dotnet/dotnet-api-docs/issues/6666 to track the new PNSEs. I think it's fine to merge this PR as-is. Thank you @MSDN-WhiteKnight !

gewarren (approved):

PR #6660, opened 2021-04-22
Detail observed behavior of AllowUnknownCertificateAuthority

stewartadam:
AllowUnknownCertificateAuthority not only ignores untrusted roots, but also partial chains. This updates the documentation to reflect this behavior.

## Summary
See https://github.com/dotnet/runtime/issues/49615 for more details. AllowUnknownCertificateAuthority has the (presently) undocumented side effect of also ignoring the PartialChain verification status, which means that there is no validation that any of the root CAs added to the trust store were actually signing the certificate under verification.

Users unaware of this behavior could therefore be lulled into a false sense of safety when verification succeeds.

Fixes #49615.

### DIFF
--- a/xml/System.Security.Cryptography.X509Certificates/X509VerificationFlags.xml
+++ b/xml/System.Security.Cryptography.X509Certificates/X509VerificationFlags.xml
@@ -131,7 +131,7 @@
       </ReturnValue>
       <MemberValue>16</MemberValue>
       <Docs>
-        <summary>Ignore that the chain cannot be verified due to an unknown certificate authority (CA) or partial chains.
+        <summary>Ignore that the chain cannot be verified due to an unknown certificate authority (CA) or partial chains.</summary>
       </Docs>
     </Member>
     <Member MemberName="IgnoreCertificateAuthorityRevocationUnknown">


xml/System.Security.Cryptography.X509Certificates/X509VerificationFlags.xml
```diff
@@ -131,7 +131,7 @@
       </ReturnValue>
       <MemberValue>16</MemberValue>
       <Docs>
-        <summary>Ignore that the chain cannot be verified due to an unknown certificate authority (CA).</summary>
+        <summary>Ignore that the chain cannot be verified due to an unknown certificate authority (CA), or a partial chain. Note that ignoring partial chain skips verification that the signing authority of the certificate under validation was actually signed by one of the known CAs, so this must be performed separately if desired.</summary>
```
bartonjs, line 134: I'm not sure what the second sentence is trying to say.  If a candidate is found but the signature doesn't work out, that produces a different error.

```suggestion
        <summary>Ignore that the chain cannot be verified due to an unknown certificate authority (CA), or a partial chain.</summary>
```

xml/System.Security.Cryptography.X509Certificates/X509VerificationFlags.xml
```diff
@@ -45,10 +45,10 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- These flags indicate the conditions under which chain verification should occur. For example, if an application does not require certificates time values in a chain to be valid, the IgnoreNotTimeValid flag can be used.  
-  
-   
-  
+ These flags indicate the conditions under which chain verification should occur. For example, if an application does not require certificates time values in a chain to be valid, the IgnoreNotTimeValid flag can be used.
+
+ Note that the behavior of certificate validation in .NET Core aims to standardize its behavior according to the underlying win32 implementation APIs (i.e. CertVerifyCertificateChainPolicy in wincrypt.h), even on other platforms. Therefore, AllowUnknownCertificateAuthority not only allows for an untrusted roots, but also permits partial chains on all platforms.
```
bartonjs, line 50: I don't think this paragraph provides much value, especially the second sentence.

```suggestion
```

xml/System.Security.Cryptography.X509Certificates/X509VerificationFlags.xml
stewartadam: I'm trying to elaborate on the consequences of not validating partial chain - in particular, that because partial chain status gets suppressed, the certificate under validation is essentially always validated:

If unknown CAs are permitted, the certificate under validation doesn't actually have to be signed by any of the certificates added to the trust store for validation to still succeed.

IMO this is a key piece of information the docs need to disclose; that users should expect to have to do their own validation of the fingerprint on the end-certificate's issuer and signing root to ensure those values are one and the same.

xml/System.Security.Cryptography.X509Certificates/X509VerificationFlags.xml
stewartadam: I think it's important to disclose this piece of information from #49615:

> On Windows, that behavior is controlled by Windows. On non-Windows we maintain parity with what Windows does since that describes the ".NET Framework behavior" for this type, which is the strong starting point for the ".NET behavior" for this type.

This helps provide context for how .NET Core is going to behave on other OSs and how users should expect these flags to behave. AFAIK it's not mentioned anywhere else that the implementation attempts to preserve the behavior of the underlying win32 calls [including quirks like ignoring partial chain, which is what bit me in dotnet/runtime/issues/49615]

Is there a way you'd prefer its phrased?

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=6660) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=6660) before we can accept your contribution.<br/><br/>:x: stewartadam [sign now](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=6660)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/dotnet-api-docs?pullRequest=6660) it.</sub>

xml/System.Security.Cryptography.X509Certificates/X509VerificationFlags.xml
bartonjs: > If unknown CAs are allowed, the certificate under validation doesn't actually have to be signed by any of the certificates added to the trust store for validation to still succeed. 

Yes, that's true by definition: By specifying this flag you've said it's OK for the chain to end in a CA you don't know about.  Whether last thing in the chain is a root authority or an intermediate (or the end-entity itself) you've said "trust doesn't matter" and you need to do something about it.

xml/System.Security.Cryptography.X509Certificates/X509VerificationFlags.xml
bartonjs: > AFAIK it's not mentioned anywhere else that the implementation attempts to preserve the behavior of the underlying win32 calls

That's generally true of .NET documentation.  We often ended up being a thin wrapper over some Win32 and now that's just "how .NET works".  I don't think it's useful to have in documentation, because it's just a starting point (not all of the OSes we run on can support all of the things in all of the same ways), and it could be interpreted as limiting us from providing additions on top of what Win32 does (e.g. if macOS did something interesting that we wanted to replicate on Windows).

xml/System.Security.Cryptography.X509Certificates/X509VerificationFlags.xml
stewartadam: > true by definition: By specifying this flag you've said it's OK for the chain to end in a CA you don't know about.

The docs don't describe this definition. They say it only ignores `UntrustedRoot` flag, which is the behavior most users are actually after; verifying a complete chain but allowing for the `UntrustedRoot` chain status flag is very common for validating certificates against a self-signed root CA. `Build()` returning verification success for any chain at all in this scenario has not been documented nor is expected.

Most other software that interacts with X.509 has similar flags or options for validating a certificate with an unknown/self-signed certificate authority, but they otherwise keeps all rules the same - i.e. requiring completed chain, and hence the user confusion around this.

The .NET implementation's unexpected behavior regarding partial chain and verification success has come up in the past (dotnet/runtime/issues/26449) and [has turned users away](https://stackoverflow.com/a/32872136).

What do you suggest?

bartonjs: > verifying a complete chain but allowing for the UntrustedRoot chain status flag is very common for validating certificates against a self-signed root CA

The correct piece of action for that scenario is that after calling Build you check that `chain.ChainElements[^1].Certificate.RawData.SequenceEquals(s_pinnedCertBytes)`.  If you're doing that, it doesn't matter if it ended in UnknownRoot or PartialChain (or a known/trusted root).  If you're not doing that, you're not actually testing it came from your own root, and is a security hole.

The only way that the two interact is if you did a check that if the last one ended in UntrustedRoot then you check pinning (which probably isn't what's desired, but maybe it is during "I plan to use a trusted CA cert soon" rollout phases), in which case merely the "or a partial chain" addition in docs covers the scenario.

stewartadam: I don't disagree with that, and it's been described well in the above linked GitHub issue discussions which is why I'm trying to move it into the docs. If you feel the examples section is a better place for it, I'm happy to add one there.

But let's put specific root CA pinning aside, and consider a more general case of a user wanting to use `chain.Build()` to verify a given certificate against the OS trust store --

> in which case merely the "or a partial chain" addition in docs covers the scenario

IMO "or a partial chain" is a far cry from its actual implications, namely that the user would **always** need to perform manual issuer verification, because ignoring partial chain implies `X509Chain.Build()` now returns `true` no matter what the contents of the trusted certificate store is. That behavior is unintuitive and these classes are in a security sensitive area, so I think it warrants being explicit for the sake of our users.

If you feel that the docs only warrant a shorter "or a partial chain" then let me know and I'll adjust my PR.

xml/System.Security.Cryptography.X509Certificates/X509VerificationFlags.xml
stewartadam: OK - Removed from PR

xml/System.Security.Cryptography.X509Certificates/X509VerificationFlags.xml
bartonjs: > But let's put specific root CA pinning aside, and consider a more general case of a user wanting to use chain.Build() to verify a given certificate against the OS trust store

Then they don't set the `X509VerificationFlags.AllowUnknownCertificateAuthority` flag, and now `PartialChain` (or an unknown root) make `Build()` return `false`, and everything's happy.

>  If you feel that the docs only warrant a shorter "or a partial chain" then let me know

I do feel that way.  Especially because of the technical errors in your statement (e.g. it'll return false if it's expired (unless the "ignore validity times" flag is also set)).

stewartadam: > Especially because of the technical errors in your statement (e.g. it'll return false if it's expired (unless the "ignore validity times" flag is also set)).

Fair enough, the first sentence of the elaboration was ambiguous. Either way, I've removed all of it at your request.

xml/System.Security.Cryptography.X509Certificates/X509VerificationFlags.xml
```diff
@@ -131,7 +131,7 @@
       </ReturnValue>
       <MemberValue>16</MemberValue>
       <Docs>
-        <summary>Ignore that the chain cannot be verified due to an unknown certificate authority (CA).</summary>
+        <summary>Ignore that the chain cannot be verified due to an unknown certificate authority (CA) or partial chains.
```
bartonjs, line 134:
```suggestion
        <summary>Ignore that the chain cannot be verified due to an unknown certificate authority (CA) or partial chains.</summary>
```

bartonjs (approved):

PR #6663, opened 2021-04-23
Improve Half docs a bit

gewarren:


xml/System/Half.xml
```diff
@@ -34,8 +34,17 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>An IEEE 754 compliant <see langword="float16" /> type.</summary>
-    <remarks>To be added.</remarks>
+    <summary>Represents a half-precision floating-point number.</summary>
+    <remarks>
+      <format type="text/markdown"><![CDATA[
+
+## RemarksThe <xref:System.Half> value type represents a half-precision 16-bit number with values ranging from negative 65,504 to positive 65,504, as well as positive or negative zero, <xref:System.Half.PositiveInfinity>, <xref:System.Half.NegativeInfinity>, and not a number (<xref:System.Half.NaN>).
```
BillWagner, line 41: I think this is missing a newline:

```suggestion
## Remarks

The <xref:System.Half> value type represents a half-precision 16-bit number with values ranging from negative 65,504 to positive 65,504, as well as positive or negative zero, <xref:System.Half.PositiveInfinity>, <xref:System.Half.NegativeInfinity>, and not a number (<xref:System.Half.NaN>).
```

BillWagner (approved):
This is better by half @gewarren 

(Sorry, bad pun).

I think I found one typo, and then this is ready to :shipit:

PR #6674, opened 2021-04-28
Fix HKDF Expand to Extract

vcsjones:
## Summary

The API documentation for `Expand` mentions that the PRK parameter is "Usually the output from the Expand step". This was probably meant to be `Extract`.

@bartonjs

### DIFF
--- a/xml/System.Security.Cryptography/HKDF.xml
+++ b/xml/System.Security.Cryptography/HKDF.xml
@@ -123,7 +123,7 @@ In situations where the input key material is already a uniformly random bit str
       </Parameters>
       <Docs>
         <param name="hashAlgorithmName">The hash algorithm used for HMAC operations.</param>
-        <param name="prk">The pseudorandom key that is at least as long as the output byte array of the specified hash algorithm (usually the output from Extract step).</param>
+        <param name="prk">The pseudorandom key that is at least as long as the output byte array of the specified hash algorithm (usually the output from the Extract step).</param>
         <param name="outputLength">The length of the output keying material.</param>
         <param name="info">The optional context and application specific information.</param>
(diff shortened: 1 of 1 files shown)


xml/System.Security.Cryptography/HKDF.xml
```diff
@@ -155,7 +155,7 @@ In situations where the input key material is already a uniformly random bit str
       </Parameters>
       <Docs>
         <param name="hashAlgorithmName">The hash algorithm used for HMAC operations.</param>
-        <param name="prk">The pseudorandom key that is at least as long as as the output byte array of the specified hash algorithm (usually the output from Expand step).</param>
+        <param name="prk">The pseudorandom key that is at least as long as as the output byte array of the specified hash algorithm (usually the output from Extract step).</param>
```
bartonjs, line 158:
```suggestion
        <param name="prk">The pseudorandom key that is at least as long as as the output byte array of the specified hash algorithm (usually the output from the Extract step).</param>
```

xml/System.Security.Cryptography/HKDF.xml
```diff
@@ -123,7 +123,7 @@ In situations where the input key material is already a uniformly random bit str
       </Parameters>
       <Docs>
         <param name="hashAlgorithmName">The hash algorithm used for HMAC operations.</param>
-        <param name="prk">The pseudorandom key that is at least as long as the output byte array of the specified hash algorithm (usually the output from Expand step).</param>
+        <param name="prk">The pseudorandom key that is at least as long as the output byte array of the specified hash algorithm (usually the output from Extract step).</param>
```
bartonjs, line 126:
```suggestion
        <param name="prk">The pseudorandom key that is at least as long as the output byte array of the specified hash algorithm (usually the output from the Extract step).</param>
```

bartonjs (commented):
Feels like it could use some definite articles.

bartonjs (approved):

PR #6678, opened 2021-04-28
Nomenclature: Prefer 'your code' over 'your app'

perlun:
I happened to notice this while reading the docs at https://docs.microsoft.com/en-us/dotnet/api/System.Text.StringBuilder?view=net-5.0&viewFallbackFrom=netcore-5.0. The text consistently talked about your "app" in multiple places.

While this might seem like a nice and modern way to express things, I do think that it conveys the wrong meaning to the reader. This is because "app" has a specific meaning - often used as synonym to _mobile application_. While a webapp could also reasonably be called an "app", probably not so much for your random console application, your Windows background service, your Linux daemon... etc. These are not "apps", but `StringBuilder` is still perfectly usable in such scenarios also.

In other words, it feels more correct in this context to talk about "application" since it has a broader scope. In other words, all _apps_ are applications, but all _applications_ are not apps... :smile:

### DIFF
--- a/xml/System.Text/StringBuilder.xml
+++ b/xml/System.Text/StringBuilder.xml
@@ -105,11 +105,11 @@
  For routines that perform extensive string manipulation (such as apps that modify a string numerous times in a loop), modifying a string repeatedly can exact a significant performance penalty. The alternative is to use <xref:System.Text.StringBuilder>, which is a mutable string class. Mutability means that once an instance of the class has been created, it can be modified by appending, removing, replacing, or inserting characters. A <xref:System.Text.StringBuilder> object maintains a buffer to accommodate expansions to the string. New data is appended to the buffer if room is available; otherwise, a new, larger buffer is allocated, data from the original buffer is copied to the new buffer, and the new data is then appended to the new buffer.
 
 > [!IMPORTANT]
->  Although the <xref:System.Text.StringBuilder> class generally offers better performance than the <xref:System.String> class, you should not automatically replace <xref:System.String> with <xref:System.Text.StringBuilder> whenever you want to manipulate strings. Performance depends on the size of the string, the amount of memory to be allocated for the new string, the system on which your app is executing, and the type of operation. You should be prepared to test your app to determine whether <xref:System.Text.StringBuilder> actually offers a significant performance improvement.
+>  Although the <xref:System.Text.StringBuilder> class generally offers better performance than the <xref:System.String> class, you should not automatically replace <xref:System.String> with <xref:System.Text.StringBuilder> whenever you want to manipulate strings. Performance depends on the size of the string, the amount of memory to be allocated for the new string, the system on which your code is executing, and the type of operation. You should be prepared to test your code to determine whether <xref:System.Text.StringBuilder> actually offers a significant performance improvement.
 
  Consider using the <xref:System.String> class under these conditions:
 
--   When the number of changes that your app will make to a string is small. In these cases, <xref:System.Text.StringBuilder> might offer negligible or no performance improvement over <xref:System.String>.
+-   When the number of changes that your code will make to a string is small. In these cases, <xref:System.Text.StringBuilder> might offer negligible or no performance improvement over <xref:System.String>.
 
 -   When you are performing a fixed number of concatenation operations, particularly with string literals. In this case, the compiler might combine the concatenation operations into a single operation.
 
@@ -117,9 +117,9 @@
 
  Consider using the <xref:System.Text.StringBuilder> class under these conditions:
 
--   When you expect your app to make an unknown number of changes to a string at design time (for example, when you are using a loop to concatenate a random number of strings that contain user input).
+-   When you expect your code to make an unknown number of changes to a string at design time (for example, when you are using a loop to concatenate a random number of strings that contain user input).
 
--   When you expect your app to make a significant number of changes to a string.
+-   When you expect your code to make a significant number of changes to a string.
 
 <a name="HowWorks"></a>
 ## How StringBuilder works
@@ -152,7 +152,7 @@
 
 -   If the eventual size of the <xref:System.Text.StringBuilder> object is likely to grow exceedingly large, typically in excess of several megabytes. In this case, there may be some performance benefit from setting the initial <xref:System.Text.StringBuilder.Capacity%2A> property to a significantly high value to eliminate the need for too many memory reallocations.
 
--   If your app is running on a system with limited memory. In this case, you may want to consider setting the <xref:System.Text.StringBuilder.MaxCapacity%2A> property to less than <xref:System.Int32.MaxValue?displayProperty=nameWithType> if your app is handling large strings that may cause it to execute in a memory-constrained environment.
+-   If your code is running on a system with limited memory. In this case, you may want to consider setting the <xref:System.Text.StringBuilder.MaxCapacity%2A> property to less than <xref:System.Int32.MaxValue?displayProperty=nameWithType> if your code is handling large strings that may cause it to execute in a memory-constrained environment.
 
 <a name="Instantiating"></a>
 ## Instantiating a StringBuilder object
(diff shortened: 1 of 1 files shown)


tdykstra:
cc @Rick-Anderson

gewarren:
The MS style guide says:

In general, both app and application are acceptable. App is preferred in names and when space is limited.

perlun:
Thanks for your reply @gewarren. Do you happen to have a URL to the referred-to style guide?

> In general, both app and application are acceptable. App is preferred in names and when space is limited.

As for the latter part, neither point is true in this particular case. I think clarity is more important than conciseness in this context.

gewarren:
> Thanks for your reply @gewarren. Do you happen to have a URL to the referred-to style guide?
> 
> > In general, both app and application are acceptable. App is preferred in names and when space is limited.
> 
> As for the latter part, neither point is true in this particular case. I think clarity is more important than conciseness in this context.

Here's the link, but I think it's internal to MS: https://styleguides.azurewebsites.net/Styleguide/Read?id=2696&topicid=27896.

I wasn't necessarily opining for one or the other, just adding info from the style guide.

adegeo:
I prefer `app`. When trying to help a friend or family member troubleshoot something on their computer, I don't use the word "application" I use "app" and they know what I'm talking about. It might have had a decade of being a synonym for "mobile application" but if you look at the word app it's been used in desktop computer for much longer.

The word "application" doesn't even appear on https://www.microsoft.com/en-us/windows/windows-10-apps

Application sounds formal these days, and probably has its place. It's the evolution of common computer terms. We don't use E-Mail anymore, it's email.

perlun:
@adegeo

> I prefer `app`. When trying to help a friend or family member troubleshoot something on their computer, I don't use the word "application" I use "app" and they know what I'm talking about.

Sure, I get that point, but I'm not sure how relevant this is in this context though. This is dealing with how the `StringBuilder` .NET class is documented and the target audience for these docs can be assumed to be slightly different than in the scenario you describe.

However, it _is_ indeed a challenge between at least the following goals:

- avoid "dumbing things down" too much, and
- avoid "speaking above the heads" of people

Both of these are obviously important goals but I would perhaps (for this context) consider the former being the more important design consideration here. Feel free to disagree of course.

> It might have had a decade of being a synonym for "mobile application" but if you look at the word app, it's been used in desktop computing prior to that.

Thanks for pointing this out and digging up a classic example from the good old days. :+1: :heart: I would _guess_ (don't have any facts on this though) that this usage was less common than "application" back then, though. I think it was largely popularized by another company whose name coincidentally also starts with "app" around the time of 2007 and onwards. :wink: 

This Stack Exchange thread was quite interesting, even though it's written a few years ago: https://english.stackexchange.com/questions/45762/appropriate-use-of-app-vs-application. Here are some quotes from that thread:

> Gartner says: “Apps and applications are two very different expressions of software. The defining characteristic of an app is its reduced functional presence. Apps do less than applications. That is their goal."

> Apple and Google talk about "Apps" even in their formal docs, because they've made a conscious choice to adopt that into their company lexicon. Is that appropriate for all companies? I think the short answer is: it depends. Consider the domain (mobile apps versus windows apps) **and the tech-savvy sophistication of the audience.** [my emphasis]

The last sentence reiterates my point above - it depends on what kind of readers of this text we are expecting and their level of technical experience.

Rick-Anderson:
> The MS style guide says:
> 
> In general, both app and application are acceptable. App is preferred in names and when space is limited.

app, application
The Microsoft technologies that customers use to get things done come in a variety of shapes and sizes. Always refer to your group's editorial style guide for details about when to use app, application, product, and service.

If possible, refer to a product by its descriptor, such as database management system or spreadsheet.

***Most of the time use app, not application***, to describe desktop apps and cloud apps for tablets and mobile devices.

Global tip The abbreviated form, app, isn't available in all languages, so allow enough space for the full translated term in localized content.

For full-featured products, such as Office and Visual Studio, it's OK to use program instead of application.

It's OK to use application in content for a technical audience, especially to refer to a set of software files that includes executable files and other components.

Don't use application program.

https://styleguides.azurewebsites.net/StyleGuide/Read?id=2700&topicid=32507

perlun:
Thanks for the style guide excerpt @Rick-Anderson. :+1: 

> _**Most of the time use app, not application**_, to describe desktop apps and cloud apps for tablets and mobile devices.

Exactly, this sums it all up in a single sentence. _"To describe desktop apps and cloud apps **for tablets and mobile devices**"._ The .NET platform in (general and `StringBuilder` in particular) is not limited only to these use cases.

I'm personally using it to write console applications targeting Linux and other command-line oriented scenarios. There's no GUI involved. It feels overly limiting to talk about "app" in that context, since it slightly implies (at least to me) a more narrow scope for this framework class than actually is the case here.

For more GUI-oriented classes (e.g. WPF, Windows Forms etc) I think it's quite fine to talk about "app". For things more generic in nature, used for both GUI and non-GUI applications alike, I still insist that "application" is better for getting the message across.

adegeo:
Just more info. I noticed today that the templates in Visual Studio for a new desktop app say

- `Windows Forms App`
- `WPF Application`

No one can decide 🤣

Rick-Anderson:
> Apple and Google talk about "Apps" even in their formal docs, because they've made a conscious choice to adopt that into their company lexicon. 

We've never had a complaint about using app for web applications.

> This Stack Exchange thread was quite interesting

SE is very opinionated. I disagree with that assertion. Goog and APL got it right.

adegeo:
_My final opinion on the matter..._

@perlun Scanning the article, there were many instances where it talked about "the app/application" but really it's talking about your code and doesn't need to call out an app. I would say if you wanted to remove app/application where it made sense and change it to a more generic term of code or program, go for it. But as @Rick-Anderson was pointing out, we probably don't need to choose one term over the other regarding app/application; people aren't going to be confused with app vs application.

perlun:
@adegeo 

> @perlun Scanning the article, there were many instances where it talked about "the app/application" but really it's talking about your code and doesn't need to call out an app. I would say if you wanted to remove app/application where it made sense and change it to a more generic term of code or program, go for it. But as @Rick-Anderson was pointing out, we probably don't need to choose one term over the other regarding app/application; people aren't going to be confused with app vs application.

Thanks, I think this might be some valid points you're making here. :+1: I edited the PR and amended it to speak about `your program`, to avoid the whole app vs application debate altogether. Please give it a look and see if you think it is an improvement from the original wording; if so, we could consider moving forward with it.

xml/System.Text/StringBuilder.xml
```diff
@@ -105,21 +105,21 @@
...
->  Although the <xref:System.Text.StringBuilder> class generally offers better performance than the <xref:System.String> class, you should not automatically replace <xref:System.String> with <xref:System.Text.StringBuilder> whenever you want to manipulate strings. Performance depends on the size of the string, the amount of memory to be allocated for the new string, the system on which your app is executing, and the type of operation. You should be prepared to test your app to determine whether <xref:System.Text.StringBuilder> actually offers a significant performance improvement.
+>  Although the <xref:System.Text.StringBuilder> class generally offers better performance than the <xref:System.String> class, you should not automatically replace <xref:System.String> with <xref:System.Text.StringBuilder> whenever you want to manipulate strings. Performance depends on the size of the string, the amount of memory to be allocated for the new string, the system on which your program is executing, and the type of operation. You should be prepared to test your program to determine whether <xref:System.Text.StringBuilder> actually offers a significant performance improvement.
```
Rick-Anderson, line 108: We consistently use app, sometimes application. We rarely use program. App is the standard we use.

adegeo: A lot of these would use code. This one specifically though I would say should remain app.

xml/System.Text/StringBuilder.xml
```diff
@@ -105,21 +105,21 @@
...
 
  Consider using the <xref:System.String> class under these conditions:
 
--   When the number of changes that your app will make to a string is small. In these cases, <xref:System.Text.StringBuilder> might offer negligible or no performance improvement over <xref:System.String>.
+-   When the number of changes that your program will make to a string is small. In these cases, <xref:System.Text.StringBuilder> might offer negligible or no performance improvement over <xref:System.String>.
```
adegeo, line 112: This type of wording I would use `code` instead. it could be a library or an app really. But in general, most of this all talks about the code you're writing.

perlun:
@adegeo and others, sorry for the delay but the PR is now updated & rebased on top of the latest `main`. Maybe this would be a slight improvement to the current wording.

xml/System.Text/StringBuilder.xml
```diff
@@ -783,7 +783,7 @@
  The <xref:System.Text.StringBuilder.Append%28System.Boolean%29> method modifies the existing instance of this class; it does not return a new class instance. Because of this, you can call a method or property on the existing reference and you do not have to assign the return value to a <xref:System.Text.StringBuilder> object, as the following example illustrates.
 
  :::code language="csharp" source="~/samples/snippets/csharp/VS_Snippets_CLR_System/system.Text.StringBuilder.Append/cs/append2.cs" interactive="try-dotnet-method" id="Snippet2":::
- :::code language="vb" source="~/samples/snippets/visualbasic/VS_Snippets_CLR_System/system.Text.StringBuilder.Append/vb/append2.vb" id="Snippet2"::: 
+ :::code language="vb" source="~/samples/snippets/visualbasic/VS_Snippets_CLR_System/system.Text.StringBuilder.Append/vb/append2.vb" id="Snippet2":::
```
perlun, line 786: (A few whitespace fixes like this slipped through, there were some trailing blanks and my VS Code is configured to trim blanks on save. Can revert these/move to a separate PR if you so prefer.)

tdykstra (approved):
Thanks, @perlun !

PR #6682, opened 2021-04-29
Fix unnecessary copy

john-h-k:
## Summary

Kinda defeats the point of showing eliding copy.. if you copy 😄

### DIFF
--- a/samples/snippets/csharp/api/system/span/slice2/Program.cs
+++ b/samples/snippets/csharp/api/system/span/slice2/Program.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Globalization;
 
 class Program
 {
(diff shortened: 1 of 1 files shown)


GrabYourPitchforks:
Nobody's going to mention the incorrect overload of `int.Parse` being used? :)

SingleAccretion:
And the trailing whitespaces...

GrabYourPitchforks:
As discussed offline, I think this entire sample needs to be rethought. There are two major problems with it: (a) it's not immediately clear that an `int` is being pulled out from the string data, because the sample isn't doing anything with it other than reflecting it back out to the output; and (b) a protocol parser necessarily involves using complicated overloads of `int.Parse`, which increases the concept count of this sample.

I think a better example would be to ask the user to enter a series of \[comma-?\]delimited integers ("4, 10, -45, 3"), then to `Console.ReadLine` this and use `ROS<char>.Slice`, `MemoryExtensions.IndexOf`, and `int.Parse` to pull out each _int_ and sum them, then write the sum to the output. This has the advantage that it teaches the developer how to use the very common `IndexOf` + `Slice` pattern, _and_ it shows performing an operation (addition) on the extracted data, _and_ it doesn't involve more complex issues like globalization.

BillWagner:
@GrabYourPitchforks @SingleAccretion @john-h-k 

Should this PR be closed in favor of a new PR or issue for the changes in the above comment?

jeffhandley:
Closed -- yes, a new PR that addresses the issues listed would be better. Thanks!

PR #6712, opened 2021-05-03
Add note that SafeCanonicalizationMethods is static-like.

vcsjones:
## Summary

This property, for historical purposes, is instance-based but is backed by a static field, so changes are observed between instances.

Fixes dotnet/runtime#46256


bartonjs:
Technical content LGTM.  @gewarren  do you have any style or language changes you'd like to see?

gewarren (approved):
LGTM. I'll merge it now; thanks!

PR #6747, opened 2021-05-13
Fix typo in comment in processstart.* - "is is".

GitHubPang:
Fix typo in comment in `processstart.*` - "is is" should be "it is".


tdykstra (approved):
Thanks for fixing this!.

PR #6761, opened 2021-05-24
Fix braces typo in IComparable csharp example

Busyrev:
Just missing newline


gewarren (approved):
This is probably not a typo, but a style preference - see https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#csharp_new_line_before_open_brace. But I think most of our other examples use this style, so we'll accept the PR.

PR #6769, opened 2021-05-26
Correction of an error in SslStreamCertificateContext documentation

dmitry-valyukov:
## Summary

Corrections of errors in SslStreamCertificateContext.Create doc.

Now it is:
**Parameters**
...
**`offline`** Boolean
`true` to indicate that the missing certificates can be downloaded from the network; `false` to indicate that only available X509Certificate stores should be searched for missing certificates.

But should be:
**`offline`** Boolean
`false` to indicate that the missing certificates can be downloaded from the network; `true` to indicate that only available X509Certificate stores should be searched for missing certificates.


ManickaP (approved):
LGTM
cc @wfurt just in case I overlooked something.

wfurt (approved):
LGTM

dmitry-valyukov:
"Only those with write access to this repository can merge pull requests."

PR #6775, opened 2021-05-28
Remove System.Drawing.Common

### LINKED ISSUE #6767: Asking if System.Drawing.Common is supported within ASP.NET
# Question
[System.Drawing Namespace - Remarks](https://docs.microsoft.com/en-us/dotnet/api/system.drawing?view=net-5.0#remarks) said that **System.Drawing namespace is not supported** within Windows services, ASP.NET Core, and ASP.NET.". Meanwhile, that said **System.Drawing.Common**, ImageSharp, SkiaSharp, and Windows Imaging Components is supported.

So, most of the System.Drawing namespaces aren't supported within ASP.NET.
However, **only System.Drawing.Common class** is supported, right? Or, is this a documentation error? 

From [System.Drawing.Common](https://www.nuget.org/packages/System.Drawing.Common/), it seems that System.Drawing.Common also depends on GDI+ graphics api.

# Docs Source File
https://github.com/dotnet/dotnet-api-docs/blob/187d891bba9291682e9bf749d3962d66c474f336/includes/drawing.md

gewarren:
Fixes #6767


ericstj (approved):
LGTM.

For those finding this PR.  System.Drawing.Common itself is what exposes System.Drawing.  This note is talking about that package as being not recommended nor supported in these scenarios.

PR #6778, opened 2021-05-28
Clarify misleading AllowAutoRedirect remarks

HebaruSan:
## Problem

This snippet is misleading in some cases:

> The Authorization header is cleared on auto-redirects and HttpWebRequest automatically tries to re-authenticate to the redirected location. In practice, this means that an application can't put custom authentication information into the Authorization header if it is possible to encounter redirection. Instead, the application must implement and register a custom authentication module. The System.Net.AuthenticationManager and related class are used to implement a custom authentication module. The AuthenticationManager.Register method registers a custom authentication module.

It's quite true that the Authorization header is cleared for an auto-redirect, but the rest of the paragraph strongly ("Instead, ...") makes it sound like you can naïvely rip out code that previously set the `Authorization` header and replace it by implementing and registering a new `IAuthenticationModule`-implementing object. I read this and happily set off to implement that solution, only to conclude a few hours later that it was a dead end in my case.

The problem is that this suggestion only works if you're dealing with a server where authentication is **mandatory**, which is not the case in all situations where the Authorization header might be sent. The GitHub API _allows_ authorization but does not _require_ it; the benefit if provided is a much higher [rate limit](https://docs.github.com/en/rest/reference/rate-limit). The GitHub API also uses redirects when a file is requested from a repo that has been moved or renamed (when the old location is requested).

If you attempt this solution for the GitHub API, **your `IAuthenticationModule` code will never be called**, because the server never demands authentication. Not only will your auto-redirects not be authorized (and therefore be subject to the GitHub API's lower anonymous rate limiting), as before attempting this solution, but _so will your **initial** requests_. It's lose-lose.

## Summary

I tried to keep the existing text intact as much as possible while adding caveats and clarification so the reader can tell more easily whether the old suggestion applies to his or her use case.

Then I added a new short paragraph explaining the impact for the optional authorization use case; as far as I can tell, there is no way to use `AllowAutoRedirect` successfully in this case, so it must be set to `false` and the redirects handled manually. If an official tech expert can correctly tell me differently, I will be delighted beyond words.

FYI to my team member @DasSkelett in case you feel like doing some proof-reading and confirming this is what you've seen in testing.

### DIFF
--- a/xml/System.Net/HttpWebRequest.xml
+++ b/xml/System.Net/HttpWebRequest.xml
@@ -1251,10 +1251,13 @@ Both <xref:System.Net.HttpWebRequest> constructors are obsolete and should not b
   
  If <xref:System.Net.HttpWebRequest.AllowAutoRedirect%2A> is set to `false`, all responses with an HTTP status code from 300 to 399 is returned to the application.  
   
- The Authorization header is cleared on auto-redirects and <xref:System.Net.HttpWebRequest> automatically tries to re-authenticate to the redirected location. In practice, this means that an application can't put custom authentication information into the Authorization header if it is possible to encounter redirection. Instead, the application must implement and register a custom authentication module. The <xref:System.Net.AuthenticationManager?displayProperty=nameWithType> and related class are used to implement a custom authentication module. The <xref:System.Net.AuthenticationManager.Register%2A?displayProperty=nameWithType> method registers a custom authentication module.  
-  
-   
-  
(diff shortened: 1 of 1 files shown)


ManickaP:
@aik-jahoda @wfurt @scalablecory @dotnet/ncl : could anyone give this a look? Is the proposed change technically OK?

aik-jahoda:
As HttpWebRequest is deprecated I would not extend documentation. Instead add reference to HttpClient. Also we should verify behaviour of HttpClient when auto redirect, however I believe it behaves same for security reason.

For the security reason I would not advise how to "fix" missing header and let people to think twice before they implement it (perhaps call the target URL directly).

PR #6781, opened 2021-06-01
Add note about System.Drawing.Common

gewarren:
Follow up to https://github.com/dotnet/dotnet-api-docs/issues/6767#issuecomment-850458124.


includes/drawing.md
```diff
@@ -1,3 +1,4 @@
...
-> - On Windows, `System.Drawing` depends on the GDI+ native library, which is shipped as part of the OS. Some Windows SKUs, like Windows Server Core or Windows Nano, don't include this native library as part of the OS. Exceptions will be thrown at run time because the library can't be loaded.
+> - On Windows, `System.Drawing` depends on the GDI+ native library, which is shipped as part of the OS. Some Windows SKUs, like Windows Server Core or Windows Nano, don't include this native library as part of the OS. If you use this namespace and the library can't be loaded, exceptions will be thrown at run time.
+> - The [System.Drawing.Common](https://www.nuget.org/packages/System.Drawing.Common/) NuGet package works on both Windows and Linux, however, there are some differences between the plaforms. Libgdiplus is not part of the Linux operating system and doesn't support all the functionality of gdiplus on Windows. There are also platforms where libgdiplus is not available at all.
```
ericstj, line 4: I believe this language needs to be stronger so that it doesn't contradict the above two remarks.  I'd prefer if we had the experts weigh in on this before making any changes.
@safern @akoeplinger @JeremyKuhne

akoeplinger: Yeah I think the whole section about `The System.Drawing namespace is not recommended, [..] Attempting to use System.Drawing classes within one of these application types may result in run-time exceptions` is a bit misleading since the primitives like Color, Size, Point, Rectangle which are in the System.Drawing **namespace**  work fine everywhere.

It's just the types that need GDI+ support like Bitmap, Font etc which are in the System.Drawing.Common NuGet package that run into the issues mentioned here.

Re: libgdiplus, I'd say something similar to:

```suggestion
> - The [System.Drawing.Common](https://www.nuget.org/packages/System.Drawing.Common/) NuGet package works on Windows, Linux and macOS. However, there are some differences between the plaforms. On Linux and macOS the GDI+ functionality is implemented by the libgdiplus library which is not installed by default in most Linux distributions and on macOS and doesn't support all the functionality of GDI+ on Windows. There are also platforms where libgdiplus is not available at all.
```

... and explain that you need to get libgdiplus separately, we have instructions on most of the dotnet install docs pages: <https://github.com/dotnet/docs/search?q=libgdiplus>

JeremyKuhne: I'd consider making the services bit a specific bullet point and link to the source info: https://docs.microsoft.com/en-us/windows/win32/gdiplus/-gdiplus-gdi-start#where-applicable

I'd also consider just having a bullet point with not recommended and recommended to start and keep the details in discrete bullet points.

Per @akoeplinger's comment it might be worth having another note saying that primitives such as Color, Size, Point, and Rectangle are safe to use.

@davidfowl do you know why ASP.Net doesn't work with GDI+? Is there something we can link to?

gewarren:
Here's a [preview](https://review.docs.microsoft.com/en-us/dotnet/api/system.drawing.drawing2d?view=dotnet-plat-ext-2.1&branch=pr-en-us-6781#remarks). How does it look?

akoeplinger (approved):
Looks good to me, thanks.

gewarren:
Thanks for the thoughtful reviews!

PR #6795, opened 2021-06-06
Added remark on Console.BackgroundColor and ForegroundColor.

Kuinox:
Console.BackgroundColor and ForegroundColor may return -1 on *nix platforms.
This behavior is intended but was only documented on its commit message.

Was discussed on https://github.com/dotnet/runtime/issues/53099.

### DIFF
--- a/xml/System/Console.xml
+++ b/xml/System/Console.xml
@@ -258,7 +258,7 @@
  :::code language="vb" source="~/samples/snippets/visualbasic/VS_Snippets_CLR_System/system.console.backgroundcolor/vb/backgroundcolor1.vb" id="Snippet1":::
 
  A get operation for a Windows-based application, in which a console does not exist, returns <xref:System.ConsoleColor.Black?displayProperty=nameWithType>.
- On Linux, because there is no general mechanism to fetch the current Console colors, <xref:System.Console.BackgroundColor%2A> will return `(ConsoleColor)-1` until it has been set.
+ Unix systems don't provide any general mechanism to fetch the current console colors. Because of that, <xref:System.Console.BackgroundColor%2A> returns `(ConsoleColor)-1` until it is set in explicit way (using the setter).
 
 
 ## Examples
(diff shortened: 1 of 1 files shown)


xml/System/Console.xml
```diff
@@ -258,7 +258,7 @@
  :::code language="vb" source="~/samples/snippets/visualbasic/VS_Snippets_CLR_System/system.console.backgroundcolor/vb/backgroundcolor1.vb" id="Snippet1":::
 
  A get operation for a Windows-based application, in which a console does not exist, returns <xref:System.ConsoleColor.Black?displayProperty=nameWithType>.
-
+ On Linux, because there is no general mechanism to fetch the current Console colors, <xref:System.Console.BackgroundColor%2A> will return `(ConsoleColor)-1` until it has been set.
```
adamsitnik, line 261: It's not specific only to Linux, it applies to all Unixes. I would reword it in the following way:

```suggestion
 Unix systems don't provide any general mechanism to fetch the current console colors. Because of that, <xref:System.Console.BackgroundColor%2A> returns `(ConsoleColor)-1` until it is set in explicit way (using the setter).
```

xml/System/Console.xml
```diff
@@ -1208,6 +1208,7 @@
 
 ## Remarks
  A get operation for a Windows-based application, in which a console does not exist, returns <xref:System.ConsoleColor.Gray>.
+ On Linux, because there is no general mechanism to fetch the current Console colors, <xref:System.Console.ForegroundColor%2A> will return `(ConsoleColor)-1` until it has been set.
```
adamsitnik, line 1211: same as above:

```suggestion
  Unix systems don't provide any general mechanism to fetch the current console colors. Because of that, <xref:System.Console.ForegroundColor%2A> returns `(ConsoleColor)-1` until it is set in explicit way (using the setter).
```

adamsitnik (approved):
LGTM, thank you for improving our docs @Kuinox !

PR #6807, opened 2021-06-09
Update IHostApplicationLifetime.xml

explorest:
## Summary

The members ApplicationStopped and ApplicationStopping both have exact same description in the table. This can't possibly be correct. Must be an editing mistake.  
Also , I have tested it on a real application. They are indeed different events. ApplicationStopping indicates that the process of stopping has commenced (e.g. it occurs BEFORE any IHostedService.StopAsync calls are made to hosted services) while ApplicationStopped indicates the process has completed ( e.g. it occurs AFTER all IHostedService.StopAsync calls  have returned).


eerhardt (approved):
LGTM. Nice catch!

PR #6809, opened 2021-06-09
Remove incorrect exception

### LINKED ISSUE #6788: Documented IOException isn't thrown
In [ZipFile.ExtractToDirectory](https://docs.microsoft.com/en-us/dotnet/api/system.io.compression.zipfile.extracttodirectory?view=net-5.0) and [ZipFileExtensions.ExtractToDirectory](https://docs.microsoft.com/en-us/dotnet/api/system.io.compression.zipfileextensions.extracttodirectory?view=net-5.0) the Exceptions section says that an IOException is thrown if the target directory exists. This is incorrect, as the method that is eventually called by these, the internal [ExtractRelativeToDirectory](https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/src/libraries/System.IO.Compression.ZipFile/src/System/IO/Compression/ZipFileExtensions.ZipArchiveEntry.Extract.cs#L87), works fine with existing directories. 

This was first asked in [this SO question](https://stackoverflow.com/questions/67806595/does-zipfile-extracttodirectory-throw-an-exception-when-the-directory-exists/67807550). I also opened an [issue in the runtime repo](https://github.com/dotnet/runtime/issues/53609)

https://github.com/dotnet/dotnet-api-docs/blob/46c62678f62c1f379d5b3073ab0738171da212f8/xml/System.IO.Compression/ZipFileExtensions.xml#L297

gewarren:
Fixes #6788


BillWagner (approved):
This LGTM @gewarren 

One the product team weighs in, you can :shipit:

PR #6814, opened 2021-06-10
Remove exception from IPAddress.TryParse(string, out IPAddress)

akoeplinger:
It returns false instead of throwing an exception in .NET Framework and .NET Core: https://github.com/dotnet/runtime/blob/43b1f2c165257faae4d6a4aa886b701b63d2d5c0/src/libraries/System.Net.Primitives/src/System/Net/IPAddress.cs#L216-L222

### DIFF
--- a/xml/System.Net/IPAddress.xml
+++ b/xml/System.Net/IPAddress.xml
@@ -2173,8 +2173,6 @@
   
  ]]></format>
         </remarks>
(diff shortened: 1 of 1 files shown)


antonfirsov (approved):
LGTM

gewarren:
The build system went crazy last night. Sorry about that!

PR #6833, opened 2021-06-21
Update ComWrappers.xml

AaronRobinsonMSFT:
## Summary

Clarify what is returned from the `GetOrCreateComInterfaceForObject()`.

/cc @eerhardt

### DIFF
--- a/xml/System.Runtime.InteropServices/ComWrappers.xml
+++ b/xml/System.Runtime.InteropServices/ComWrappers.xml
@@ -230,7 +230,7 @@ If the object cannot be created and `null` is returned, the call to <xref:System
         <param name="instance">The managed object to expose outside the .NET runtime.</param>
         <param name="flags">Flags used to configure the generated interface.</param>
         <summary>Creates a COM representation of the supplied object that can be passed to a non-managed environment.</summary>
-        <returns>The generated COM interface, set to the <code>IUnknown</code> vtable, that can be passed outside the .NET runtime.</returns>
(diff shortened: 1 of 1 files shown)


xml/System.Runtime.InteropServices/ComWrappers.xml
```diff
@@ -230,7 +230,7 @@ If the object cannot be created and `null` is returned, the call to <xref:System
         <param name="instance">The managed object to expose outside the .NET runtime.</param>
         <param name="flags">Flags used to configure the generated interface.</param>
         <summary>Creates a COM representation of the supplied object that can be passed to a non-managed environment.</summary>
-        <returns>The generated COM interface that can be passed outside the .NET runtime.</returns>
+        <returns>The generated COM interface, set to the <code>IUnknown</code> vtable, that can be passed outside the .NET runtime.</returns>
```
jkoritzinsky, line 233:
```suggestion
        <returns>The generated COM interface, as a pointer to the <code>IUnknown</code> interface implementation, that can be passed outside the .NET runtime.</returns>
```

I don't like the wording of "pointer to the vtable" since it's a pointer to the COM object, not the vtable itself.

eerhardt (approved):
Thanks for putting this up there. Hopefully it saves someone else time 😄

xml/System.Runtime.InteropServices/ComWrappers.xml
AaronRobinsonMSFT: Great point. Thanks.

jkoritzinsky (approved):

PR #6838, opened 2021-06-22
Escape backslash

### LINKED ISSUE #5299: AppContext doc has incorrectly formatted registry key
The AppContext [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.appcontext?view=netcore-3.1#remarks), under subheading "AppContext for library consumers", describes several ways to set a switch. The third method uses the registry and mentions two registry keys.  Both keys are missing a backslash between "Microsoft" and ".NET":  

**HKLM\SOFTWARE\Microsoft.NETFramework\AppContext** should be
**HKLM\SOFTWARE\Microsoft\\.NETFramework\AppContext**

**HKLM\SOFTWARE\Wow6432Node\Microsoft.NETFramework\AppContext** should be
**HKLM\SOFTWARE\Wow6432Node\Microsoft\\.NETFramework\AppContext**

gewarren:
Fixes #5299.

Preview links:

- [AppContext](https://review.docs.microsoft.com/en-us/dotnet/api/system.appcontext?view=net-6.0&branch=pr-en-us-6838#appcontext-for-library-consumers)
- [SetSwitch method](https://review.docs.microsoft.com/en-us/dotnet/api/system.appcontext.setswitch?view=net-6.0&branch=pr-en-us-6838#remarks)

### DIFF
--- a/xml/System/AppContext.xml
+++ b/xml/System/AppContext.xml
@@ -137,9 +137,9 @@
     > [!NOTE]
     > For information on the switches defined by the .NET Framework, see the [\<AppContextSwitchOverrides> element](/dotnet/framework/configure-apps/file-schema/runtime/appcontextswitchoverrides-element).
 
-- By adding an entry to the registry. Add a new string value to the **HKLM\SOFTWARE\Microsoft\.NETFramework\AppContext** subkey. Set the name of the entry to the name of the switch. Set its value to one of the following options: `True`, `true`, `False`, or `false`. If the runtime encounters any other value, it ignores the switch.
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @gewarren 

I'm curious why you only needed to escape the `\` before the `.NET`. I would have thought all would need to be escaped.

gewarren:
> I'm curious why you only needed to escape the `\` before the `.NET`. I would have thought all would need to be escaped.

It seems that the single backslash escapes the "." which must be a special character.

PR #6841, opened 2021-06-23
Verb remarks added

robertmuehsig:
Based on this [issue](https://github.com/dotnet/runtime/issues/52960) I think the remarks section should include this information.

## Summary

The resulting verbs are not always correct and I thought this was a bug, but it seems this is "more or less" by design.


adamsitnik (approved):
LGTM, thank you @robertmuehsig

PR #6843, opened 2021-06-24
WinForms Timer is not .Net Framework only

svick:
Since .Net Core 3, WinForms are not .Net Framework only, so the documentation shouldn't claim that the WinForms `Timer` is .Net Framework only.


kouvel (approved):

gewarren:
Thank you @svick

PR #6849, opened 2021-06-28
Add remarks for StandardInputEncoding

imba-tjd:
Do reword my content if necessary.

## Reason

This is a C Program that reads by UTF16:

```c
#include <io.h>
#include <fcntl.h>
#include <stdio.h>
#include <wchar.h>

int main() {
    _setmode(_fileno(stdin), _O_U16TEXT);
    _setmode(_fileno(stdout), _O_U16TEXT);

    wchar_t input[10] = {};
    fgetws(input, 10, stdin);
    wprintf(L"%X\n", input[0]);
    fputws(input, stdout);
}
```

When executing it from Console, it works normally. For example type `asdf`:

```
asdf # Type
61
asdf
```

But when executing it from C# with `Encoding.Unicode`, it won't work normally:

```c#
using System;
using System.Text;
using System.Diagnostics;

class Program {
    static void Main() {
        ProcessStartInfo psi = new() {
            FileName = "tr.exe",
            UseShellExecute = false,
            RedirectStandardInput = true,
            StandardInputEncoding = Encoding.Unicode,
            // StandardInputEncoding = new UnicodeEncoding(false, false),
            RedirectStandardOutput = true,
            StandardOutputEncoding = Encoding.Unicode
        };
        var p = Process.Start(psi);
        p.StandardInput.WriteLine("asdf");
        p.WaitForExit();
        string s = p.StandardOutput.ReadToEnd();
        Console.WriteLine(s);
    }
}
```

The output:

```
FEFF
?asdf
```

`FEFF` is UTF16 BE BOM. I think it's because C is LE, the actual bytes are `FF, FE`, UTF16 LE, but when outputing the FE is interpreted as high part.

The correct way is to use `new UnicodeEncoding(false, false)`.

### DIFF
--- a/xml/System.Diagnostics/ProcessStartInfo.xml
+++ b/xml/System.Diagnostics/ProcessStartInfo.xml
@@ -1579,7 +1579,7 @@ You can use asynchronous read operations to avoid these dependencies and their d
 ## Remarks  
  If the value of the <xref:System.Diagnostics.ProcessStartInfo.StandardOutputEncoding%2A> property is `null`, the process uses the default standard output encoding for the standard output. The <xref:System.Diagnostics.ProcessStartInfo.StandardOutputEncoding%2A> property must be set before the process is started. Setting this property does not guarantee that the process will use the specified encoding. The application should be tested to determine which encodings the process supports.  
 
-> [!IMPORTANT]
+> [!NOTE]
 >  When interacting with applications that cannot handle BOM correctly consider using BOM-less encoding. For example use `new UnicodeEncoding(false, false)` rather than `Encoding.Unicode`.
   
  ]]></format>


xml/System.Diagnostics/ProcessStartInfo.xml
```diff
@@ -1578,6 +1578,9 @@ You can use asynchronous read operations to avoid these dependencies and their d
   
 ## Remarks  
  If the value of the <xref:System.Diagnostics.ProcessStartInfo.StandardOutputEncoding%2A> property is `null`, the process uses the default standard output encoding for the standard output. The <xref:System.Diagnostics.ProcessStartInfo.StandardOutputEncoding%2A> property must be set before the process is started. Setting this property does not guarantee that the process will use the specified encoding. The application should be tested to determine which encodings the process supports.  
+
+> [!IMPORTANT]
+>  Use encodings that don't have BOM. For example use `new UnicodeEncoding(false, false)` rather than `Encoding.Unicode`.
```
krwq, line 1583: I think this remark without this issue as a context is very confusing.

krwq:
```suggestion
>  When interacting with applications that cannot handle BOM correctly consider using BOM-less encoding. For example use `new UnicodeEncoding(false, false)` rather than `Encoding.Unicode`.
```

krwq:
cc: @adamsitnik @carlossanlop @jozkee

PR #6851, opened 2021-06-30
Update IDisposable.xml

### LINKED ISSUE #5035: Very wrong advice about implementing IDisposable
At https://docs.microsoft.com/en-us/dotnet/api/system.idisposable?view=netcore-3.1#implementing-idisposable it reads:

> You should implement `IDisposable` only if your type uses unmanaged resources directly.

This advice is very wrong.
You should implement `IDisposable` if you use unmanaged resources directly or if you use disposable objects yourself.

Which is why the Disposable pattern (correctly) generated by Visual Studio goes:

```cs
protected virtual void Dispose(bool disposing)
{
    if (disposing)
    {
        // get rid of managed resources
    }   
    // get rid of unmanaged resources
}
```

steveberdy:
## Summary

Line 94 fixes issue #5035 by being more specific about usages of the `IDisposable.Dispose` method, as well as providing a link to the article about the `Dispose(bool)` method overload.

Fixes #5035


xml/System/IDisposable.xml
```diff
@@ -91,7 +91,7 @@
...
+ You should implement <xref:System.IDisposable> if your type uses unmanaged resources directly or if you wish to use unmanaged resources yourself. The consumers of your type can call your <xref:System.IDisposable.Dispose%2A?displayProperty=nameWithType> implementation to free resources when the instance is no longer needed. To handle cases in which they fail to call <xref:System.IDisposable.Dispose%2A>, you should either use a class derived from <xref:System.Runtime.InteropServices.SafeHandle> to wrap the unmanaged resources, or you should override the <xref:System.Object.Finalize%2A?displayProperty=nameWithType> method for a reference type. In either case, you use the <xref:System.IDisposable.Dispose%2A> method to perform whatever cleanup is necessary after using the unmanaged resources, such as freeing, releasing, or resetting the unmanaged resources. For more information about implementing <xref:System.IDisposable.Dispose%2A?displayProperty=nameWithType>, see [the Dispose(bool) method overload](/dotnet/standard/garbage-collection/implementing-dispose#the-disposebool-method-overload).
```
BillWagner, line 94: One nit to fix here:

```suggestion
 You should implement <xref:System.IDisposable> if your type uses unmanaged resources directly or if you wish to use disposable resources yourself. The consumers of your type can call your <xref:System.IDisposable.Dispose%2A?displayProperty=nameWithType> implementation to free resources when the instance is no longer needed. To handle cases in which they fail to call <xref:System.IDisposable.Dispose%2A>, you should either use a class derived from <xref:System.Runtime.InteropServices.SafeHandle> to wrap the unmanaged resources, or you should override the <xref:System.Object.Finalize%2A?displayProperty=nameWithType> method for a reference type. In either case, you use the <xref:System.IDisposable.Dispose%2A> method to perform whatever cleanup is necessary after using the unmanaged resources, such as freeing, releasing, or resetting the unmanaged resources. For more information about implementing <xref:System.IDisposable.Dispose%2A?displayProperty=nameWithType>, see [the Dispose(bool) method overload](/dotnet/standard/garbage-collection/implementing-dispose#the-disposebool-method-overload).   
```

BillWagner (changes requested):
Thanks for making this update @steveberdy 

I have one word change, and then this is ready to :shipit:

Thanks for making your first contribution to docs! 🎆

BillWagner (approved):

PR #6857, opened 2021-07-02
Fix method signature format and clarify functional complements

### LINKED ISSUE #5152: Linq Take() and Skip() are stated as Functinally complement, while it is not always true
Honestly, I am not sure if it is Documentation issue of code bug.
Documentation states:
The `Take` and `Skip` methods are functional complements. Given a sequence `coll` and an integer `n`, concatenating the results of `coll.Take(n)` and `coll.Skip(n)` yields the same sequence as `coll`. 
https://github.com/dotnet/dotnet-api-docs/blob/f33aa309d83192d0563b9bb35c70c176238f8f85/xml/System.Linq/Enumerable.xml#L9591 (skip)
https://github.com/dotnet/dotnet-api-docs/blob/f33aa309d83192d0563b9bb35c70c176238f8f85/xml/System.Linq/Enumerable.xml#L11251 (take)

However, that is not always true, Take and Skip sometimes silently throw away elements, and taken.Concat(skipped) is not equal to original:
```csharp
class TakeSkipEnumerableFail
    {
        private static readonly string[] array = Enumerable.Range(1, 7).Select(i => $"{i}").ToArray();
        private static int position = 0;
        private static IEnumerable<string> GetIEnumerable()
        {
            while (position < array.Length)
            {
                yield return array[position];
                position++;
            }
        }
        static void Main()
        {
            var sequence = GetIEnumerable();
            var before = sequence.Take(3);
            var after = sequence.Skip(3);
            var concat = before.Concat(after); // should be 1,2,3,4,5,6,7
            var result = $"concat: {string.Join(",", concat)}"; // but actually it is 1,2,3,6,7
            Console.WriteLine(result);
        }
    }
```
Most probably it is due to nature of this very IEnumerable (it is true for enumerations that are not repeatable), but it makes sense to mention this in documentation.

### LINKED ISSUE #6790: Member signatures in C# don't follow the official formatting rules
E.g. the GC.AddMemoryPressure has the following signature format:

https://github.com/dotnet/dotnet-api-docs/blob/54e27a3d78dcbaedc38a38478cdaefa0be962936/xml/System/GC.xml#L112

Notice the space before the parentheses. This is not how C# signatures are formatted in the rest of the repo or samples. Instead of 

```
public static void AddMemoryPressure (long bytesAllocated);
```

This should be:

```
public static void AddMemoryPressure(long bytesAllocated);
```

steveberdy:
### Edits to GC.xml (issue #6790)

This fixes the space before the parentheses in the method signature, making it comply with official formatting rules.

### Edits to Enumerable.xml (issue #5152)

These fixes add the slight detail that the `Take` and `Skip` methods in `System.Linq` are complements ***given first*** that the enumerable is not repeatable.

Fixes #6790
Fixes #5152

### DIFF
--- a/samples/snippets/cpp/VS_Snippets_CLR/Generic.SortedList/cpp/source.cpp
+++ b/samples/snippets/cpp/VS_Snippets_CLR/Generic.SortedList/cpp/source.cpp
@@ -107,7 +107,7 @@ public:
         IList<String^>^ ilistValues = openWith->Values;
 
         // The elements of the list are strongly typed with the
-        // type that was specified for the SorteList values.
+        // type that was specified for the SortedList values.
         Console::WriteLine();
         for each( String^ s in ilistValues )
         {
--- a/samples/snippets/csharp/VS_Snippets_CLR/Generic.SortedList/CS/source.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR/Generic.SortedList/CS/source.cs
@@ -104,7 +104,7 @@ public class Example
         IList<string> ilistValues = openWith.Values;
 
         // The elements of the list are strongly typed with the
-        // type that was specified for the SorteList values.
+        // type that was specified for the SortedList values.
         Console.WriteLine();
         foreach( string s in ilistValues )
         {


krwq:
@eiriktsarpalis @layomia can you review?

xml/System.Linq/Enumerable.xml
```diff
@@ -9902,7 +9902,7 @@ Only unique elements are returned.
   
  If `source` contains fewer than `count` elements, an empty <xref:System.Collections.Generic.IEnumerable%601> is returned. If `count` is less than or equal to zero, all elements of `source` are yielded.  
   
- The <xref:System.Linq.Enumerable.Take%2A> and <xref:System.Linq.Enumerable.Skip%2A> methods are functional complements. Given a sequence `coll` and an integer `n`, concatenating the results of `coll.Take(n)` and `coll.Skip(n)` yields the same sequence as `coll`.  
+ The <xref:System.Linq.Enumerable.Take%2A> and <xref:System.Linq.Enumerable.Skip%2A> methods are functional complements. Given an unrepeatable sequence `coll` and an integer `n`, concatenating the results of `coll.Take(n)` and `coll.Skip(n)` yields the same sequence as `coll`.
```
eiriktsarpalis, line 9905: I'm not sure I understand the meaning of the phrase "unrepeatable sequence". Could you clarify?

steveberdy: An unrepeating sequence; one who's items only are iterated over once in the iteration of the sequence as a whole.

eiriktsarpalis: > one who's items only are iterated over once in the iteration of the sequence as a whole.

I see your point, but I think the purpose of this comment is to provide an illustration on the semantics of `Take` and `Skip`, rather than state a universal invariant. I don't think that the particular change adds clarity to the statement. How about something like "Given a collection sequence"?

steveberdy: That sounds better. I'm not able to edit this any more since I deleted the fork a week ago. Could you add in this change?

adegeo: You can't update this without the original fork. I've cloned this PR and made the change. Would you @steveberdy and @eiriktsarpalis like me to open a new PR with the change?

steveberdy: I don't mind. Just mention this PR in your pull request so I can close it when yours gets merged.

steveberdy: @adegeo I'll submit another PR for GC.xml. I'm assuming your PR just has changes to Enumerable.xml?

adegeo: I'll clone this PR so it will have all the changes. Is that OK?

steveberdy: Yes

adegeo:
I'll close this in favor of #6883

PR #6868, opened 2021-07-07
Clarify memory load

gewarren:
See https://github.com/dotnet/runtime/issues/55126#issuecomment-875246868.

### DIFF
--- a/xml/System/GCMemoryInfo.xml
+++ b/xml/System/GCMemoryInfo.xml
@@ -75,9 +75,9 @@ A garbage collection (GC) is identified by its <xref:System.GCMemoryInfo.Index>,
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets a value that indicates if this is a background or concurrent GC or not.</summary>
+        <summary>Gets a value that indicates if this is a concurrent GC (background GC) or not.</summary>
         <value>
-          <see langword="true" /> if this is a background or concurrent GC; <see langword="false" /> otherwise.</value>
+          <see langword="true" /> if this is a concurrent GC (background GC); <see langword="false" /> otherwise.</value>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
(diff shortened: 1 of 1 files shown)


xml/System/GCMemoryInfo.xml
```diff
@@ -75,9 +75,9 @@ A GC is identified by its <xref:System.GCMemoryInfo.Index>, which starts from 1
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets a value that indicates if this is a concurrent GC (BGC) or not.</summary>
+        <summary>Gets a value that indicates if this is a background or concurrent GC or not.</summary>
```
Maoni0, line 78: background GCs are concurrent. so I would keep the original text; or if you want to make it more explicit you could say 

Gets a value that indicates if this is a background (which is concurrent) or not.

xml/System/GCMemoryInfo.xml
```diff
@@ -279,9 +279,9 @@ The memory after `OBJ_D` is not considered part of the `FragmentedBytes` but is
         <ReturnType>System.Int64</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets the memory load when the last garbage collection occurred.</summary>
-        <value>The memory load when the last garbage collection occurred.</value>
-        <remarks>To be added.</remarks>
+        <summary>Gets the physical memory load when the last garbage collection occurred.</summary>
+        <value>The physical memory load, in bytes, when the last garbage collection occurred.</value>
+        <remarks>If you don't touch the pages of an object, the memory load doesn't change significantly.</remarks>
```
Maoni0, line 284: I would probably say this -

"Data is only brought into physical memory on first touch so if you allocated a big object but doesn't actually use it, most of its memory will not be in physical memory so you could see the memory load change much."

xml/System/GCMemoryInfo.xml
gewarren, line 78:
```suggestion
        <summary>Gets a value that indicates if this is a concurrent GC (background GC) or not.</summary>
```

xml/System/GCMemoryInfo.xml
```diff
@@ -75,9 +75,9 @@ A GC is identified by its <xref:System.GCMemoryInfo.Index>, which starts from 1
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets a value that indicates if this is a concurrent GC (BGC) or not.</summary>
+        <summary>Gets a value that indicates if this is a background or concurrent GC or not.</summary>
         <value>
-          <see langword="true" /> if this is a concurrent GC (BGC); <see langword="false" /> otherwise.</value>
+          <see langword="true" /> if this is a background or concurrent GC; <see langword="false" /> otherwise.</value>
```
gewarren, line 80:
```suggestion
          <see langword="true" /> if this is a concurrent GC (background GC); <see langword="false" /> otherwise.</value>
```

xml/System/GCMemoryInfo.xml
gewarren, line 284:
```suggestion
        <remarks>Data is only brought into physical memory on first touch. If you allocated a big object but haven't actually used it, most of its memory isn't in physical memory. In this case, the allocation won't affect the memory load significantly.</remarks>
```

Maoni0 (approved):
thanks so much!

PR #6883, opened 2021-07-13
Fix method signature format and clarify functional complements

### LINKED ISSUE #5152: Linq Take() and Skip() are stated as Functinally complement, while it is not always true

### LINKED ISSUE #6790: Member signatures in C# don't follow the official formatting rules

adegeo:
## Summary

**Originally by @steveberdy**

### Edits to GC.xml (issue #6790)

This fixes the space before the parentheses in the method signature, making it comply with official formatting rules.

### Edits to Enumerable.xml (issue #5152)

These fixes add the slight detail that the `Take` and `Skip` methods in `System.Linq` are complements ***given first*** that the enumerable is not repeatable.

Fixes #6790
Fixes #5152

### DIFF
--- a/xml/System.Linq/Enumerable.xml
+++ b/xml/System.Linq/Enumerable.xml
@@ -9902,7 +9902,7 @@ Only unique elements are returned.
   
  If `source` contains fewer than `count` elements, an empty <xref:System.Collections.Generic.IEnumerable%601> is returned. If `count` is less than or equal to zero, all elements of `source` are yielded.  
   
- The <xref:System.Linq.Enumerable.Take%2A> and <xref:System.Linq.Enumerable.Skip%2A> methods are functional complements. Given an unrepeatable sequence `coll` and an integer `n`, concatenating the results of `coll.Take(n)` and `coll.Skip(n)` yields the same sequence as `coll`.  
+ The <xref:System.Linq.Enumerable.Take%2A> and <xref:System.Linq.Enumerable.Skip%2A> methods are functional complements. Given a collection sequence `coll` and an integer `n`, concatenating the results of `coll.Take(n)` and `coll.Skip(n)` yields the same sequence as `coll`.  
   
  In Visual Basic query expression syntax, a `Skip` clause translates to an invocation of <xref:System.Linq.Enumerable.Skip%2A>.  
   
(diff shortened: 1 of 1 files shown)


adegeo:
@eiriktsarpalis this just has the updated wording you suggested in the other PR. Please approve. Cheers!

eiriktsarpalis (approved):

eiriktsarpalis:
Thanks!

gewarren:
The method signature is autogenerated, so that issue will have to be fixed in our tooling, not through manual edits.

adegeo:
@gewarren where would that be?

gewarren:
@adegeo I think here: https://github.com/mono/api-doc-tools

adegeo:
@gewarren Oh, ha, I was thinking you meant the description needed that. I'll file a bug in that repo.

PR #6886, opened 2021-07-15
Update ChannelReader`1.xml

### LINKED ISSUE #6751: ChannelReader<T>.ReadAllAsync(CancellationToken) - important cancellation details missing in docs
The documentation about ChannelReader<T>.ReadAllAsync(CancellationToken) does not clearly state that a cancellation request through the passed token will not take effect until all present items in the channel are extracted, and given no others are being written to the channel at the same time.

The text "The cancellation token to use to cancel the enumeration." should be altered to point out the cancellation will occur after all items in the channel are read (and not on next iteration after the token was triggered, as it might be expected), in order to conform with the actual behavior of the ReadAllAsync method.

steveberdy:
## Summary

Update `cancellationToken` parameter documentation in `ChannelReader.ReadAllAsync` to  specify when the cancellation occurs.

### Reference
- .NET Runtime [code](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Threading.Channels/src/System/Threading/Channels/ChannelReader.netstandard21.cs)

Fixes #6751

### DIFF
--- a/xml/System.Threading.Channels/ChannelReader`1.xml
+++ b/xml/System.Threading.Channels/ChannelReader`1.xml
@@ -139,7 +139,7 @@
         <Parameter Name="cancellationToken" Type="System.Threading.CancellationToken" Index="0" FrameworkAlternate="dotnet-plat-ext-3.0;netcore-3.0;netcore-3.1;net-5.0;net-6.0" />
       </Parameters>
       <Docs>
-        <param name="cancellationToken">The cancellation token to use to cancel the enumeration. Cancellation occurs after reading is finished.</param>
+        <param name="cancellationToken">The cancellation token to use to cancel the enumeration. If data is immediately ready for reading, then that data may be yielded even after cancellation has been requested.</param>
         <summary>Creates an <see cref="T:System.Collections.Generic.IAsyncEnumerable`1" /> that enables reading all of the data from the channel.</summary>
         <returns>The created async enumerable.</returns>
         <remarks>


steveberdy:
@BillWagner I'm not sure how the automation works, but if this PR gets merged, will the comments in the [dotnet/runtime](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Threading.Channels/src/System/Threading/Channels/ChannelReader.netstandard21.cs) file get updated or should I send a separate PR for that repo as well? The same would also apply for my previous PRs in this repo.

steveberdy:
@BillWagner please review.

steveberdy:
Can we review this for merging?

xml/System.Threading.Channels/ChannelReader`1.xml
```diff
@@ -139,7 +139,7 @@
         <Parameter Name="cancellationToken" Type="System.Threading.CancellationToken" Index="0" FrameworkAlternate="dotnet-plat-ext-3.0;netcore-3.0;netcore-3.1;net-5.0;net-6.0" />
       </Parameters>
       <Docs>
-        <param name="cancellationToken">The cancellation token to use to cancel the enumeration.</param>
+        <param name="cancellationToken">The cancellation token to use to cancel the enumeration. Cancellation occurs after reading is finished.</param>
```
adamsitnik, line 142: I am not sure if "Cancellation occurs after reading is finished." is accurate description. Perhaps we should write something like this:

```suggestion
        <param name="cancellationToken">The cancellation token to use to cancel the enumeration. If there's already data available for reading immediately, cancellation occurs after reading is finished.</param>
```

@stephentoub do you have any opinion on that?

steveberdy: Hi @adamsitnik your changes look reasonable to me.

stephentoub: > cancellation occurs after reading is finished.

This part isn't quite right.  Cancellation might never occur, and I also don't want the docs to imply that it might read the value and then discard the value due to being canceled.

I think the sentiment this is trying to express is more like:
"If data is immediately available for reading, an implementation may prefer to yield that data immediately, even if cancellation has been requested."
Or something like that.

steveberdy: Done!

carlossanlop (approved):
LGTM after applying [Stephen's suggestion](https://github.com/dotnet/dotnet-api-docs/pull/6886#discussion_r685263907).

PR #6892, opened 2021-07-15
Mark exceptions as .NET Framework only

### LINKED ISSUE #6630: Documentation for System.Version operators is inconsistent with implementation
The documentation about the comparison operators of the `System.Version` class mentions that an `ArgumentNullException` is thrown if one of the provided parameters is `null` (See for example [Version.LessThan](https://docs.microsoft.com/en-us/dotnet/api/system.version.op_lessthan?view=net-5.0)).

In the reference sources for .NET framework this is indeed the case (as can be seen [here](https://github.com/microsoft/referencesource/blob/5697c29004a34d80acdaf5742d7e699022c64ecd/mscorlib/system/version.cs#L450)), but for the implementation for .NET it does not seem to be the case, and actually allows the arguments to be `null` (as can be seen [here](https://github.com/dotnet/runtime/blob/3cbbadee12cc95bd62c70786d5408a2277a21e0a/src/libraries/System.Private.CoreLib/src/System/Version.cs#L387)).

gewarren:
Fixes #6630

(Hide whitespace changes.)

### DIFF
--- a/xml/System/Version.xml
+++ b/xml/System/Version.xml
@@ -76,88 +76,88 @@
   <Docs>
     <summary>Represents the version number of an assembly, operating system, or the common language runtime. This class cannot be inherited.</summary>
     <remarks>
-      <format type="text/markdown"><![CDATA[  
-  
-## Remarks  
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

gewarren:
Thanks Bill!

PR #6894, opened 2021-07-16
Mark SocketException versions

### LINKED ISSUE #6658: Remove SokcetException from Begin*** methods for 6.0
There is an ongoing refactor in 6.0 (https://github.com/dotnet/runtime/issues/43845), resulting in the `Socket.Begin***` methods no longer throwing `SocketException`. We should remove `SocketException` from the exception list of the methods being refactored for 6.0, hopefully all of them.

gewarren:
Fixes #6658


gewarren:
@antonfirsov Can you review? Also, it seems like `BeginConnect (string host, int port, AsyncCallback? requestCallback, object? state)` never threw a SocketException?

antonfirsov:
@gewarren on vacation now, if no one will be faster than me, will review by end of next week.

gewarren:
@antonfirsov Looks like no one else picked this up, so could you review it?

antonfirsov (approved):
LGTM! Apologies for forgetting about this, hope we can still merge this on time.

PR #6900, opened 2021-07-18
Using more modern syntax for try parse

WhiteBlackGoose:
It would be more relevant so that newcomers don't pull old techniques

I'm using string interpolation instead of old Console WriteLine's API. Also replaced built-in types by the corresponding keywords

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR_System/system.Int32.TryParse/cs/TryParse1.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR_System/system.Int32.TryParse/cs/TryParse1.cs
@@ -11,7 +11,7 @@ public class Example
       {
          int number;
 
-         bool success = int.TryParse(value, out number);   
+         bool success = int.TryParse(value, out number);
          if (success)
          {
             Console.WriteLine($"Converted '{value}' to {number}.");


WhiteBlackGoose:
Please, let me know if it's useful. If so, I will add more PRs replacing
1) Types to keywords when appropriate
2) Convert -> Parse or ToString when appropriate
3) String/WriteLine format -> string interpolation

WhiteBlackGoose:
Also, is it intentend to use 2-3 space indentations instead of the standard 4-space one?

BillWagner (approved):
Thanks for making these changes @WhiteBlackGoose We appreciate it!

I've reviewed the changes, and I'll :shipit: now. Thank you. You should see the changes on our live site in the next day or two, on our regular publishing cycle.

To answer your questions:

> Please, let me know if it's useful. If so, I will add more PRs replacing
>
> 1. Types to keywords when appropriate
> 1. Convert -> Parse or ToString when appropriate
> 1. String/WriteLine format -> string interpolation

Yes please. We do prefer those. Note that we should be careful on where to replace `Convert` with another method. We should not replace the samples on the `Convert` API pages.

> Also, is it intended to use 2-3 space indentations instead of the standard 4-space one?

Usually we use 4 spaces. We use 2-3 when that's needed to a better web page presentation.

Thanks again. 🎆

PR #6901, opened 2021-07-18
Documenting important FileInfo.Extension behavior directly in <summary>

daiplusplus:
I keep on forgetting if `FileInfo.Extension`'s value contains the leading dot character or not - this isn't mentioned in the `<summary>` documentation so we can't know from just VS's IntelliSense documentation popup without looking up the full documentation (so I end up doing `myFile.Extension.Equals(".foobar") || myFile.Extension.Equals("foobar")` just to be safe, which is silly.

So this is just a tiny documentation change with a real quality-of-life benefit for everyone.

I also documented .NET Core's `FileInfo` behavior w.r.t. common edge-cases of extensions - I'm sure we've all wondered what `.gitignore`'s `Extension` is...

### DIFF
--- a/xml/System.IO/FileSystemInfo.xml
+++ b/xml/System.IO/FileSystemInfo.xml
@@ -669,7 +669,7 @@ The `Extension` property returns the <xref:System.IO.FileSystemInfo> extension, 
 * For a file `c:\.gitignore`, this property returns `".gitignore"`.
 * For a file `c:\noextension`, this property returns an empty string `""`.
 * For a file `c:\a.b.c`, this property returns `".c"`.
-* When the `FileInfo` is constructed from a file path with a trailing dot, that trailing dot is trimmed, and `new FileInfo("foo.bar.").Extension` returns `".bar"`.
+* When the `FileInfo` is constructed from a file path with a trailing dot, then that trailing dot is trimmed, so `new FileInfo("foo.bar.").Extension` returns `".bar"`.
   
  The following table lists examples of other typical or related I/O tasks.  
   


Jehoel:
Okay, I'm stumped as to why the `cref` to `FileSystemInfo.Name` isn't resolving...

xml/System.IO/FileSystemInfo.xml
```diff
@@ -657,13 +657,18 @@ On Unix platforms that do not support creation or birth time, this property retu
...
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The `Extension` property returns the <xref:System.IO.FileSystemInfo> extension, including the period (.). For example, for a file c:\NewFile.txt, this property returns ".txt".  
+ The `Extension` property returns the <xref:System.IO.FileSystemInfo> extension, which starts at, and includes, the last (`'.'`) dot in the file's full-name. For example:
+* For a file `c:\NewFile.txt`, this property returns `".txt"`.
+* For a file `c:\.gitignore`, this property returns `".gitignore"`.
+* For a file `c:\noextension`, this property returns an empty string `""`.
+* For a file `c:\a.b.c`, this property returns `".c"`.
+* Note that when a `FileInfo` is constructed from a file path with a trailing dot then those dots will be trimmed, so `new FileInfo("foo.bar.").Extension` will return `".bar"`.
```
gewarren, line 671:
```suggestion
* When the `FileInfo` is constructed from a file path with a trailing dot, that trailing dot is trimmed, and `new FileInfo("foo.bar.").Extension` returns `".bar"`.
```

xml/System.IO/FileSystemInfo.xml
```diff
@@ -657,13 +657,18 @@ On Unix platforms that do not support creation or birth time, this property retu
...
       <Docs>
-        <summary>Gets the string representing the extension part of the file.</summary>
+        <summary>Gets the extension part of the file name, including the leading dot <c>'.'</c> - or an empty string if no extension is present - or the file's <see cref="FileSystemInfo.Name"/> if dot-prefixed with no other dots present.</summary>
         <value>A string containing the <see cref="T:System.IO.FileSystemInfo" /> extension.</value>
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The `Extension` property returns the <xref:System.IO.FileSystemInfo> extension, including the period (.). For example, for a file c:\NewFile.txt, this property returns ".txt".  
+ The `Extension` property returns the <xref:System.IO.FileSystemInfo> extension, which starts at, and includes, the last (`'.'`) dot in the file's full-name. For example:
```
gewarren, line 666:
```suggestion
The `Extension` property returns the <xref:System.IO.FileSystemInfo> extension, which starts at, and includes, the last dot (`.`) in the file's full name. For example:
 
```

xml/System.IO/FileSystemInfo.xml
```diff
@@ -657,13 +657,18 @@ On Unix platforms that do not support creation or birth time, this property retu
         <ReturnType>System.String</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets the string representing the extension part of the file.</summary>
+        <summary>Gets the extension part of the file name, including the leading dot <c>'.'</c> - or an empty string if no extension is present - or the file's <see cref="FileSystemInfo.Name"/> if dot-prefixed with no other dots present.</summary>
```
gewarren, line 660:
```suggestion
        <summary>Gets the extension part of the file name, including the leading dot <c>.</c>, or an empty string if no extension is present.</summary>
```

Since the dot-prefixed file name is covered in the remarks and it makes the summary perhaps too verbose, I removed it here.

gewarren (commented):
@Jehoel Thanks for adding these improvements! I left some suggestions.

xml/System.IO/FileSystemInfo.xml
Jehoel: How about "Gets the extension part of the file name, including the leading dot <c>.</c> even if it is the entire file name, or an empty string if no extension is present." ?

Jehoel: I've made the summary much shorter but included the bit about matching the whole file-name without being verbose.

gewarren (approved):

PR #6902, opened 2021-07-19
Converts and formatting to be replaced with string interpolation

WhiteBlackGoose:
Continuation of #6900. Replacing `Convert.To` and `string.Format` and Console WriteLine's format with string interpolation in `csharp/VS_Snippets_CLR_System` snippets.

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR_System/system.threading.tasks.taskfactory.startnew/cs/startnew1.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR_System/system.threading.tasks.taskfactory.startnew/cs/startnew1.cs
@@ -33,7 +33,7 @@ public class Example
       // Handle only non-control characters.
       if (! Char.IsControl(value, 0)) {
          foreach (var ch in value)
-            hexString += $"0x{(ushort)ch} ";
+            hexString += $"0x{(ushort)ch:X} ";
       }   
       return hexString.Trim();
    }


samples/snippets/csharp/VS_Snippets_CLR_System/system.threading.tasks.taskfactory.startnew/cs/startnew1.cs
```diff
@@ -33,7 +33,7 @@ private static string ShowHex(string value)
       // Handle only non-control characters.
       if (! Char.IsControl(value, 0)) {
          foreach (var ch in value)
-            hexString += String.Format("0x{0} ", Convert.ToUInt16(ch));
+            hexString += $"0x{(ushort)ch} ";
```
WhiteBlackGoose, line 36: I made equivalent code like everywhere before, but I feel like we should do `:X` for hex format, otherwise it's misleading, since it's outputting decimal format. Or am I wrong?

samples/snippets/csharp/VS_Snippets_CLR_System/system.Console.WriteLine/CS/WriteLine6.cs
```diff
@@ -8,8 +8,8 @@ public static void Main()
       Random rnd = new Random();
       // Generate five random Boolean values.
       for (int ctr = 1; ctr <= 5; ctr++) {
-         Boolean bln = Convert.ToBoolean(rnd.Next(0, 2));
-         Console.WriteLine("True or False: {0}", bln);
+         bool bln = rnd.Next(0, 2) == 1;
```
WhiteBlackGoose, line 11: I could've done it with `is 1`, but I assumed that it would raise additional questions among the newcomers

samples/snippets/csharp/VS_Snippets_CLR_System/system.Linq.Enumerable/CS/enumerable.cs
```diff
@@ -341,7 +341,7 @@ static void AverageEx3()
             // <Snippet16>
             string[] numbers = { "10007", "37", "299846234235" };
 
-            double average = numbers.Average(num => Convert.ToInt64(num));
+            double average = numbers.Average(num => long.Parse(num));
```
WhiteBlackGoose, line 344: I could've made it a method group, but like previous time, this one is probably more understandable for newcomers

samples/snippets/csharp/VS_Snippets_CLR_System/system.globalization.charunicodeinfo.getnumericvalue/cs/getnumericvalue1.cs
```diff
@@ -15,8 +15,7 @@ private static void Overload1()
       int utf32 = 0x10107;       // AEGEAN NUMBER ONE
       string surrogate = Char.ConvertFromUtf32(utf32);
       foreach (var ch in surrogate)
-         Console.WriteLine("U+{0:X4}: {1}    ", Convert.ToUInt16(ch),
-                           System.Globalization.CharUnicodeInfo.GetNumericValue(ch));
+         Console.WriteLine($"U+{(ushort)ch:X4}: {System.Globalization.CharUnicodeInfo.GetNumericValue(ch)}    ");
```
WhiteBlackGoose, line 18: We can do `using` here, but I'm not sure what's preferred for docs

WhiteBlackGoose:
> Also, this is currently about the max size for a PR that's reasonable to review. We're happy to see multiple PRs over time of this size. That's a great way to make progress on these concerns.

Noted, thanks for the info

WhiteBlackGoose:
I think the bot doesn't work correctly. Or it doesn't validate whether the snippets are buildable? Because otherwise, this is my second commit
image

So in the first one this should not compile, since there's no `tr` variable

samples/snippets/csharp/VS_Snippets_CLR_System/system.threading.tasks.taskfactory.startnew/cs/startnew1.cs
BillWagner: I would prefer the `:x` as well. That would be more resilient.

samples/snippets/csharp/VS_Snippets_CLR_System/system.Console.WriteLine/CS/WriteLine6.cs
BillWagner: I'd leave this one as is.

samples/snippets/csharp/VS_Snippets_CLR_System/system.Linq.Enumerable/CS/enumerable.cs
BillWagner: I like the current change.

samples/snippets/csharp/VS_Snippets_CLR_System/system.globalization.charunicodeinfo.getnumericvalue/cs/getnumericvalue1.cs
BillWagner: This one should stay with the fully qualified namespace.

In docs, we generally prefer the `using`, but will use the fully qualified namespace to support runnable snippets in the browser. `System.Globalization` isn't added by default when the snippet is plugged into the (not visible) `Main` method.

BillWagner:
> Commenting so it stays in my review queue.

I reviewed all the changes @WhiteBlackGoose Once we make the one change, this is ready to merge.  Thanks again.

BillWagner (approved):
Thanks again @WhiteBlackGoose 

WhiteBlackGoose:
@BillWagner hey, one question. I *think* I might be able to replace all string formatting with string interpolation (only where appropriate, of course!) and other refactorings, but it's gonna be thousands of LoC changed, and you of course don't want to review it. But even after splitting it into many PRs, it would still be quite a lot to review, right? So, is it needed or I should abandon this idea?

BillWagner:
@WhiteBlackGoose That's why we haven't made as much progress on this. It's one of the areas we want to improve, but the amount of benefit doesn't justify the work when weighed against other priorities. 

That said, an ongoing stream of mid-sized PRs would be very much appreciated. Reviews are generally quick, and the CI build will catch most coding errors in PRs of this type.

Thanks for all your help.

P.S, I'll be on vacation next week, so reviews may be slowed.

gewarren:
> ... and the CI build will catch most coding errors in PRs of this type.

Unfortunately Snippets 5000 isn't enabled in this repo, so build errors won't be caught. I know I've asked @adegeo about this before and I think there's a reason it's not enabled. Andy?

adegeo:
Just to close the loop in this issue thread, we deployed it today, woohoo!

PR #6906, opened 2021-07-20
Name `disposed` parameter for `Dispose` method

xtqqczze:


BillWagner (approved):
Thanks for submitting this fix @xtqqczze 

We appreciate it.

You should see the changes on our live site in the next day or two, on our regular publishing cycle.

Thanks again!

PR #6908, opened 2021-07-20
Clarify cancellation behavior of ChannelReader.ReadAsync

stephentoub:
Clarify the behavior of ChannelReader.ReadAsync in the face of cancellation.
Fixes https://github.com/dotnet/runtime/issues/1710


adamsitnik (approved):
LGTM (it's been a while since I've seen Future Perfect in use)

PR #6917, opened 2021-07-23
Is it a useful change, to replace built-in types with keywords?

WhiteBlackGoose:
I'm not spamming now. I'm trying to learn what's a good change and what is not worth reviewing. I didn't make this change before, so I need to know if I should make it. So far I know what I can do:

1) String formatting -> string interpolation (where appropriate)
2) Convert.To -> casting/parsing etc. (where appropriate)

Please tell me if I should ping someone to review my PRs. Also, if this gets merged silently, would it be fair to consider this as "yes" to my initial question?

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR_System/system.activator.createinstance/cs/CreateInstance5.cs
+++ b/samples/snippets/csharp/VS_Snippets_CLR_System/system.activator.createinstance/cs/CreateInstance5.cs
@@ -5,7 +5,7 @@ public class Example
 {
    public static void Main()
    {
-        // Initialize array of characters from a to z.
+      // Initialize array of characters from a to z.
       char[] chars = new char[26];
       for (int ctr = 0; ctr < 26; ctr++)
          chars[ctr] = (char) (ctr + 0x0061);


MSDN-WhiteKnight:
It's best to open issue first if you want to submit a set of changes over multiple files or when you are not sure that changes will be accepted. See contributing guide: https://docs.microsoft.com/en-us/contribute/dotnet/dotnet-contribute

>DON'T surprise us with large pull requests. Instead, file an issue and start a discussion so we can agree on a direction before you invest a large amount of time.

As for these changes, replacing type identifiers with language keywords is good, it was discussed recently here: https://github.com/dotnet/docs/pull/24985 The system automatically pings the docs team, so your PR will be reviewed soon.

WhiteBlackGoose:
> DON'T surprise us with large pull requests. Instead, file an issue and start a discussion so we can agree on a direction before you invest a large amount of time.

That's exactly why I did NOT open a big PR. So this PR is more of a question, but phrased as a PR :). The problem is that this change *alone* might take *tens or hundreds* of PRs, so I'm a bit worried that you don't want it.

> It's best to open issue first if you want to submit a set of changes over multiple files or when you are not sure that changes will be accepted.

Right, but opening a PR immediately shows what I'm up to. I'm still a bit unsure whether it's good idea to fix so many things

BillWagner (approved):
Thanks for doing this @WhiteBlackGoose 

I've reviewed these changes, and I'll :shipit: now.

BillWagner:
Thanks for the conversation @WhiteBlackGoose and @MSDN-WhiteKnight 

This is a good way to start, and over time, improving our samples is a goal.

We'll continue to watch for and merge PRs that are roughly this size for these updates that bring our samples in line with our current stated coding standard.

That said, reviews and merging may occasionally be delayed depending on release cycles (including preview releases). Thanks for understanding.

Issue #6920, opened 2021-07-24
Replacing built-in types with keywords in one big PR


WhiteBlackGoose:
As answered [here](https://github.com/dotnet/dotnet-api-docs/pull/6917#issuecomment-885807494), it is a good idea to replace built-in types with keywords, when those exist (e. g. `System.String` -> `string`). It is a very pritimitive change, so here's my question: can I do it in one PR?

WhiteBlackGoose:
Alternative: I can create a number (20-200) PRs of moderate size to perform these changes. Another alternative: not to do it at all (but it should be done eventually!)

BillWagner:
@WhiteBlackGoose 

> Alternative: I can create a number (20-200) PRs of moderate size to perform these changes.

If you're interested in doing that, we'll watch for them and merge them. Thank you!

WhiteBlackGoose:
Oops XD

WhiteBlackGoose:
Alright, I started opening PRs.
Current ones:
https://github.com/dotnet/dotnet-api-docs/pull/6925
https://github.com/dotnet/dotnet-api-docs/pull/6926
https://github.com/dotnet/dotnet-api-docs/pull/6927
https://github.com/dotnet/dotnet-api-docs/pull/6928
https://github.com/dotnet/dotnet-api-docs/pull/6929

WhiteBlackGoose:
I want to triple check. Like I said, it's gonna be a number of PRs. Is it still fine? At the moment it will be at least 50+ PRs. Tell me if I should actually stop

WhiteBlackGoose:
I opened 9 PRs so far. Please tell me, if I should keep going or stop and open a big PR making all these changes at once

WhiteBlackGoose:
Alright, the program's finished. It will be 170 PRs in total (like I said). Is it still better to keep a lot of PRs instead of one big?

safern:
FYI: @gewarren

PR #6925, opened 2021-07-26
Names -> Keywords batch 1

WhiteBlackGoose:

Files affected: service.cs, service.cs, program.cs, order.cs, source.cs, services.cs, services.cs, snippets.cs.

See #6920


BillWagner (approved):
Thanks @WhiteBlackGoose 

PR #6928, opened 2021-07-26
Names -> Keywords batch 4

WhiteBlackGoose:

Files affected: source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs.

See #6920


BillWagner (approved):
LGTM.

PR #6930, opened 2021-07-26
Names -> Keywords batch 6

WhiteBlackGoose:

Files affected: source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, source.cs, action.cs, Action3.cs, Anon.cs, Delegate.cs, Lambda.cs, Action4.cs.

See #6920


BillWagner (approved):
LGTM

PR #6952, opened 2021-07-30
Fix escaping

### LINKED ISSUE #6941: Wrong output example for escaped XML with XmlTextWriter.WriteString
[This page](
https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmltextwriter.writestring?view=net-5.0#remarks) lists an example how the text `test<item>test` would be encoded to XML. The remarks section indicates that some special characters are like `<` and `>` are escaped. But then the docs page notes that the text is written unescaped. I would expect that the page reads: 

> For example, this input string `test<item>test` is written as
> `test&lt;item&gt;test`

gewarren:
Fixes #6941

### DIFF
--- a/xml/System.Xml/XmlTextWriter.xml
+++ b/xml/System.Xml/XmlTextWriter.xml
@@ -2460,11 +2460,7 @@ writer.WriteStartElement("root");
   
 -   If `WriteString` is called in the context of an attribute value, double and single quotes are replaced with `&quot;` and `&apos;` respectively.  
   
- For example, this input string `test<item>test` is written as  
-  
-```
-test\&lt;item\&gt;test
-```
+ For example, this input string `test<item>test` is written as `test&lt;item&gt;test`.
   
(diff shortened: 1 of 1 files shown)


xml/System.Xml/XmlTextWriter.xml
```diff
@@ -2462,13 +2462,11 @@ writer.WriteStartElement("root");
   
  For example, this input string `test<item>test` is written as  
   
-```  
-test&lt;item&gt;test  
-```  
+```
+test\&lt;item\&gt;test
```
buyaa-n, line 2466: i think the example code should be escaped not this one

buyaa-n: Ah it was the example, i thought examples are in a different file, never mind

buyaa-n (approved):

gewarren:
Thank you Buyaa!

PR #6962, opened 2021-08-03
Doc updates for StringInfo

danmoseley:
Used doc porter tool.

Fixes https://github.com/dotnet/runtime/issues/54876

### DIFF
--- a/xml/System.Globalization/StringInfo.xml
+++ b/xml/System.Globalization/StringInfo.xml
@@ -494,7 +494,15 @@ This example uses the <xref:System.Globalization.StringInfo.GetTextElementEnumer
         <param name="str">The input span to analyze.</param>
         <summary>Returns the length of the first text element (extended grapheme cluster) that occurs in the input span.</summary>
         <returns>The length (in chars) of the substring corresponding to the first text element within <paramref name="str" />, or 0 if <paramref name="str" /> is empty.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+A grapheme cluster is a sequence of one or more Unicode code points that should be treated as a single unit.
+
+          ]]></format>
+        </remarks>
       </Docs>
     </Member>
     <Member MemberName="GetNextTextElementLength">
@@ -528,7 +536,15 @@ This example uses the <xref:System.Globalization.StringInfo.GetTextElementEnumer
         <param name="str">The input string to analyze.</param>
         <summary>Returns the length of the first text element (extended grapheme cluster) that occurs in the input string.</summary>
         <returns>The length (in chars) of the substring corresponding to the first text element within <paramref name="str" />, or 0 if <paramref name="str" /> is empty.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+A grapheme cluster is a sequence of one or more Unicode code points that should be treated as a single unit.
+
+          ]]></format>
+        </remarks>
         <exception cref="T:System.ArgumentNullException">
           <paramref name="str" /> is <see langword="null" />.</exception>
       </Docs>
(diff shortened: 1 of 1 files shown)


xml/System.Globalization/StringInfo.xml
```diff
@@ -491,9 +491,9 @@ This example uses the <xref:System.Globalization.StringInfo.GetTextElementEnumer
         <Parameter Name="str" Type="System.ReadOnlySpan&lt;System.Char&gt;" Index="0" FrameworkAlternate="net-6.0" />
       </Parameters>
       <Docs>
-        <param name="str">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="str">The input span to analyze.</param>
```
tarekgh, line 494: >The input span to analyze. [](http://example.com/codeflow?start=26&length=26)

The input char span from which to get the first text element length.

xml/System.Globalization/StringInfo.xml
```diff
@@ -491,9 +491,9 @@ This example uses the <xref:System.Globalization.StringInfo.GetTextElementEnumer
...
+        <summary>Returns the length of the first text element (extended grapheme cluster) that occurs in the input span.</summary>
```
tarekgh, line 495: >(extended grapheme cluster) [](http://example.com/codeflow?start=62&length=27)

this part can be confusing as the first element in the text could be not a grapheme cluster and can bee a normal single character.

Maybe remove this part and add extra sentence telling  the first element could be grapheme cluster or surrogate pair which can have length greater than one character.

tarekgh:
CC @GrabYourPitchforks to validate the comments I added to this review.

tarekgh (approved):
add a couple of comments. LGTM otherwise.

GrabYourPitchforks:
> this part can be confusing as the first element in the text could be not a grapheme cluster and can be a normal single character.

A grapheme cluster might consist of just a single char, just like how a string or a `ROS<char>` can have a 1-char length if you really want. I wonder if we should reword the grapheme cluster docs to clarify this?

GrabYourPitchforks (approved):

tarekgh:
> A grapheme cluster might consist of just a single char, just like how a string or a ROS<char> can have a 1-char length if you really want. I wonder if we should reword the grapheme cluster docs to clarify this?

Anyone reading `grapheme cluster` will not understand that :-) few people familiar with that. Clarification would be super helpful. at least we add something in the remarks explaining what `grapheme cluster` means.

GrabYourPitchforks:
@tarekgh The _Remarks_ section at https://docs.microsoft.com/dotnet/api/system.globalization.stringinfo starts off with this:

> .NET defines a text element as a unit of text that is displayed as a single character, that is, a grapheme. A text element can be a base character, a surrogate pair, or a combining character sequence.

You thinking that we should perhaps pull this same text in to these docs?

tarekgh:
> You thinking that we should perhaps pull this same text in to these docs?

Yes, that will be helpful. I would prefer the definition `A grapheme cluster is a sequence of one or more Unicode code points that should be treated as a single unit` to avoid tying it to display behaviors only. But I am fine with either definition.

danmoseley:
Thanks both. Note that this is just mechanically produced from the sources. I will turn around and make a PR in dotnet/runtime to update this text then regenerate this PR.

In general, it's surprising how "ready" the XML doc text in the sources is. It's certainly valuable to get it "right" there first time, in the PR process, so that ideally these PR's need no comment.

xml/System.Globalization/StringInfo.xml
danmoseley: I ended up leaving this, because the summary explains what it does. If I repeat the summary on this parameter you just get to read both. Note there are many of these here eg `The input string to analyze`

tarekgh (approved):
Thanks @danmoseley. LGTM.

danmoseley:
Updated. I added the comment to all 5 methods in the code, but only 3 ported because the GetNextTextElement overloads already have remarks. Those remarks are even better, so maybe should port back at some point.

I wonder if there's a flag on the doc porter tool to flag when it bails on porting remarks because they're present and different.

PR #6972, opened 2021-08-04
Update GC.xml

mangod9:
## Summary

Add clarification for GetTotalAllocatedBytes that The returned value also does not include any native allocations.

Fixes https://github.com/dotnet/runtime/issues/49174


gewarren (approved):
LG with one minor suggestion. Thanks for adding this!

PR #6973, opened 2021-08-04
Names -> Keywords batch 10

WhiteBlackGoose:

Files affected: operations1.cs, parse2.cs, parse3.cs, tryparseex.cs, bcopy.cs, getbyte.cs, setbyte.cs, systembyte.cs, tostring.cs, charstructure.cs.

See #6920

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR_System/system.Buffer.BlockCopy/CS/QuackTest.csproj
+++ b/samples/snippets/csharp/VS_Snippets_CLR_System/system.Buffer.BlockCopy/CS/QuackTest.csproj
@@ -1,0 +1,41 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5</TargetFramework>
+    <LangVersion>9.0</LangVersion>
+    <StartupObject>buffer</StartupObject>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Platforms>AnyCPU;x86</Platforms>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <DocumentationFile></DocumentationFile>
+  </PropertyGroup>
+
+</Project>
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+


BillWagner:
closing and reopening to trigger the samples CI build

WhiteBlackGoose:
@BillWagner I just want to 100% verify. Is it really the way you want me to do this? Is it convenient for you or other reviewers? Because there are 150 more branches to merge, those were only 20 or about that.

BillWagner:
Tagging @adegeo

The snippets 5000 CI check failed because the nupkg isn't installed in the dependencies folder. Can you open a new PR since that was missed in #6982?

BillWagner:
@WhiteBlackGoose Thanks for reaching out.  Here's a detailed update:

Yes, we do like these changes. Making them is also driving us to add a check that we've had in the docs repo for some time now: a CI build on all sample code.  Briefly, the *snippets 5000* checks any PR that changes source code to ensure the code changed is included in a project, and if so, it builds that project.

For historical reasons, the sample code in this repo typically didn't have a project file. It was migrated from an internal system that did verification builds. Our expectation is that these changes will fail in the *snippets 5000* build because most of these folders won't have a project file. 

The fix is reasonably simple: add a project file for a console app. Right now, the project file can use .NET Core 3.1 (latest LTS), .NET 5 (latest release) or .NET 6 (preview).  Our version checker will flag any version that's out of support on a monthly basis.

One challenge we've had in some folders is where there are multiple `Main` methods in different files. These are quite common, because we have many examples that are full programs. Those are used to support C# interactive, where we run many of the snippets in the browser on docs. To make a project file that passes the CI build, specify a startup object, and add a new `Main` method that calls the existing `Main` method(s) in turn.  You can see an example here: https://github.com/dotnet/docs/tree/main/docs/csharp/language-reference/builtin-types/snippets/shared

Let us know how that works, if you are are comfortable doing these extra steps on these PRs. It will really help us validate the quality of these samples.

WhiteBlackGoose:
@BillWagner so, I have to add a project file to each folder, right? Then here's a question. Why not to add it to *each file*, instead of each folder? Iirc all snippets are runnable independently, hence we can do it.

adegeo:
> so, I have to add a project file to each folder, right? Then here's a question. Why not to add it to _each file_, instead of each folder? Iirc all snippets are runnable independently, hence we can do it.

Currently the build system wants a single project per folder. To do multiple projects you would have to move the files to sub folders and add projects to each folder. Then you would have to update the articles that reference the sample, it's a bit more work.

Adding the project file is generally simple, as Bill pointed out. However, with multiple main objects, you may actually need to specify the `<StartupObject>class method</StartupObject>` to indicate which main is the main object. The snippet system doesn't actually RUN the code, it just checks for compile. So (@BillWagner) you don't need to make a Main that calls the other Mains, it will be pointless. You just need the compiler to pass.

BillWagner:
> you don't need to make a Main that calls the other Mains, it will be pointless. You just need the compiler to **pass.**

That's true. I do find that I like being able to run the samples locally. That's why I've added that.

BillWagner:
triggering a fresh build.

BillWagner:
@WhiteBlackGoose The last *snippets 5000* build shows the errors we'd expect from missing project files.

tannergooding (approved):

pgovind (approved):
Code changes look fine. Will leave the CI/build issues to the docs folks

WhiteBlackGoose:
@pgovind @adamsitnik  I just didn't have time yet to add the necessary files, as per Bill Wagner's comment. Sorry, should've converted the PR to draft

BillWagner:

gewarren:
I'm going to close this PR since the snippets have all moved to new folders. Feel free to continue this work in the new snippet folder structure. Also, most snippets should have associated project files now, so you won't need to add too many of those.

PR #6976, opened 2021-08-04
Names -> Keywords batch 13

WhiteBlackGoose:

Files affected: source2.cs, program.cs, program.cs, program.cs.

See #6920


WhiteBlackGoose:

PR #6979, opened 2021-08-04
Names -> Keywords batch 16

WhiteBlackGoose:

Files affected: remarks.cs, nvc.cs, comparisont1.cs, backgroundcolor1.cs, example3.cs, normalize1.cs, unicode1.cs, clear1.cs, error1.cs.

See #6920


WhiteBlackGoose:

PR #6981, opened 2021-08-04
Names -> Keywords batch 18

WhiteBlackGoose:

Files affected: changetype_nullable.cs, changetype00.cs, changetype01.cs, changetype03.cs, ToBase64String.cs.

See #6920


WhiteBlackGoose:

Issue #6986, opened 2021-08-05
Address HTML validation


gewarren:
Recently, HTML validation was added to the build reports. It's flagged a number of problems in this repo, for example, `<related>` tags nested in a `<summary>` tag. This issue tracks the work to fix these problems.

PR #6992, opened 2021-08-06
HTML tag cleanup

gewarren:
Contributes to #6986


BillWagner (approved):
These all LGTM @gewarren 

You can :shipit: when ready.

PR #7003, opened 2021-08-07
Update CompilerServices docs

danmoseley:
Used porting tool.

Contributes to https://github.com/dotnet/runtime/issues/54879. There is still an issue with some members listed there that come from an ilproj. But otherwise it fixes that as well as adds more goodness elsewhere.

### DIFF
--- a/xml/System.Runtime.CompilerServices/DefaultInterpolatedStringHandler.xml
+++ b/xml/System.Runtime.CompilerServices/DefaultInterpolatedStringHandler.xml
@@ -278,7 +278,7 @@ This is intended to be called only by compiler-generated code. Arguments are not
         <Parameter Name="value" Type="T" />
       </Parameters>
       <Docs>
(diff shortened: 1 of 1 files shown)


xml/System.Runtime.CompilerServices/DefaultInterpolatedStringHandler.xml
```diff
@@ -255,8 +279,8 @@
       </Parameters>
       <Docs>
         <typeparam name="T">To be added.</typeparam>
-        <param name="value">To be added.</param>
-        <summary>To be added.</summary>
+        <param name="value">The value to write.</param>
```
krwq, line 282: should we also write description for typeparam above?

krwq: same comment for other places like this

krwq (approved):
LGTM, thank you - only comment about typeparam

PR #7019, opened 2021-08-09
Add explanation for serializing a DataMember with EmitDefaultValue as false.

LI-Yanting:
## Summary

Explain explicitly that a DataMember with a non default value will still be serialized despite its EmitDefaultValue = false.


adamsitnik:
cc @StephenMolloy @HongGit

BillWagner:
ping @carlossanlop to route for review.

Issue #6957, opened 2021-08-02
Remove all Quic docs


danmoseley:
System.Net.Quic will not ship as public API in .NET 6.0. We need to remove all the docs for it: https://docs.microsoft.com/en-us/dotnet/api/system.net.quic?view=net-6.0 and below.

cc @ManickaP @karelz

danmoseley:
@gewarren  these API no longer appear in the reference assembly but are still public in the code (for ASP.NET). Given that, if we simply delete the topics here will they reappear later?

gewarren:
If they've been removed from the reference assembly then I *think* they will be deleted the next time we update the .NET 6 APIs (since each preview version overwrites the last published version instead of creating a new moniker). But if you don't want to wait for that, you could probably just delete them. They'll also need to be removed from xml/FrameworksIndex/net-6.0.xml. Is https://docs.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.quicimplementationprovider?view=net-6.0#System_Net_Http_SocketsHttpHandler_QuicImplementationProvider removed too?

ManickaP:
`QuicImplementationProvider` is removed as well. Should I rather do the removal manually now or are we ok with waiting and letting automation handle it?

carlossanlop:
@ManickaP xml files should not be deleted manually from dotnet-api-docs.

The referenced Prev7 PR should automatically address this issue.

PR #7032, opened 2021-08-13
CI Update - .NET 6 Preview 7

gewarren:
Fixes https://github.com/dotnet/dotnet-api-docs/issues/6957


xml/System.Runtime.InteropServices/NativeMemory.xml
```diff
@@ -0,0 +1,302 @@
...
+          <AttributeName Language="C#">[System.CLSCompliant(false)]</AttributeName>
+          <AttributeName Language="F#">[&lt;System.CLSCompliant(false)&gt;]</AttributeName>
+        </Attribute>
+      </Attributes>
+      <ReturnValue>
+        <ReturnType>System.Void*</ReturnType>
+      </ReturnValue>
+      <Parameters>
+        <Parameter Name="ptr" Type="System.Void*" />
+        <Parameter Name="byteCount" Type="System.UIntPtr" />
```
AaronRobinsonMSFT, line 104: I think these should be `nuint`. Technically this is the same thing but from an API the `nuint` is the definition. @tannergooding

tannergooding: This is likely a place where the docs tooling doesn't understand certain newer language features. Off the top of my head:
* nint/nuint types
* in parameters
* readonly members
* function pointers
* ref and ref readonly returns
* records
* tuple names
* init only setters

There are possibly more as well and its possible that a few of these have already been fixed.

xml/FrameworksIndex/net-6.0.xml
```diff
@@ -85,7 +86,6 @@
     <Assembly Name="System.Net.NetworkInformation" Version="6.0.0.0" />
     <Assembly Name="System.Net.Ping" Version="6.0.0.0" />
     <Assembly Name="System.Net.Primitives" Version="6.0.0.0" />
-    <Assembly Name="System.Net.Quic" Version="6.0.0.0" />
```
gewarren, line 88: @danmoseley Looks like this Preview 7 update did what we wanted.

xml/System.Runtime.InteropServices/NativeMemory.xml
gewarren: @TianqiZhang Can you comment on this?

gewarren:
LGTM. I don't see any new /// comments being brought in, so I guess those haven't changed for the assemblies where they were provided.

image

xml/FrameworksIndex/net-6.0.xml
danmoseley: will this magically fix https://github.com/dotnet/dotnet-api-docs/issues/6957 ? @ManickaP

carlossanlop: I think it will, @danmoseley. If you go to the Files tab and scroll, you'll see the Quic xml files are being deleted.

BTW, these deleted files are causing a conflict with the main branch @gewarren .

carlossanlop:
> LGTM. I don't see any new /// comments being brought in, so I guess those haven't changed for the assemblies where they were provided.

@gewarren is that screenshot complete? There should be 4 *.xml files:

- System.IO.Compression
- System.IO.Compression.ZipFile
- System.IO.Compression.Brotli
- System.Numerics.Vectors

Those are the 4 assemblies that use triple slash as source of truth, currently.

carlossanlop (approved):
LGTM. The questions can be addressed separately.

PR #7005, opened 2021-08-07
Update IO docs

danmoseley:
Using porting tool.

Fixes https://github.com/dotnet/runtime/issues/54860 (System.IO.BinaryReader and System.IO.BinaryWriter members are missing code comments, added directly here instead)

PR #7035, opened 2021-08-16
Fix FileStream xrefs

gewarren:
Follow up to #7005


adamsitnik (approved):
LGTM, thank you @gewarren 

cc @carlossanlop is this something that we should apply as well in the docs tool?

PR #7039, opened 2021-08-17
Fix unescaped backslash

GitHubPang:
Backslash wasn't escaped, so it wasn't rendered in
https://docs.microsoft.com/en-us/dotnet/api/system.io.enumeration.filesystemname.matcheswin32expression

See also #4136


MihaZupan (approved):
Thanks!

PR #7052, opened 2021-08-18
Port System.Net triple slash to docs

carlossanlop:
Ported using DocsPortingTool.

Found a small bunch of descriptions that could be ported automatically.


ManickaP (approved):
LGTM, thank you!

PR #7062, opened 2021-08-20
Add remaining missing docs to System.Runtime.CompilerServices

krwq:
Fixes: https://github.com/dotnet/runtime/issues/54879

Adding remaining missing docs for System.Runtime.CompilerServices.Unsafe from manual xml file (for ilproj) from dotnet/runtime.

### DIFF
--- a/xml/System.Runtime.CompilerServices/Unsafe.xml
+++ b/xml/System.Runtime.CompilerServices/Unsafe.xml
@@ -922,9 +922,9 @@
       <Docs>
         <typeparam name="T">The type of the reference.</typeparam>
         <param name="source">The reference to check.</param>
-        <summary>Returns if a given reference to a value of type <typeparamref name="T"/> is a null reference.</summary>
+        <summary>Determines if a given reference to a value of type <typeparamref name="T"/> is a null reference.</summary>
         <returns><see langword="true" /> if <paramref name="source" /> is a null reference; otherwise, <see langword="false" />.</returns>
-        <remarks>This check is conceptually similar to "(void*)(&amp;source) == nullptr".</remarks>
+        <remarks>This check is conceptually similar to <code>(void*)(&amp;source) == nullptr</code>.</remarks>
       </Docs>
     </Member>
     <Member MemberName="NullRef&lt;T&gt;">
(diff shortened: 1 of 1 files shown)


xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -920,11 +920,11 @@
...
       </Parameters>
       <Docs>
-        <typeparam name="T">To be added.</typeparam>
-        <param name="source">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <typeparam name="T">The type of the reference.</typeparam>
+        <param name="source">The reference to check.</param>
+        <summary>Returns if a given reference to a value of type <typeparamref name="T"/> is a null reference.</summary>
```
gewarren, line 925:
```suggestion
        <summary>Determines if a given reference to a value of type <typeparamref name="T"/> is a null reference.</summary>
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -920,11 +920,11 @@
...
+        <returns><see langword="true" /> if <paramref name="source" /> is a null reference; otherwise, <see langword="false" />.</returns>
+        <remarks>This check is conceptually similar to "(void*)(&amp;source) == nullptr".</remarks>
```
gewarren, line 927:
```suggestion
        <remarks>This check is conceptually similar to <code>(void*)(&amp;source) == nullptr</code>.</remarks>
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -948,9 +948,9 @@
       </TypeParameters>
       <Parameters />
       <Docs>
-        <typeparam name="T">To be added.</typeparam>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <typeparam name="T">The type of the reference.</typeparam>
+        <summary>Returns a reference to a value of type <typeparamref name="T"/> that is a null reference.</summary>
+        <returns>Reference to a value of type <typeparamref name="T"/> that is a null reference.</returns>
```
gewarren, line 953:
```suggestion
        <returns>A reference to a value of type <typeparamref name="T"/> that is a null reference.</returns>
```

gewarren (approved):
Not sure if `&amp;` should just be `&` if it's code-fenced.

krwq:
`&amp;` must work because it's an XML syntax. `&` might not work depending on the tool and parser setting. I do not believe `<code>` will make any difference. I think it's safer to stick with `&amp;`

PR #7064, opened 2021-08-20
Fix bad character

gewarren:
Introduced by #6998


tdykstra (approved):
The three dots in one character are hard to make out - you could change to three dot characters.

gewarren:
Thanks Tom. In the interest of time, because I need to kick off a CI build once I merge this, I'm going to leave as-is for now.

PR #7068, opened 2021-08-23
Update String.xml docs to reflect usage of `CultureInfo` arguments

### LINKED ISSUE #7059: `string.ToLower(CultureInfo? culture)` does not throw `ArgumentNullException` for a null culture
When a null culture is passed into `string.ToLower(CultureInfo? culture)`, no `ArgumentNullException` is thrown, despite what the docs say: `ToLower()` documentation: https://docs.microsoft.com/en-us/dotnet/api/system.string.tolower?view=net-5.0

This is due to a null check that assigns the null culture to `CultureInfo.CurrentCulture`: https://github.com/dotnet/runtime/blob/37d6d80d40b8fafd876d95d5eb07f676ac7b380c/src/libraries/System.Private.CoreLib/src/System/String.Manipulation.cs#L1905-L1909

steveberdy:
## Summary

The documentation for `string` methods stated that there were instances in which an `System.ArgumentNullException` can be thrown in the case of a `null` `CultureInfo` being passed as an argument. This is not the case, as with every `string` method that takes in a `CultureInfo? culture`, that is checked and replaced with `CultureInfo.CurrentCulture` if it is null.

I have not checked if the documentation is incorrect at other locations for `CultureInfo` arguments, but this is definitely true for the `string` documentation. 

Fixes #7059

### DIFF
--- a/xml/System/String.xml
+++ b/xml/System/String.xml
@@ -1946,7 +1946,7 @@ The following example demonstrates how you can create an instance of the <xref:S
         <param name="strB">The second string to compare.</param>
         <param name="ignoreCase">
           <see langword="true" /> to ignore case during the comparison; otherwise, <see langword="false" />.</param>
-        <param name="culture">An object that supplies culture-specific comparison information.</param>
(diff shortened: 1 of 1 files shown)


tarekgh (approved):
LGTH, thanks @steveberdy

Issue #7060, opened 2021-08-20
string.StartsWith(char) and string.EndsWith(char) remarks are incorrect


slozier:
Both of these methods have the following remark:
>This method performs an ordinal (case-sensitive and culture-insensitive) comparison.

but I believe the comparison is actually ordinal?

MihaZupan:
You are correct, these are all doing ordinal comparisons.

`Contains(char)` has the correct remark `This method performs an ordinal (case-sensitive and culture-insensitive) comparison.`, but [`StartsWith(char)`](https://docs.microsoft.com/en-us/dotnet/api/system.string.startswith?view=net-5.0#System_String_StartsWith_System_Char_) and [`EndsWith(char)`](https://docs.microsoft.com/en-us/dotnet/api/system.string.endswith?view=net-5.0#System_String_EndsWith_System_Char_) do not.

danmoseley:
Do you want to offer a quick PR @slozier ?

slozier:
Sure, I can do that.

PR #7073, opened 2021-08-24
Fix string.StartsWith(char) and string.EndsWith(char) remarks

slozier:
## Summary

Corrects the remarks to state that string.StartsWith(char) and string.EndsWith(char) are ordinal comparisons.

Fixes https://github.com/dotnet/dotnet-api-docs/issues/7060


MihaZupan (approved):
Thanks!

PR #7078, opened 2021-08-28
Update examples to c:\WINDOWS rather than obsolete c:\WINNT

danmoseley:
It's not WINNT any more. For some reason the underlying Windows API returns the path with capital WINDOWS even though it's likely c:\Windows on disk.


BillWagner (approved):
This LGTM @danmoseley 

I'll :shipit: now.

PR #7132, opened 2021-09-09
Generate XML files from most recent RC1 WPF source and update docs with DocsPortingTool. 

ryalanms:
Generate XML files from most recent WPF source and update documentation from XML using the DocsPortingTool.  The remaining WPF .NET 6.0 APIs will need to be updated by manually.  

/cc @dotnet/wpf-developers 

Edit:  The System.Windows.Forms files will be removed in the next iteration.


carlossanlop:
I asked @ryalanms to re-run the tool with the latest bits, to correct the bug causing the encoding to change. Also, we are going to see if more comments are caught by the tool, because we only see remarks and one exception, which is unexpected.

ryalanms:
With @carlossanlop's guidance, the tool updated the documentation for 5665 WPF APIs.  

This will need to be split into multiple PRs per assembly.

carlossanlop:
Thanks for opening the rest of the PRs. I'm closing this one, @ryalanms .

PR #7134, opened 2021-09-10
Filter System.Text.Json.SourceGeneration namespace

gewarren:
Per request from @eerhardt.

### DIFF
--- a/xml/_filter.xml
+++ b/xml/_filter.xml
@@ -149,7 +149,7 @@
       <typeFitler name="*" expose="false" />
     </namespaceFilter>
     <namespaceFilter name="System.Text.Json.SourceGeneration">
-      <typeFitler name="*" expose="false" />
+      <typeFitler name="JsonSourceGenerator" expose="false" />
     </namespaceFilter>
     <namespaceFilter name="Polly">
       <typeFilter name="*" expose="false" />


xml/_filter.xml
```diff
@@ -148,6 +148,9 @@
     <namespaceFilter name="StackExchange.Redis">
       <typeFitler name="*" expose="false" />
     </namespaceFilter>
+    <namespaceFilter name="System.Text.Json.SourceGeneration">
```
eerhardt, line 151: Is it possible to filter on assembly name?

gewarren: Not that I know. @TianqiZhang can you confirm? Eric, what is the concern?

gewarren: Well we can just remove the assembly from the reference assemblies that are imported.

eerhardt: > Eric, what is the concern?

The concern is someone puts a public API in this namespace in a different assembly. Filtering by namespace would exclude it.

> we can just remove the assembly from the reference assemblies that are imported.

That would be great!

eerhardt: There is another assembly that is in the same situation as this one: `Microsoft.Extensions.Logging.Generators`. If it is also imported, could you exclude that assembly as well?

gewarren: > There is another assembly that is in the same situation as this one: `Microsoft.Extensions.Logging.Generators`. If it is also imported, could you exclude that assembly as well?

I'm not seeing Microsoft.Extensions.Logging.Generators.dll in the assembly drop.

xml/_filter.xml
```diff
@@ -148,6 +148,9 @@
...
+      <typeFitler name="JsonSourceGenerator" expose="false" />
```
gewarren, line 152: I changed this PR to specifically filter the `JsonSourceGenerator` class instead of the whole namespace. So we could either keep this PR, or @safern could exclude the System.Text.Json.SourceGeneration.dll assembly from the drop. Which do you think is better Eric and Santi?

carlossanlop: @gewarren I'll take care of the next DLL drop. The next time I send you one, I can manually remove the System.Text.Json.SourceGeneration.dll .

xml/_filter.xml
eerhardt: ok, that's good to know.

xml/_filter.xml
eerhardt: My preference would be to exclude the System.Text.Json.SourceGeneration.dll assembly.

gewarren: @carlossanlop Is there a clean way of removing this assembly/namespace/type from the net6-rc1 branch? Or should we regenerate without that assembly?

carlossanlop: It may also be possible to automatically exclude it, but I don't know enough about how MSBuild works (Santi is the expert). We use this repo to collect the ref assemblies we send you:

https://github.com/dotnet/api-catalog-infra

There's a file that explicitly mentions the includes for the droplayout, but I don't see any excludes, and I also don't see any `System.Text.Json` there:
https://github.com/dotnet/api-catalog-infra/blob/release/6.0-rc1/eng/droplayout.props

@eerhardt maybe this file has the answer for which we are seeing `Microsoft.Extensions.Logging.Generators` and `System.Text.Json.SourceGenerators` unexpectedly?

carlossanlop: > Is there a clean way of removing this assembly/namespace/type from the net6-rc1 branch? Or should we regenerate without that assembly?

@gewarren I can generate the ref assemblies again, no problem. I can try to get these Source Generation assemblies explicitly excluded, or I can manually remove the DLLs. The only question I have is: if we are going to overwrite RC1 API layout, will we lose all the documentations commits we've done?

gewarren: It shouldn't overwrite if we uncheck the option to import XML doc comments.

gewarren: You don't need to generate the drop again. I'll just remove that assembly from what I have already.

carlossanlop: two assemblies*

gewarren: I didn't see the other assembly in the drop. But I'll look again.

gewarren:
We're going to remove the assembly from the drop instead of filtering it.

xml/_filter.xml
carlossanlop: Ah apologies. I re-read @eerhardt 's other comment, and he simply mentioned that assembly was also autogenerated, but didn't say it was being included as well. It's not being included in dotnet-api-docs, I'm not seeing any files with that assembly in the net6-rc1 branch.

Issue #2518, opened 2019-05-29
Thread.VolatileRead/Write remarks and volatile keyword


MSDN-WhiteKnight:
In [Thread.VolatileRead](https://docs.microsoft.com/en-us/dotnet/api/system.threading.thread.volatileread?view=netframework-4.7.2#System_Threading_Thread_VolatileRead_System_SByte__) remarks:

>In C#, using the volatile modifier on a field guarantees that all access to that field uses VolatileRead or VolatileWrite

This is incorrect, because `volatile` keyword does not actually use these methods, but only causes all field access instructions to be volatile operations. I was going to submit a pull request similar to #2503 , but then i realized it won't really help, because this article is still missing the real use cases for Thread.VolatileRead/Write and how are these methods related to similar methods of [Volatile](https://docs.microsoft.com/en-us/dotnet/api/system.threading.volatile?view=netframework-4.8) class. Given that this statement is duplicated many times, i decided to open issue first to disscuss how it should be changed. 

As I understood, In .NET Core Thread.VolatileRead/Write are just shortcuts to Volatile.Read/Write, and in .NET Framework they are older inefficient variant of them (use full memory barrier instead of half-barrier). Should documentation mark these methods as not recommended for use and point to Volatile class? 

Related: https://github.com/dotnet/dotnet-api-docs.ru-ru/issues/22

rpetrusha:
@kouvel, can you comment on this issue and how the documentation should be changed?

PR #7137, opened 2021-09-10
Clarify/fix some threading API docs

kouvel:
- Clarify terms used in `LazyThreadSafetyMode` field names. Relevant to https://github.com/dotnet/docs/issues/12214.
- Clarify documentation of the `Volatile` class. Relevant to https://github.com/dotnet/docs/issues/24318.
- Update docs for `Thread.VolatileRead` and `Thread.VolatileWrite`. Fixes https://github.com/dotnet/dotnet-api-docs/issues/2518.
- Clarify something about `ManualResetEventSlim`'s perf benefit over `ManualResetEvent`. Fixes https://github.com/dotnet/dotnet-api-docs/issues/3323.


xml/System.Threading/Volatile.xml
```diff
@@ -982,7 +985,7 @@
       </Parameters>
       <Docs>
         <param name="location">The field where the value is written.</param>
-        <param name="value">The value to write. The value is written immediately so that it is visible to all processors in the computer.</param>
```
mangod9, line 985: is the extra clarification not required, or has the behavior now changed?

kouvel: The behavior hasn't changed, just removing the statement since it's incorrect

xml/System.Threading/LazyThreadSafetyMode.xml
```diff
@@ -104,7 +104,7 @@
...
...may be published and used by all threads. If the initialization method (or the parameterless constructor, if there is no initialization method) uses locks internally, deadlocks can occur. If you use a <see cref="T:System.Lazy`1" /> constructor that specifies an initialization method (<paramref name="valueFactory" /> parameter), and if that initialization method throws an exception (or fails to handle an exception) the first time you call the <see cref="P:System.Lazy`1.Value" /> property, then the exception is cached and thrown again on subsequent calls to the <see cref="P:System.Lazy`1.Value" /> property. If you use a <see cref="T:System.Lazy`1" /> constructor that does not specify an initialization method, exceptions that are thrown by the parameterless constructor for <paramref name="T" /> are not cached. In that case, a subsequent call to the <see cref="P:System.Lazy`1.Value" /> property might successfully initialize the <see cref="T:System.Lazy`1" /> instance. If the initialization method recursively accesses the <see cref="P:System.Lazy`1.Value" /> property of the <see cref="T:System.Lazy`1" /> instance, an <see cref="T:System.InvalidOperationException" /> is thrown.</summary>
```
gewarren, line 107:
```suggestion
        <summary>Locks are used to ensure that only a single thread can initialize a <see cref="T:System.Lazy`1" /> instance in a thread-safe manner. Effectively, the initialization method is executed in a thread-safe manner (referred to as `Execution` in the field name). `Publication` of the initialized value is also thread-safe in the sense that only one value may be published and used by all threads. If the initialization method (or the parameterless constructor, if there is no initialization method) uses locks internally, deadlocks can occur. If you use a <see cref="T:System.Lazy`1" /> constructor that specifies an initialization method (<paramref name="valueFactory" /> parameter), and if that initialization method throws an exception (or fails to handle an exception) the first time you call the <see cref="P:System.Lazy`1.Value" /> property, then the exception is cached and thrown again on subsequent calls to the <see cref="P:System.Lazy`1.Value" /> property. If you use a <see cref="T:System.Lazy`1" /> constructor that does not specify an initialization method, exceptions that are thrown by the parameterless constructor for <paramref name="T" /> are not cached. In that case, a subsequent call to the <see cref="P:System.Lazy`1.Value" /> property might successfully initialize the <see cref="T:System.Lazy`1" /> instance. If the initialization method recursively accesses the <see cref="P:System.Lazy`1.Value" /> property of the <see cref="T:System.Lazy`1" /> instance, an <see cref="T:System.InvalidOperationException" /> is thrown.</summary>
```

xml/System.Threading/LazyThreadSafetyMode.xml
```diff
@@ -186,7 +186,7 @@
...
...f="T:System.Lazy`1" /> instance (this is referred to as `Publication` in the field names). That value is returned to any other threads that were simultaneously running the initialization method, unless the initialization method throws exceptions on those threads. Any instances of <paramref name="T" /> that were created by the competing threads are discarded. Effectively, the publication of the initialized value is thread-safe in the sense that only one of the initialized values may be published and used by all threads. If the initialization method throws an exception on any thread, the exception is propagated out of the <see cref="P:System.Lazy`1.Value" /> property on that thread. The exception is not cached. The value of the <see cref="P:System.Lazy`1.IsValueCreated" /> property remains <see langword="false" />, and subsequent calls to the <see cref="P:System.Lazy`1.Value" /> property, either by the thread where the exception was thrown or by other threads, cause the initialization method to run again. If the initialization method recursively accesses the <see cref="P:System.Lazy`1.Value" /> property of the <see cref="T:System.Lazy`1" /> instance, no exception is thrown.</summary>
```
gewarren, line 189:
```suggestion
        <summary>When multiple threads try to initialize a <see cref="T:System.Lazy`1" /> instance simultaneously, all threads are allowed to run the initialization method (or the parameterless constructor, if there is no initialization method). The first thread to complete initialization sets the value of the <see cref="T:System.Lazy`1" /> instance. This is referred to as `Publication` in the field names. That value is returned to any other threads that were simultaneously running the initialization method, unless the initialization method throws exceptions on those threads. Any instances of <paramref name="T" /> that were created by the competing threads are discarded. Effectively, the publication of the initialized value is thread-safe in the sense that only one of the initialized values may be published and used by all threads. If the initialization method throws an exception on any thread, the exception is propagated out of the <see cref="P:System.Lazy`1.Value" /> property on that thread. The exception is not cached. The value of the <see cref="P:System.Lazy`1.IsValueCreated" /> property remains <see langword="false" />, and subsequent calls to the <see cref="P:System.Lazy`1.Value" /> property, either by the thread where the exception was thrown or by other threads, cause the initialization method to run again. If the initialization method recursively accesses the <see cref="P:System.Lazy`1.Value" /> property of the <see cref="T:System.Lazy`1" /> instance, no exception is thrown.</summary>
```

xml/System.Threading/ManualResetEventSlim.xml
```diff
@@ -56,7 +56,9 @@
 ## Remarks  
  You can use this class for better performance than <xref:System.Threading.ManualResetEvent> when wait times are expected to be very short, and when the event does not cross a process boundary. <xref:System.Threading.ManualResetEventSlim> uses busy spinning for a short time while it waits for the event to become signaled. When wait times are short, spinning can be much less expensive than waiting by using wait handles. However, if the event does not become signaled within a certain period of time, <xref:System.Threading.ManualResetEventSlim> resorts to a regular event handle wait.  
   
-   
+> [!NOTE]
+>  In .NET Core and .NET 5+, the default spin-waiting duration is very short, in the order of 10s of microseconds depending on platform and processor. If wait times are expected to be much longer than that, this class can still be used instead of <xref:System.Threading.ManualResetEvent> (perhaps configured with less or no spin-waiting), but the performance benefit would likely be only marginal.
```
gewarren, line 60:
```suggestion
> In .NET Core and .NET 5+, the default spin-waiting duration is short: on the order of 10s of microseconds, depending on platform and processor. If you expect wait times to be much longer than that, you can still use this class instead of <xref:System.Threading.ManualResetEvent> (perhaps configured with less or no spin-waiting). However, the performance benefit would likely be only marginal.  
```

xml/System.Threading/Volatile.xml
```diff
@@ -40,7 +40,10 @@
...
+  
+> [!NOTE]
+>  On a multiprocessor system, a volatile read operation does not guarantee to obtain the latest value written to that memory location by any processor. Similarly, a volatile write operation does not guarantee that the value written would be immediately visible to other processors.  
   
 > [!NOTE]
 >  On a uniprocessor system, volatile reads and writes ensure that a value is read or written to memory and not cached (for example, in a processor register). Thus, you can use these operations to synchronize access to a field that can be updated by another thread or by hardware.
```
gewarren, line 49:
```suggestion
> - On a multiprocessor system, a volatile read operation is not guaranteed to obtain the latest value written to that memory location by any processor. Similarly, a volatile write operation does not guarantee that the value written would be immediately visible to other processors.
> - On a uniprocessor system, volatile reads and writes ensure that a value is read or written to memory and not cached (for example, in a processor register). Thus, you can use these operations to synchronize access to a field that can be updated by another thread or by hardware.  
```

xml/System.Threading/Volatile.xml
```diff
@@ -40,7 +40,10 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- On a multiprocessor system, a volatile write operation ensures that a value written to a memory location is immediately visible to all processors. A volatile read operation obtains the very latest value written to a memory location by any processor. These operations might require flushing processor caches, which can affect performance.  
+ On a multiprocessor system, due to performance optimizations in the compiler or processor, regular memory operations may appear to be reordered when multiple processors are operating on the same memory. Volatile memory operations prevent certain types of reordering with respect to the operation. A volatile write operation prevents earlier memory operations from being reordered to occur after the volatile write. A volatile read operation prevents later memory operations from being reordered to occur before the volatile read. These operations might involve memory barriers on some processors, which can affect performance.  
+
```
gewarren, line 44:
```suggestion
```

xml/System.Threading/Volatile.xml
```diff
@@ -40,7 +40,10 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- On a multiprocessor system, a volatile write operation ensures that a value written to a memory location is immediately visible to all processors. A volatile read operation obtains the very latest value written to a memory location by any processor. These operations might require flushing processor caches, which can affect performance.  
+ On a multiprocessor system, due to performance optimizations in the compiler or processor, regular memory operations may appear to be reordered when multiple processors are operating on the same memory. Volatile memory operations prevent certain types of reordering with respect to the operation. A volatile write operation prevents earlier memory operations from being reordered to occur after the volatile write. A volatile read operation prevents later memory operations from being reordered to occur before the volatile read. These operations might involve memory barriers on some processors, which can affect performance.  
+  
+> [!NOTE]
```
gewarren, line 45:
```suggestion
```

xml/System.Threading/Volatile.xml
```diff
@@ -40,7 +40,10 @@
...
+>  On a multiprocessor system, a volatile read operation does not guarantee to obtain the latest value written to that memory location by any processor. Similarly, a volatile write operation does not guarantee that the value written would be immediately visible to other processors.
```
gewarren, line 46:
```suggestion
```

gewarren (approved):

carlossanlop (approved):

PR #7156, opened 2021-09-13
Fix weird char causing IntelliSense CI to fail

carlossanlop:
The CI job that we use for IntelliSense generation from dotnet-api-docs is failing in main due to an unexpected indent character.


gewarren:
This is already [fixed in net6-rc1](https://github.com/dotnet/dotnet-api-docs/commit/6b85304e833e9b3c3619105d20b546b8aad7acf1), which will be merged to main today or tomorrow, so I don't think we need this PR. I had to fix it to run the CI job when I reimported the assemblies without the source generator one.

PR #7169, opened 2021-09-13
Try to fix first batch of moniker warnings

gewarren:


carlossanlop:
@gewarren shouldn't the Docs build system do this automatically?

gewarren:
This is not the right fix. Closing.

PR #7190, opened 2021-09-15
Filter System.Text.Json.Serialization.Metadata APIs

carlossanlop:
@eiriktsarpalis @layomia @steveharter 

MS Docs should not show any APIs from System.Text.Json.Serialization.Metadata because they are only meant to be consumed by source generation code.

They don't have their own assembly: they come included in the System.Text.Json assembly, which is why we cannot exclude these from the ref assemblies we send to Docs.

Pending confirming that the namespace itself is _also_ excluded from MS Docs.

### DIFF
--- a/xml/_filter.xml
+++ b/xml/_filter.xml
@@ -145,6 +145,11 @@
       <typeFilter name="DependencyInjectionSpecificationTests" expose="false" />
       <typeFilter name="DependencyInjectionSpecificationTests.CreateInstanceFunc" expose="false" />
     </namespaceFilter>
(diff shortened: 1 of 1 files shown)


carlossanlop:
@eiriktsarpalis an interesting set of warnings showed up after I excluded S.T.J.Serialization.Metadata: Those APIs that were supposed to be ignored, are being referenced in several places.

carlossanlop:
I'll have to close this, @eiriktsarpalis . There are several MemberSignatures that depend on APIs that live in S.T.J.Serialization.Metadata, so the namespace cannot be ignored.

I think instead we should clarify in the namespace description that the APIs in inside it are not supposed to be called by users.

PR #7196, opened 2021-09-16
Revert moniker changes

gewarren:
I need to revert these changes because the ECMAXML is actually correct. The issue seems to be that _the assembly that a type is forwarded to_ is not always present in the correct directory in the binaries repo when we import the APIs.


BillWagner (approved):
This LGTM @gewarren 

Once you and the product team members are happy, you can :shipit:

PR #7200, opened 2021-09-17
Update Application.xml

RussKie:
## Summary

Fix incorrect rendering of the remarks
image

### DIFF
--- a/xml/System.Windows.Forms/Application.xml
+++ b/xml/System.Windows.Forms/Application.xml
@@ -1625,7 +1625,7 @@ static class Program
 ## Remarks
 
 The system text scale factor is applied to the font. For example, if the default font is set to "Calibri, 11f"
-             and the text scale factor is set to 150% the resulting default font will be set to "Calibri, 16.5f".
+and the text scale factor is set to 150%, the resulting default font is set to "Calibri, 16.5f".
 
 You can adjust text scale with the **Make text bigger** slider on the **Settings** > **Ease of Access** > **Vision/Display** screen.
 


xml/System.Windows.Forms/Application.xml
```diff
@@ -1585,89 +1585,83 @@ static class Program
...
              The system text scale factor will be applied to the font, i.e. if the default font is set to "Calibri, 11f"
              and the text scale factor is set to 150% the resulting default font will be set to "Calibri, 16.5f".
-            
-
-
-             Users can adjust text scale with the Make text bigger slider on the Settings -&gt; Ease of Access -&gt; Vision/Display screen.
-            
-
 
+             Users can adjust text scale with the Make text bigger slider on the Settings -> Ease of Access -> Vision/Display screen.
```
gewarren, line 1630:
```suggestion
You can adjust text scale with the **Make text bigger** slider on the **Settings** > **Ease of Access** > **Vision/Display** screen.
```

xml/System.Windows.Forms/Application.xml
```diff
@@ -1585,89 +1585,83 @@ static class Program
...
+      <Docs>
+        <param name="font">The font to be used as a default across the application.</param>
+        <summary>Sets the default <see cref="T:System.Drawing.Font" /> for the process.</summary>
+        <remarks>
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
-
              The system text scale factor will be applied to the font, i.e. if the default font is set to "Calibri, 11f"
```
gewarren, line 1627:
```suggestion
The system text scale factor is applied to the font. For example, if the default font is set to "Calibri, 11f"
```

xml/System.Windows.Forms/Application.xml
```diff
@@ -1585,89 +1585,83 @@ static class Program
...
              and the text scale factor is set to 150% the resulting default font will be set to "Calibri, 16.5f".
```
gewarren, line 1628:
```suggestion
and the text scale factor is set to 150%, the resulting default font is set to "Calibri, 16.5f".
```

gewarren (approved):
The problem is the indentation.

RussKie:
> The problem is the indentation.

This is [our xmldocs](https://github.com/dotnet/winforms/blob/08b29bcd36af441b653b89a32a9a4ce285a35ae9/src/System.Windows.Forms/src/System/Windows/Forms/Application.cs#L1245-L1263):
image

Feels like we have some kind of bug in the transformation pipeline which creates such wild indentations. Would you know where to raise this?

PR #7209, opened 2021-09-19
UintPtr & IntPtr struct – rewording

deep-outcome:
## Summary
Microfix of bad wording.


MSDN-WhiteKnight:
Why is this change needed? Is the current wording incorrect grammar, unclear, or considered bad writing style?

Uzivatel919:
> Why is this change needed? Is the current wording incorrect grammar, unclear, or considered bad writing style?

It has bad formula. Data are not referenced by language. At least for me, it seems confusing. The more that `UintPtr`/`IntPtr` will likely be used only in one of them. So between as conjunction of meaning in the middle is slightly inaccurate.

Sense is clear in common language but in technical vocabulary I mean not.

MSDN-WhiteKnight:
This PR likely conflicted with https://github.com/dotnet/dotnet-api-docs/pull/8132

PR #7211, opened 2021-09-21
Fix description for RegexOptions.Compiled

Catalan-1ap:
## Summary

Fixed description for Regex.Options.Compiled enum flag.
With the reference to https://www.youtube.com/watch?v=RSFiiKUvzLI

### DIFF
--- a/xml/System.Text.RegularExpressions/RegexOptions.xml
+++ b/xml/System.Text.RegularExpressions/RegexOptions.xml
@@ -117,7 +117,7 @@
       </ReturnValue>
       <MemberValue>8</MemberValue>
       <Docs>
-        <summary>Specifies that the regular expression will compile IL code on the fly, not compile time, runtime to pass string for that specific expression. This value should not be assigned to the <see cref="P:System.Text.RegularExpressions.RegexCompilationInfo.Options" /> property when calling the <see cref="M:System.Text.RegularExpressions.Regex.CompileToAssembly(System.Text.RegularExpressions.RegexCompilationInfo[],System.Reflection.AssemblyName)" /> method. For more information, see the "Compiled Regular Expressions" section in the <see href="/dotnet/standard/base-types/regular-expression-options">Regular Expression Options</see> article.</summary>
+        <summary>Specifies that the regular expression is compiled to MSIL code, instead of being interpreted. Compiled regular expressions maximize run-time performance at the expense of initialization time. This value should not be assigned to the <see cref="P:System.Text.RegularExpressions.RegexCompilationInfo.Options" /> property when calling the <see cref="M:System.Text.RegularExpressions.Regex.CompileToAssembly(System.Text.RegularExpressions.RegexCompilationInfo[],System.Reflection.AssemblyName)" /> method. For more information, see the "Compiled Regular Expressions" section in the <see href="/dotnet/standard/base-types/regular-expression-options">Regular Expression Options</see> article.</summary>
       </Docs>
     </Member>
     <Member MemberName="CultureInvariant">


xml/System.Text.RegularExpressions/RegexOptions.xml
```diff
@@ -117,7 +117,7 @@
...
+        <summary>Specifies that the regular expression is compile IL code on the fly, not compile time, runtime to pass string for that specific expression. This value should not be assigned to the <see cref="P:System.Text.RegularExpressions.RegexCompilationInfo.Options" /> property when calling the <see cref="M:System.Text.RegularExpressions.Regex.CompileToAssembly(System.Text.RegularExpressions.RegexCompilationInfo[],System.Reflection.AssemblyName)" /> method. For more information, see the "Compiled Regular Expressions" section in the <see href="/dotnet/standard/base-types/regular-expression-options">Regular Expression Options</see> article.</summary>
```
eerhardt, line 120: I'm not sure this sentence is grammatically correct.

> the regular expression is compile IL code on the fly

xml/System.Text.RegularExpressions/RegexOptions.xml
```diff
@@ -117,7 +117,7 @@
...
+        <summary>Specifies that the regular expression will compile IL code on the fly, not compile time, runtime to pass string for that specific expression. This value should not be assigned to the <see cref="P:System.Text.RegularExpressions.RegexCompilationInfo.Options" /> property when calling the <see cref="M:System.Text.RegularExpressions.Regex.CompileToAssembly(System.Text.RegularExpressions.RegexCompilationInfo[],System.Reflection.AssemblyName)" /> method. For more information, see the "Compiled Regular Expressions" section in the <see href="/dotnet/standard/base-types/regular-expression-options">Regular Expression Options</see> article.</summary>
```
eerhardt, line 120: > This yields faster execution but increases startup time

I think this is still important to keep in the docs.

eerhardt, line 120: I'm not sure I understand this part:

> on the fly, not compile time, runtime to pass string for that specific expression

1. I'm not sure "runtime" makes sense in this sentence.
2. "to pass string for that specific expression". I don't think that is grammatically correct.

Catalan-1ap: But tests don't prove it. This flag does not speed up execution, but on the contrary slows it down.

Catalan-1ap: > I'm not sure I understand this part:
> 
> > on the fly, not compile time, runtime to pass string for that specific expression
> 
> 1. I'm not sure "runtime" makes sense in this sentence.
> 2. "to pass string for that specific expression". I don't think that is grammatically correct.

How about this? - "This string will be passed runtime"

eerhardt: It does speed up execution. You just need to cache the Regex instance, and not create a new instance every time you call it. That's what it means by "increases startup time".

eerhardt: > How about this? - "This string will be passed runtime"

I still don't understand what it means.

Catalan-1ap: > It does speed up execution. You just need to cache the Regex instance, and not create a new instance every time you call it. That's what it means by "increases startup time".

In that case
this line - "This yields faster execution but increases startup time"
should be replaced by this - "This yields faster execution only if expression cached, but increases startup time"

Catalan-1ap: > > How about this? - "This string will be passed runtime"
> 
> I still don't understand what it means.

This may clarify that that this expression will be executed at runtime anyway.

gewarren:
Close/reopen for fresh build.

eerhardt (commented):
@Catalan-1ap, thanks for your help here. Please take a look at my wording suggestion. I believe it addresses your concerns.

xml/System.Text.RegularExpressions/RegexOptions.xml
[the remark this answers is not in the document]
Catalan-1ap: Yeah, it’s a great solution, thanks for your help

eerhardt (approved):
Thanks @Catalan-1ap for the contribution!

PR #7222, opened 2021-09-24
Fix mscorlib in .NET 5 APIs

gewarren:
Remove mscorlib shim from net-5.0 and dependencies/dotnet-plat-ext-5.0 directories so forwarded types in that assembly don't appear as applicable to .NET 5.


safern (approved):
LGTM, thanks!

PR #7224, opened 2021-09-24
Fix CODEOWNERS syntax

gewarren:
Folder/file owners aren't getting in tagged in PRs, which means there's a problem with the syntax here.


carlossanlop (approved):
Oops, my bad. Thanks for fixing!

PR #7252, opened 2021-09-29
Separate out Windows Desktop moniker

gewarren:
Separates out the assemblies that ship as part of the [WindowsDesktop ref pack](https://www.nuget.org/packages/Microsoft.WindowsDesktop.App.Ref/). This PR covers .NET Core 3.1, .NET 5, and .NET 6. If .NET Core 3.0 should be split out as well, I can add that.


gewarren:
Replaces #7241

safern:
> If .NET Core 3.0 should be split out as well, I can add that.

I would only do that for historical reasons but I don't think it is a big deal as 3.0 is already out of support.

PR #7253, opened 2021-09-29
Separate out Windows Desktop moniker

gewarren:
Separates Windows Desktop assemblies from shared framework assemblies for .NET Core 3.0, .NET Core 3.1, .NET 5, and .NET 6.

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/api/system.drawing.brush?view=dotnet-plat-ext-2.1&branch=pr-en-us-7253#applies-to).

Replaces #7252.


safern (approved):
I skimmed through and looks good. I don't know if I would name the moniker that the user sees ".NET Windows Desktop" rather than just "Windows Desktop".

safern:
cc: @terrajobst @RussKie

terrajobst:
How will this show up in the picker?

gewarren:
> How will this show up in the picker?

Like this:

image

safern:
Looking at the picture I think using ".NET" before "Windows Desktop" makes more sense so that the moniker is part of the .NET group.

RussKie:
I agree "**.NET** Windows Desktop" would visualise it belongs to .NET group.

terrajobst:
The picker is generally somewhat unfortunate in that it's not aligned with the mechanics of how we ask people to think about this when creating projects. Sadly, that is also complicated as the mechanics aren't uniform either:

1. For ".NET" and ".NET Framework" it's simple -- just set `<TargetFramework>` to `netX.Y`
2. For "ASP.NET Core" it's a combination of (1) plus using the Web SDK (or by manually adding `<FrameworkReference>` elements)
3. For "Windows Desktop" it's (1) plus setting `<UseWindowsForms>` and/or `<UseWpf>`.
4. etc.

Prefixing everything with ".NET" isn't the right fix in this dropdown because it's all .NET. And since UWP/WinUI is already separate from Windows Desktop, leaving it as "Windows Desktop" is probably fine.

terrajobst (approved):

gewarren:
Will rerun the CI due to the conflicts.

PR #7258, opened 2021-09-30
Revert "Disable recommendations - merge to live"

gewarren:
Reverts dotnet/dotnet-api-docs#7257

The SiteHelp team needs to repro the issues I logged.


gewarren:
This should go through main first, otherwise it will just get overwritten tonight.

PR #7259, opened 2021-09-30
Fix codeowners file

gewarren:
Fixes team names. I also had to explicitly give Write permissions to each of these teams.

"The people you choose as code owners must have write permissions for the repository. When the code owner is a team, that team must be visible and it must have write permissions, even if all the individual members of the team already have write permissions directly, through organization membership, or through another team membership."


gewarren:
Ping @carlossanlop for review. I think we need to get this merged before it will start working again.

PR #7266, opened 2021-10-03
Clarify `SymmetricAlgorithm` `.Dispose()` vs. `.Clear()`

dosentmatter:
## Summary

**EDIT:**
`Aes` code snippets don't use `.Clear()` while the docs say that you should call `.Clear()` to zero out sensitive data. `.Clear()` actually just calls `.Dispose()`, so a call to `.Clear()` in addition to disposing is unnecessary. Clarify the docs to describe that `.Clear()` calls `.Dispose()` and that developers should dispose instead of using `.Clear()`.

**OLD:**
Update `Aes` Code Snippets to use `.Clear()`.

In the [`SymmetricAlgorithm`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.symmetricalgorithm?view=net-5.0#remarks) and [`SymmetricAlgorithm.Clear`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.symmetricalgorithm.clear?view=net-5.0#remarks) docs, it says to use `.Clear()` before `.Dispose()` to zero out sensitive data:

> Note that when using derived classes, it is not enough, from a security perspective, to simply force a garbage collection after you have finished using the object. You must explicitly call the Clear method on the object to zero out any sensitive data within the object before it is released. Note that garbage collection does not zero out the contents of collected objects but simply marks the memory as available for reallocation. Thus the data contained within a garbage collected object may still be present in the memory heap in unallocated memory. In the case of cryptographic objects, this data could contain sensitive information such as key data or a block of plain text.
> 
> All cryptographic classes in the .NET Framework that hold sensitive data implement a Clear method. When called, the Clear method overwrites all sensitive data within the object with zeros and then releases the object so that it can be safely garbage collected. When the object has been zeroed and released, you should then call the Dispose method with the disposing parameter set to True to dispose of all managed and unmanaged resources associated with the object.

### DIFF
--- a/xml/System.Security.Cryptography/SymmetricAlgorithm.xml
+++ b/xml/System.Security.Cryptography/SymmetricAlgorithm.xml
@@ -58,7 +58,7 @@
 
  Note that when using derived classes, it is not enough, from a security perspective, to simply force a garbage collection after you have finished using the object. You must dispose the object to zero out any sensitive data within the object before it is released. Note that garbage collection does not zero out the contents of collected objects, and the data contained within a garbage collected object may still be present in the memory. In the case of cryptographic objects, this data could contain sensitive information such as key data or a block of plain text.
 
- All cryptographic classes in the .NET Framework that hold sensitive data implement the <xref:System.IDisposable> interface. Disposing overwrites all sensitive data within the object with zeros so that it can be safely garbage collected.
+ All cryptographic classes in the .NET Framework that hold sensitive data implement the <xref:System.IDisposable> interface. Disposing overwrites all sensitive data within the object with zeros so that it can be garbage collected safely.
 
 
 


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=7266) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=7266) before we can accept your contribution.<br/><br/>:x: dosentmatter [sign now](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=7266)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/dotnet-api-docs?pullRequest=7266) it.</sub>

gfoidl:
`SymmetricAlgorithm.Clear` just calls [Dispose](https://github.com/dotnet/runtime/blob/355a8cf12b21d81248aa8f5fc7ca2544e4a63db5/src/libraries/System.Security.Cryptography.Primitives/src/System/Security/Cryptography/SymmetricAlgorithm.cs#L192-L195), so by following the using-pattern to dispose the object properly, a additional call to Clear adds no benefit.

Maybe the docs should be clarified in that concern?

dosentmatter:
You are right.

Seems like the [`SymmetricAlgorithm.Clear()` was added in Nov 2016](https://github.com/dotnet/runtime/commit/7067aa592e451bc739914cdd4a8577d11820ce4c#diff-9776ffb4b35099e38ae75ab2375c64491ac26e68be330f76a947036ba4aeb2bf) while the [docs about needing to manually call `Clear` were added in March 2017](https://github.com/dotnet/dotnet-api-docs/commits/ca7d94d93ac693ef3a5d234cbceaf445cdc9ed35/xml/System.Security.Cryptography/SymmetricAlgorithm.xml). Note that the GitHub diff for the second link is too large to view, but you can [see it in this blame](https://github.com/dotnet/dotnet-api-docs/blame/5b92d7e84820fa8713b379b2a97fb3bb4a44059a/xml/System.Security.Cryptography/SymmetricAlgorithm.xml#L54-L61).

I think `SymmetricAlgorithm.xml` is the only [doc that mentions using `.Clear()`](https://github.com/dotnet/dotnet-api-docs/search?q=%22You+must+explicitly+call+the%22+%22clear%22+method+%22zero+out%22). I guess we can delete that section and for the docs on `.Clear()`, we can just mimic the [`Stream` docs](https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.IO/Stream.xml#L618) by telling the user to use `.Dispose()` instead:

> Closes the current stream and releases any resources (such as sockets and file handles) associated with the current stream. Instead of calling this method, ensure that the stream is properly disposed.

gfoidl:
> Does that sound okay?

For me this sounds good, but I defer to other people on that decision.

PR #7282, opened 2021-10-10
Adds reference to IAsyncDisposable interface

strobelt:
## Summary

Added reference to IAsyncDisposable interface in ObjectDisposedException Class Remarks

Fixes https://github.com/dotnet/docs/issues/25968

### DIFF
--- a/xml/System/ObjectDisposedException.xml
+++ b/xml/System/ObjectDisposedException.xml
@@ -62,7 +62,7 @@
      :::code language="csharp" source="~/samples/snippets/csharp/VS_Snippets_CLR_System/system.objectdisposedexception/cs/dispose1.cs" id="Snippet1":::
      :::code language="vb" source="~/samples/snippets/visualbasic/VS_Snippets_CLR_System/system.objectdisposedexception/vb/dispose1.vb" id="Snippet1":::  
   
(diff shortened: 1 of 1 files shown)


strobelt:
@Uzivatel919 I added some references to `IAsyncDisposable` to the Remarks section. What do you think?
I thought about adding an example, but didn't know about any `IAsyncDisposable` implementations to use in it. Maybe it's not necessary?

Uzivatel919:
@strobelt, I think it is fine now. [System.Threading.Timer](https://docs.microsoft.com/en-us/dotnet/api/system.threading.timer?view=net-5.0) used in example also implements `IAsyncDisposable` interface. If you find `async Task Main` example beneficial in this case, you can go on with it or create some perhpaps by choosing some other implementation.

https://dotnetfiddle.net/n5LIxq

```csharp
using System;
using System.Threading;
using System.Threading.Tasks;
					
public class Example
{
   public static async Task Main()
   {
      Timer t = new Timer(TimerNotification, null,
                         100, Timeout.Infinite);
      Thread.Sleep(2000);
      await t.DisposeAsync();

      t.Change(200, 1000);
      Thread.Sleep(3000);
   }

   private static void TimerNotification(Object obj)
   {
      Console.WriteLine("Timer event fired at {0:F}", DateTime.Now);
   }
}
```

Uzivatel919:
@strobelt, I agree.

BillWagner (approved):
Thanks for submitting these changes @strobelt 

We appreciate it.

I've reviewed the changes and I'll :shipit: now.

You should see the changes on our live site in the next day or two, on our regular publishing cycle.

PR #7294, opened 2021-10-15
EnableBroadcast property description is incorrect - controls only sending broadcast packets

rmiller14:
## Summary

The description for the EnableBroadcast property of the Socket and UdpClient classes states that it controls "whether the [class] can send or receive broadcast packets."  This is incorrect.  That property controls only sending to broadcast packets, as its underlying implementation involves using [SO_BROADCAST](https://docs.microsoft.com/en-us/windows-hardware/drivers/network/so-broadcast), which "determines whether broadcast messages can be transmitted over a datagram socket."

### DIFF
--- a/xml/System.Net.Sockets/Socket.xml
+++ b/xml/System.Net.Sockets/Socket.xml
@@ -4815,9 +4815,9 @@ Duplication of the socket reference failed.</exception>
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets or sets a <see cref="T:System.Boolean" /> value that specifies whether the <see cref="T:System.Net.Sockets.Socket" /> can send or receive broadcast packets.</summary>
+        <summary>Gets or sets a <see cref="T:System.Boolean" /> value that specifies whether the <see cref="T:System.Net.Sockets.Socket" /> can send broadcast packets.</summary>
         <value>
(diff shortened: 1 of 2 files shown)


rmiller14:
> Before making this change, we should make sure there is a test case for this in System.Net.Sockets. @rmiller14 is this something you want to do?
> 
> Part of me feels like `SO_BROADCAST` was in some cases needed for receiving, but I haven't used it in so long and memory is foggy.

I don't have too many spare cycles to help out with a test.  Also, what's the motivation behind adding such a test?  System.Net.Sockets.Socket.SetSocketOption is well covered already, and the broadcast option itself is already covered by the OS itself.

Plus, I can guarantee that SO_BROADCAST is not needed for receiving.  My team owns socket options :)

rmiller14:
Ping?

rmiller14:
Can we get this checked in please?  This is causing customer confusion and has already cost support time for my team.  It's a simple doc change.

antonfirsov (approved):
> Plus, I can guarantee that SO_BROADCAST is not needed for receiving. My team owns socket options :)

With this I'm fine merging.

rmiller14:
Thanks @antonfirsov.  Who do I need to ping to get this merged?

rmiller14:
Can we please get this merged?

PR #7304, opened 2021-10-20
Correct typo, CTL -> CTRL

Lucas7yoshi:
## Summary

Corrects references to the CTRL key as CTL in System.Windows.Forms.Control.DoDragDrop
I was not able to find any information that would suggest this is correct.


adegeo:
Oh, whoops. I thought you were looking for why the snippets build was failing. 😁

It's just updating some code comments using CTRL to identify the keyboard key.

@gewarren @BillWagner  Are we currently requiring minor updates like this to be recreated with full code examples?

gewarren:
I think we can let this change go without adding a project file since it's literally just updating code comments.

PR #7308, opened 2021-10-20
Networking settings - clarify .NET version availability

karelz:
Clarify availability of switch `UseSocketsHttpHandler` on .NET Core 2.1-3.1 only.

### DIFF
--- a/xml/System/AppContext.xml
+++ b/xml/System/AppContext.xml
@@ -363,8 +363,8 @@ In .NET 5.0 and later versions, for bundled assemblies, the value returned is th
 
 |Switch|Values|Description|
 |--|--|--|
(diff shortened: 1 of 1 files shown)


karelz:
Actually, the more I look at the section [`AppContext.SetSwitch and .NET Core`](https://docs.microsoft.com/en-us/dotnet/api/system.appcontext.setswitch?view=net-5.0#appcontextsetswitch-and-net-core), I think it is incorrect. These switches are not programmatic-only (they can come from config or env vars). And there are more of such switches - at least 1 more in Networking (Latin1) and most likely more outside of Networking.
We should probably remove the entire section in favor of https://docs.microsoft.com/en-us/dotnet/core/run-time-config/networking?source=docs

@dotnet/area-system-runtime any thoughts here?

PR #7319, opened 2021-10-24
Update ProductInfoHeaderValue.xml

### LINKED ISSUE #3274: Document ProductInfoHeaderValue constructors
I'm sometimes making HTTP requests to various APIs. Quite often, they require specifying [the User-Agent header](https://en.wikipedia.org/wiki/User_agent#Use_in_HTTP). To do that with `HttpClient`, I have to create `ProductInfoHeaderValue`, e.g.:

```c#
ProductInfoHeaderValue value = …;
var client = new HttpClient { DefaultRequestHeaders = { UserAgent = { value } } };
```

My problem is that creating valid `ProductInfoHeaderValue` is not obvious and [the documentation for its constructors](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.headers.productinfoheadervalue.-ctor) does not explain which constructor I should use or what are the requirements for the `string` parameters.

As an example, if I try `new ProductInfoHeaderValue("svick's awesome tool")`, it throws "FormatException: The format of value 'svick's awesome tool' is invalid." This doesn't help me figure out what the valid values are, which is why I think documenting this is important.

[This article](http://bizcoder.com/the-much-maligned-user-agent-header) explains the rules for User-Agent values and might be helpful when resolving this issue.

wenz:
## Summary

Taking a shot at better describing System.Net.Http.Headers.ProductInfoHeaderValue

Fixes #3274

### DIFF
--- a/xml/System.Net.Http.Headers/ProductInfoHeaderValue.xml
+++ b/xml/System.Net.Http.Headers/ProductInfoHeaderValue.xml
@@ -44,7 +44,7 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Net.Http.Headers.ProductInfoHeaderValue> class provides support for the product and comment tokens values used in  a User-Agent header as defined in [RFC 2616](https://go.microsoft.com/fwlink/?LinkID=241642) by the IETF.  
+ The <xref:System.Net.Http.Headers.ProductInfoHeaderValue> class provides support for the product and comment tokens values used in a User-Agent header as defined in [RFC 2616](https://go.microsoft.com/fwlink/?LinkID=241642) by the IETF.  
   
  The User-Agent field contains information about the user agent originating the request. The field can contain multiple product tokens and comments identifying the user agent and any sub-products.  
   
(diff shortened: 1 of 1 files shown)


xml/System.Net.Http.Headers/ProductInfoHeaderValue.xml
```diff
@@ -44,7 +44,7 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Net.Http.Headers.ProductInfoHeaderValue> class provides support for the product and comment tokens values used in  a User-Agent header as defined in [RFC 2616](https://go.microsoft.com/fwlink/?LinkID=241642) by the IETF.  
+ The <xref:System.Net.Http.Headers.ProductInfoHeaderValue> class provides support for the product and comment tokens values used in a User-Agent header as defined in [RFC 2616](https://go.microsoft.com/fwlink/?LinkID=241642) by the IETF.
```
ManickaP, line 47: Why link it as fwlink?
Also 2616 has been obsoleted, twice already, the newest RFC is here: https://www.rfc-editor.org/rfc/rfc9110#field.user-agent

PR #7328, opened 2021-10-27
Update x500DistinguishedName snippets to include relevant code

### LINKED ISSUE #7327: X500DistinguishedNameFlags examples do not contain relevant snippets
The documentation page for [X500DistinguishedNameFlags ](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x500distinguishednameflags?view=netcore-3.1) contains an example code snippet that does not reference `X500DistinguishedNameFlags` at all. 

Ideally the snippet on this page would show an example of how the flags are intended to be used.

bderusha:
## Summary

Adds x500DistinguishedNameFlag example to the general x500DistinguishedName snippets making the snippet more relevant to the [X500DistinguishedNameFlags documentation page](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x500distinguishednameflags?view=netcore-3.1).

Fixes #7327

### DIFF
--- a/xml/System.Security.Cryptography.X509Certificates/X500DistinguishedName.xml
+++ b/xml/System.Security.Cryptography.X509Certificates/X500DistinguishedName.xml
@@ -60,41 +60,38 @@
   <Docs>
     <summary>Represents the distinguished name of an X509 certificate. This class cannot be inherited.</summary>
     <remarks>
-      <format type="text/markdown"><![CDATA[  
-  
-## Remarks  
- This class is like an extension to the <xref:System.Security.Cryptography.X509Certificates.X509Certificate2.SubjectName%2A> or <xref:System.Security.Cryptography.X509Certificates.X509Certificate2.IssuerName%2A> property, which is the name of the person or entity that the certificate is being issued to. X.500 is an international standard for distributed directory services. The distinguished name uses the following format:  
-  
- [X500:/C=CountryCode/O=Organization/OU=OrganizationUnit/CN=CommonName]  
-  
- The following field length limits apply:  
-  
-|Field|Length Limit|  
-|-----------|------------------|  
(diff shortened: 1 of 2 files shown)


jimbobmcgee:
Just a two-cents opinion from an end-user who discovered this issue whilst looking at the documentation for `X500DistinguishedNameFlags` but, while updating the snippet to actually _use_ the enum is a great first step, it isn't hard to intuit how to use it from the `X500DistinguishedName` constructor itself, if you are familiar with the language.

What _would_ be useful is if the documentation indicated whether using the enum affected how the input string was parsed or validated, or just set some property of the resulting `X500DistinguishedName` object.

For instance, one of the flags values is `DoNotUsePlusSign`, and the description for this is _"The distinguished name does not use the plus sign."_.  If the input string to `new X500DistinguishedName(...)` did in fact contain a plus sign, what happens?  An exception?  Some automatic re-encoding of the input string?  Some automatic re-encoding of the `.ToString()` output?  It is not immediately clear from either the main description of `X500DistinguishedNameFlags` or the `X500DistinguishedName(string, X500DistinguishedNameFlags)` constructor.

If the example code instead used some literal input strings, and described the resulting output, e.g. something of the form...

```
(new X509DistinguishedName("CN=foo,O=Fabrikam+Co.,C=US", X509DistinguishedNameFlags.Reversed).ToString();    // returns "C=US,O=Fabrikam+Co.,CN=foo"
(new X509DistinguishedName("CN=foo,O=Fabrikam+Co.,C=US", X509DistinguishedNameFlags.DoNotUsePlusSign).ToString();    // throws SomethingException
```

...(but accurate), that might be more demonstrative of this enum.

gewarren (approved):

Issue #834, opened 2018-08-16
Doesn't work with Zip64 files


EbrahimMajdey:
The ZipArchive can't read any entry that is larger than 4,294,967,295 bytes .. which is the old limit for zip files although the OS can read this.. would appreciate if such capabilities is enabled instead of using other packages

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3a46d025-34f1-9ac4-6d29-a8e527cc5f91
* Version Independent ID: 5642d6cc-6abc-e662-bc03-d284ca9cb438
* Content: [ZipArchive Class (System.IO.Compression)](https://docs.microsoft.com/en-us/dotnet/api/system.io.compression.ziparchive?redirectedfrom=MSDN&view=netframework-4.7.2)
* Content Source: [xml/System.IO.Compression/ZipArchive.xml](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System.IO.Compression/ZipArchive.xml)
* Product: **dotnet-api**
* GitHub Login: @mairaw
* Microsoft Alias: **mairaw**

carlossanlop:
I think this was answered a long time ago, so we should close it.

PR #7333, opened 2021-10-28
Clarify that ZipArchive supports ZIP64.

ericmutta:
I spent a great deal of time trying to figure out whether the `ZipArchive` class supports the [ZIP64 format](https://en.wikipedia.org/wiki/ZIP_(file_format)#ZIP64) and could thus create archives that are larger than 4GB in size.  I don't see any official documentation that says it can, even though the code shows that [ZIP64 is in fact supported](https://github.com/dotnet/runtime/blob/main/src/libraries/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.cs#L464), so I added a sentence to the summary of the `ZipArchive` class.


ericmutta:
Any word on this?

carlossanlop:
Thank you for your patience, and apologies for the delay.

I agree with you, we should have some clarification about zip64 somewhere in our zip docs. 

We used to have another issue asking about zip64 but I closed it since there was no request to modify the docs.  In hindsight, I should've proposed the docs improvement. 
 https://github.com/dotnet/dotnet-api-docs/issues/834#issuecomment-425173006

I think eventually we should also mention the conditions required for zip64 to work:  We support zips larger than 4gb when they're encoded in supported methods and decoded in Read Mode.  But I can add that info in a separate PR after I confirm all the info we need to mention.

Can you please move your new text to the remarks, please? The summary does not seem the appropriate location.

ericmutta:
Hi @carlossanlop many thanks for getting back to me!

As per your suggestion, I have removed the text from the summary and added a short sentence to the remarks which simply says:

> **Zip packages larger than 4GB are supported.**

I believe that covers the two questions I had personally when looking into this:

1. Can I _create_ a zip archive larger than 4GB?
2. Can I _read_ a zip archive larger than 4GB?

One final question a person might have is: **Can I _add/remove_ entries in an existing archive that is larger than 4GB?** This is not too important in my use case but if the answer is known, I will be happy to make further edits to the remarks to add the info :+1:

BillWagner:
Ping @carlossanlop to answer above question.

ericmutta:
Closing this as it doesn't look like a priority to the docs team (it's been _two months_) :disappointed:

PR #7335, opened 2021-10-29
F# Snippets for System.Random

albert-du:
## Add F# Code Snippets for System.Random

Describe your changes here.

Adds a few F# snippets for System.Random. Snippets are equivalent to C# and VB versions with a little refactoring to make them more idiomatic.

### DIFF
--- a/samples/snippets/fsharp/VS_Snippets_CLR_System/system.Random.Next/FS/next.fs
+++ b/samples/snippets/fsharp/VS_Snippets_CLR_System/system.Random.Next/FS/next.fs
@@ -10,27 +10,27 @@ let noBoundsRandoms seed =
 
     // Generate six random integers from 0 to int.MaxValue.
     for _ = 1 to 6 do
-        printf "%11i " (randObj.Next())
+        printf $"%11i{randObj.Next()} "
     printfn ""
 
 // Generate random numbers with an upper bound specified.
 let upperBoundRandoms seed upper = 
-    printfn "\nRandom object, seed = %i, upper bound = %i" seed upper
+    printfn $"\nRandom object, seed = %i{seed}, upper bound = %i{upper}"
     let randObj = Random seed
 
     // Generate six random integers from 0 to the upper bound.
     for _ = 1 to 6 do
-        printf "%11i " (randObj.Next(upper))
+        printf $"%11i{randObj.Next upper} "
     printfn ""
 
 // Generate random numbers with both bounds specified.
 let bothBoundRandoms seed lower upper =
-    printfn "\nRandom object, seed = %i, lower = %i, upper = %i: " seed lower upper
+    printfn $"\nRandom object, seed = %i{seed}, lower = %i{lower}, upper = %i{upper}: "
     let randObj = Random seed
 
     // Generate six random integers from the lower to upper bounds.
     for _ = 1 to 6 do 
-        printf "%11i " (randObj.Next(lower,upper))
+        printf $"%11i{randObj.Next(lower,upper)} "
     printfn ""
 
 printfn "This example of the Random.Next() methods\ngenerates the following.\n"


albert-du:
should be ready now

BillWagner:
Thanks @albert-du 

We've added a CI build for all our snippets. Can you include the .fsproj file for building these samples? Then, that step should pass, and this would be ready to merge.  (I did a reasonable review, and once the build is clean, it's ready).

albert-du:
Awesome, I'll add one when I get a chance

eerhardt:
@KathleenDollard - can you review?

albert-du:
@dsyme

eerhardt:
cc @baronfel as well, in case you are interested in reviewing.

samples/snippets/fsharp/VS_Snippets_CLR_System/system.Random.Ctor/FS/ctor.fs
```diff
@@ -0,0 +1,99 @@
...
+
+// Code added to show how to initialize Random objects with the
+// same timer value that will produce unique random number sequences.
+let createEngineWithSameTimer () =
+// <Snippet3>
+    let randomEngines = 
+        Array.init 4 (fun i ->
+            Operators.(>>>) DateTime.Now.Ticks i
+            |> int
+            |> Random )
```
baronfel, line 96: There's a lot of syntax here for a new reader, I suggest something a bit simpler-looking:

```suggestion
        [| for i in 0 .. 3 -> Random(int (DateTime.Now.Ticks >>> i)) |]
```

Array comprehensions are pretty straightforward in my opinion, though again this is a stylistic thing and not a functional thing.

baronfel (approved):
Wow what a set of changes!  By and large they look correct, and idiomatic. I left a couple stylistic suggestions, and if you want you could go with using interpolated strings only, but I'm very happy with this overall.  Thank you @albert-du!


Content notes about this type:

> The following example uses the C# lock Statement and the Visual Basic SyncLock statement to ensure that a single random number generator is accessed by 11 threads in a thread-safe manner. Each thread generates 2 million random numbers, counts the number of random numbers generated and calculates their sum, and then updates the totals for all threads when it finishes executing.

It would be good to add a reference and link to the `lock` function for F# here.

> A lock (the lock statement in C# and the SyncLock statement in Visual Basic) protects access to the variables for the total count and sum of all random numbers generated on all threads.

It would be good to have a mention of the `lock` function in F# here as well, though no link is necessary as with the other languages.

> The NextBytes(Byte[], Byte, Byte) method wraps a call to the Next(Int32, Int32) method and specifies the minimum value and one greater than the maximum value (in this case, 0 and 101) that we want returned in the byte array. Because we are sure that the integer values returned by the Next method are within the range of the Byte data type, we can safely cast them (in C#) or convert them (in Visual Basic) from integers to bytes.

Similarly a brief statement about how we are safely casting using `byte` in F#

samples/snippets/fsharp/VS_Snippets_CLR_System/system.Random.Ctor/FS/ctor.fs
albert-du: Yeah I used `Operators.(>>>)` because for some reason the C# version used an `unchecked` context though it shouldn't make a difference, I'll change it.

albert-du:
@baronfel 👍

samples/snippets/fsharp/VS_Snippets_CLR_System/system.Random.Next/FS/next.fs
```diff
@@ -0,0 +1,89 @@
...
+    let randObj = Random seed
+
+    // Generate six random integers from 0 to the upper bound.
+    for _ = 1 to 6 do
+        printf "%11i " (randObj.Next(upper))
+    printfn ""
+
+// Generate random numbers with both bounds specified.
+let bothBoundRandoms seed lower upper =
+    printfn "\nRandom object, seed = %i, lower = %i, upper = %i: " seed lower upper
```
dsyme, line 28: Use interpolation

samples/snippets/fsharp/VS_Snippets_CLR_System/system.Random.Next/FS/next.fs
```diff
@@ -0,0 +1,89 @@
...
+    let randObj = Random seed
+
+    // Generate six random integers from 0 to int.MaxValue.
+    for _ = 1 to 6 do
+        printf "%11i " (randObj.Next())
+    printfn ""
+
+// Generate random numbers with an upper bound specified.
+let upperBoundRandoms seed upper = 
+    printfn "\nRandom object, seed = %i, upper bound = %i" seed upper
```
dsyme, line 18: Use string interpolation

dsyme (approved):
Looks great.  Just a minor request to use string interpolation consistently throughout

albert-du:
ready!

albert-du:
@BillWagner ready

BillWagner (approved):
Thanks again @albert-du 

I'll :shipit: now.

PR #7336, opened 2021-10-29
Fake docs for ephemeral `ApplicationConfiguration` API

RussKie:


xml/FrameworksIndex/windowsdesktop-6.0.xml
```diff
@@ -30087,6 +30087,9 @@
       <Member Id="P:System.Windows.Forms.Application.UseWaitCursor" />
       <Member Id="P:System.Windows.Forms.Application.VisualStyleState" />
     </Type>
+    <Type Name="System.Windows.Forms.ApplicationConfiguration" Id="T:System.Windows.Forms.ApplicationConfiguration" />
+      <Member Id="M:System.Windows.Forms.ApplicationConfiguration.Initialize" />
+    </Type>
```
RussKie, line 30092: 🤔  I'm not 100% sure if this is correct - this `ApplicationConfiguration` is not part of System.Windows.Forms namespace and it doesn't exist in System.Windows.Forms.dll...

PR #7352, opened 2021-11-04
Add F# snippets for System.Action

albert-du:
## Summary

Adds F# code snippets for [System.Action](https://docs.microsoft.com/en-us/dotnet/api/system.action?view=net-5.0)

The behaviors are identical to the other languages 

I didn't translate the anonymous method snippet as they aren't a thing in F# as far as I know

The article itself might benefit from a remark about F# first class functions as an alternative to delegates but I didn't include anything of the sort here

### DIFF
--- a/xml/System/Action.xml
+++ b/xml/System/Action.xml
@@ -62,7 +62,7 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- You can use this delegate to pass a method as a parameter without explicitly declaring a custom delegate. The encapsulated method must correspond to the method signature that is defined by this delegate. This means that the encapsulated method must have no parameters and no return value. (In C#, the method must return `void`. In Visual Basic, it must be defined by the `Sub`…`End Sub` construct. It can also be a method that returns a value that is ignored.) Typically, such a method is used to perform an operation.  
+ You can use this delegate to pass a method as a parameter without explicitly declaring a custom delegate. The encapsulated method must correspond to the method signature that is defined by this delegate. This means that the encapsulated method must have no parameters and no return value. (In C#, the method must return `void`. In F# the function or method must return `unit`. In Visual Basic, it must be defined by the `Sub`…`End Sub` construct. It can also be a method that returns a value that is ignored.) Typically, such a method is used to perform an operation.  
   
 > [!NOTE]
 >  To reference a method that has no parameters and returns a value, use the generic <xref:System.Func%601> delegate instead.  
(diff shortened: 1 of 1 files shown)


eerhardt:
@KathleenDollard - can you review?

albert-du:
@dsyme

eerhardt:
cc @baronfel as well, in case you are interested in reviewing.

samples/snippets/fsharp/VS_Snippets_CLR_System/system.Action/fs/Action.fs
```diff
@@ -0,0 +1,21 @@
+module Action
+
+// <Snippet2>
+open System
+open System.Windows.Forms
+
+type Name(name) =
+    let instanceName = name
```
baronfel, line 8: In F# we don't need the explicit local field set; in `DisplayToConsole` and `DisplayToWindow` we are able to use the `name` directly. I'm guessing that this form what chosen to maintain symmetry with the C# samples? I'm not sure that's necessary to be honest.

samples/snippets/fsharp/VS_Snippets_CLR_System/system.Action/fs/Delegate.fs
```diff
@@ -0,0 +1,21 @@
+module Delegate
+
+// <Snippet1>
+open System.Windows.Forms
+
+type ShowValue = delegate of unit -> unit
+
+type Name(name) =
+    let instanceName = name
```
baronfel, line 9: Same comment here, I think we can be a bit more slim, plus it helps highlight and show off the scoping available to F# classes!

samples/snippets/fsharp/VS_Snippets_CLR_System/system.Action/fs/Lambda.fs
```diff
@@ -0,0 +1,22 @@
+module Lambda
+
+// <Snippet4>
+open System
+open System.Windows.Forms
+
+type Name(name) =
+    let instanceName = name
```
baronfel, line 8: Third time's the charm :)

baronfel (approved):
These all look perfectly good, just a style comment.

Separately, the doc page for Action calls out C# and VB in particular in a few places:

>  (In C#, the method must return void. In Visual Basic, it must be defined by the Sub…End Sub construct. It can also be a method that returns a value that is ignored.) 

This could be expanded to mention the `unit -> unit` relationship you mention in your comment!

> You can also use the Action delegate with anonymous methods in C#, as the following example illustrates. (For an introduction to anonymous methods, see Anonymous Methods.)

It might be nice to have a parallel to lambas in F# and a link here?

dsyme (approved):

albert-du:
@BillWagner this is ready!

BillWagner:
I apologize for the delay @albert-du This dropped off my radar.

Merging now.

BillWagner (approved):
LGTM

PR #7360, opened 2021-11-06
Add ArgumentNullException to AddProvider

gewarren:
Contributes to dotnet/docs#26780.

### DIFF
--- a/xml/Microsoft.Extensions.Logging/LoggerFactory.xml
+++ b/xml/Microsoft.Extensions.Logging/LoggerFactory.xml
@@ -29,7 +29,7 @@
     </Interface>
   </Interfaces>
   <Docs>
-    <summary>Produces instances of <see cref="T:Microsoft.Extensions.Logging.ILogger" /> classes based on the given providers.</summary>
+    <summary>Produces instances of <see cref="T:Microsoft.Extensions.Logging.ILogger" /> classes based on the specified providers.</summary>
     <remarks>To be added.</remarks>
   </Docs>
   <Members>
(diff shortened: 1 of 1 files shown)


tarekgh (approved):

xml/Microsoft.Extensions.Logging/LoggerFactory.xml
```diff
@@ -29,7 +29,7 @@
     </Interface>
   </Interfaces>
   <Docs>
-    <summary>Produces instances of <see cref="T:Microsoft.Extensions.Logging.ILogger" /> classes based on the given providers.</summary>
+    <summary>Produces instances of <see cref="T:Microsoft.Extensions.Logging.ILogger" /> classes based on the specified providers.</summary>
```
eerhardt, line 32: @gewarren @carlossanlop - for some reason I thought the triple-slash doc comments in dotnet/runtime were now the "master". Shouldn't we be updating them instead?

https://github.com/dotnet/runtime/blob/a8ac67ae5373dce4d337012d2e8c875d32ffa638/src/libraries/Microsoft.Extensions.Logging/src/LoggerFactory.cs#L12

safern: They are not yet. We have only moved over 3 projects and we stopped because the porting tool has some issues. However we will try and make it the main story for .NET 7

eerhardt: Is the plan to port these changes to the triple-slash doc comments in the code?

safern: We should with the doc porting tool once that is ready.

PR #7375, opened 2021-11-10
Clarify behavior of OfType() with null values

matthiaslischka:
`OfType<T>` does not take all elements that *can be cast*. It skips `null`.
And consequently the hint that `Cast<T>` would *instead* throw is a bit misleading too.

```c#
new List<string> {"Mango", null, "Orange"}.OfType<string>();
// --> ["Mango", "Orange"]
	
new List<string> {"Mango", null, "Orange"}.Cast<string>();
// --> ["Mango", null, "Orange"]
```

BR Matthias

### DIFF
--- a/snippets/csharp/System.Linq/Enumerable/AggregateTSource/Enumerable.csproj
+++ b/snippets/csharp/System.Linq/Enumerable/AggregateTSource/Enumerable.csproj
@@ -3,7 +3,6 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net8.0</TargetFramework>
-    <LangVersion>11</LangVersion>
   </PropertyGroup>
 
 </Project>


BillWagner:
ping @KathleenDollard

matthiaslischka:
Any updates on that? Not sure what's up with the Snippets 5000 check - not related to the change.

BillWagner:
closing and reopening for a fresh build (I Think the out-of-support versions have been addressed)

BillWagner:
Hi @matthiaslischka 

I looked everything over. I'll approve the changes once we have a clean build. 

The issue with the snippets build is that it's using `netcoreapp2.2`, which is out of support. Can you update that to `net6.0`? These samples would still compile, and then I'll merge it.

gewarren (approved):

PR #7376, opened 2021-11-11
Remove outdated references to Windows 8.x Store

stevemonaco:
## Summary

PipeReader references Windows 8.x Store and "desktop app" in an awkward sense on multiple occasions. This PR changes the wording to a more generic "applications" to also encompass mobile.

### DIFF
--- a/xml/System.IO.Pipes/PipeStream.xml
+++ b/xml/System.IO.Pipes/PipeStream.xml
@@ -1557,7 +1557,7 @@ The pipe handle has not been set. (Did your <see cref="T:System.IO.Pipes.PipeStr
 
 ## Remarks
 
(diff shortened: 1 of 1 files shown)


carlossanlop:
Thank you for your PR! The change makes sense, we just need the Docs team to confirm which is the approved string for the replacement. @gewarren can you please help?

gewarren:
I like this change.

The guidance I got from the Windows team is:

"Windows Store apps" is an old term that effectively refers to "Windows 8.x apps" now. Direct references to the "Windows Store" must now be "Microsoft Store".

But in this case, that would just change the existing text to "in a Windows 8.x app or desktop app", which is probably not inclusive enough.

MSDN-WhiteKnight:
There seems to be a lot of "Windows 8.x Store" mentions in other async IO APIs, for example, in Stream: https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.readasync?view=net-5.0 It might be needed to do a mass update once a replacement term is figured out. The majority of these don't have anything Windows 8 or Windows Store specific.

gewarren (approved):

gewarren:
@MSDN-WhiteKnight Can you create an issue to track that related work?

MSDN-WhiteKnight:
> @MSDN-WhiteKnight Can you create an issue to track that related work?

Done: https://github.com/dotnet/dotnet-api-docs/issues/7406

PR #7381, opened 2021-11-11
Fix typo in HostingAbstractionsHostExtensions.xml

ericmutta:
The following sentence:

> Runs an application and block the calling thread...

Should read:

> Runs an application and **blocks** the calling thread...


tarekgh (approved):

tarekgh:
Thanks @ericmutta

ericmutta:
Many thanks for following up @tarekgh 👍

PR #7384, opened 2021-11-12
Refresh APIs for .NET 6 GA

gewarren:
This refresh includes the following key changes:

- Exclude source generator APIs
- Add 3 assemblies that were previously omitted (System.Numerics.Vectors.dll, System.IO.Compression.Brotli.dll, System.Formats.Cbor.dll)
- Add System.Runtime.dll assembly from the System.Runtime.Experimental package to the extensions moniker

***

 Build.Reason:Manual by Genevieve Warren
 Build.Url:https://apidrop.visualstudio.com/Content%20CI/_build/results?buildId=264012&view=results
 source_repo.branch:net6-ga


xml/System/Int32.xml
```diff
@@ -54,27 +59,165 @@
     <BaseTypeName>System.ValueType</BaseTypeName>
   </Base>
   <Interfaces>
+    <Interface FrameworkAlternate="dotnet-plat-ext-6.0">
```
tannergooding, line 62: These interfaces and the new members that come from them are all `RequiresPreviewFeatures`. Do we have something (even just displaying the attribute) to help display to users the APIs are "preview only" and not available for use in production?

AraHaan:
> This refresh includes the following key changes:
> 
>     * Exclude source generator APIs
Why? People use these apis, why not document them here.

>     * Add assemblies for 3 assemblies where previously they were omitted (System.Numerics.Vectors.dll, System.IO.Compression.Brotli.dll, System.Formats.Cbor.dll)
Did you mean to word this as ``Add 3 assemblies that where previously they were omitted``?

>     * Add System.Runtime.dll assembly from the System.Runtime.Experimental package to the extensions moniker
I wonder why there is no ``net-experimental`` tfm instead of needing to reference them by packages.

eerhardt:
>> This refresh includes the following key changes:
>> 
>>     * Exclude source generator APIs
> Why? People use these apis, why not keem them documented here.

This means to exclude the public APIs in the library that implements the source generator itself. No one calls into the source generator library - just the compiler. External developers don't call methods on the `JsonSourceGenerator` class.

gewarren:
> Did you mean to word this as `Add 3 assemblies that where previously omitted`?

Yes, I did. I've updated it, thanks.

xml/System/Int32.xml
gewarren: Yes, we display the attribute in the syntax box at least:

image

[Preview](https://review.docs.microsoft.com/en-us/dotnet/api/system.iadditionoperators-3?view=dotnet-plat-ext-6.0&branch=pr-en-us-7384)

tannergooding: Does it also display for the methods? Like `struct Int32 { int IAdditionOperators<int, int, int>.operator +(int left, int right); }` ?

tannergooding: Hmmm, it actually looks like those methods aren't visible at all: <https://review.docs.microsoft.com/en-us/dotnet/api/system.int32?view=dotnet-plat-ext-6.0&branch=pr-en-us-7384>

I'd expect them to be under `Explicit Interface Implementations` and have the `RequiresPreviewFeatures` attribute

gewarren:
Two interesting things to note for types like Char:

- Syntax block is crowded now:

  image

- It shows as being applicable to the .NET platform extensions moniker (this is due to the version of System.Runtime.dll that's in the System.Runtime.Experimental package).

  image

Are we okay with this? Or should I try to fix it somehow?

tannergooding:
> Syntax block is crowded now:

Notably, the type only directly implements:
```
public readonly struct Char : IComparable, IComparable<char>, IEquatable<char>, IConvertible, ISpanFormattable
#if FEATURE_GENERIC_MATH
        , IBinaryInteger<char>,
          IMinMaxValue<char>,
          IUnsignedNumber<char>
#endif // FEATURE_GENERIC_MATH
```

I wonder if there would be some benefit to not displaying the transitive closure in the block here and instead only showing the "top level" set of interfaces. The transitive closure could then be displayed elsewhere...

Notably, we also get:
image

Where those `TSelf` and `TOther` are actually `Char`. Likewise, this list is much bigger than strictly necessary. Most of the interfaces listed are "indirectly" implemented and perhaps we could have `Implements` and `Indirectly Implements` or something?

xml/System/Int32.xml
gewarren: It looks like there is some work to do here then. I wonder if we shouldn't published the APIs from that package just yet? cc @huangmin-ms @mimisasouvanh.

I do see that for existing operators and methods that newly implement an interface method in the experimental package, at least that "Implements" information shows up, e.g.: https://review.docs.microsoft.com/en-us/dotnet/api/system.double.op_lessthan?view=dotnet-plat-ext-6.0&branch=pr-en-us-7384

gewarren:
Closing in favor of #7387

xml/System/Int32.xml
gewarren: Logged https://dev.azure.com/ceapex/Engineering/_queries/edit/518460/?triage=true.

huangmin-ms: @gewarren @tannergooding @mimisasouvanh the preview feature - static abstract members in interfaces is not yet supported in mdoc.

PR #7393, opened 2021-11-16
Converted the markdown alert block

JeffInChrist:
To render an alert (Caution, Important, Note, Tip or Warning) on Docs for dotnet contents, we can make it in a tricky way. Just put an alert statement in the following format:
image
Please note that the class value is in upper case.

Related Work Item:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/413290

### DIFF
--- a/xml/System/UriBuilder.xml
+++ b/xml/System/UriBuilder.xml
@@ -1118,7 +1118,7 @@ In either case, the display string does not contain the <xref:System.UriBuilder.
         </remarks>
         <exception cref="T:System.UriFormatException">The <see cref="T:System.UriBuilder" /> instance has a bad password.
 
-<div class="Note"><p>NOTE</p><p>In <a href="/previous-versions/br230232(v=vs.110)">.NET for Windows Store apps</a> or the <a href="/dotnet/standard/cross-platform/cross-platform-development-with-the-portable-class-library">Portable Class Library</a>, catch the base class exception, <a cref="T:System.FormatException" />, instead.</p></div>
+<div class="NOTE"><p>NOTE</p><p>In <see href="/previous-versions/br230232(v=vs.110)">.NET for Windows Store apps</see> or the <see href="/dotnet/standard/cross-platform/cross-platform-development-with-the-portable-class-library">Portable Class Library</see>, catch the base class exception, <see cref="T:System.FormatException" />, instead.</p></div>
         </exception>
       </Docs>
     </Member>


JeffInChrist:
@gewarren I've figured out a workaround to render markdown alerts on Docs. Can you please approve it?

gewarren:
@meganbradley Is this use of HTML okay in the ECMAXML docs for API ref?

JeffInChrist:
@gewarren @meganbradley It occurred to me that I used to resolve a similar issue in native API Ref in the same way. And the note block is working very well on Docs now. Probably it is applicable to dotnet API ref, too. 

Here is the PR for your reference:https://github.com/MicrosoftDocs/win32-pr/pull/1139

gewarren:
@mimisasouvanh Are you okay with using these HTML tags to create Note blocks?

mimisasouvanh:
@gewarren @JeffInChrist If we are to use these HTML tags to create Note blocks in the Ref docs, please document that somewhere so that its consistent elsewhere when needed?

JeffInChrist:
> @gewarren @JeffInChrist If we are to use these HTML tags to create Note blocks in the Ref docs, please document that somewhere so that its consistent elsewhere when needed?

@mimisasouvanh @gewarren Shall we document it on this page? 
https://docs.microsoft.com/en-us/contribute/markdown-reference#alerts-note-tip-important-caution-warning

xml/System/UriBuilder.xml
```diff
@@ -1118,7 +1118,7 @@ In either case, the display string does not contain the <xref:System.UriBuilder.
         </remarks>
         <exception cref="T:System.UriFormatException">The <see cref="T:System.UriBuilder" /> instance has a bad password.
 
-Note: In <see href="/previous-versions/br230232(v=vs.110)">.NET for Windows Store apps</see> or the <see href="/dotnet/standard/cross-platform/cross-platform-development-with-the-portable-class-library">Portable Class Library</see>, catch the base class exception, <see cref="T:System.FormatException" />, instead.
+<div class="NOTE"><p>NOTE</p><p>In <see href="/previous-versions/br230232(v=vs.110)">.NET for Windows Store apps</see> or the <see href="/dotnet/standard/cross-platform/cross-platform-development-with-the-portable-class-library">Portable Class Library</see>, catch the base class exception, <see cref="T:System.FormatException" />, instead.</p></div>
```
gewarren, line 1121: @carlossanlop Do you see any issues with having HTML tags like this in the /// source? For example, if we were to backport this to dotnet/runtime.

PR #7111, opened 2021-09-07
Sort fields by its enum value in LogLevel.

sviatoslav-vilkovych:
## Summary

Instead sorting fields by name, fields of LogLevel sorted by its value.

PR #7399, opened 2021-11-17
fix: sort list by level instead of name

ruant:
## Summary

Almost all other sites documenting logging levels, have them sorted by severity level and not alphabetically.

### DIFF
--- a/xml/Microsoft.Extensions.Logging/LogLevel.xml
+++ b/xml/Microsoft.Extensions.Logging/LogLevel.xml
@@ -25,13 +25,13 @@
     <remarks>To be added.</remarks>
   </Docs>
   <Members>
-    <Member MemberName="Critical">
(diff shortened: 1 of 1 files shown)


tarekgh (approved):

MSDN-WhiteKnight:
These Member tags are auto-generated by monodoc, so this change might be overwritten by next CI update. It was already proposed to reorder enum fields, but rejected: https://github.com/dotnet/dotnet-api-docs/pull/7111

tarekgh:
@MSDN-WhiteKnight when the auto-generation by monodoc occur? Is it auto generated from the source?

MSDN-WhiteKnight:
Auto-generation occurs on each .NET Release, to my understanding. CI update scans reference assemblies for new release and adds/edits XML metadata from added/changed APIs. Not sure whether it processes everything or just files that have API changes in that release.

PR #7414, opened 2021-11-22
Added more helpful text for return value of Remove

JeroenOortwijn:
## Summary
Changed the return value text of `HttpHeaders.Remove()` to be a bit more helpful than "Returns bool."
Largely taken from `List.Remove()`.


ManickaP (approved):
Thanks for the contribution!

JeroenOortwijn:
I just stumbled upon #7402. 😄

It's a strange coincidence that two Dutch guys happen to change the same line in such a small timespan...
(No, I don't know @bramhaag 😉.)

ManickaP:
What the heck?! And I thought I did something wrong when I saw and approved the same PR twice yesterday and it was 2 PRs :rofl:

karelz:
Wow, that is called co-incidence. Thank you both @JeroenOortwijn and @bramhaag!

PR #7417, opened 2021-11-23
Fix #7415 System.Net.Sockets.Socket.Receive code example does not work.

### LINKED ISSUE #7415: Socket.Receive code example does not work
`System.Net.Sockets.Socket` class, `Receive` method document (https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.receive):

> Examples
> 
> The following code example specifies a data buffer, an offset, a size, and a socket flag before receiving data on a connected Socket.
> 
> ```cs
> // Displays sending with a connected socket
> // using the overload that takes a buffer, offset, message size, and socket flags.
> public static int SendReceiveTest4(Socket server)
> {
>     byte[] msg = Encoding.UTF8.GetBytes("This is a test");
>     byte[] bytes = new byte[256];
>     try
>     {
>         // Blocks until send returns.
>         int byteCount = server.Send(msg, 0, msg.Length, SocketFlags.None);
>         Console.WriteLine("Sent {0} bytes.", byteCount);
> 
>         // Get reply from the server.
>         byteCount = server.Receive(bytes, 0, server.Available,
>                                    SocketFlags.None);
> 
>         if (byteCount > 0)
>             Console.WriteLine(Encoding.UTF8.GetString(bytes));
>     }
>     catch (SocketException e)
>     {
>         Console.WriteLine("{0} Error code: {1}.", e.Message, e.ErrorCode);
>         return (e.ErrorCode);
>     }
>     return 0;
> }
> ```

This code example (located in [samples/snippets/csharp/VS_Snippets_Remoting/Socket_Sync_Send_Receive/CS/source.cs](https://github.com/dotnet/dotnet-api-docs/blob/main/samples/snippets/csharp/VS_Snippets_Remoting/Socket_Sync_Send_Receive/CS/source.cs)) does not work. The problem is:

```cs
>         byteCount = server.Receive(bytes, 0, server.Available,
>                                    SocketFlags.None);
```

Here `server.Available` is 0 initially, and it causes `server.Receive` not write anything to the buffer array `bytes`, and return 0. We have to wait until the data receiving is done, then `server.Available` will return the expected number.

The right example to use this `Receive` method should be:

```cs
public static int SendReceiveTest4(Socket server)
{
    try
    {
        byte[] msg = Encoding.UTF8.GetBytes("This is a test");
        // Blocks until send returns.
        int sentTotalLength = server.Send(msg, 0, msg.Length, SocketFlags.None);
        Console.WriteLine("Sent {0} bytes.", sentTotalLength);

        const int bufferLength = 256;
        byte[] received = new byte[bufferLength];
        int receivedTotalLength = 0;
        // Get reply from the server.
        while (true)
        {
            int receivedLength = server.Receive(received, receivedTotalLength, bufferLength, SocketFlags.None);
            if (receivedLength == 0)
            {
                break;
            }

            receivedTotalLength += receivedLength;
            Array.Resize(ref received, receivedTotalLength + bufferLength);
        }

        if (receivedTotalLength > 0)
            Console.WriteLine(Encoding.UTF8.GetString(received, 0, receivedTotalLength));
    }
    catch (SocketException e)
    {
        Console.WriteLine("{0} Error code: {1}.", e.Message, e.ErrorCode);
        return (e.ErrorCode);
    }
    return 0;
}
```

Dixin:
## Summary

Fix `System.Net.Sockets.Socket` class's `Receive` method code example (https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.receive).

The original example does not work because it uses `Socket.Available`, which is 0 initially. You have to wait until the data is received, then `Socket.Available` will be the correct value.

Fixes #7415


samples/snippets/csharp/VS_Snippets_Remoting/Socket_Sync_Send_Receive/CS/source.cs
```diff
@@ -123,20 +123,31 @@ public static int SendReceiveTest3(Socket server)
...
+            const int bufferLength = 256;
+            byte[] received = new byte[bufferLength];
+            int receivedTotalLength = 0;
             // Get reply from the server.
-            byteCount = server.Receive(bytes, 0, server.Available,
-                                       SocketFlags.None);
+            while (true)
+            {
+                int receivedLength = server.Receive(received, receivedTotalLength, bufferLength, SocketFlags.None);
+                if (receivedLength == 0)
```
antonfirsov, line 140: This will keep draining data and resizing the `received` array until the other endpoint closes it's socket, which is not clear from the code. This example now differs from the rest of the samples which intend to demonstrate a single call without deeper context on how do the peers cooperate.

Note that there is also `SendReceiveTest3`, which suffers from the same issue, but can't be refactored in the same way, considering the way it's being used in the `main` method.

Can't we instead change the sample in the following way?

```diff
- byteCount = server.Receive(bytes, 0, server.Available, SocketFlags.None);
+ byteCount = server.Receive(bytes, 0, bytes.Length, SocketFlags.None);
```

Dixin: > Can't we instead change the sample in the following way?

Yes, `bytes.Length` will work.

> This will keep draining data

The original code uses `server.Available`, which I guess intends to drain all data available from server. That's why I created the loop.

scalablecory (approved):
lgtm, thanks.

scalablecory:
@dotnet/docs what needs to be done re: this build failure?

gewarren:
We recently added a requirement for a project file so we can build the snippet. Simply add a .csproj file to the directory with the snippet and add the following content:

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net6.0</TargetFramework>
  </PropertyGroup>

</Project>
```

gewarren (approved):
Thanks a lot @Dixin !

PR #7423, opened 2021-11-24
Correcting the ArgumentNullException.xml string constructor example text

martinstenhoff:
## Summary

Changing arg01 to arg1 to match the "For example" text and correcting the message to match the actually message seen in .NET (not .NET Framework).
In .NET Framework the message is "Value cannot be null.\r\nParameter name: arg1", which also does not match what is currently in the doc as the linefeed newline has been stripped.

### DIFF
--- a/xml/System/ArgumentNullException.xml
+++ b/xml/System/ArgumentNullException.xml
@@ -191,7 +191,7 @@
 |Property|Value|  
 |--------------|-----------|  
 |<xref:System.Exception.InnerException%2A>|A null reference (`Nothing` in Visual Basic).|  
-|<xref:System.ArgumentException.Message%2A>|A localized error message string that identifies the null argument. For example, if the `paramName` argument is "arg1", the English language message string is (using .NET):<br /><br /> `Value cannot be null. (Parameter name: 'arg1')`|  
+|<xref:System.ArgumentException.Message%2A>|A localized error message string that identifies the null argument. For example, if the `paramName` argument is "arg1", the English language message string is:<br /><br />On .NET 5+ and .NET Core: `Value cannot be null. (Parameter name: 'arg1')`<br /><br />On .NET Framework: `Value cannot be null.\r\nParameter name: arg1`|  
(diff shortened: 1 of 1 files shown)


xml/System/ArgumentNullException.xml
```diff
@@ -191,7 +191,7 @@
 |Property|Value|  
 |--------------|-----------|  
 |<xref:System.Exception.InnerException%2A>|A null reference (`Nothing` in Visual Basic).|  
-|<xref:System.ArgumentException.Message%2A>|A localized error message string that identifies the null argument. For example, if the `paramName` argument is "arg1", the English language message string is:<br /><br /> `Value cannot be null.Parameter name: arg01`|  
+|<xref:System.ArgumentException.Message%2A>|A localized error message string that identifies the null argument. For example, if the `paramName` argument is "arg1", the English language message string is (using .NET):<br /><br /> `Value cannot be null. (Parameter name: 'arg1')`|
```
gewarren, line 194:
```suggestion
|<xref:System.ArgumentException.Message%2A>|A localized error message string that identifies the null argument. For example, if the `paramName` argument is "arg1", the English language message string is:<br /><br />On .NET 5+ and .NET Core: `Value cannot be null. (Parameter name: 'arg1')`<br /><br />On .NET Framework: `Value cannot be null.\r\nParameter name: arg1`|  
```

gewarren: It's probably best to include the default message for both .NET Framework and .NET Core/5+.

gewarren (commented):
Thanks for fixing this!

xml/System/ArgumentNullException.xml
martinstenhoff: Yes having both is ideal. I couldn't figure out how one can build the docs site to preview the changes (the PR build preview is only accessible to MS I take it) so I didn't want to make more changes than necessary.

gewarren: Yes, the preview is internal only, unfortunately. I'll check it out and make sure it looks okay.

gewarren (approved):

PR #7441, opened 2021-11-30
Removed exceptions in SerialPort.IsOpen docs

reijerh:
Removed the two exceptions in the docs for property `System.IO.Ports.SerialPort.IsOpen`, as it's a get-only property (the exceptions talk about passing a value).


adamsitnik (approved):
LGTM, thank you for improving the docs @reijerh !

PR #7449, opened 2021-12-02
Default has been changed to 1 minute.

ManickaP:
https://github.com/dotnet/runtime/pull/52687 follow up

### DIFF
--- a/xml/System.Net.Http/SocketsHttpHandler.xml
+++ b/xml/System.Net.Http/SocketsHttpHandler.xml
@@ -684,7 +684,7 @@ For example, if the value is 64, then 65,536 bytes are allowed for the maximum r
       </ReturnValue>
       <Docs>
         <summary>Gets or sets how long a connection can be idle in the pool to be considered reusable.</summary>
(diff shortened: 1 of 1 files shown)


karelz (approved):
I would suggest to switch the order and start with latest .NET version.

xml/System.Net.Http/SocketsHttpHandler.xml
```diff
@@ -684,7 +684,7 @@ For example, if the value is 64, then 65,536 bytes are allowed for the maximum r
       </ReturnValue>
       <Docs>
         <summary>Gets or sets how long a connection can be idle in the pool to be considered reusable.</summary>
-        <value>The maximum idle time for a connection in the pool. The default value for this property is 2 minutes.</value>
+        <value>The maximum idle time for a connection in the pool. The default value for this property is 2 minutes in .NET Core and .NET 5; from .NET 6 on, the default value is 1 minute.</value>
```
karelz, line 687:
```suggestion
        <value>The maximum idle time for a connection in the pool. The default value for this property is 1 minute in .NET 6 and later versions; the default value is 2 minutes in .NET Core and .NET 5</value>
```

karelz (approved):
I would suggest to switch the order and start with latest .NET version.

PR #7450, opened 2021-12-02
BeginReceive --> BeginSendTo in remarks for EndSendTo

Regelink:
## Summary

Replaced in the remarks for System.Net.Sockets.Socket.EndSendTo: "... is called by the system after BeginReceive returns" with "... is called by the system after BeginSendTo returns"


antonfirsov (approved):
Thanks! 👍

scalablecory (approved):

PR #7457, opened 2021-12-05
System.Web: Demote .NET Framework

yecril71pl:
## Summary

The description for the namespace was carried over from version 4.8 but the mentioned items do not exist any more.


gewarren (dismissed):
This namespace description is not versioned, so it's shared with all versions. For that reason, I think we should keep the old info, but add something about the differences between .NET Framework versions and .NET 5+ (Core) versions.

https://docs.microsoft.com/en-us/dotnet/api/system.web?view=netframework-4.8

gewarren (approved):

PR #7466, opened 2021-12-08
Fix Type.Namespace remarks

### LINKED ISSUE #7340: `Type.Namespace` property doesn't seem to ever return null if instance represents a generic parameter
Documentation of [`Type.Namespace` property](https://docs.microsoft.com/en-us/dotnet/api/system.type.namespace?view=net-5.0#remarks) states:

> If the current Type represents a constructed generic type, this property returns the 
namespace that contains the generic type definition. Similarly, **if the current Type represents a generic parameter T** this property returns the namespace that contains the generic type definition that defines T.
>
> **If the current Type object represents a generic parameter**, this property returns null.

Initially I raised this issue on StackOverflow [here](https://stackoverflow.com/questions/69786564/when-does-type-namespace-return-null-if-instance-represents-a-generic-parameter). This post [answer](https://stackoverflow.com/a/69786923/2304432) and [comment](https://stackoverflow.com/questions/69786564/when-does-type-namespace-return-null-if-instance-represents-a-generic-parameter#comment123356287_69786923) strenghtened my believe this might be an issue in documentation.

The code that demonstrates the cases where .NET framework (.NET 5.0) itself says a type is generic parameter via `Type.IsGenericParameter property`, but actually returns a namespace:

```
using System;

namespace TypeDemo
{
    internal static class Program
    {
        public static void Main()
        {
            Console.WriteLine(typeof(Foo.C<>).GetGenericArguments()[0].IsGenericParameter);
            Console.WriteLine(typeof(Foo.C<>).GetGenericArguments()[0].Namespace);
            Console.WriteLine(typeof(Foo.C).GetMethod("Foo").GetGenericArguments()[0].IsGenericParameter);
            Console.WriteLine(typeof(Foo.C).GetMethod("Foo").GetGenericArguments()[0].Namespace);
        }
    }
}

namespace Foo
{
    class C<T>
    {
        public void Foo()
        {
        }
    }

    class C
    {
        public void Foo<T>()
        {
        }
    }
}
```

The statements in bold from cited documentation seem to contradict each other, and I think the second one in bold is not correct. Or I got it wrong and is there actually a case when `Type.Namespace` returns null if `Type` object represents a generic parameter?

MSDN-WhiteKnight:
Fix conditions when property returns null

@buyaa-n for review

Fixes #7340

### DIFF
--- a/xml/System/Type.xml
+++ b/xml/System/Type.xml
@@ -15587,7 +15587,7 @@ If the value of this property is `true` for an array type, it can be used to cre
   
  If the current <xref:System.Type> represents a constructed generic type, this property returns the namespace that contains the generic type definition. Similarly, if the current <xref:System.Type> represents a generic parameter `T`, this property returns the namespace that contains the generic type definition that defines `T`.  
   
- If the current <xref:System.Type> object represents a generic parameter and a generic type definition is not available, such as for a signature type returned by <xref:System.Type.MakeGenericMethodParameter%2A> method, this property returns `null`.
+ If the current <xref:System.Type> object represents a generic parameter and a generic type definition is not available, such as for a signature type returned by <xref:System.Type.MakeGenericMethodParameter%2A>, this property returns `null`.
   
    
   


buyaa-n (approved):
LGTM, thank you @MSDN-WhiteKnight!

xml/System/Type.xml
```diff
@@ -15587,7 +15587,7 @@ If the value of this property is `true` for an array type, it can be used to cre
   
  If the current <xref:System.Type> represents a constructed generic type, this property returns the namespace that contains the generic type definition. Similarly, if the current <xref:System.Type> represents a generic parameter `T`, this property returns the namespace that contains the generic type definition that defines `T`.  
   
- If the current <xref:System.Type> object represents a generic parameter, this property returns `null`.  
+ If the current <xref:System.Type> object represents a generic parameter and a generic type definition is not available, such as for a signature type returned by <xref:System.Type.MakeGenericMethodParameter%2A> method, this property returns `null`.
```
gewarren, line 15590:
```suggestion
 If the current <xref:System.Type> object represents a generic parameter and a generic type definition is not available, such as for a signature type returned by <xref:System.Type.MakeGenericMethodParameter%2A>, this property returns `null`.
```

PR #7476, opened 2021-12-09
Remove version-dependant marker for ArgumentException thrown by Copy methods

miloszkukla:
The code for .NET 6: https://source.dot.net/#System.Private.CoreLib/File.cs,57 indicates that ArgumentException will be thrown for zero-lenght strings so the information seems to be invalid. Other methods in this class also contain this information but I didn't check whether it's a case for them.


adamsitnik (approved):
LGTM, thank you @miloszkukla !

PR #7487, opened 2021-12-11
Remove float exception for x.Equals(x)

svick:
Fixes https://github.com/dotnet/dotnet-api-docs/issues/771.

I couldn't find any case where `x.Equals(x)` would return `false` for floating point types. I have checked this for `float` (and `Half`) using the following code:

```c#
using System;

int i = int.MinValue;
do
{
    var f = BitConverter.Int32BitsToSingle(i);
    if (!f.Equals(f))
        Console.WriteLine(f);
} while (i++ < int.MaxValue);
```

I expect this to also be the case for `double`.

Note that `float.NaN == float.NaN` will return `false`, but even then `float.NaN.Equals(float.NaN)` is still `true`.

In case I'm wrong and there is some case where `x.Equals(x)` returns `false` for floating point types, I think this paragraph needs to be clarified: a link to an outdated version of a 70-page paid document is not enough.

### DIFF
--- a/xml/System/Object.xml
+++ b/xml/System/Object.xml
@@ -270,7 +270,7 @@
   
  The following statements must be true for all implementations of the <xref:System.Object.Equals%28System.Object%29> method. In the list, `x`, `y`, and `z` represent object references that are not **null**.  
   
--   `x.Equals(x)` returns `true`, except in cases that involve floating-point types. See [ISO/IEC/IEEE 60559:2011, Information technology -- Microprocessor Systems -- Floating-Point arithmetic](https://www.iso.org/standard/57469.html).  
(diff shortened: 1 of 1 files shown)


BillWagner:
Tagging @tannergooding for review.

tannergooding:
For `float`/`double` themselves this is correct.

However, there are still many user-defined types involving `float` that may use `==` accidentally and break the contract here.

BillWagner (approved):
Thanks @svick and @tannergooding 

I'll approve this and merge it.  

We scope our docs to the types that are part of the .NET product. In addition, these rules provide guidance for type authors implementing `==`.

PR #7492, opened 2021-12-15
Update Math.xml

RVerheggen:
Typing error in Note: availble -> available

## Summary

There was a typing error in the first note (about MidpointRounding) : `availble` is changed to `available`


gewarren (approved):
Thank you @RVerheggen!

PR #7496, opened 2021-12-16
Clarify Visible property

### LINKED ISSUE #7458: Explanation of Control.Visible attribute is confuse or incorrect !
Hello,

on [docs.microsoft.com](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.visible?view=windowsdesktop-6.0#System_Windows_Forms_Control_Visible) page, explanation of Visible=True is (for me) not correct.

For information, it is written

```
Property Value
Boolean
true if the control and all its child controls are displayed; otherwise, false. 
```

CHILD must be replaced by PARENT or CONTAINER !!!

I have written an answer on StackOverflow describing this issue (my username is "schlebe").

[When is Control.Visible = true turns out to be false?](https://stackoverflow.com/questions/6061535/when-is-control-visible-true-turns-out-to-be-false/70241950#70241950
)

I know that I can change documentation myself (I think) but I'm not sure (at 100%) that my proposal is correct because my natural language is french and not english.

For me a child control is a control contained in another control that is a container. And this control/container is parent of child control.

gewarren:
Fixes #7458. Source code is [here](https://github.com/dotnet/winforms/blob/76dedba0e22119fb170098b51911a83ab004c1ee/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs#L3869).

### DIFF
--- a/xml/System.Windows.Forms/Control.xml
+++ b/xml/System.Windows.Forms/Control.xml
@@ -19868,14 +19868,16 @@ Note: This property will always return <see langword="true" /> for an instance o
       <Docs>
         <summary>Gets or sets a value indicating whether the control and all its child controls are displayed.</summary>
         <value>
-          <see langword="true" /> if the control and all its child controls are displayed; otherwise, <see langword="false" />. The default is <see langword="true" />.</value>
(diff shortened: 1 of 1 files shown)


gewarren:
@RussKie Could you check this one too please?

RussKie (approved):

RussKie:
Thank you

Issue #7451, opened 2021-12-02
Add See Also links to API ref content


gewarren:
We've heard feedback that code examples for newer APIs exist mostly in conceptual docs. But web searches don't always produce those conceptual docs. In general, API ref is lacking code examples. This issue aims to address this problem with a first step of adding See Also links from high page view (but low CSAT) API ref articles to conceptual articles that have xref links to the API.

As a next step, we can look at expanding the See Also links to include links to:

- articles with code snippets that use the API
- code samples that use the API

gewarren:
Related: #3369

gewarren:
Of the APIs with page views > 5000 per month and bounce rate > 7.5%, the only one without an example is JsonSerializer class. However, that page does have a remarks section with a link to the conceptual docs. I've added a See also link as well.

gewarren:
New(er) APIs that could benefit with links/examples/remarks:

- [ ] Microsoft.Win32.SafeHandles
  - [ ] SafeFileHandle
  - [ ] SafePipeHandle
  - [ ] SafeProcessHandle
  - [ ] SafeWaitHandle 
- [ ] File.OpenHandle
- [ ] Environment.ProcessId
- [ ] Environment.ProcessPath
- [ ] RandomNumberGenerator.GetBytes(Int32)
- [ ] Parallel.ForEachAsync
- [ ] ConfigurationExtensions.GetRequiredSection
- [ ] Enumerable.Chunk/MaxBy/MinBy/DistinctBy/ExceptBy/IntersectBy/UnionBy - can link to examples added [here](https://github.com/dotnet/docs/pull/25979)
- [ ] BitOperations.RoundUpToPowerOf2
- [ ] BitOperations.IsPow2
- [ ] Task.WaitAsync
- [ ] ArgumentNullException.ThrowIfNull
- [ ] NativeMemory class
- [ ] System.Diagnostics.Metrics - entire namespace
- [ ] PeriodicTimer class
- [ ] IntPtr.Parse and UIntPtr.Parse (also TryParse and TryFormat and methods like Math.Abs(IntPtr))
- [ ] BitConverter.ToHalf
- [ ] Random.NextInt64
- [ ] ActivityStatusCode
- [ ] Uri.UriScheme*
- [ ] StringInfo.GetNextTextElementLength
- [ ] BrotliStream.ReadByte and BrotliStream.WriteByte
- [ ] ZLibStream
- [ ] Socket - Async methods like ConnectAsync
[...]

PR #7510, opened 2021-12-20
Add see also links to conceptual docs

gewarren:
Contributes to #7451.

### DIFF
--- a/xml/System.Diagnostics.Metrics/ObservableGauge`1.xml
+++ b/xml/System.Diagnostics.Metrics/ObservableGauge`1.xml
@@ -58,7 +58,7 @@ This class supports only the following generic parameter types: <xref:System.Byt
       <Parameters />
       <Docs>
         <summary>Fetches the current measurements being tracked by this observable counter.</summary>
-        <returns>To be added.</returns>
+        <returns>An enumeration of measurements of type `<typeparamref name="T"/>` that are being tracked by this observable counter.</returns>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>


xml/System.Diagnostics.Metrics/ObservableGauge`1.xml
```diff
@@ -57,7 +57,7 @@ This class supports only the following generic parameter types: <xref:System.Byt
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>Observe() fetches the current measurements being tracked by this observable counter.</summary>
+        <summary>Fetches the current measurements being tracked by this observable counter.</summary>
         <returns>To be added.</returns>
```
carlossanlop, line 61: Any objections to adding the missing returns value, since we're already here?:

```suggestion
        <returns>An enumeration of measurements of type `<typeparamref name="T"/>` that are being tracked by this observable counter.</returns>
```

carlossanlop (approved):
The proposed changes LGTM. I just added a missing remarks, if you don't mind including it in this PR. Feel free to reword it if necessary.

xml/System.Diagnostics.Metrics/ObservableGauge`1.xml
gewarren: Since `T` isn't listed as a `typeparam`, presumably because it's not specified by the caller, do you think it's okay to use `typeparamref`?

gewarren: I guess because this is a sealed class, these custom comments don't show up anyway. Only inherited members are shown.

https://review.docs.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.observablegauge-1?view=net-6.0&branch=pr-en-us-7510

PR #7513, opened 2021-12-21
Stream.Length does not always throw for non-seekable streams

adamsitnik:
Fixes https://github.com/dotnet/runtime/issues/63040

### DIFF
--- a/xml/System.IO/Stream.xml
+++ b/xml/System.IO/Stream.xml
@@ -1677,7 +1677,7 @@
         <summary>When overridden in a derived class, gets the length in bytes of the stream.</summary>
         <value>A long value representing the length of the stream in bytes.</value>
         <remarks>To be added.</remarks>
-        <exception cref="T:System.NotSupportedException">A class derived from <see langword="Stream" /> does not support seeking.</exception>
+        <exception cref="T:System.NotSupportedException">A class derived from <see langword="Stream" /> does not support seeking and the length is unknown.</exception>
         <exception cref="T:System.ObjectDisposedException">Methods were called after the stream was closed.</exception>
         <related type="Article" href="/dotnet/standard/io/">File and Stream I/O</related>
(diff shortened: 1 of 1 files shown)


xml/System.IO/Stream.xml
```diff
@@ -1677,7 +1677,7 @@
         <summary>When overridden in a derived class, gets the length in bytes of the stream.</summary>
         <value>A long value representing the length of the stream in bytes.</value>
         <remarks>To be added.</remarks>
-        <exception cref="T:System.NotSupportedException">A class derived from <see langword="Stream" /> does not support seeking.</exception>
+        <exception cref="T:System.NotSupportedException">A class derived from <see langword="Stream" /> does not support seeking and the length is unknown.</exception>
```
carlossanlop, line 1680: @adamsitnik Just a quick clarification: I assume this exception is referring to the _default_ behavior of a stream that derives from `Stream`, with no overriden behaviors?

I ask because from the issue discussion, the stream that is involved in a `ZipArchive` with an unseekable stream is `SubReadStream`, which is an internal stream that _does not_ allow setting a position, but allows getting it because _we overrode_ the `Position` property:

https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.IO.Compression/src/System/IO/Compression/ZipCustomStreams.cs#L250-L264

```cs
        public override long Position
        {
            get
            {
                ThrowIfDisposed();
                return _positionInSuperStream - _startInSuperStream;
            }
            set
            {
                ThrowIfDisposed();
                throw new NotSupportedException(SR.SeekingNotSupported);
            }
        }
```

carlossanlop (approved):
Assuming the description is good enough to explain that the exception was thrown because the user derived a stream without overriding `Position`, then LGTM.

gewarren:
Is this good to merge?

stephentoub (approved):

PR #7521, opened 2021-12-23
Replace method name Ceiling -> Floor

Adriansillo:
The correct method in this doc is Floor and it had Celling originally

## Summary

This PR updates method name for 

FROM

Dim d1 As Double = 1043.75133
Dim i1 As Integer = CInt(Math.Ceiling(d1))        ' Result: 1043

Dim d2 As Double = 7968.4136
Dim i2 As Integer = CInt(Math.Ceiling(d2))        ' Result: 7968

TO 

Dim d1 As Double = 1043.75133
Dim i1 As Integer = CInt(Math.`Floor`(d1))        ' Result: 1043

Dim d2 As Double = 7968.4136
Dim i2 As Integer = CInt(Math.`Floor`(d2))        ' Result: 7968

### DIFF
--- a/xml/System/Math.xml
+++ b/xml/System/Math.xml
@@ -3277,10 +3277,10 @@ Starting with Visual Basic 15.8, the performance of Double-to-integer conversion
    
 ```vb
 Dim d1 As Double = 1043.75133
-Dim i1 As Integer = CInt(Math.Ceiling(d1))        ' Result: 1043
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks for making these changes @Adriansillo 

We appreciate it.

I've reviewed the changes and I'll :shipit: now.

PR #7529, opened 2021-12-25
Binary Reader sample: using should protect File.Open too

yecril71pl:
## Summary

If a new reader could not be constructed from a stream, the stream itself would leak.


gewarren:
Thanks @yecril71pl! Would you mind also adding a simple project file to the directory, similar to https://github.com/dotnet/dotnet-api-docs/blob/main/samples/snippets/csharp/api/system.io/path/combine2/combine.csproj?

jeffhandley:
@gewarren I think this is ready for re-review.

BillWagner (approved):
Thanks @yecril71pl This looks great, and I'll :shipit: now.

Thanks again.

PR #7551, opened 2022-01-05
BinaryReader.Dispose releases unless configured otherwise

yecril71pl:
## Summary

`BinaryReader.Dispose` releases the `BaseStream` unless configured otherwise.


adamsitnik (approved):
LGTM, thank you @yecril71pl !

PR #7556, opened 2022-01-06
Fix issue: AbandandMutexException for .net 6

TsuyoshiUshio:
## Summary

The Mutex sample doesn't work.  I investigate the issue. 
The .NET Core and .NET6 , and .Net Framework behave differently. 

looks　the Main() method thread is different from the thread executes L8. 
this fix is one of the examples. The other fix will be creating a static constructor, then the thread runs Main() and private static Mutex mut = new Mutex(true); thread will become the same in .NET 6.

### DIFF
--- a/samples/snippets/csharp/VS_Snippets_CLR_System/system.Threading.Mutex 1Arg Ctor Example/CS/class1.csproj
+++ b/samples/snippets/csharp/VS_Snippets_CLR_System/system.Threading.Mutex 1Arg Ctor Example/CS/class1.csproj
@@ -1,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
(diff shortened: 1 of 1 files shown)


buyaa-n (approved):
Thanks @TsuyoshiUshio, the fix looks good to me, but seems you need to add a project file (seems doc samples now require project files something like [this](https://github.com/dotnet/dotnet-api-docs/blob/main/samples/snippets/csharp/VS_Snippets_Remoting/ClassicTcpClient.PublicMethodsAndPropertiesExample/CS/TcpClient.csproj))

kouvel (approved):
LGTM, thanks! Looks like the build requires a project file as mentioned above, seems like it would satisfy the build to copy the project file linked above alongside this file and name it class1.csproj.

Issue #7554, opened 2022-01-06
Reorganize C# snippets based on API name


gewarren:
Organize C# snippets into subfolders based on namespace and type.

Folders to organize:

- [x] samples/snippets/csharp/api
- [x] samples/snippets/csharp/VS_Snippets_CLR
- [x] samples/snippets/csharp/VS_Snippets_CLR_Classic
- [x] samples/snippets/csharp/VS_Snippets_CLR_System
- [x] samples/snippets/csharp/VS_Snippets_CodeAnalysis
- [x] samples/snippets/csharp/VS_Snippets_Data
- [x] samples/snippets/csharp/VS_Snippets_Misc
- [x] samples/snippets/csharp/VS_Snippets_Remoting
- [x] samples/snippets/csharp/VS_Snippets_VBCSharp

Consider for later:

- [x] samples/snippets/csharp/VS_Snippets_WinForms
- [x] samples/snippets/csharp/VS_Snippets_WPF

PR #7573, opened 2022-01-11
Reorganize System.Diagnostics snippets

gewarren:
Contributes to #7554. It's too much work to add projects and make them buildable, so for now, I'm just reorganizing the snippets.

### DIFF
--- a/samples/snippets/csharp/System.Diagnostics/Trace/Assert/source.cs
+++ b/samples/snippets/csharp/System.Diagnostics/Trace/Assert/source.cs
@@ -1,6 +1,4 @@
-﻿using System;
-using System.Data;
-using System.Diagnostics;
+﻿using System.Diagnostics;
 
 // <Snippet1>
 // Create an index for an array.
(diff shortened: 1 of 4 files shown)


samples/snippets/csharp/System.Diagnostics/BooleanSwitch/Overview/remarks.cs
```diff
@@ -1,21 +1,21 @@
-﻿using System;
-using System.Diagnostics;
-
-public class SomeClass
-{
-// <Snippet2>
```
safern, line 6: Overview has two Main methods, so I guess compiling this code snippet will fail?

gewarren: Yes, it would fail. Adding the project files and fixing compilation errors will have to happen in a separate PR. There's a lot of manual effort involved in that.

BillWagner (approved):
Thanks @gewarren 

I'm fine with shipping this.

(I think you'll need to reconfigure the snippets 5000 check as not required, then reconfigure it after you merge)

PR #7574, opened 2022-01-11
Update SortedList`2.xml

NetMage:
Corrected remarks for RemoveAt.

## Summary

Removed inaccurate discussion of binary search in Remarks.

### DIFF
--- a/xml/System.Collections.Generic/SortedList`2.xml
+++ b/xml/System.Collections.Generic/SortedList`2.xml
@@ -1445,7 +1445,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- This method performs a binary search; however, the elements are moved up to fill in the open spot, so this method is an O(`n`) operation, where `n` is <xref:System.Collections.Generic.SortedList%602.Count%2A>.  
+ The elements are moved up to fill in the open spot, so this method is an O(`n`) operation, where `n` is <xref:System.Collections.Generic.SortedList%602.Count%2A>.  
   
  ]]></format>
         </remarks>


xml/System.Collections.Generic/SortedList`2.xml
```diff
@@ -1445,7 +1445,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- This method performs a binary search; however, the elements are moved up to fill in the open spot, so this method is an O(`n`) operation, where `n` is <xref:System.Collections.Generic.SortedList%602.Count%2A>.  
+ The elements are moved up to fill in the open spot, so this method is an O(`n`) operation, where `n` is <xref:System.Collections.Generic.SortedList%602.Count%2A>.
```
buyaa-n, line 1448: > Removed inaccurate discussion of binary search in Remarks.

I think it is trying to say that it does binary search for finding the element, however after removal elements are moved up to fill in the open spot causing it run `O(n)`, so i don't think `binary search` part should be removed

NetMage: What does RemoveAt use to search for as the element?

buyaa-n: I do not know fo sure, from the remarks above i would think it uses binary search, that is why i do not like to remove that part, but i will let the area owners review this @eiriktsarpalis @krwq

eiriktsarpalis: It clearly isn't using binary search:

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Collections/src/System/Collections/Generic/SortedList.cs#L700-L719

So this change looks good to me.

PR #7578, opened 2022-01-13
Reorganize snippets - VS_Snippets_CLR_Classic

gewarren:
Contributes to #7554. This PR organizes the remainder of the /samples/snippets/csharp/VS_Snippets_CLR_Classic/ directory.

Snippets 5000 fails because most directories don't have project files. That work (adding projects and fixing build errors) will happen at a later point.


ericstj:
@gewarren what sort of review were you looking for here?  If these are all just moves I think it should be OK.  If not can you call out what you'd like reviewers to do?

gewarren:
> @gewarren what sort of review were you looking for here? If these are all just moves I think it should be OK. If not can you call out what you'd like reviewers to do?

@ericstj Basically it just needs an approval to be able to merge it. It's just a move - none of the code changed.

ericstj (approved):

PR #7580, opened 2022-01-13
Update to use mdoc 5.8.6

gewarren:
Baseline to new mdoc version 5.8.6.

Two main changes:

1) `in` parameters are now shown correctly instead of as `ref`.
2) Types are no longer shown as applicable to a framework version if they're forwarded to an assembly in a different framework.


BillWagner (approved):
This is a great step forward.

It LGTM @gewarren 

Let's :shipit:

PR #7583, opened 2022-01-14
Remove unneeded package reference

MSDN-WhiteKnight:
I've made a mistake when creating project files for these samples, System.Reflection.Metadata is inbox on all .NET Core versions and it's unneeded to reference it as package.


BillWagner (approved):
Thanks @MSDN-WhiteKnight 

I've reviewed the changes and the build, and I'll :shipit: now.

PR #7594, opened 2022-01-15
FileInfo.OpenRead: Unauthorized: it is secret

yecril71pl:
## Summary

A secret file cannot be opened for reading.

### DIFF
--- a/xml/System.IO/FileInfo.xml
+++ b/xml/System.IO/FileInfo.xml
@@ -1948,7 +1948,7 @@ The following example demonstrates moving a file to a different location and ren
  ]]></format>
         </remarks>
         <exception cref="T:System.UnauthorizedAccessException">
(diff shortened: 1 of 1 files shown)


MSDN-WhiteKnight:
What is exactly a secret file? Shouldn't this be something like "current user does not have read permissions for the file" instead, as i'd expect from exception name?

yecril71pl:
> What is exactly a secret file? Shouldn't this be something like "current user does not have read permissions for the file" instead, as i'd expect from exception name?

A secret file is a file most users cannot read.  We can replace *secret* with *classified*, we can also recite your full-blown definition if you prefer.  Please tell me what to do.

MSDN-WhiteKnight:
It's up to maintainers / area owners to decide on the final form of changes that will get merged. I personally think that "secret" or "classified" are not common terms and might be confusing for documentation.

PR #7601, opened 2022-01-17
Update StackTraceHiddenAttribute.xml

SteveDunn:
## Summary

Changed `public` to `internal` to match current visibility.


gfoidl:
Actually it is [`public`](https://github.com/dotnet/runtime/blob/ae2f60c950cc021921fce83c796cbcb5ff96c658/src/libraries/System.Private.CoreLib/src/System/Diagnostics/StackTraceHiddenAttribute.cs#L11) since https://github.com/dotnet/runtime/pull/42632 merged on 24 Sep 2020 (so for .NET 6 onwards).

Did you look at a former version of .NET?

MSDN-WhiteKnight:
And in any case members metadata (anything except `<Docs>` tags) should not be edited manually, because it's autogenerated by monodoc.

PR #7617, opened 2022-01-20
Update Directory.Move remarks

gewarren:
Added clarification that the old directory is deleted.

### DIFF
--- a/xml/System.IO/Directory.xml
+++ b/xml/System.IO/Directory.xml
@@ -4015,7 +4015,7 @@ An I/O error occurred.</exception>
 
 This method creates a new directory with the name specified by `destDirName` and moves the contents of `sourceDirName`, including files and directories, to the newly created destination directory. It then deletes the `sourceDirName` directory.
 
-If you try to move a directory to a directory that already exists, an <xref:System.IO.IOException> will occur. For example, an exception will occur if you try to move c:\mydir to c:\public, and c:\public already exists. Alternatively, you could specify "c:\\\public\\\mydir" as the `destDirName` parameter, provided that "mydir" does not exist under "c:\\\public", or specify a new directory name such as "c:\\\newdir".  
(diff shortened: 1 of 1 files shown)


xml/System.IO/Directory.xml
```diff
@@ -4006,27 +4006,29 @@ An I/O error occurred.</exception>
...
         <summary>Moves a file or a directory and its contents to a new location.</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
-## Remarks  
- This method creates a new directory with the name specified by `destDirName` and moves the contents of `sourceDirName` to the newly created destination directory. If you try to move a directory to a directory that already exists, an <xref:System.IO.IOException> will occur. For example, an exception will occur if you try to move c:\mydir to c:\public, and c:\public already exists. Alternatively, you could specify "c:\\\public\\\mydir" as the `destDirName` parameter, provided that "mydir" does not exist under "c:\\\public", or specify a new directory name such as "c:\\\newdir".  
+## Remarks
+
+This method creates a new directory with the name specified by `destDirName` and moves the contents of `sourceDirName`, including files and directories, to the newly created destination directory. It then deletes the `sourceDirName` directory.
```
adamsitnik, line 4016: I am not a native speaker, but to me the best word that describes what this method does is "rename".

```suggestion
The method renames the directory using provided paths. It creates a new directory with the name specified by `destDirName` and moves the contents of `sourceDirName`, including files and directories, to the newly created destination directory. It then deletes the `sourceDirName` directory.
```

@gewarren what do you think?

gewarren: @adamsitnik I don't see it as a straight rename. For example, you could move c:\mydir to c:\sub1\newfolder, in which case it's moved *and* renamed.

adamsitnik (approved):
LGTM, thank you @gewarren !

Issue #7625, opened 2022-01-24
FileSystemWatcher docs should describe the inotify limits


adamsitnik:
The `FileSystemWatcher` users (including majority of ASP.NET Core users) are frequently getting an error similar to:

> The configured user limit ($ABC) on the number of inotify instances has been reached

The [remarks section](https://docs.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher?view=net-6.0#remarks) of the `FileSystemWatcher` type should contain an explanation how to properly modify this limit, as it's not obvious that the settings stored in the "/proc/sys/fs/inotify/max_user_watches" file are not persisted (https://github.com/dotnet/runtime/pull/62096#issuecomment-981662587)

Related:

* https://stackoverflow.com/questions/43469400/asp-net-core-the-configured-user-limit-128-on-the-number-of-inotify-instance
* https://github.com/dotnet/runtime/pull/62096

yecril71pl:
> The [remarks section](https://docs.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher?view=net-6.0#remarks) of the `FileSystemWatcher` type should contain an explanation how to properly modify this limit, as it's not obvious that the settings stored in the "/proc/sys/fs/inotify/max_user_watches" file are not persisted ([dotnet/runtime#62096 (comment)](https://github.com/dotnet/runtime/pull/62096#issuecomment-981662587))

duplicate of #7409

PR #7627, opened 2022-01-24
FSWatcher.EnableRaisingEvents doc += IOException

### LINKED ISSUE #7409: FileSystemWatcher. EnableRaisingEvents throws IOException
&hellip; and it is &rarr; [undocumented](https://docs.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher.enableraisingevents).

> Wystąpił wyjątek: CLR/System.IO.IOException
> Wystąpił nieobsługiwany wyjątek typu „System.IO.IOException” w System.IO.FileSystem.Watcher.dll: 'The configured user limit (128) on the number of inotify instances has been reached, or the per-process limit on the number of open file descriptors has been reached.'
>    w System.IO.FileSystemWatcher.StartRaisingEvents()
>    w System.IO.FileSystemWatcher.StartRaisingEventsIfNotDisposed()
>    w System.IO.FileSystemWatcher.set_EnableRaisingEvents(Boolean value)

yecril71pl:
## Summary

1. Add [IOException](https://docs.microsoft.com/en-us/dotnet/api/system.io.ioexception) to the list of exceptions thrown by setting [FileSystemWatcher.EnableRaisingEvents](https://docs.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher.enableraisingevents#system-io-filesystemwatcher-enableraisingevents).
2. Explain how volatile it is and how to handle it (fingers crossed).

Fixes #7409, #7625 and dotnet/AspNetCore.Docs#19814.

### DIFF
--- a/xml/System.IO/FileSystemWatcher.xml
+++ b/xml/System.IO/FileSystemWatcher.xml
@@ -862,7 +862,7 @@ handle_mui_error 'The configured user limit (128) on the number of inotify insta
         <exception cref="T:System.ObjectDisposedException">The <see cref="T:System.IO.FileSystemWatcher" /> object has been disposed.</exception>
         <exception cref="T:System.PlatformNotSupportedException">The current operating system is not Microsoft Windows NT or later.</exception>
         <exception cref="T:System.IO.FileNotFoundException">The directory specified in <see cref="P:System.IO.FileSystemWatcher.Path" /> could not be found.</exception>
-        <exception cref="T:System.IO.IOException">Unix-based operating systems only: The configured user limit on the number of <c>inotify</c> instances has been reached, or the per-process limit on the number of open file descriptors has been reached.</exception>
+        <exception cref="T:System.IO.IOException">Linux-based operating systems only: The configured user limit on the number of <c>inotify</c> instances has been reached, or the per-process limit on the number of open file descriptors has been reached.</exception>
         <exception cref="T:System.ArgumentException">
           <see cref="P:System.IO.FileSystemWatcher.Path" /> has not been set or is invalid.</exception>
       </Docs>


xml/System.IO/FileSystemWatcher.xml
```diff
@@ -830,13 +830,42 @@
   
  :::code language="cpp" source="~/samples/snippets/cpp/VS_Snippets_CLR_Classic/classic NotifyFilters Example/CPP/source.cpp" id="Snippet1":::
  :::code language="csharp" source="~/samples/snippets/csharp/System.IO/FileSystemWatcher/Overview/source.cs" id="Snippet1":::
- :::code language="vb" source="~/samples/snippets/visualbasic/VS_Snippets_CLR_Classic/classic NotifyFilters Example/VB/source.vb" id="Snippet1":::  
+ :::code language="vb" source="~/samples/snippets/visualbasic/VS_Snippets_CLR_Classic/classic NotifyFilters Example/VB/source.vb" id="Snippet1":::
+
+## Troubleshooting
```
MSDN-WhiteKnight, line 835: Better place this in Remarks section above example, it's not in API Docs convention to create third `Troubleshooting` section

xml/System.IO/FileSystemWatcher.xml
```diff
@@ -830,13 +830,42 @@
...
+ If an exception <see cref="T:System.IO.IOException" /> is thrown
```
MSDN-WhiteKnight, line 836:
```suggestion
 If an exception <xref:System.IO.IOException> is thrown
```
This is a proper syntax for links in markdown

xml/System.IO/FileSystemWatcher.xml
```diff
@@ -830,13 +830,42 @@
...
+```
+
+ Each execution of this script doubles the system-wide limit, except when run in parallel.
+ This script requires privilege elevation, so either run it as superuser
+ or be prepared to authenticate using [pkexec]((https://www.freedesktop.org/software/polkit).
+ This is a global setting that can be modified by another user at any time.
+ Use the configuration tools provided by your operating system to change the initial value of this setting at boot time.
+ This script does not attempt to fix the other possible reason for this exception,
+ which is *the per-process limit on the number of open file descriptors has been reached*
+ (see your operating system documentation for error `EMFILE`).
```
MSDN-WhiteKnight, line 861: Is this information only relevant on Unix systems? If yes it should be mentioned.

xml/System.IO/FileSystemWatcher.xml
```diff
@@ -830,13 +830,42 @@
   
  :::code language="cpp" source="~/samples/snippets/cpp/VS_Snippets_CLR_Classic/classic NotifyFilters Example/CPP/source.cpp" id="Snippet1":::
  :::code language="csharp" source="~/samples/snippets/csharp/System.IO/FileSystemWatcher/Overview/source.cs" id="Snippet1":::
- :::code language="vb" source="~/samples/snippets/visualbasic/VS_Snippets_CLR_Classic/classic NotifyFilters Example/VB/source.vb" id="Snippet1":::  
+ :::code language="vb" source="~/samples/snippets/visualbasic/VS_Snippets_CLR_Classic/classic NotifyFilters Example/VB/source.vb" id="Snippet1":::
+
+## Troubleshooting
+ If an exception <see cref="T:System.IO.IOException" /> is thrown
+ telling you that the *the configured user limit on the number of inotify instances has been reached*,
```
MSDN-WhiteKnight, line 837:
```suggestion
 telling you that the *the configured user limit on the number of `inotify` instances has been reached*,
```
To prevent localization

xml/System.IO/FileSystemWatcher.xml
```diff
@@ -830,13 +830,42 @@
...
+ This script does not attempt to fix the other possible reason for this exception,
+ which is *the per-process limit on the number of open file descriptors has been reached*
+ (see your operating system documentation for error `EMFILE`).
   
  ]]></format>
         </remarks>
         <exception cref="T:System.ObjectDisposedException">The <see cref="T:System.IO.FileSystemWatcher" /> object has been disposed.</exception>
         <exception cref="T:System.PlatformNotSupportedException">The current operating system is not Microsoft Windows NT or later.</exception>
         <exception cref="T:System.IO.FileNotFoundException">The directory specified in <see cref="P:System.IO.FileSystemWatcher.Path" /> could not be found.</exception>
+        <exception cref="T:System.IO.IOException">The configured user limit on the number of inotify instances has been reached, or the per-process limit on the number of open file descriptors has been reached.</exception>
```
MSDN-WhiteKnight, line 868:
```suggestion
        <exception cref="T:System.IO.IOException">The configured user limit on the number of <c>inotify</c> instances has been reached, or the per-process limit on the number of open file descriptors has been reached.</exception>
```

xml/System.IO/FileSystemWatcher.xml
yecril71pl: I am sorry, I have no idea.  I have seen it on Linux but it does not mean it cannot appear elsewhere.

xml/System.IO/FileSystemWatcher.xml
```diff
@@ -823,20 +823,46 @@
   
  The <xref:System.IO.FileSystemWatcher.WaitForChanged%2A> method allows event handlers to be invoked to respond to file changes even if this property is set to `false`.  
   
-   
-  
+ If an exception <xref:System.IO.IOException> is thrown
```
adamsitnik, line 826: In general we never add scripts to the docs. Moreover it does not solve the problem as it modifies the file that is not persisted between system reboots.

We should rather describe where the setting comes from and mention the file (`/etc/sysctl.conf`) and name (` fs.inotify.max_user_instances`) of the setting that needs to be modified.

cc @tmds

adamsitnik (commented):
@yecril71pl thank you for your contribution! PTAL at my comment

xml/System.IO/FileSystemWatcher.xml
yecril71pl: As I already explained, persisting this value between system reboots does not solve the problem.

Jozkee: "exception IOException" sounds redundant.
```suggestion
 If an <xref:System.IO.IOException> is thrown
```

Jozkee: I agree that "mention a persistent solution + applying the configuration to the running kernel" is a more valuable hint to users that read this doc page.

Instead of changing `/proc/sys/fs/inotify/max_user_instances` can you please mention the solution where you change `/etc/sysctl.conf` and then apply the change with `sudo sysctl -p`?

Jozkee (changes requested):

xml/System.IO/FileSystemWatcher.xml
yecril71pl: This is not a solution.  This parameter can be changed live at any time.

xml/System.IO/FileSystemWatcher.xml
```diff
@@ -823,20 +823,46 @@
...
- :::code language="vb" source="~/samples/snippets/visualbasic/VS_Snippets_CLR_Classic/classic NotifyFilters Example/VB/source.vb" id="Snippet1":::  
-  
+ :::code language="vb" source="~/samples/snippets/visualbasic/VS_Snippets_CLR_Classic/classic NotifyFilters Example/VB/source.vb" id="Snippet1":::
+
  ]]></format>
         </remarks>
         <exception cref="T:System.ObjectDisposedException">The <see cref="T:System.IO.FileSystemWatcher" /> object has been disposed.</exception>
         <exception cref="T:System.PlatformNotSupportedException">The current operating system is not Microsoft Windows NT or later.</exception>
         <exception cref="T:System.IO.FileNotFoundException">The directory specified in <see cref="P:System.IO.FileSystemWatcher.Path" /> could not be found.</exception>
+        <exception cref="T:System.IO.IOException">The configured user limit on the number of <c>inotify</c> instances has been reached, or the per-process limit on the number of open file descriptors has been reached.</exception>
```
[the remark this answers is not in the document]
gewarren: I don't know of any "rules" but we typically do it [like this](https://github.com/dotnet/dotnet-api-docs/blob/bf815f4e86a4b5311e2ee1e5453226feef139c38/xml/System.IO/Directory.xml#L237). So in this case it would be:

```suggestion
        <exception cref="T:System.IO.IOException">Unix-based operating systems only: The configured user limit on the number of <c>inotify</c> instances has been reached, or the per-process limit on the number of open file descriptors has been reached.</exception>
```

xml/System.IO/FileSystemWatcher.xml
Jozkee: Editing /etc/sysctl.conf as siggested [here](https://github.com/dotnet/aspnetcore/issues/8449#issuecomment-512275929) changes the parameter instantly + every time the OS is booted. 
Do you mind explaining why that is not a solution (or point me to where you explained it since you say already did) or why your suggested script is better?

yecril71pl: 1. Editing `/etc/sysctl.conf` changes the parameter at the next reboot.  The kernel does not observe changes to this configuration file.
2. The latter command applies the change to the kernel but does not work on BSD systems, as discussed there.
3. The number 524288 is arbitrary and potentially harmful.  256 is enough for my Linux system.  But every system is different, so a more adaptive approach should be recommended, such as the one implemented in my script.

yecril71pl: > Tagging @tmds as he might know some ways to improve [this script](https://review.docs.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher.enableraisingevents?view=net-5.0&branch=pr-en-us-7627#remarks).

This is an open source project now.

xml/System.IO/FileSystemWatcher.xml
```diff
@@ -837,6 +862,7 @@
         <exception cref="T:System.ObjectDisposedException">The <see cref="T:System.IO.FileSystemWatcher" /> object has been disposed.</exception>
         <exception cref="T:System.PlatformNotSupportedException">The current operating system is not Microsoft Windows NT or later.</exception>
         <exception cref="T:System.IO.FileNotFoundException">The directory specified in <see cref="P:System.IO.FileSystemWatcher.Path" /> could not be found.</exception>
+        <exception cref="T:System.IO.IOException">Unix-based operating systems only: The configured user limit on the number of <c>inotify</c> instances has been reached, or the per-process limit on the number of open file descriptors has been reached.</exception>
```
tmds, line 865: This exception is Linux specific. `Unix` -> `Linux`.

xml/System.IO/FileSystemWatcher.xml
tmds: This is all Linux specific.

xml/System.IO/FileSystemWatcher.xml
```diff
@@ -823,8 +823,33 @@
   
  The <xref:System.IO.FileSystemWatcher.WaitForChanged%2A> method allows event handlers to be invoked to respond to file changes even if this property is set to `false`.  
   
-   
-  
+ If an <xref:System.IO.IOException> is thrown
```
tmds, line 826: It would be good to mention why the limits are there, and how to persist the setting.

Suggestion (to be improved):

On Linux the number of watchers and directories that can be watched is limited. These limits protect the amount of kernel memory that is available for watchers. When a limit is reached `FileSystemWatcher` will throw an `IOException` or report it via the `FileSystemWatcher.Error` event. The limits can be read from `/proc/sys/fs/inotify/max_user_instances` and `/proc/sys/fs/inotify/max_user_watches`.

The files can be used to set the kernel values, but the changes will not persist across reboots. On most distributions, you can persist the settings by adding the following to `/etc/sysctl.conf`.
```
fs.inotify.max_user_instances=<value>
fs.inotify.max_user_watches=<value>
```
As values you can use double of the values reported by the `proc` files. After changing the file, you can run `sysctl -p` to apply the settings.

xml/System.IO/FileSystemWatcher.xml
yecril71pl: We have seen reports of this happening on BSD systems.

tmds: Does BSD support `inotify`? I thought it was Linux specific.

yecril71pl: I have no first-hand experience but that is what other people complain about.  I suppose Apple may have ported it.

tmds: To be sure, do you have a reference to such a report?

yecril71pl: @Jozkee shared it with me and then deleted their comment.

Jozkee: @yecril71pl I linked back to it in my next reply, but here's again https://github.com/dotnet/aspnetcore/issues/8449#issuecomment-514089725.

tmds: > I linked back to it in my next reply, but here's again https://github.com/dotnet/aspnetcore/issues/8449#issuecomment-514089725.

This is about Linux running in a VM on Mac for docker. It is not about BSD.

tmds: Looking at the [csproj](https://github.com/dotnet/runtime/blob/135a9608c411f89f947837701ab1fd7ff1d3a5d6/src/libraries/System.IO.FileSystem.Watcher/src/System.IO.FileSystem.Watcher.csproj#L72-L75) there are two platforms that can throw this exception: Linux and FreeBSD.

The FreeBSD part was added in https://github.com/dotnet/corefx/pull/42488. It uses libinotify.

libinotify has a limit of [2147483646](https://github.com/libinotify-kqueue/libinotify-kqueue/blob/ea7835fcafc3cee2a0d6c0e3c8034962c48f6afe/sys/inotify.h#L35) instances by default.

yecril71pl: > This is about Linux running in a VM on Mac for docker. It is not about BSD.

The next comment is about BSD.

tmds: > The next comment is about BSD.

I don't see it, and when I search BSD on that page my browser doesn't see it either.

Based on the code, besides Linux only FreeBSD can throw the exception, and given the 2G default limit in libinotify, it isn't really limited.

yecril71pl: > > The next comment is about BSD.

> I don't see it, and when I search BSD on that page my browser doesn't see it either.

> `sysctl` doesn't seem to work the same way on mac

tmds: This exception doesn't get thrown on mac, so we don't need to provide instructions for it.

yecril71pl: Forgive my being dumb&mdash;but for what other reason would the commenter complain about `sysctl` in a context where it is recommended to be used to deal with this exception?

tmds: The exception occurs on a mac host that runs a Linux VM.
The commands need to be executed in the Linux VM, and not on the host (where they give an error).

PR #7628, opened 2022-01-24
Update wording in 'List<T>.RemoveAll(Predicate<T>) Method'

woeterman94:
I didn't really know how to formulate it better , but the docs don't really make it clear that no exceptions is thrown if no matching elements are found in the list.

See: https://stackoverflow.com/questions/12260889/list-removeall-argument-null-exception-not-thrown

Low prio

Edit: Just to clearify the section 'match is null' under exceptions made me think that an exception is thrown if the match is null (no match).

### DIFF
--- a/xml/System.Collections.Generic/List`1.xml
+++ b/xml/System.Collections.Generic/List`1.xml
@@ -3318,7 +3318,7 @@ Public Function StartsWith(e As Employee) As Boolean
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Predicate%601> is a delegate to a method that returns `true` if the object passed to it matches the conditions defined in the delegate.  The elements of the current <xref:System.Collections.Generic.List%601> are individually passed to the <xref:System.Predicate%601> delegate, and the elements that match the conditions are removed from the <xref:System.Collections.Generic.List%601>. If no matching elements are found no items are removed from the list.  
+ The <xref:System.Predicate%601> is a delegate to a method that returns `true` if the object passed to it matches the conditions defined in the delegate. The elements of the current <xref:System.Collections.Generic.List%601> are individually passed to the <xref:System.Predicate%601> delegate, and the elements that match the conditions are removed from the <xref:System.Collections.Generic.List%601>. If no matching elements are found, no items are removed from the list.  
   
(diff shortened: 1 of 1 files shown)


xml/System.Collections.Generic/List`1.xml
```diff
@@ -3318,7 +3318,7 @@ Public Function StartsWith(e As Employee) As Boolean
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Predicate%601> is a delegate to a method that returns `true` if the object passed to it matches the conditions defined in the delegate.  The elements of the current <xref:System.Collections.Generic.List%601> are individually passed to the <xref:System.Predicate%601> delegate, and the elements that match the conditions are removed from the <xref:System.Collections.Generic.List%601>.  
+ The <xref:System.Predicate%601> is a delegate to a method that returns `true` if the object passed to it matches the conditions defined in the delegate.  The elements of the current <xref:System.Collections.Generic.List%601> are individually passed to the <xref:System.Predicate%601> delegate, and the elements that match the conditions are removed from the <xref:System.Collections.Generic.List%601>. If no matching elements are found no items are removed from the list.
```
gewarren, line 3321:
```suggestion
 The <xref:System.Predicate%601> is a delegate to a method that returns `true` if the object passed to it matches the conditions defined in the delegate. The elements of the current <xref:System.Collections.Generic.List%601> are individually passed to the <xref:System.Predicate%601> delegate, and the elements that match the conditions are removed from the <xref:System.Collections.Generic.List%601>. If no matching elements are found, no items are removed from the list.  
```

gewarren (approved):
Thanks @woeterman94

xml/System.Collections.Generic/List`1.xml
eiriktsarpalis, line 3321: The added clarification follows trivially from the previous sentence. I don't think it is needed.

eiriktsarpalis (changes requested):

xml/System.Collections.Generic/List`1.xml
woeterman94: I'm all open for suggestions.

woeterman94:
@eiriktsarpalis how would you formulate it?

eiriktsarpalis:
> how would you formulate it?

As mentioned, I don't believe that this clarification is necessary since it follows trivially from the previous sentence.

PR #7635, opened 2022-01-24
Add link to Configure Await info

### LINKED ISSUE #1254: Add explanation or link on why this method exists
It is not self-evident. Ta.

---
#### Document Details

* ID: e069a261-9adf-92e6-9b16-8a701968a36d
* Version Independent ID: a12bf93b-1690-1604-20b0-2bf5ca44b372
* Content: [Task.ConfigureAwait(Boolean) Method (System.Threading.Tasks)](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.configureawait?view=netframework-4.7.2)
* Content Source: [xml/System.Threading.Tasks/Task.xml](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System.Threading.Tasks/Task.xml)
* Product: **dotnet-api**
* GitHub Login: @dotnet-bot

gewarren:
Fixes #1254.

### DIFF
--- a/xml/System.Threading.Tasks/Task.xml
+++ b/xml/System.Threading.Tasks/Task.xml
@@ -791,7 +791,7 @@ To retrieve the object's data, cast it back to the original type.
 
 ## Remarks
 
-When an asynchronous method awaits a <xref:System.Threading.Tasks.Task> directly, continuation usually occurs in the same thread that created the task, depending on the async context. This behavior can be costly in terms of performance and can result in a deadlock on the UI thread. To signal your intention for continuation, call <xref:System.Threading.Tasks.Task.ConfigureAwait(System.Boolean)?displayProperty=nameWithType>. For more information, see [ConfigureAwait FAQ](https://devblogs.microsoft.com/dotnet/configureawait-faq/).
+When an asynchronous method awaits a <xref:System.Threading.Tasks.Task> directly, continuation usually occurs in the same thread that created the task, depending on the async context. This behavior can be costly in terms of performance and can result in a deadlock on the UI thread. To avoid these problems, call `Task.ConfigureAwait(false)`. For more information, see [ConfigureAwait FAQ](https://devblogs.microsoft.com/dotnet/configureawait-faq/).
 
        ]]></format>
         </remarks>


xml/System.Threading.Tasks/Task.xml
```diff
@@ -791,7 +786,16 @@
           <see langword="true" /> to attempt to marshal the continuation back to the original context captured; otherwise, <see langword="false" />.</param>
         <summary>Configures an awaiter used to await this <see cref="T:System.Threading.Tasks.Task" />.</summary>
         <returns>An object used to await this task.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+When an asynchronous method awaits a <xref:System.Threading.Tasks.Task> directly, continuation usually occurs in the same thread that created the task, depending on the async context. This behavior can be costly in terms of performance and can result in a deadlock on the UI thread. To signal your intention for continuation, call <xref:System.Threading.Tasks.Task.ConfigureAwait(System.Boolean)?displayProperty=nameWithType>. For more information, see [ConfigureAwait FAQ](https://devblogs.microsoft.com/dotnet/configureawait-faq/).
```
buyaa-n, line 794: Seems `ConfigureAwait` with `false` is mainly suggested, awaitng with `ConfigureAwait(true)` has no difference not using it at all, therefore i think the remarks should suggest to use `ConfigureAwait(false)`, something like:
```suggestion
When an asynchronous method awaits a <xref:System.Threading.Tasks.Task> directly, continuation usually occurs in the same thread that created the task, depending on the async context. This behavior can be costly in terms of performance and can result in a deadlock on the UI thread. To avoid this call <xref:System.Threading.Tasks.Task.ConfigureAwait(false)?displayProperty=nameWithType>. For more information, see [ConfigureAwait FAQ](https://devblogs.microsoft.com/dotnet/configureawait-faq/).
```

gewarren, line 794:
```suggestion
When an asynchronous method awaits a <xref:System.Threading.Tasks.Task> directly, continuation usually occurs in the same thread that created the task, depending on the async context. This behavior can be costly in terms of performance and can result in a deadlock on the UI thread. To avoid these problems, call `Task.ConfigureAwait(false)`. For more information, see [ConfigureAwait FAQ](https://devblogs.microsoft.com/dotnet/configureawait-faq/).
```

gewarren: The xref link won't work with false, so I made an alternative suggestion. How does that look @buyaa-n?

buyaa-n: I was not sure if it would work, yes, yours looks good, thanks!

buyaa-n (approved):

PR #7649, opened 2022-01-26
Fix mix-up between IEnumerable and IEnumerator

abduelhamit:
## Summary

The method `BlockingCollection<T>.GetConsumingEnumerable` provides an `IEnumerable<T>`, but in some cases the doc says that it's an `IEnumerator<T>`, which is wrong.

This pull request fixes that mix-up.


eiriktsarpalis (approved):
Thanks!

PR #7651, opened 2022-01-26
Refresh readme for repo

gewarren:


README.md
```diff
@@ -1,8 +1,7 @@
 # .NET API Reference Docs
 
-This repo contains documentation for the .NET API Reference.
+This repo contains documentation for the .NET API reference. We track all work for this repository using [GitHub issues](https://github.com/dotnet/dotnet-api-docs/issues). The documentation for APIs is built from the text and code snippets in this repo, and from the samples in the [dotnet/samples](https://github.com/dotnet/samples) repository.
```
BillWagner, line 3: Would this be a reasonable location to add a comment about regarding the move to use `///` comments in the runtime repo as the source of truth, over time? I'm not sure. It could be useful, but details on which namespaces have been converted could quickly become out of date.

BillWagner (approved):
This is great @gewarren 

I had one question to consider, and then this is ready to merge.

README.md
gewarren: That's a good point, but I think we should only add that note once we've fully made the switch - so as not to confuse people.

PR #7659, opened 2022-01-27
Update Path.xml

doterik:
`TryJoin(ReadOnlySpan<Char>, ReadOnlySpan<Char>, ReadOnlySpan<Char>, Span<Char>, Int32)`

Attempts to concatenate ~~two~~ three path components...


gewarren (approved):
Thanks @doterik

PR #7663, opened 2022-01-28
Update summary for two `MemoryExtensions.AsSpan(this string? text …)` overloads

cston:
## Summary

Switch the summaries of two `MemoryExtensions.AsSpan(this string? text …)` overloads, to correspond to the method signatures.


gewarren (approved):
Thanks @cston!

PR #7673, opened 2022-02-01
Update FrameworkName.xml

RussKie:
Remove stray spaces that don't exist in the implementation (see https://github.com/dotnet/runtime/blob/c6fa5b2f318c22dbdbe1a84ca0638092a185ae3d/src/libraries/System.Private.CoreLib/src/System/Runtime/Versioning/FrameworkName.cs#L49-L62).

Relates to https://github.com/dotnet/winforms-designer/pull/4077#discussion_r795394814


BillWagner (approved):
This LGTM. I'll :shipit: now.

PR #7675, opened 2022-02-01
ObservableCollection.CollectionChanged clarification.

peter-dye:
## Summary

The summary makes it sound like the event will fire if a collection item property changes but this is not the case. Added a remark to make this more clear.

### DIFF
--- a/xml/System.Collections.ObjectModel/ObservableCollection`1.xml
+++ b/xml/System.Collections.ObjectModel/ObservableCollection`1.xml
@@ -461,7 +461,7 @@
         <ReturnType>System.Collections.Specialized.NotifyCollectionChangedEventHandler</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Occurs when an item is added, removed, moved, or the entire list is refreshed.</summary>
+        <summary>Occurs when an item is added, removed, or moved, or the entire list is refreshed.</summary>
         <remarks>To be added.</remarks>
         <altmember cref="T:System.Collections.Specialized.INotifyCollectionChanged" />
         <altmember cref="E:System.Collections.Specialized.INotifyCollectionChanged.CollectionChanged" />


xml/System.Collections.ObjectModel/ObservableCollection`1.xml
```diff
@@ -461,8 +461,8 @@
         <ReturnType>System.Collections.Specialized.NotifyCollectionChangedEventHandler</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Occurs when an item is added, removed, changed, moved, or the entire list is refreshed.</summary>
-        <remarks>To be added.</remarks>
+        <summary>Occurs when an item is added, removed, changed, or moved, or the entire list is refreshed.</summary>
+        <remarks>Does not fire when a property of a collection item changes.</remarks>
```
eiriktsarpalis, line 465: Wouldn't this suffice?

```suggestion
        <summary>Occurs when an item is added, removed, moved, or the entire list is refreshed.</summary>
        <remarks>To be added.</remarks>
```

peter-dye: My understanding was that the "changed" wording referred to a situation such as:
```
collection[0] = item1;
collection[0] = item2;
```
But that could be considered an Add and a Remove even though `.Add()` and `.Remove()` aren't used.

eiriktsarpalis: I took the wording from description of the type itself: https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.observablecollection-1?view=net-6.0#definition

I don't think a clarification on mutation of individual collection items is warranted, for generic collections it is generally not possible to achieve anyway.

xml/System.Collections.ObjectModel/ObservableCollection`1.xml
```diff
@@ -461,8 +461,8 @@
         <ReturnType>System.Collections.Specialized.NotifyCollectionChangedEventHandler</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Occurs when an item is added, removed, changed, or moved, or the entire list is refreshed.</summary>
-        <remarks>Does not fire when a property of a collection item changes.</remarks>
+        <summary>Occurs when an item is added, removed, moved, or the entire list is refreshed.</summary>
```
gewarren, line 464:
```suggestion
        <summary>Occurs when an item is added, removed, or moved, or the entire list is refreshed.</summary>
```

gewarren (approved):
Thank you!

eiriktsarpalis (approved):
Thanks!

PR #7678, opened 2022-02-02
Add example for BlobEncoder.MethodSignature

MSDN-WhiteKnight:
Add examples that generate different shapes of method signatures with BlobEncoder. 

Note: if you want to verify that encoded signatures are valid, you can use [this sample](https://github.com/MSDN-WhiteKnight/CodeSamples/blob/master/System.Reflection.Metadata/MethodSignature/MethodSignatureSnippets.cs) which does similar things, but produce a complete assembly that could be inspected with ildasm. I'm not adding a complete sample here, because MetadataBuilder already has it.

### DIFF
--- a/xml/System.Reflection.Metadata.Ecma335/BlobEncoder.xml
+++ b/xml/System.Reflection.Metadata.Ecma335/BlobEncoder.xml
@@ -249,7 +249,7 @@
 
 ## Remarks
 
-Method signature blobs are used when defining or referencing methods in .NET metadata. For a complete example of emitting a method definition see the <xref:System.Reflection.Metadata.Ecma335.MetadataBuilder> class remarks.
+Method signature blobs are used when defining or referencing methods in .NET metadata. For a complete example of emitting a method definition, see the <xref:System.Reflection.Metadata.Ecma335.MetadataBuilder> class remarks.
 
 ## Examples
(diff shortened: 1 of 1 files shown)


xml/System.Reflection.Metadata.Ecma335/BlobEncoder.xml
```diff
@@ -244,7 +244,20 @@
           <see langword="true" /> to encode an instance method signature, <see langword="false" /> to encode a static method signature.</param>
         <summary>Encodes method signature blob.</summary>
         <returns>An encoder of the rest of the signature including return value and parameters.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
+        <format type="text/markdown"><![CDATA[  
+
+## Remarks
+
+Method signature blobs are used when defining or referencing methods in .NET metadata. For a complete example of emitting a method definition see the <xref:System.Reflection.Metadata.Ecma335.MetadataBuilder> class remarks.
```
gewarren, line 252:
```suggestion
Method signature blobs are used when defining or referencing methods in .NET metadata. For a complete example of emitting a method definition, see the <xref:System.Reflection.Metadata.Ecma335.MetadataBuilder> class remarks.
```

buyaa-n (approved):
LGTM, thanks!

PR #7683, opened 2022-02-03
Remove outdated COM reference

perlun:
I discovered this while browsing the docs. From what I can tell, this reference is no longer relevant, but please help me verify. Here's the current 6.0.1 version of the (Windows) implementation of this method: https://github.com/dotnet/runtime/blob/v6.0.1/src/libraries/Common/src/Interop/Windows/Kernel32/Interop.ExpandEnvironmentStrings.cs

As far as I understand it, this is a regular P/Invoke call and not anything COM-related. I _guess_ that this is a remnant from the .NET Framework times, but as noted, please don't take my word for it - help me verify if I'm reading this correctly or not. :slightly_smiling_face:

### DIFF
--- a/xml/System/Environment.xml
+++ b/xml/System/Environment.xml
@@ -487,8 +487,6 @@ Invalid argument
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- COM interop is used to retrieve the environment variables from the operating system. If the environment variables cannot be retrieved due to a COM error, the HRESULT that explains the cause of the failure is used to generate one of several possible exceptions; that is, the exception depends on the HRESULT. For more information about how the HRESULT is processed, see the Remarks section of the <xref:System.Runtime.InteropServices.Marshal.ThrowExceptionForHR%2A?displayProperty=nameWithType> method.
-
  Replacement only occurs for environment variables that are set. For example, suppose `name` is "MyENV = %MyENV%". If the environment variable, MyENV, is set to 42, this method returns "MyENV = 42". If MyENV is not set, no change occurs; this method returns "MyENV = %MyENV%".
 
  The size of the return value is limited to 32K.


MSDN-WhiteKnight:
.NET Framework does not use COM either, but it does convert Win32 error to HRESULT and then throw exception with that HRESULT. Not sure if this is important fact or not.

perlun:
> .NET Framework does not use COM either, but it does convert Win32 error to HRESULT and then throw exception with that HRESULT. Not sure if this is important fact or not.

Hmm, that's a good point @MSDN-WhiteKnight. Here's the code where the method is being called (Windows only): https://github.com/dotnet/runtime/blob/3a25a7f1cc446b60678ed25c9d829420d6321eba/src/libraries/System.Private.CoreLib/src/System/Environment.Windows.cs#L65-L81

So perhaps it does make a bit of sense the way it's phrased (for Windows); it just needs to be rephrased a bit to better reflect reality. 🤔 Any suggestions for how we could phrase this in a platform-generic way?

MSDN-WhiteKnight:
We couldn't phrase it in platform specific way, the behaviour is clearly different between operating systems and between .NET Framework and .NET/.NET Core. In current .NET, looking at the code you linked, it could throw IOException or its subclasses. In .NET Framework (https://referencesource.microsoft.com/#mscorlib/system/environment.cs,495) it could throw some specific exceptions for known HRESULTs or ComException for others. And knowing possible thrown exceptions is important for readers, because uncaught exception could be a reliability issue.

I think remarks would need to specify behaviour for 3 cases, .NET Framework, .NET Core on Windows and .NET Core on Unix/WASM. But it's also good to gather CLR maintainers/area owners opinions on that.

BillWagner:
closing as abandoned

PR #7684, opened 2022-02-03
Fix .NET Core -> .NET references

perlun:
There are still a few references in this file to .NET Core even after this change, but I believe they should remain. .NET Core releases <= 3.1 were actually named .NET Core, it's only .NET (Core) 5 and 6 that has dropped the "Core" part. Hence, it makes sense to retain it in contexts where older .NET Core versions are being referred to.

### DIFF
--- a/xml/System/Environment.xml
+++ b/xml/System/Environment.xml
@@ -916,7 +916,7 @@ On macOS and Linux, the environment block of the current process includes the fo
       </Parameters>
       <Docs>
         <param name="variable">The name of an environment variable.</param>
-        <param name="target">One of the <see cref="T:System.EnvironmentVariableTarget" /> values. Only <see cref="F:System.EnvironmentVariableTarget.Process" /> is supported on .NET running on Unix-bases systems.</param>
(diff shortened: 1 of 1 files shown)


xml/System/Environment.xml
```diff
@@ -916,7 +916,7 @@ On macOS and Linux, the environment block of the current process includes the fo
       </Parameters>
       <Docs>
         <param name="variable">The name of an environment variable.</param>
-        <param name="target">One of the <see cref="T:System.EnvironmentVariableTarget" /> values. Only <see cref="F:System.EnvironmentVariableTarget.Process" /> is supported on .NET Core running on Unix-bases systems.</param>
+        <param name="target">One of the <see cref="T:System.EnvironmentVariableTarget" /> values. Only <see cref="F:System.EnvironmentVariableTarget.Process" /> is supported on .NET running on Unix-bases systems.</param>
```
gewarren, line 919:
```suggestion
        <param name="target">One of the <see cref="T:System.EnvironmentVariableTarget" /> values. Only <see cref="F:System.EnvironmentVariableTarget.Process" /> is supported on .NET running on Unix-based systems.</param>
```

gewarren (approved):
One typo that isn't yours.

gewarren:
Thanks @perlun !

PR #7686, opened 2022-02-03
Fix FileSystemInfo.CreationTimeUtc Exceptions typo

EatonZ:
There is a typo in the [Exceptions docs for FileSystemInfo.CreationTimeUtc](https://docs.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo.creationtimeutc?view=net-6.0#exceptions).
The documentation for the `ArgumentOutOfRangeException` says:
_The caller attempts to set an invalid access time._

This is the **creation** time, not access time. The [documentation for CreationTime (non-UTC)](https://docs.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo.creationtime?view=net-6.0#exceptions) is correct, and this PR simply fixes this discrepancy between the two.


gewarren (approved):
Thanks @EatonZ

PR #7687, opened 2022-02-04
Improve Code Safety

CoalChris:
This API looks dangerous. What is wrong with trying QueueUserWorkItem()?

https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadpool.unsafequeueuserworkitem?view=net-6.0 

Using UnsafeQueueUserWorkItem could inadvertently open up a security hole. Code access security bases its permission checks on the permissions of all the callers on the stack. When work is queued on a thread pool thread using UnsafeQueueUserWorkItem, the stack of the thread pool thread will not have the context of the actual callers. Malicious code might be able exploit this to avoid permission checks.

## Summary

Describe your changes here.

### DIFF
--- a/snippets/csharp/System.Threading.Tasks/TaskScheduler/Overview/limitex1.cs
+++ b/snippets/csharp/System.Threading.Tasks/TaskScheduler/Overview/limitex1.cs
@@ -104,7 +104,7 @@ public class LimitedConcurrencyLevelTaskScheduler : TaskScheduler
    // Inform the ThreadPool that there's work to be executed for this scheduler.
    private void NotifyThreadPoolOfPendingWork()
    {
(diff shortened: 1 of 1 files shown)


stephentoub:
There is no arbitrary user code that's executed here without the ExecutionContext. The only user code is in the Task that's executed with TryExecuteTask, and that Task itself flows its own ExevutionContext. Doing so for the outer wrapper that just dequeues the next task is superfluous. What exactly is the security concern?

CoalChris:
Why _not_ use the safe method? While it might be interesting to argue why the unsafe method works, it is not following best practices and can cause a security hole in the future, so that is good enough reason to make the switch.

stephentoub:
> Why not use the safe method?

For this use case it's more expensive without providing any safety benefits.

PR #7698, opened 2022-02-08
Improve String.Contains(String) remarks

chwarr:
## Summary

Improve String.Contains(String) remarks

* Redirect to the `String.Contains(String, StringComparison)` overload
  instead of the char one. That overload makes for sense when the needle
  is a string.
* Include the demo extension method snippet as well as the example
  showing what it returns.


gewarren (approved):
LGTM, thanks @chwarr!

PR #7725, opened 2022-02-16
System.Exception F# snippets

albert-du:
Adds F# snippets for Exception


BillWagner (approved):
Thanks again @albert-du 

These all LGTM.  I'll :shipit: now.

PR #7726, opened 2022-02-16
System.FlagsAttribute F# snippets

albert-du:
Adds F# snippets for FlagsAttribute


BillWagner (approved):
Thanks again @albert-du 

This LGTM. I'll :shipit: now.

PR #7729, opened 2022-02-16
System.Guid F# snippets

albert-du:
Adds F# snippets for Guid


BillWagner (approved):
Thanks again @albert-du 

These are great, and I'll :shipit: now.

PR #7730, opened 2022-02-16
System.HashCode F# snippets

albert-du:
Adds F# snippets for HashCode


BillWagner (approved):
Thanks again @albert-du 

This also looks great. I'll :shipit: now.

PR #7737, opened 2022-02-17
Assembly.LoadFrom (.NET Framework) behaviour clarifications

vuplea:
The current docs can lead to confusion; they can be understood to say that an assembly with the same identity can only be loaded once - in fact it can only be loaded once in the same context, but the same identity can be both in the load and load-from contexts.

Also added a caveat with the influence of binding redirects.  Will remove if deemed too obscure.

### DIFF
--- a/xml/System.Reflection/Assembly.xml
+++ b/xml/System.Reflection/Assembly.xml
@@ -4652,7 +4652,7 @@ Starting with .NET Framework 4, the ability to execute code in assemblies loaded
 
 -   An assembly can be loaded in the load-from context even though an assembly with the same identity exists in the load context. Interoperability between the two assemblies will not work, leading to errors such as <xref:System.InvalidCastException>, <xref:System.MissingMethodException>, or other unexpected behavior.
 
--   Calling <xref:System.Reflection.Assembly.LoadFrom%2A> with a location which is in the probing path will load the assembly in the load context and not in the load-from context.
+-   Calling <xref:System.Reflection.Assembly.LoadFrom%2A> with a location that's in the probing path will load the assembly in the load context and not in the load-from context.
 
 -   If an assembly file whose identity is goverened by a binding redirect policy is passed to <xref:System.Reflection.Assembly.LoadFrom%2A>, the policy will be applied and the assembly will be loaded from the probing path in the load context.
 


joperezr:
@steveharter could you please take a look at this one?

steveharter (approved):

xml/System.Reflection/Assembly.xml
```diff
@@ -4648,11 +4648,15 @@ Starting with .NET Framework 4, the ability to execute code in assemblies loaded
...
 
--   If an assembly is loaded with <xref:System.Reflection.Assembly.LoadFrom%2A>, and later an assembly in the load context attempts to load the same assembly by display name, the load attempt fails. This can occur when an assembly is de-serialized.
+-   An assembly can be loaded in the load-from context even though an assembly with the same identity exists in the load context. Interoperability between the two assemblies will not work, leading to errors such as <xref:System.InvalidCastException>, <xref:System.MissingMethodException>, or other unexpected behavior.
 
--   If an assembly is loaded with <xref:System.Reflection.Assembly.LoadFrom%2A>, and the probing path includes an assembly with the same identity but a different location, an <xref:System.InvalidCastException>, <xref:System.MissingMethodException>, or other unexpected behavior can occur.
+-   Calling <xref:System.Reflection.Assembly.LoadFrom%2A> with a location which is in the probing path will load the assembly in the load context and not in the load-from context.
```
gewarren, line 4655:
```suggestion
-   Calling <xref:System.Reflection.Assembly.LoadFrom%2A> with a location that's in the probing path will load the assembly in the load context and not in the load-from context.
```

xml/System.Reflection/Assembly.xml
```diff
@@ -4648,11 +4648,15 @@ Starting with .NET Framework 4, the ability to execute code in assemblies loaded
...
+
+-   If an assembly file whose identity is goverened by a binding redirect policy is passed to <xref:System.Reflection.Assembly.LoadFrom%2A>, the policy will be applied and the assembly will be loaded from the probing path in the load context.
+
+-   If an assembly is loaded in the load-from context, and later an assembly in the load context attempts to load the same assembly by display name, the load attempt fails. This can occur when an assembly is de-serialized.
```
gewarren, line 4659:
```suggestion
-   If an assembly is loaded in the load-from context, and later an assembly in the load context attempts to load the same assembly by display name, the load attempt fails. This can occur when an assembly is deserialized.
```

steveharter (approved):

PR #7750, opened 2022-02-18
Add F# examples for System.String.IsNullOrEmpty

brenoafb:
## Summary

Add F# examples for documention on [System.Strings.IsNullOrEmpty](https://docs.microsoft.com/en-us/dotnet/api/system.string.isnullorempty?view=net-6.0)


BillWagner (approved):
Thanks for adding these @brenoafb 

PR #7757, opened 2022-02-21
Update ArrayList.xml

wenhx:
Add the missing interface implementation declaration on the TypeSignature.

## Summary

Describe your changes here.


MSDN-WhiteKnight:
In sources ArrayList is declared to implement exactly 2 interfaces: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Collections/ArrayList.cs#L29

    public class ArrayList : IList, ICloneable

These TypeSignature tags are automatically generated by monodoc. I think it by design does not consider implicitly implemented interfaces, so the decompiled declaration is close to typical source declaration.

wenhx:
> In sources ArrayList is declared to implement exactly 2 interfaces: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Collections/ArrayList.cs#L29
> 
> ```
> public class ArrayList : IList, ICloneable
> ```
> 
> These TypeSignature tags are automatically generated by monodoc. I think it by design does not consider implicitly implemented interfaces, so the decompiled declaration is close to typical source declaration.

Thanks @MSDN-WhiteKnight , that may be true, but let's look at List&lt;T&gt;, whose TypeSignature explicitly lists implicitly implemented interfaces that the source code doesn't.
document: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1?view=net-6.0 
source: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs
Shouldn't they be consistent?

MSDN-WhiteKnight:
Good point. Yes, they should be consistent. My guess this distinction is related to generic types. Here's the code in monodoc that attempts to exclude inherited interfaces from type declaration: https://github.com/mono/api-doc-tools/blob/main/mdoc/Mono.Documentation/Updater/DocUtils.cs#L436 It clearly intends to exclude all inherited interfaces, but it compares them by type name, and on generic types it probably runs into some bug when genericness is not considered for comparison. 

Anyway, this cannot be fixed by editing members metadata, because, as i wrote, it's autogenerated and such changes will be undone on next CI update. This probably should be reported to Monodoc repo if it's really a bug and was not already reported.

wenhx:
Thanks for the explanation @MSDN-WhiteKnight .
Since this is automatically generated, I'll close my pull request.

PR #7759, opened 2022-02-21
Update RSA.xml

devzilenas:
Method is for a public key

### DIFF
--- a/xml/System.Security.Cryptography/RSA.xml
+++ b/xml/System.Security.Cryptography/RSA.xml
@@ -860,7 +860,7 @@ The encryption operation failed.</exception>
  The password bytes are passed directly into the Key Derivation Function (KDF) used by the algorithm indicated by the EncryptedPrivateKeyInfo contents.
  This enables compatibility with other systems which use a text encoding other than UTF-8 when processing passwords with PBKDF2 (Password-Based Key Derivation Function 2).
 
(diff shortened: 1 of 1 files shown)


xml/System.Security.Cryptography/RSA.xml
```diff
@@ -860,7 +860,7 @@ The encryption operation failed.</exception>
  The password bytes are passed directly into the Key Derivation Function (KDF) used by the algorithm indicated by the EncryptedPrivateKeyInfo contents.
  This enables compatibility with other systems which use a text encoding other than UTF-8 when processing passwords with PBKDF2 (Password-Based Key Derivation Function 2).
 
- This method only supports the binary (BER/CER/DER) encoding of EncryptedPrivateKeyInfo.
+ This method only supports the binary (BER/CER/DER) encoding of EncryptedPublicKeyInfo.
```
vcsjones, line 863: I don't think this change is correct. This documentation change is for the remarks of `ImportEncryptedPkcs8PrivateKey`, which handles a PKCS#8 `EncryptedPrivateKeyInfo` per [RFC 5208](https://datatracker.ietf.org/doc/html/rfc5208). 

There is also no such thing as `EncryptedPublicKeyInfo`.

Did you perhaps mean to edit a different part of this document, or need assistance with `ImportEncryptedPkcs8PrivateKey`?

PR #7763, opened 2022-02-25
System.IConvertible F# snippet

albert-du:
Adds an F# snippet for IConverible


BillWagner (approved):
Thanks again @albert-du 

This is great, and I'll :shipit: now.

PR #7767, opened 2022-02-26
System.IEquatable<T> F# snippets

albert-du:
Adds F# snippets for IEquatable<T>


BillWagner (approved):
Thanks again @albert-du 

This is great work, as always. I'll :shipit: now.

PR #7769, opened 2022-02-26
System.IFormattable F# snippets

albert-du:
Adds F# snippets for IFormattable


BillWagner (approved):
Thanks again @albert-du 

I've reviewed this and I'll :shipit: now.

PR #7772, opened 2022-02-26
System.InvalidCastException F# snippets

albert-du:
Adds F# snippets for InvalidCastException


BillWagner (approved):
Thanks again @albert-du 

I'll :shipit: now.

PR #7774, opened 2022-02-27
Path.GetFullPath has different behavior on macOS and Linux

pedrobsaila:
Fixes dotnet/runtime#65683


adamsitnik:
@jrdodds does the suggested fix is good enough from end user perspective?

jrdodds:
> @jrdodds does the suggested fix is good enough from end user perspective?

Yes. I'm happy with the suggested change. Thanks

adamsitnik (approved):
LGTM, thank you @pedrobsaila !

PR #7775, opened 2022-02-28
System.Lazy<T> F# snippets

albert-du:
Adds F# snippets for Lazy<T>


BillWagner (approved):
Thanks again @albert-du This is awesome!

I'll :shipit: now.

PR #7778, opened 2022-03-01
System.MarshalByRefObject F# snippets

albert-du:
Adds F# snippets for MarshalByRefObject


BillWagner (approved):
These are also all great @albert-du 

I'll :shipit: now.

PR #7794, opened 2022-03-05
System.MulticastDelegate F# snippet

albert-du:
Adds an F# snippet for MulticastDelegate


BillWagner (approved):
Thanks again @albert-du 

I'll :shipit: now.

PR #7796, opened 2022-03-05
System.NonSerializedAttribute F# snippet

albert-du:
Adds an F# snippet for NonSerializedAttribute


BillWagner (approved):
This also LGTM @albert-du 

I'll :shipit: now.

PR #7797, opened 2022-03-05
System.NotImplementedException F# snippet

albert-du:
Adds an F# for NotImplementedException


BillWagner (approved):
These also look great @albert-du 

I'll :shipit: now

PR #7803, opened 2022-03-07
The RemoveAll predicate should not throw exception

theodorzoulias:
## Summary

Include in the documentation that the `match` predicate of the [`List<T>.RemoveAll`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.removeall) method should not throw an exception, otherwise the state of the `List<T>` might break beyond repair.

Fixes: dotnet/runtime#66255

### DIFF
--- a/xml/System.Collections.Generic/List`1.xml
+++ b/xml/System.Collections.Generic/List`1.xml
@@ -3320,6 +3320,8 @@ Public Function StartsWith(e As Employee) As Boolean
 ## Remarks  
  The <xref:System.Predicate%601> is a delegate to a method that returns `true` if the object passed to it matches the conditions defined in the delegate.  The elements of the current <xref:System.Collections.Generic.List%601> are individually passed to the <xref:System.Predicate%601> delegate, and the elements that match the conditions are removed from the <xref:System.Collections.Generic.List%601>.  
   
+ The <xref:System.Predicate%601> delegate should not throw an exception. Otherwise, the <xref:System.Collections.Generic.List%601> is no longer in a valid state and it should be discarded.  
+
  This method performs a linear search; therefore, this method is an O(*n*) operation, where *n* is <xref:System.Collections.Generic.List%601.Count%2A>.
   
 ## Examples  


xml/System.Collections.Generic/List`1.xml
```diff
@@ -3320,6 +3320,8 @@ Public Function StartsWith(e As Employee) As Boolean
 ## Remarks  
  The <xref:System.Predicate%601> is a delegate to a method that returns `true` if the object passed to it matches the conditions defined in the delegate.  The elements of the current <xref:System.Collections.Generic.List%601> are individually passed to the <xref:System.Predicate%601> delegate, and the elements that match the conditions are removed from the <xref:System.Collections.Generic.List%601>.  
   
+ The <xref:System.Predicate%601> delegate should not throw an exception. Otherwise, the <xref:System.Collections.Generic.List%601> is no longer in a valid state and it should be discarded.
```
AraHaan, line 3323:
```suggestion
 The <xref:System.Predicate%601> delegate should not throw an exception. Otherwise, the <xref:System.Collections.Generic.List%601> is no longer in a valid state and it should be discarded (is corrupted).  
```

theodorzoulias: @AraHaan I think that "corrupted" conveys the same meaning with "not valid state", while being a bit more strong and potentially unsettling term. The people who might by puzzled by the "not valid state" term, might also be puzzled by the "corrupted" as well.

krwq: I'd go with some more generic statement, something along the lines of:

```
This API takes a predicate. Predicates throwing exception may cause undefined behavior including corrupted state.
```

eiriktsarpalis: 👍 on @krwq's choice of words. FWIW I think this clarification should apply to any mutation operation on a collection that involves inverted control. Even though certain types (such as Dictionary) happen to be exception safe at the moment, there's no guarantee future implementations are going to be preserving that invariant.

If anything I think lack of exception safety in this context is likely the norm: you were right to point out that PriorityQueue suffers from this, and it would seem like other ordered structures like `SortedSet<T>` might as well (although I have no clear evidence for that at the moment). In any case, I think that there should be wider guidance concerning exception safety of delegates/comparers when used to drive collections mutation.

theodorzoulias: > I'd go with some more generic statement, something along the lines of:
> 
> This API takes a predicate. Predicates throwing exception may cause undefined behavior including corrupted state.

@krwq this might be too generic IMHO. It could provoke an overly defensively stance against predicate-based APIs that do not modify their arguments, let alone corrupt them. Like the LINQ for example. Quoting [a comment](https://github.com/dotnet/runtime/issues/66255#issuecomment-1060322995) by @teo-tsirpanis:

> LINQ does not modify the input collection. You can safely use `ids` in the `catch` block.

Personally I an aware of no other API in .NET that accepts a predicate and in case of exception corrupts the state of the instance, apart from the  `List<T>.RemoveAll`.

krwq:
@theodorzoulias IMO and also https://github.com/dotnet/runtime/issues/66255#issuecomment-1060203234 the doc change seems optional. Even though we can't see other predicates which can potentially corrupt I'd still make the sentence generic because it could apply to other places as well and you cannot reasonably expect exception in such place to work fine. Please update the wording and we can merge, otherwise I'm ok with closing this PR.

AraHaan:
I would be ok with saying all ``predicates`` can potentially corrupt and leave it at that.

theodorzoulias:
@krwq FWIW my opinion is that the `List<T>.RemoveAll` method should either be non-corruptive (preferably), or should be documented that it's corruptive (as a last resort). Doing neither makes this method a trap. I am assuming that no one has fallen in this trap till now, and hopefully no one ever will, because it's not a commonly used API, and it's typically used with trivial lambdas that never throw. But if it ever happens, it has the potential of becoming a story for [TheDailyWTF](https://thedailywtf.com). More so because Microsoft generally is not known for releasing half-baked software.

PR #7810, opened 2022-03-10
Update AmbiguousMatchException.xml

ghost:
In the description of the example for AmbiguousMatchException(String, Exception), the term 'class' or 'classes' is used when it should be 'method' or 'methods'.  The proposed changed is to replace 'class' with 'method' and 'classes' with 'methods'

## Summary

Describe your changes here.


xml/System.Reflection/AmbiguousMatchException.xml
```diff
@@ -1,4 +1,4 @@
-<Type Name="AmbiguousMatchException" FullName="System.Reflection.AmbiguousMatchException">
+d<Type Name="AmbiguousMatchException" FullName="System.Reflection.AmbiguousMatchException">
```
gewarren, line 1:
```suggestion
<Type Name="AmbiguousMatchException" FullName="System.Reflection.AmbiguousMatchException">
```

gewarren (approved):
Thank you!

PR #7811, opened 2022-03-11
System.Object F# snippets

albert-du:
Adds F# snippets for Object


BillWagner (approved):
Thanks again @albert-du 

This is fantastic work.  I'll :shipit: now.

PR #7812, opened 2022-03-11
System.ObjectDisposedException F# snippets

albert-du:
Adds F# snippets for ObjectDisposedException


BillWagner (approved):
Thanks @albert-du 

This looks great as well. I'll :shipit: now.

PR #7821, opened 2022-03-13
System.PlatformID F# snippet

albert-du:
Adds an F# snippet for PlatformID


BillWagner (approved):
Thanks again @albert-du 

This is great.

I'll :shipit: now.

PR #7822, opened 2022-03-13
System.Predicate<T> F# snippets

albert-du:
Adds F# snippets for Predicate<T>


BillWagner (approved):
This looks great as well @albert-du 

I'll :shipit: now.

PR #7832, opened 2022-03-16
Update docs for SslCertificateTrust with changes in .NET 7

rzikm:
Reflects changes implemented in https://github.com/dotnet/runtime/pull/65195

### DIFF
--- a/xml/System.Net.Security/SslCertificateTrust.xml
+++ b/xml/System.Net.Security/SslCertificateTrust.xml
@@ -55,7 +55,7 @@
 
 ## Remarks
 
-If the `sendTrustInHandshake` argument is `true`, the client can use the list of trusted certificate authorities from the server to select an appropriate client certificate. Sending trusted issuers list is not supported for `SslCertificateTrust` instances created using the `SslCertificaetTrust.CreateForX509Collection` is not supported in .NET 6.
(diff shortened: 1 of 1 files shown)


xml/System.Net.Security/SslCertificateTrust.xml
```diff
@@ -55,13 +55,18 @@
 
 ## Remarks
 
-If the `sendTrustInHandshake` argument is `true`, the client can use the list of trusted certificate authorities from the server to select an appropriate client certificate. In .NET 6, the list is only sent on Windows, and it depends on a registry setting.
+If the `sendTrustInHandshake` argument is `true`, the client can use the list of trusted certificate authorities from the server to select an appropriate client certificate. Sending trusted issuers list is not supported for `SslCertificateTrust` instances created using the `SslCertificaetTrust.CreateForX509Collection` is not supported in .NET 6.
```
ManickaP, line 58:
```suggestion
If the `sendTrustInHandshake` argument is `true`, the client can use the list of trusted certificate authorities from the server to select an appropriate client certificate. Sending trusted issuers list is not supported for `SslCertificateTrust` instances created using the `SslCertificaetTrust.CreateForX509Collection` in .NET 6.
```

BTW, was this supported in .NET 6 on Windows? That's what I get from the previous wording.

ManickaP (approved):
LGTM

xml/System.Net.Security/SslCertificateTrust.xml
rzikm: no, it wasn't https://github.com/dotnet/runtime/blob/release/6.0/src/libraries/System.Net.Security/src/System/Net/Security/SslCertificateTrust.cs#L44-L48

Issue #1326, opened 2018-12-10
Clarify between "only a single instance should be created for an entire application" and "derived classes"


DeadX07:
The guidance here is confusing because it implies that by deriving classes you would might have multiple instances, but the other guidance states this should not be the case. Which is it? Let's say you do derive from HttpClient to have a FacebookHttpClient and GoogleHttpClient, but they share the same static instance, how would this work throughout a application, especially if you need to execute concurrent http requests? What if you need to switch between using default credentials or not? I do understand about using all the sockets, connection pooling, and HttpClient acting as a session, but the guidance should be more clear if multiple are in fact allowed in a derived scenario, and guidance on how to best implement that. It would seem to me that there are scenarios where more than one HttpClient instance is going to be required by an application?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: cbd5007c-82e5-a0c1-2156-7df4bbbcc2da
* Version Independent ID: eba98e03-3cc2-bebe-3dcd-45eda9c9c378
* Content: [HttpClient Class (System.Net.Http)](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netframework-4.7.2)
* Content Source: [xml/System.Net.Http/HttpClient.xml](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System.Net.Http/HttpClient.xml)
* Product: **dotnet-api**
* GitHub Login: @karelz
* Microsoft Alias: **ncldev**

rasitha1:
This documentation is misleading and definitely not the best practice in .net core world.
https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-6.0#remarks

![image](https://user-images.githubusercontent.com/3803499/158664160-027825bd-0536-407b-90d8-b94877bdfd65.png)

Shouldn't the documentation be based on this, if you are talking about .NET [Core], esp. recommending typed-clients?

https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests

![image](https://user-images.githubusercontent.com/3803499/158664754-a082d4a3-7289-401c-bf3d-90a201f1f5f6.png)

PR #7833, opened 2022-03-16
Update static usage remarks

rasitha1:
## Summary

Static usage is not the only best option and can be problematic for long running apps. Provide awareness of the `HttpClientFactory` to the reader and other safe ways of instantiating HttpClient along with using DI.

Fixes dotnet/dotnet-api-docs#1326

### DIFF
--- a/xml/System.Net.Http/HttpClient.xml
+++ b/xml/System.Net.Http/HttpClient.xml
@@ -70,7 +70,11 @@
 
 10. <xref:System.Net.Http.HttpClient.PatchAsync%2A>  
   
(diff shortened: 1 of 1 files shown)


CarnaViire:
Hi @rasitha1 
The team's current view is that `HttpClientFactory` is not a recommended way for a general case. Especially if the app doesn't use DI and has no need in it.
**The recommended way to overcome DNS problems is to use `PooledConnectionLifetime` property of `SocketsHttpHandler`.**
Unfortunately the latter is not well advertised in the moment.
We have an issue in our pipeline to rework current `HttpClient` documentation: https://github.com/dotnet/docs/issues/26510. Please feel free to discuss your suggestion in that issue.

PR #7854, opened 2022-03-22
Update for obsoletions

gewarren:
Contributes to dotnet/docs#66292

### DIFF
--- a/xml/System.Net.Security/EncryptionPolicy.xml
+++ b/xml/System.Net.Security/EncryptionPolicy.xml
@@ -82,7 +82,7 @@ The <xref:System.Net.Security.EncryptionPolicy> can be set to require, allow, or
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>Prefer that full encryption be used, but allow a NULL cipher (no encryption) if the server agrees. Starting in .NET 7, this field is obsolete.</summary>
+        <summary>Prefer that full encryption be used, but allow a NULL cipher (no encryption) if the server agrees. Starting in .NET 7, this member is obsolete.</summary>
       </Docs>
     </Member>
     <Member MemberName="NoEncryption">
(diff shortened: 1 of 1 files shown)


xml/System.Net.Security/EncryptionPolicy.xml
```diff
@@ -119,7 +118,7 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>Allow no encryption and request that a NULL cipher be used if the other endpoint can handle a NULL cipher.</summary>
+        <summary>Allow no encryption and request that a NULL cipher be used if the other endpoint can handle a NULL cipher. Starting in .NET 7, this field is obsolete.</summary>
```
rzikm, line 121: is "field" the right word here? I would use "value" personally, but if we call it field consistently in other docs, feel free to ignore this comment.

gewarren: Good question...it sounded a little weird to me too. What about using "member" instead? I'm not sure about "value" - is that commonly used when talking about enum members?

One reason to use "field" is that the heading in docs is "Fields", for example: https://docs.microsoft.com/en-us/dotnet/api/security.sslprotocolside?view=xamarin-ios-sdk-12#fields.

wfurt: It does seems like all the enum description is under "fields" header
https://docs.microsoft.com/en-us/dotnet/api/system.net.security.encryptionpolicy?view=net-6.0

General documentation seems to use terms "members" when talking about the set and "value" when talking about specific one
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum

gewarren: The C# spec says "Each enum member has an associated constant value." So I'll go with "member" in this case. Might be worth modifying the name of the heading from "Fields" to "Members" too cc @mimisasouvanh.

xml/System.Net.Security/EncryptionPolicy.xml
```diff
@@ -83,7 +82,7 @@
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>Prefer that full encryption be used, but allow a NULL cipher (no encryption) if the server agrees.</summary>
+        <summary>Prefer that full encryption be used, but allow a NULL cipher (no encryption) if the server agrees. Starting in .NET 7, this field is obsolete.</summary>
```
gewarren, line 85:
```suggestion
        <summary>Prefer that full encryption be used, but allow a NULL cipher (no encryption) if the server agrees. Starting in .NET 7, this member is obsolete.</summary>
```

xml/System.Net.Security/EncryptionPolicy.xml
gewarren, line 121:
```suggestion
        <summary>Allow no encryption and request that a NULL cipher be used if the other endpoint can handle a NULL cipher. Starting in .NET 7, this member is obsolete.</summary>
```

rzikm (approved):
LGTM

PR #7857, opened 2022-03-22
Should be format string in examples

DmitryMashkov-DM:
According to https://github.com/dotnet/roslyn-analyzers/issues/5626#issuecomment-1018003922

## Summary

Changed {Address} and {User} to indexes to not cause CA2254

### DIFF
--- a/xml/Microsoft.Extensions.Logging/LoggerExtensions.xml
+++ b/xml/Microsoft.Extensions.Logging/LoggerExtensions.xml
@@ -66,14 +66,14 @@
       </Parameters>
       <Docs>
         <param name="logger">The <see cref="T:Microsoft.Extensions.Logging.ILogger" /> to create the scope in.</param>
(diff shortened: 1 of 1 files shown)


eerhardt:
@maryamariyan - thoughts here? I thought we parsed the parameter name out of the message.

maryamariyan:
@DmitryMashkov-DM notice the comment link you shared is removing the `$` string interpolation. `{Address}` and `{User}` in those examples are message template arguments. We don't want to be using indexes.

maryamariyan:
maybe refer to `AppDescription` in comment

https://github.com/dotnet/roslyn-analyzers/issues/5626#issuecomment-1033240500

DmitryMashkov-DM:
> @DmitryMashkov-DM notice the comment link you shared is removing the `$` string interpolation. `{Address}` and `{User}` in those examples are message template arguments. We don't want to be using indexes.

Ahh, my bad. Thank you!

PR #7867, opened 2022-03-24
Add Ping exceptions

gewarren:
Contributes to dotnet/docs#28720


xml/System.Net.NetworkInformation/Ping.xml
```diff
@@ -793,12 +790,13 @@
   
  -or-  
   
- <paramref name="buffer" /> is <see langword="null" />, or the <paramref name="buffer" /> size is greater than 65500 bytes.</exception>
+ <paramref name="buffer" /> is <see langword="null" />.</exception>
```
gewarren, line 793: It seems like the upper size limit of buffer is covered by the ArgumentException, so I removed it as part of the ArgumentNullException causes. Please correct me if this is wrong.

rzikm (approved):

PR #7868, opened 2022-03-24
Fix return value formatting

gewarren:
Currently the line breaks aren't rendered, so it's hard to read.

[Clamp preview](https://review.docs.microsoft.com/en-us/dotnet/api/system.math.clamp?view=net-7.0&branch=pr-en-us-7868#system-math-clamp(system-byte-system-byte-system-byte))
[Table preview](https://review.docs.microsoft.com/en-us/dotnet/api/system.math.round?view=net-7.0&branch=pr-en-us-7868#which-method-do-i-call)


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

PR #7879, opened 2022-03-25
Clarify the meaning of MaxDegreeOfParallelism = -1

theodorzoulias:
The documentation of the [`ParallelOptions.MaxDegreeOfParallelism`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.paralleloptions.maxdegreeofparallelism) states that -1 means "no limit", which is true for all methods of the `Parallel` class except from the [`Parallel.ForEachAsync`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.parallel.foreachasync). This method treats the value -1 as the default (as omitting the `parallelOptions` parameter), which is `Environment.ProcessorCount`.

From [a comment](https://stackoverflow.com/questions/71610854/how-do-i-schedule-work-on-the-threadpool-without-causing-thread-starvation#comment126574403_71610854) in a recent StackOverflow question:

> However, as you point out, using -1 isn't unbounded (despite the docs)

### DIFF
--- a/xml/System.Threading.Tasks/ParallelOptions.xml
+++ b/xml/System.Threading.Tasks/ParallelOptions.xml
@@ -170,7 +170,7 @@ By default, methods on the <xref:System.Threading.Tasks.Parallel> class attempt 
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Threading.Tasks.ParallelOptions.MaxDegreeOfParallelism%2A> property affects the number of concurrent operations run by <xref:System.Threading.Tasks.Parallel> method calls that are passed this <xref:System.Threading.Tasks.ParallelOptions> instance. A positive property value limits the number of concurrent operations to the set value. If it is -1,  there is no limit on the number of concurrently running operations.  
+ The <xref:System.Threading.Tasks.ParallelOptions.MaxDegreeOfParallelism%2A> property affects the number of concurrent operations run by <xref:System.Threading.Tasks.Parallel> method calls that are passed this <xref:System.Threading.Tasks.ParallelOptions> instance. A positive property value limits the number of concurrent operations to the set value. If it is -1,  there is no limit on the number of concurrently running operations (with the exception of the <xref:System.Threading.Tasks.Parallel.ForEachAsync%2A> method, where -1 means <xref:System.Environment.ProcessorCount%2A>).  
   
  By default, <xref:System.Threading.Tasks.Parallel.For%2A> and <xref:System.Threading.Tasks.Parallel.ForEach%2A> will utilize however many threads the underlying scheduler provides, so changing <xref:System.Threading.Tasks.ParallelOptions.MaxDegreeOfParallelism%2A> from the default only limits how many concurrent tasks will be used.  
   


kouvel:
It looks to me like they all work the same way when -1 is specified:

https://github.com/dotnet/runtime/blob/9047692595c6bef787e93883e92603388339c1c1/src/libraries/System.Threading.Tasks.Parallel/src/System/Threading/Tasks/Parallel.cs#L920-L922

https://github.com/dotnet/runtime/blob/9047692595c6bef787e93883e92603388339c1c1/src/libraries/System.Threading.Tasks.Parallel/src/System/Threading/Tasks/Parallel.ForEachAsync.cs#L69-L72

If -1 is specified, it first checks if the specified scheduler has a specific max concurrency level specified, and if so uses that. Otherwise, it defaults to `Environment.ProcessorCount`.

https://github.com/dotnet/runtime/blob/9047692595c6bef787e93883e92603388339c1c1/src/libraries/System.Threading.Tasks.Parallel/src/System/Threading/Tasks/Parallel.cs#L111-L117

So it seems the "no limit" part in the docs should be generally updated to more clearly describe the behavior.

theodorzoulias:
> It looks to me like they all work the same way when -1 is specified

@kouvel the program below demostrates the effect of `MaxDegreeOfParallelism = -1` to the four `Parallel` methods: `For`, `ForEach`, `Invoke` and `ForEachAsync`. It's clear that the `ForEachAsync` behaves differently than the other three, provided that the `ThreadPool` can create enough threads immediately on demand, so that the `ThreadPool` availability does not become a limiting factor.

```C#
using System;
using System.Linq;
using System.Threading;
using System.Threading.Tasks;

static class Program
{
    static void Main()
    {
        object locker = new object();
        int currentConcurrency = 0;
        int maximumConcurrency = 0;
        Action action = new Action(() =>
        {
            lock (locker)
            {
                currentConcurrency++;
                maximumConcurrency = Math.Max(currentConcurrency, maximumConcurrency);
            }
            Thread.Sleep(50);
            lock (locker) currentConcurrency--;
        });

        const int itemsCount = 100;
        ThreadPool.SetMinThreads(itemsCount, itemsCount);
        var options = new ParallelOptions() { MaxDegreeOfParallelism = -1 };
        Console.WriteLine($"Items.Count: {itemsCount}");
        Console.WriteLine($"MaxDegreeOfParallelism: {options.MaxDegreeOfParallelism}");
        Console.WriteLine($"Environment.ProcessorCount: {Environment.ProcessorCount}");
        ThreadPool.GetMinThreads(out var workerThreads, out _);
        Console.WriteLine($"ThreadPool.GetMinThreads: {workerThreads}");
        Console.WriteLine();
        Console.WriteLine($"Method                 Maximum Concurrency");

        currentConcurrency = maximumConcurrency = 0;
        Parallel.For(0, itemsCount, options, _ => action());
        Console.WriteLine($"Parallel.For           {maximumConcurrency}");

        currentConcurrency = maximumConcurrency = 0;
        Parallel.ForEach(Enumerable.Range(0, itemsCount), options, _ => action());
        Console.WriteLine($"Parallel.ForEach       {maximumConcurrency}");

        currentConcurrency = maximumConcurrency = 0;
        Parallel.Invoke(options, Enumerable.Range(0, itemsCount).Select(_ => action).ToArray());
        Console.WriteLine($"Parallel.Invoke        {maximumConcurrency}");

        currentConcurrency = maximumConcurrency = 0;
        Parallel.ForEachAsync(Enumerable.Range(0, itemsCount), options, (_, __) =>
        {
            action(); return ValueTask.CompletedTask;
        }).Wait();
        Console.WriteLine($"Parallel.ForEachAsync  {maximumConcurrency}");
    }
}
```

Output:

```
Items.Count: 100
MaxDegreeOfParallelism: -1
Environment.ProcessorCount: 4
ThreadPool.GetMinThreads: 100

Method                 Maximum Concurrency
Parallel.For           100
Parallel.ForEach       100
Parallel.Invoke        100
Parallel.ForEachAsync  4
```

([Live demo](https://dotnetfiddle.net/Vd0Wb1))

theodorzoulias:
Something else that I've noticed just today is that the unbounded `Parallel.ForEach` no longer saturates the `ThreadPool` as it did in the previous .NET version (.NET 5). It uses all the `ThreadPool` threads that are available at the moment it started, but it doesn't induce the creation of more threads while it's running, beyond the limit specified by the `SetMinThreads` configuration. So the `ThreadPool` is able to satisfy concurrent requests from elsewhere (from `Timer` event handlers for example), which is a good thing IMHO. As far as I understand, effectively the `MaxDegreeOfParallelism = -1` configuration now means *"use all the threads that are currently available, and potentially create more up to the `SetMinThreads` limit, but no more"*.

In any case, these observations are probably beyond the scope of this PR. The focus here is on the new `Parallel.ForEachAsync` API.

kouvel:
Thanks, yes I see that although `Parallel.For` splits the range into `Environment.ProcessorCount` subranges by default for `-1`, the task replicator keeps replicating so other thread pool worker threads will also participate.

kouvel:
The behavior seems reasonable to me despite the slight difference, since more parallelization it seems would cause more contention. The change looks good to me based on that. @dotnet/area-system-threading-tasks, is this the expected parallelization behavior of `Parallel.ForEachAsync`?

theodorzoulias:
@kouvel take a look at [this comment](https://github.com/dotnet/runtime/issues/1946#issuecomment-581472891) by @stephentoub, in response to @GSPP:

> In general, what's optimal is going to depend on many factors, some (most?) of which aren't known in most cases statically. You argue that `Environment.ProcessorCount` and unbounded are going to be wrong. I argue that it's going to be close to impossible for a developer to determine and pass in the "optimal" dop, except in very constrained cases where they're intimately aware of everything in the application, the hardware, the network, etc. If we force a dop to be provided, I can almost guarantee you we're going to see the vast majority of call sites passing in either `Environment.ProcessorCount` (or a constant factor of it) or some arbitrary number based on little more than sticking one's index finger in the air to see which way the wind is blowing. Allowing a dop to be provided is great, but I'm not seeing how not requiring one is any more dangerous than taking one; in fact, whether we take advantage of it or not, not taking one gives us some flexibility we lose if a strict bound is provided.

Personally I agree with this opinion. I only hope that the same value (`Environment.ProcessorCount`) was the default for the other three `Parallel` methods as well!

buyaa-n:
cc @stephentoub

stephentoub:
This looks fine.  For/ForEach/Invoke all invoke operations synchronously, thus relying on threads, and thus relying on the thread pool to gate concurrency.  ForEachAsync is different: the majority of the time, work items shouldn't be on thread pool threads, since the expectation is most of the time is spent await'ing for I/O and the like.  If -1 were to mean "schedule an unbounded number of work items and rely on the thread pool to provide backpressure", ForEachAsync would often very quickly end up running every work item concurrently. A source of a million URLs passed to `await Parallel.ForEachAsync(urls, url => Process(await http.GetAsync(url))` would likely result in it trying to process a million requests concurrently.

PR #7881, opened 2022-03-25
Update ColorChannelFlag.xml

ghost:
The believe the "Remarks" are incorrect.  The signature for 'ColorChannelFlag' does not have the FlagsAttribute applied.  Also, it would not make sense.  The enumeration numbers are not powers of 2.  For example, combining ColorChannelM (1) and ColorChannelY (2) would not produce ColorChannelK (3).

## Summary

Describe your changes here.


gewarren:
Thanks @nick1941 - it looks like you are correct: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Drawing.Common/src/System/Drawing/Imaging/ColorChannelFlags.cs. We should just remove the remarks section entirely (which I've done).

gewarren (approved):

PR #7888, opened 2022-03-28
Fix `Socket.Close` `timeout` parameter doc

deeprobin:
## Summary

As described in https://github.com/dotnet/runtime/pull/64860#discussion_r836680375, this PR provides a fix for `seconds` -> `milliseconds`, since `setsockopt` expects a milliseconds value.

/cc @danmoseley 
/cc @stephentoub


danmoseley (approved):
thanks. Please ensure that the XML comments in the code also say milliseconds.

deeprobin:
> thanks. Please ensure that the XML comments in the code also say milliseconds.

In the current code this is undocumented.

PR #7896, opened 2022-03-30
Update equals_ex3.cs

IvanStoychev:
## Summary

Fixed wrong output shown in commented part of snippet.

Upon running the snippet code an output, different from the one in the documentation, is produced.


gewarren:
Closely related to #7897.

Thanks for catching this. This change is likely due to https://docs.microsoft.com/en-us/dotnet/core/compatibility/globalization/5.0/icu-globalization-api, because if you run the code in a .NET Core 3.1 project, the output matches the existing text.

In this case, a larger overhaul of the example and corresponding intro text is needed. The F# example will also need to be updated.

IvanStoychev:
I was not aware of the ICU change, I agree it's probably it.

In my opinion, the best option would be to have different versions of the snippet for the 3.1 and 6 versions, which would be toggled from the version dropdown in the upper-left corner.

Can I help with the overhaul or is this something that needs to be done internally?

gewarren:
> In my opinion, the best option would be to have different versions of the snippet for the 3.1 and 6 versions, which would be toggled from the version dropdown in the upper-left corner.

Unfortunately we don't have the ability to version snippets like that in the API reference docs.

> Can I help with the overhaul or is this something that needs to be done internally?

You can definitely help with it. For this example, I'd like to show two strings that compare as equal in one culture but not equal in another culture. (Instead of the "ae" examples.)

IvanStoychev:
> Unfortunately we don't have the ability to version snippets like that in the API reference docs.

Would you like me to look into it?

> You can definitely help with it. For this example, I'd like to show two strings that compare as equal in one culture but not equal in another culture. (Instead of the "ae" examples.)

I'm looking into this. When I have something I will update the pull request.

IvanStoychev:
I have been trying to find a string that will be differently equal in two cultures for two days now and I have absolutely nothing to show for it.

In .Net 6 even the characters '1' and '๑' are equal under "en-US" culture when using "CurrentCulture" comparison.

I've tried:
- ‘ß’ (\u00df) and "ss" in "en-US" and "de" cultures,
- "\u00e5" and "a" in "en-US" and "da" cultures
- “a\u0325\u0300” and “a\u0300\u0325” after I read [this section of the ICU string search model](https://unicode-org.github.io/icu/userguide/collation/string-search.html#icu-string-search-model)

The only partial success I've had was with "i" - "İ" and "ı" - "I", which are equal under "tr-TR" culture when using "CurrentCultureIgnoreCase", but not "CurrentCulture". The same strings are not equal under "en-US" culture.

So far I have found nothing that is equal under one culture when using "CurrentCulture" but not under another.

What are your thoughts, @gewarren ?

gewarren:
@tarekgh Do you know of two strings that compare equal using `String.Equals` in one culture but not in another culture using ICU globalization? We'd like to show it in an example.

tarekgh:
@gewarren here are two examples:

```C#
            CultureInfo.CurrentCulture = CultureInfo.GetCultureInfo("th-TH"); // Thai Culture
            Console.WriteLine(string.Equals("a", "a-", StringComparison.CurrentCulture)); // print true
            CultureInfo.CurrentCulture = CultureInfo.GetCultureInfo("en-US"); // English Culture
            Console.WriteLine(string.Equals("a", "a-", StringComparison.CurrentCulture)); // print false

            CultureInfo.CurrentCulture = CultureInfo.GetCultureInfo("tr-TR"); // Turkish Culture
            Console.WriteLine(string.Equals("i", "\u0130", StringComparison.CurrentCultureIgnoreCase)); // print true
            CultureInfo.CurrentCulture = CultureInfo.GetCultureInfo("en-US"); // English Culture
            Console.WriteLine(string.Equals("i", "\u0130", StringComparison.CurrentCultureIgnoreCase)); // print false
```

gewarren:
Thank you so much @tarekgh!

@IvanStoychev Do you want to update the example with Tarek's suggestion?

gewarren:
Thank you @IvanStoychev! Can you also update the intro text in:

- https://docs.microsoft.com/en-us/dotnet/api/system.string.equals?view=net-6.0#system-string-equals(system-string-system-string-system-stringcomparison)
- https://docs.microsoft.com/en-us/dotnet/api/system.stringcomparison?view=net-6.0#examples

xml/System/String.xml
```diff
@@ -5592,7 +5592,7 @@ Invalid sequences are represented in the enumeration by <xref:System.Text.Rune.R
  The `comparisonType` parameter indicates whether the comparison should use the current or invariant culture, honor or ignore the case of the two strings being compared, or use word or ordinal sort rules.
 
 ## Examples
- The following example compares four sets of words by using each member of the <xref:System.StringComparison> enumeration.  The comparisons use the conventions of the English (United States) and Sami (Upper Sweden) cultures. Note that the strings "encyclopædia" and "encyclopaedia" are considered equivalent in the en-US culture but not in the Sami (Northern Sweden) culture.
+ The following example compares three sets of strings by using each member of the <xref:System.StringComparison> enumeration. The comparisons use the conventions of the English (United States), Thai (Thailand) and Turkish (Turkey) cultures. Note that the strings "a" and "a-" are considered equivalent in the "th-TH" culture but not in the others, while "i" and "İ" are considered equivalent in the "tr-TR" culture when case is ignored but not in the other cultures.
```
gewarren, line 5595:
```suggestion
 The following example compares three sets of strings by using each member of the <xref:System.StringComparison> enumeration. The comparisons use the conventions of the English (United States), Thai (Thailand), and Turkish (Turkey) cultures. Note that the strings "a" and "a-" are considered equivalent in the "th-TH" culture but not in the others, while "i" and "İ" are considered equivalent in the "tr-TR" culture when case is ignored but not in the other cultures.
```

xml/System/StringComparison.xml
```diff
@@ -69,7 +69,7 @@
...
   
 ## Examples  
- The following example compares four sets of words by using each member of the <xref:System.StringComparison> enumeration.  The comparisons use the conventions of the English (United States) and Sami, Northern (Sweden) cultures. Note that the strings "encyclopædia" and "encyclopedia" are considered equivalent in the en-US culture but not in the Sami, Northern (Sweden) culture.  
+ The following example compares three sets of strings by using each member of the <xref:System.StringComparison> enumeration. The comparisons use the conventions of the English (United States), Thai (Thailand) and Turkish (Turkey) cultures. Note that the strings "a" and "a-" are considered equivalent in the "th-TH" culture but not in the others, while "i" and "İ" are considered equivalent in the "tr-TR" culture when case is ignored but not in the other cultures.
```
gewarren, line 72:
```suggestion
 The following example compares three sets of strings by using each member of the <xref:System.StringComparison> enumeration. The comparisons use the conventions of the English (United States), Thai (Thailand), and Turkish (Turkey) cultures. Note that the strings "a" and "a-" are considered equivalent in the "th-TH" culture but not in the others, while "i" and "İ" are considered equivalent in the "tr-TR" culture when case is ignored but not in the other cultures.
```

gewarren (approved):

PR #7897, opened 2022-03-30
Update equals_ex4.cs

IvanStoychev:
## Summary

Fixed wrong output shown in commented part of snippet.

Upon running the snippet code an output, different from the one in the documentation, is produced.

### DIFF
--- a/snippets/fsharp/System/String/Equals/fs.fsproj
+++ b/snippets/fsharp/System/String/Equals/fs.fsproj
@@ -7,7 +7,6 @@
   <ItemGroup>
     <Compile Include="equalsex1.fs" />
     <Compile Include="eqcmp.fs" />
-    <Compile Include="equals_ex4.fs" />
     <Compile Include="equals_ex3.fs" />
     <Compile Include="equals.fs" />
   </ItemGroup>


IvanStoychev:
I was not aware of the ICU change, I agree it's probably it.

In my opinion, the best option would be to have different versions of the snippet for the 3.1 and 6 versions, which would be toggled from the version dropdown in the upper-left corner.

Can I help with the overhaul or is this something that needs to be done internally?

gewarren:
For this example, it doesn't add anything above the first example in the Examples section, so I suggest we delete the snippet files and intro text. Would you like to help with that @IvanStoychev ?

IvanStoychev:
Absolutely, @gewarren !

Just committed the changes. Take a look and let me know if it's OK, since I'm not sure how the docs are put together or how to debug / visualise them.

On an unrelated note - should I mention people in comments?
It is my understanding participating in a pull request / thread subscribes a person to it.

gewarren (approved):
Great, thank you!

gewarren:
> Just committed the changes. Take a look and let me know if it's OK, since I'm not sure how the docs are put together or how to debug / visualise them.

Unfortunately there's no way for external contributors to visualize the changes, but it looks good, and the build will catch any issues if there are other places that reference the deleted snippets.

> On an unrelated note - should I mention people in comments? It is my understanding participating in a pull request / thread subscribes a person to it.

Yes, if you participate then you're automatically notified (usually - it may be configurable).

gewarren:
Looks like we need to remove the F# source file from the F# project.

IvanStoychev:
Yes, was just looking into that.
Would you like me to do it?

PR #7695, opened 2022-02-06
Update enumerable.cs

HamedMoghadasi:
## Summary
The _**Clump<T>**_ class bracket was closed in the wrong position, so because of that one of its methods, **_AsEnumerableEx1()_**, was out of class.

PR #7905, opened 2022-04-01
Fix up Enumerable snippet directories

gewarren:
Moves a project and source files to a subdirectory so there's not a nested project file.
Also updates snippet tag names to be more intuitive.

Fixes project error in #7695.


BillWagner (approved):
These look great @gewarren 
You can :shipit: when ready.

gewarren:
Thanks Bill!

PR #7912, opened 2022-04-02
BlockingCollection<T>.BoundedCapacity correction

theodorzoulias:
The value of the [`BoundedCapacity`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.blockingcollection-1.boundedcapacity) property of an unbounded `BlockingCollection<T>` is -1, not `int.MaxValue`.

Mentioned in the [source code](https://github.com/dotnet/runtime/blob/2eddb3dd24b0191af2a9d3e0bc6bf25f7cdf63f5/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/BlockingCollection.cs#L61):

```C#
/// <value>The bounded capacity of this collection, or -1 if no bound was supplied.</value>
```

Fixes [#25747](https://github.com/dotnet/runtime/issues/25747)

### DIFF
--- a/xml/System.Collections.Concurrent/BlockingCollection`1.xml
+++ b/xml/System.Collections.Concurrent/BlockingCollection`1.xml
@@ -606,7 +606,7 @@
       </ReturnValue>
       <Docs>
         <summary>Gets the bounded capacity of this <see cref="T:System.Collections.Concurrent.BlockingCollection`1" /> instance.</summary>
-        <value>The bounded capacity of this collection, or int.MaxValue if no bound was supplied.</value>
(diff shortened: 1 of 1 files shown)


gewarren (approved):
That looks correct to me based on https://github.com/dotnet/runtime/blob/2eddb3dd24b0191af2a9d3e0bc6bf25f7cdf63f5/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/BlockingCollection.cs#L49.

eiriktsarpalis (approved):

eiriktsarpalis:
Thanks!

PR #7914, opened 2022-04-03
System.AppDomain F# snippets

albert-du:
Adds F# snippets for AppDomain, AppDomainSetup, and AppDomainUnloadedException


BillWagner (approved):
Thanks again @albert-du 

These look great. I'll :shipit: now.

PR #7924, opened 2022-04-04
Clarify EventLevel.LogAlways

noahfalk:
Fixes https://github.com/dotnet/diagnostics/issues/2445

Using EventLevel.LogAlways to enable events is a special case. The standard pattern says
that events of level <= the filter level should be enabled which would mean only events at level
LogAlways would be included. However specifying LogAlways enables events of all levels.


noahfalk:
cc @dotnet/dotnet-diag

gewarren (approved):

hoyosjs (approved):
Sad approval

noahfalk:
Thanks for review!
@gewarren - Could you hit the merge button for me? I don't have permission to do that.

PR #7934, opened 2022-04-08
correct position of example in text

trivalik:
Correct the direction word for where the example could be found.

### DIFF
--- a/xml/System.Windows.Documents/InlineUIContainer.xml
+++ b/xml/System.Windows.Documents/InlineUIContainer.xml
@@ -33,7 +33,7 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- A <xref:System.Windows.UIElement> element (i.e. a <xref:System.Windows.Controls.Button>) may be embedded directly in flow content by enclosing it in a <xref:System.Windows.Documents.InlineUIContainer> (see example below).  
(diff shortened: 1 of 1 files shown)


xml/System.Windows.Documents/InlineUIContainer.xml
```diff
@@ -33,7 +33,7 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- A <xref:System.Windows.UIElement> element (i.e. a <xref:System.Windows.Controls.Button>) may be embedded directly in flow content by enclosing it in a <xref:System.Windows.Documents.InlineUIContainer> (see example below).  
+ A <xref:System.Windows.UIElement> element (i.e. a <xref:System.Windows.Controls.Button>) may be embedded directly in flow content by enclosing it in a <xref:System.Windows.Documents.InlineUIContainer> (see example above).
```
gewarren, line 36:
```suggestion
 A <xref:System.Windows.UIElement> element (for example, a <xref:System.Windows.Controls.Button>) may be embedded directly in flow content by enclosing it in a <xref:System.Windows.Documents.InlineUIContainer> (see the [example](#example)).
```

gewarren: It's better not to use a directional word at all, but good catch.

PR #7938, opened 2022-04-08
Update code to match comment

3141592654:
Right now, the example referred to in "The following example sets MaxReceivedMessageSize to 1,000,000 bytes" sets MaxReceivedMessageSize to 131,072 bytes.

### DIFF
--- a/xml/System.ServiceModel/BasicHttpBinding.xml
+++ b/xml/System.ServiceModel/BasicHttpBinding.xml
@@ -758,8 +758,7 @@ The following example sets MaxReceivedMessageSize to 1,000,000 bytes.
 ```csharp
 BasicHttpBinding binding = new BasicHttpBinding();
 
-// Use double the default value
-binding.MaxReceivedMessageSize = 65536 * 2;
+binding.MaxReceivedMessageSize = 1000000;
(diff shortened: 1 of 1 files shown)


3141592654:
@mconnew Can I get a review on this?

mconnew:
This change is correct based on the source documentation xml. I'm having difficulty finding the actual documentation page though. Clicking the MaxReceivedMessageSize property in the list of properties takes me to the documentation for the base class HttpBindingBase.

mconnew (approved):

3141592654:
> ...I'm having difficulty finding the actual documentation page though...

https://docs.microsoft.com/en-us/dotnet/api/system.servicemodel.basichttpbinding.maxreceivedmessagesize?view=netframework-4.0

If you go to the Preview URL and set the Version to ".NET Framework 4.0", you will see the page. (This bit of the documentation is a little old).

mconnew:
@3141592654, thanks. I didn't realize that the class hierarchy changed between 4.0 and 4.5 as it was an in-place replacement, and I wouldn't have expected moving a property to a base class would have been binary compatible. Thanks.

PR #7940, opened 2022-04-08
Fix WriteEventCore string guidance

noahfalk:
I made some mistakes when adding recent guidance to EventSource WriteEventCore docs that were pointed out in feedback.
This addresses those issues + eliminates a line of code that was unused I noticed while making the change.

### DIFF
--- a/xml/System.Diagnostics.Tracing/EventSource.xml
+++ b/xml/System.Diagnostics.Tracing/EventSource.xml
@@ -3579,11 +3579,12 @@ desc.DataPointer = (IntPtr)(&fileTime);
 desc.Size = 8;
 
 // string arg
-fixed(byte* ptr = arg)
+fixed(char* ptr = arg)
 {
-    int length = arg.Length;
-    desc.DataPointer = ptr;
-    desc.Size = (arg.Length + 1) * 2;  // strings use 2 byte per char UTF16 encoding and a null terminator at the end
+    desc.DataPointer = (IntPtr)ptr;
+    // strings use 2 byte per char UTF16 encoding and a null terminator at the end
+    // only strings without embedded null characters are supported
+    desc.Size = (arg.Length + 1) * 2;  
 }
 
 // byte[] arg


noahfalk:
cc @dotnet/dotnet-diag

xml/System.Diagnostics.Tracing/EventSource.xml
```diff
@@ -3579,11 +3579,12 @@ desc.DataPointer = (IntPtr)(&fileTime);
...
 // string arg
-fixed(byte* ptr = arg)
+fixed(char* ptr = arg)
 {
-    int length = arg.Length;
-    desc.DataPointer = ptr;
-    desc.Size = (arg.Length + 1) * 2;  // strings use 2 byte per char UTF16 encoding and a null terminator at the end
+    desc.DataPointer = (IntPtr)ptr;
+    // strings use 2 byte per char UTF16 encoding and a null terminator at the end
+    // only strings without embedded null characters are supported
```
josalem, line 3586: We can say they are unsupported but technically they will still be used. Certain parsers (all of the ones I'm aware of have this behavior) just won't know what to do when they encounter them.

KalleOlaviNiemitalo: Does ETW give [PEVENT_RECORD_CALLBACK](https://docs.microsoft.com/windows/win32/api/evntrace/nc-evntrace-pevent_record_callback) any information about the sizes of the individual data items, or does it just concatenate all of them to EVENT_RECORD::UserData?

noahfalk: My understanding is that the encoder concatenates Size bytes from each DataPointer into a single contiguous buffer and then throws away the size information. The parser has to reconstruct the argument boundaries on its own. The code in TraceEvent which does that reconstruction is here: https://github.com/microsoft/perfview/blob/f57006741703fcb33bcb5bd9ae6835b95b33fe31/src/TraceEvent/DynamicTraceEventParser.cs#L1251

@josalem - do you think there is a better comment we should make? 
If someone did this it is impossible for the parser to know which how to reconstruct it correctly in the general case:
```
WriteEvent("hi\0there", "jo\0salem");
WriteEvent("hi\0there\0jo", "salem");
```
The parser sees a buffer "hi\0there\0jo\0salem\0", in both cases and has no other information to know which split is the correct one.

josalem: After some thought, I don't think going into detail adds anything. Stating it's unsupported is sufficient to me. If we encounter enough questions down the road about why it's unsupported or the repercussions of doing it anyway, we can document further.

josalem (approved):

BillWagner (approved):
Thanks @noahfalk 

This looks great. I'll :shipit: now.

Issue #7777, opened 2022-02-28
Guidance for ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult Method is too absolute


JimDaly:
The remarks in this reference topic:  [ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult Method](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.configuredtaskawaitable-1.configuredtaskawaiter.getresult) says: 

`This method is intended for compiler use; do not use it directly in your code.`

While this is generally a good recommendation when working with asynchronous modern web applications, the absolute statement makes developers feel it is _never_ appropriate. 

For the use case in our product, we recommend people use this. See [ Microsoft Dataverse Plug-ins > Troubleshoot plug-ins > Unhandled exception in plugin ](https://docs.microsoft.com/en-us/powerapps/developer/data-platform/troubleshoot-plug-in#unhandled-exception-in-plugin). The code they are writing is single-threaded, synchronous using HttpClient.
The GetResult method allows better method to get the specific error caused by a failure.

Recommend description be changed to something like '_This method isn’t recommended for modern applications designed to use asynchronous patterns_.'

joperezr:
cc: @stephentoub

JimDaly:
Thanks. Agree.
We had a confirmation of this in another issue. I'm happy to close this.

PR #7941, opened 2022-04-08
Update ConfiguredTaskAwaitable`1+ConfiguredTaskAwaiter.xml

JimDaly:
Please see https://github.com/dotnet/dotnet-api-docs/issues/7777

## Summary

Changed the Remarks for the GetResult method.

### DIFF
--- a/xml/System.Runtime.CompilerServices/ConfiguredTaskAwaitable`1+ConfiguredTaskAwaiter.xml
+++ b/xml/System.Runtime.CompilerServices/ConfiguredTaskAwaitable`1+ConfiguredTaskAwaiter.xml
@@ -123,7 +123,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- This method is intended for compiler use; do not use it directly in your code.  
+This method isn’t recommended for modern applications designed to use asynchronous patterns.
   
  ]]></format>
         </remarks>


BillWagner:
ping @stephentoub for comments.

stephentoub:
No one should ever be calling `.ConfigureAwait(anything).GetAwaiter().GetResult()`; there's literally zero benefit to using that over `.GetAwaiter().GetResult()`.  The purpose of `ConfigureAwait` is to influence how the `{Unsafe}OnCompleted` method that the compiler binds to with `await` behaves, and it has no impact on GetResult.  So we shouldn't be updating the docs for the configured awaiters here; they are accurate, they are intended only for compiler use.

Now, there's the related question of "what about `GetAwaiter().GetResult()` on a `Task` rather than on a configured awaiter, as those docs also say the same thing".  When all of this support was introduced, the intent was in fact that no one should ever be using these directly, that they were purely for compiler consumption.  However, there's a behavioral difference between `task.Wait()` and `task.GetAwaiter().GetResult()`, which is that, while functionally identical in terms of how they wait, the latter propagates the first exception that faulted/canceled the task whereas the former wraps the one or more exceptions into an AggregateException and throws that wrapper.  As such, developers that want the original exception propagated often use `.GetAwaiter().GetResult()` instead of `Wait`.  At this point, the right thing for `TaskAwaiter` and `TaskAwaiter<T>` is probably to just strike those sentence from the docs, though some extra verbiage around the distinction from Task.Wait would be welcome.  No one should be using `GetAwaiter().GetResult()` on any of the `ValueTask` variants for all the reasons discussed in https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/.

JimDaly:
Closing this as @stephentoub  mentioned it shouldn't be used.

PR #7946, opened 2022-04-11
BlockingCollection<T>.Take method corrections, related to exceptions

theodorzoulias:
Contrary to what the docs say, the public [`Take()`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.blockingcollection-1.take#system-collections-concurrent-blockingcollection-1-take) method of the [`BlockingCollection<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.blockingcollection-1) class doesn't throw `OperationCanceledException`.

Corrections made according to the embedded XML documentation in the [source code](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/BlockingCollection.cs) of the class.

Fixes [#67812](https://github.com/dotnet/runtime/pull/67812)


krwq (approved):

krwq:
Thanks @theodorzoulias!

PR #7947, opened 2022-04-11
System.Tuple<T1,T2,T3,T4> F# snippets

albert-du:
Adds F# snippets for Tuple<T1,T2,T3,T4>


BillWagner (approved):
Thanks again @albert-du 

This is great work, as always.  I'll :shipit: now.

PR #7948, opened 2022-04-11
System.Tuple<T1,T2,T3,T4,T5> F# snippets

albert-du:
Adds F# snippets for Tuple<T1,T2,T3,T4,T5>


BillWagner (approved):
Thanks again @albert-du 

These look great. I'll :shipit: now.

PR #7950, opened 2022-04-12
Update AppDomainSetup.xml

lindexi:
## Summary

The AppDomainSetup.ApplicationBase is get only in dotnet 6


gewarren (approved):
Thanks @lindexi

PR #7954, opened 2022-04-13
RunAsync waits for hosted services to stop

pedrobsaila:
## Summary

This change makes explicit the fact that Run/RunAsync waits for all the hosted services to be stopped, before returning.

Fixes dotnet/runtime#58675

### DIFF
--- a/xml/Microsoft.Extensions.Hosting/HostingAbstractionsHostExtensions.xml
+++ b/xml/Microsoft.Extensions.Hosting/HostingAbstractionsHostExtensions.xml
@@ -50,7 +50,7 @@
       </Parameters>
       <Docs>
         <param name="host">The <see cref="T:Microsoft.Extensions.Hosting.IHost" /> to run.</param>
-        <summary>Runs an application and blocks the calling thread until host shutdown.</summary>
+        <summary>Runs an application and blocks the calling thread until host shutdown is triggered and all <see cref="T:Microsoft.Extensions.Hosting.IHostedService" /> instances are stopped.</summary>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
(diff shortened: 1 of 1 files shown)


xml/Microsoft.Extensions.Hosting/HostingAbstractionsHostExtensions.xml
```diff
@@ -50,7 +50,7 @@
       </Parameters>
       <Docs>
         <param name="host">The <see cref="T:Microsoft.Extensions.Hosting.IHost" /> to run.</param>
-        <summary>Runs an application and blocks the calling thread until host shutdown.</summary>
+        <summary>Runs an application and blocks the calling thread until host shutdown is triggered and all <see cref="T:Microsoft.Extensions.Hosting.IHostedService" /> instances are stopped.</summary>
```
maryamariyan, line 53: The issue you linked is for `RunAsync` not `Run`. Let's only update the doc over `RunAsync`.

Do you have a code repro showing the same holds true for `Run`?

cc @eerhardt

maryamariyan (commented):
The issue you linked is for RunAsync not Run. Let's only update the doc over RunAsync.

Do you have a code repro showing the same holds true for Run?

cc @eerhardt

xml/Microsoft.Extensions.Hosting/HostingAbstractionsHostExtensions.xml
pedrobsaila: Run calls RunAsync

https://github.com/dotnet/runtime/blob/2be87c9c976c0cd21b66ea12ae406a559aacaac0/src/libraries/Microsoft.Extensions.Hosting.Abstractions/src/HostingAbstractionsHostExtensions.cs#L44-L51

maryamariyan (approved):

PR #7965, opened 2022-04-17
Update DbDataReader.xml

Grauenwolf:
The buffer parameter is marked as nullable, but the documentation doesn't say why.

In researching SqlDataReader, I discovered this additional behavior.

## Summary

Add information on what happens when buffer is null.

### DIFF
--- a/xml/System.Data.Common/DbDataReader.xml
+++ b/xml/System.Data.Common/DbDataReader.xml
@@ -628,11 +628,11 @@
       <Docs>
         <param name="ordinal">The zero-based column ordinal.</param>
         <param name="dataOffset">The index within the row from which to begin the read operation.</param>
-        <param name="buffer">The buffer into which to copy the data.</param>
+        <param name="buffer">The buffer into which to copy the data or null.</param>
         <param name="bufferOffset">The index with the buffer to which the data will be copied.</param>
         <param name="length">The maximum number of bytes to read.</param>
         <summary>When overridden in a derived class, reads a specified number of bytes from the specified column starting at a specified index and writes them to a buffer starting at a specified position in the buffer.</summary>
-        <returns>The actual number of bytes read.</returns>
+        <returns>The actual number of bytes read. If buffer is null, the number of available bytes is returned instead.</returns>
         <remarks>To be added.</remarks>
         <exception cref="T:System.IndexOutOfRangeException">The column index is out of range.</exception>
         <exception cref="T:System.InvalidCastException">The specified cast is not valid.</exception>


xml/System.Data.Common/DbDataReader.xml
```diff
@@ -628,11 +628,11 @@
       <Docs>
         <param name="ordinal">The zero-based column ordinal.</param>
         <param name="dataOffset">The index within the row from which to begin the read operation.</param>
-        <param name="buffer">The buffer into which to copy the data.</param>
+        <param name="buffer">The buffer into which to copy the data or null.</param>
         <param name="bufferOffset">The index with the buffer to which the data will be copied.</param>
         <param name="length">The maximum number of bytes to read.</param>
         <summary>When overridden in a derived class, reads a specified number of bytes from the specified column starting at a specified index and writes them to a buffer starting at a specified position in the buffer.</summary>
-        <returns>The actual number of bytes read.</returns>
+        <returns>The actual number of bytes read. If buffer is null, the number of available bytes is returned instead.</returns>
```
roji, line 635: This is somewhat ambiguous - note the remark in the SqlClient docs:

> If you pass a buffer that is null, [GetBytes](https://docs.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqldatareader.getbytes?view=sqlclient-dotnet-standard-4.1) returns the length of the entire field in bytes, not the remaining size based on the buffer offset parameter.

```suggestion
        <returns>The actual number of bytes read. If buffer is <see langword="null" />, returns the length of the entire field in bytes.</returns>
```

xml/System.Data.Common/DbDataReader.xml
```diff
@@ -628,11 +628,11 @@
       <Docs>
         <param name="ordinal">The zero-based column ordinal.</param>
         <param name="dataOffset">The index within the row from which to begin the read operation.</param>
-        <param name="buffer">The buffer into which to copy the data.</param>
+        <param name="buffer">The buffer into which to copy the data or null.</param>
```
roji, line 631:
```suggestion
        <param name="buffer">The buffer into which to copy the data. If <see langword="null" />, the method returns the length of the entire field in bytes.</param>
```

roji (commented):
Thanks, good catch - I suspect this was originally a SqlClient feature not necessarily supported elsewhere; but because of NRT annotation, the parameter had to be made nullable (and at least Npgsql supports this behavior as well).

Apart from the comments below, could you please also add the comment from the SqlClient docs, i.e.:

> If you pass a buffer that is null, [GetBytes](https://docs.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqldatareader.getbytes?view=sqlclient-dotnet-standard-4.1) returns the length of the entire field in bytes, not the remaining size based on the buffer offset parameter.

In addition, the same modifications need to be done for the GetChars method.

Grauenwolf:
Actually, this behavior is covered in IDataRecord. So you should probably cancel this PR and copy it's text instead.

roji:
@Grauenwolf good point - is this something you want to do (you could force-push to this PR)?

PR #7966, opened 2022-04-17
Updated docs on CopyTo(Stream, Int32)

Shailosingh:
## Summary
This method reads a specified number of bytes from the current stream, advances the read position and then writes them to the destination stream, advancing the write position of that as well.

The old documentation did not make it clear that the write or read positions were advanced at all. This can lead to a lot of confusion for programmers when they attempt to use this method.

### DIFF
--- a/xml/System.IO/Stream.xml
+++ b/xml/System.IO/Stream.xml
@@ -805,7 +805,7 @@
         <AssemblyVersion>4.1.0.0</AssemblyVersion>
       </AssemblyInfo>
       <Docs>
-        <summary>Asynchronously reads the bytes from the current stream and writes them to another stream.</summary>
+        <summary>Asynchronously reads the bytes from the current stream and writes them to another stream. Both streams positions are advanced by the number of bytes copied.</summary>
       </Docs>
     </MemberGroup>
     <Member MemberName="CopyToAsync">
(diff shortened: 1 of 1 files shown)


gewarren (commented):
If we update one overload, we should update all of the overloads. But I'd like the SMEs to weigh in first.

xml/System.IO/Stream.xml
```diff
@@ -775,7 +775,7 @@
       <Docs>
         <param name="destination">The stream to which the contents of the current stream will be copied.</param>
         <param name="bufferSize">The size of the buffer. This value must be greater than zero. The default size is 81920.</param>
-        <summary>Reads the bytes from the current stream and writes them to another stream, using a specified buffer size.</summary>
+        <summary>Reads the bytes from the current stream, advances the current stream's read position, writes the bytes to the destination stream and advances the destination stream's write position, using a specified buffer size.</summary>
```
adamsitnik, line 778:
```suggestion
        <summary>Reads the bytes from the current stream and writes them to another stream, using a specified buffer size. Both streams positions are advanced by the number of bytes copied.</summary>
```

adamsitnik (commented):
@Shailosingh big thanks for your contribution!

Overall, it looks good to me, but personally I would simplify it. Please take a look at my suggestion and let me know what do you think.

xml/System.IO/Stream.xml
Shailosingh: Thank you very much! The simplification makes it much easier to understand.

gewarren (dismissed):

Shailosingh:
@gewarren Thank you very much for the feedback! I think I updated the overloads and also updated the definition of Stream.CopyTo as well. I'm still pretty new to submitting pull requests to big repositories, so please let me know if something is wrong.

adamsitnik (commented):
@Shailosingh the synchronous overloads look good to me. Could you please add it to async one too? You can find them by searching for `<Member MemberName="CopyToAsync">`

again, thank you for your contribution!

Shailosingh:
@adamsitnik Thank you for the suggestion. I have now adjusted the CopyToAsync overloads and definition 😊.

adamsitnik (approved):
LGTM, thank you @Shailosingh !

PR #7972, opened 2022-04-19
Fix SortedList ctor time complexity documentation

leaomatheus:
Based on the dotnet/runtime implementation of SortedList (https://github.com/dotnet/runtime/blob/a9b3898ee4b568cba09236cc71a63faf3ec52eea/src/libraries/System.Collections/src/System/Collections/Generic/SortedList.cs), the SortedList constructor taking a IDictionary as parameter does an Array.Sort operation under the hood, which makes its complexity as slow as the sort (which should be O(N log N)).

The current documentation states that
> If the data in `dictionary` are sorted, this constructor is an O(`n`) operation, where `n` is the number of elements in `dictionary`. Otherwise it is an O(`n`*`n`) operation.

This statement is not correct -- if it was the case, this would be significantly slower. After reading the documentation, I was surprised by this fact and was looking into removing the usage of this constructor from my team codebase, until I learned this was not the case. I am proposing this change to avoid confusion for others too.

## Summary

Changing the documentation for the constructors below

SortedList<TKey,TValue>(IDictionary<TKey,TValue>)
SortedList<TKey,TValue>(IDictionary<TKey,TValue>, IComparer<TKey>)


eiriktsarpalis:
> Looking at the code _O(n log n)_ seems right. Would you like to submit a PR correcting this?

As you probably can tell, I failed to notice that this was a PR. Should have had coffee before reading my github notifications :-)

PR #7975, opened 2022-04-19
Update Task`1.xml

JimDaly:
## Summary

Describe your changes here.

Per comments in https://github.com/dotnet/dotnet-api-docs/pull/7941 , removing remarks that say 'This method is intended for compiler use rather than for use in application code.'

> At this point, the right thing for TaskAwaiter and TaskAwaiter is probably to just strike those sentence from the docs, though some extra verbiage around the distinction from Task.Wait would be welcome


gewarren:
Closing in favor of https://github.com/dotnet/dotnet-api-docs/pull/7968.

PR #7980, opened 2022-04-21
System.Type F# snippets

albert-du:
Adds F# snippets for Type


BillWagner (approved):
Thanks again @albert-du 

This is great work, as usual.  I'll :shipit: now.

PR #7981, opened 2022-04-21
System.Tuple<T1,T2,T3,T4,T5,T6> F# snippet

albert-du:
Adds F# snippets for Tuple<T1,T2,T3,T4,T5,T6>


BillWagner (approved):
Thanks again @albert-du 

This LGTM as well. I'll :shipit: now.

PR #7983, opened 2022-04-21
Fix see elements of BitOperations.RoundUpToPowerOf2

xparadoxical:
## Summary

Two `see` xmldoc elements in `summary` of both overloads of BitOperations.RoundUpToPowerOf2 are using the `langword` attribute to refer to types, this PR changes them to use `cref`.


gewarren (approved):
Thanks @xparadoxical!

PR #7985, opened 2022-04-21
Mdoc 5.8.9 pre release testing

huangmin-ms:
Build.Reason:Manual by Min Huang
 Build.Url:https://apidrop.visualstudio.com/Content%20CI/_build/results?buildId=289256&view=results
 source_repo.branch:master

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Runtime.InteropServices.ObjectiveC/ObjectiveCMarshal+UnhandledExceptionPropagationHandler.xml
+++ b/xml/System.Runtime.InteropServices.ObjectiveC/ObjectiveCMarshal+UnhandledExceptionPropagationHandler.xml
@@ -1,10 +1,10 @@
 <Type Name="ObjectiveCMarshal+UnhandledExceptionPropagationHandler" FullName="System.Runtime.InteropServices.ObjectiveC.ObjectiveCMarshal+UnhandledExceptionPropagationHandler">
-  <TypeSignature Language="C#" Value="public delegate method ObjectiveCMarshal.UnhandledExceptionPropagationHandler(Exception exception, RuntimeMethodHandle lastMethod, out IntPtr context);" />
+  <TypeSignature Language="C#" Value="public delegate delegate* unmanaged&lt;IntPtr, void&gt; ObjectiveCMarshal.UnhandledExceptionPropagationHandler(Exception exception, RuntimeMethodHandle lastMethod, out IntPtr context);" />
   <TypeSignature Language="ILAsm" Value=".class nested public auto ansi sealed ObjectiveCMarshal/UnhandledExceptionPropagationHandler extends System.MulticastDelegate" />
   <TypeSignature Language="DocId" Value="T:System.Runtime.InteropServices.ObjectiveC.ObjectiveCMarshal.UnhandledExceptionPropagationHandler" />
-  <TypeSignature Language="VB.NET" Value="Public Delegate Function ObjectiveCMarshal.UnhandledExceptionPropagationHandler(exception As Exception, lastMethod As RuntimeMethodHandle, ByRef context As IntPtr) As method " />
-  <TypeSignature Language="F#" Value="type ObjectiveCMarshal.UnhandledExceptionPropagationHandler = delegate of Exception * RuntimeMethodHandle * nativeint -&gt; method" />
-  <TypeSignature Language="C++ CLI" Value="public: delegate method ^ ObjectiveCMarshal::UnhandledExceptionPropagationHandler(Exception ^ exception, RuntimeMethodHandle lastMethod, [Runtime::InteropServices::Out] IntPtr % context);" />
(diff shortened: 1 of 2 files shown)


xml/System.Runtime.InteropServices.ObjectiveC/ObjectiveCMarshal+UnhandledExceptionPropagationHandler.xml
```diff
@@ -1,10 +1,10 @@
...
-  <TypeSignature Language="F#" Value="type ObjectiveCMarshal.UnhandledExceptionPropagationHandler = delegate of Exception * RuntimeMethodHandle * nativeint -&gt; method" />
-  <TypeSignature Language="C++ CLI" Value="public: delegate method ^ ObjectiveCMarshal::UnhandledExceptionPropagationHandler(Exception ^ exception, RuntimeMethodHandle lastMethod, [Runtime::InteropServices::Out] IntPtr % context);" />
+  <TypeSignature Language="VB.NET" Value="Public Delegate Function ObjectiveCMarshal.UnhandledExceptionPropagationHandler(exception As Exception, lastMethod As RuntimeMethodHandle, ByRef context As IntPtr) As delegate* unmanaged&lt; As IntPtr, Void&gt; " />
+  <TypeSignature Language="F#" Value="type ObjectiveCMarshal.UnhandledExceptionPropagationHandler = delegate of Exception * RuntimeMethodHandle * nativeint -&gt; delegate* unmanaged&lt;nativeint, unit&gt;" />
+  <TypeSignature Language="C++ CLI" Value="public: delegate delegate* unmanaged&lt;IntPtr , void&gt; ^ ObjectiveCMarshal::UnhandledExceptionPropagationHandler(Exception ^ exception, RuntimeMethodHandle lastMethod, [Runtime::InteropServices::Out] IntPtr % context);" />
```
AaronRobinsonMSFT, line 7: I don't think this is valid C++/CLI syntax nor valid for VB.NET or F#. In fact, I believe this API is only usable from C#. Perhaps from C++/CLI but since C++/CLI isn't supported on macOS, it doesn't really matter. Can we indicate in some way this API is only usable in C#?

huangmin-ms: Thanks Aaron for your question, please check my email for clarification.

huangmin-ms: > I don't think this is valid C++/CLI syntax nor valid for VB.NET or F#. In fact, I believe this API is only usable from C#. Perhaps from C++/CLI but since C++/CLI isn't supported on macOS, it doesn't really matter. Can we indicate in some way this API is only usable in C#?

Thanks Aaron for your question, please check my email for clarification.

AaronRobinsonMSFT: Alright. I'd like to prioritize this ability. I say that because there seems to be a push for C# to evolve much faster than the other languages and that means it is a very real possibility that the scope of impact grows. Although my view could be heavily biased since I come from the interop world where everything is a special case :-/

/cc @stephentoub @jaredpar

AaronRobinsonMSFT (approved):

xml/System.Runtime.InteropServices.ObjectiveC/ObjectiveCMarshal+UnhandledExceptionPropagationHandler.xml
jaredpar: This is definitely going to apply to more places. Consider that generic math APIs are going to be, at least initially, unusable in F#, C++/CLI, VB, etc ... There is simply no way to express the APIs in those languages. Rather than give an incorrect representation would prefer we not show it at all and give some sort of unsupported message. 

Curious: how do we say handle `Span<T>` today? It's not supported in a few languages already. It's probably not as big of an issue cause we can provide a reasonable representation of the type in any language.

huangmin-ms: > This is definitely going to apply to more places. Consider that generic math APIs are going to be, at least initially, unusable in F#, C++/CLI, VB, etc ... There is simply no way to express the APIs in those languages. Rather than give an incorrect representation would prefer we not show it at all and give some sort of unsupported message.
> 
> Curious: how do we say handle `Span<T>` today? It's not supported in a few languages already. It's probably not as big of an issue cause we can provide a reasonable representation of the type in any language.

This is the page link for [Span](https://docs.microsoft.com/en-us/dotnet/api/system.span-1?view=net-6.0)
The multi-language representations were generated by multi language formatter processing the same piece of IL code.

huangmin-ms: Thanks @AaronRobinsonMSFT  and @jaredpar for your suggestions, we will have an internal discussion with our PM @mimisasouvanh  and will let you know if any updates.

gewarren:
5.8.9 has been released. I think we can close this PR.

PR #7979, opened 2022-04-20
Config files are .NET Framework only (System.Diagnostics)

gewarren:
This PR annotates remarks that refer to settings in application configuration files as being specific to .NET Framework.

Contributes to #7970.

[*Hide whitespace changes in diff.*](https://github.com/dotnet/dotnet-api-docs/pull/7979/files?diff=split&w=1)

PR #7994, opened 2022-04-25
Updates around config files

gewarren:
Related to #7979. Updates remarks that refer to app.config files, to note that they are specific to .NET Framework apps.


BillWagner (approved):
This LGTM @gewarren 

Once you and the product team are happy, you can :shipit:.

PR #7996, opened 2022-04-25
Add note about open angle brackets and nulls in string parameters

josalem:
CC @dotnet/dotnet-diag 

fix dotnet/runtime#50892

### DIFF
--- a/xml/System.Diagnostics.Tracing/EventSource.xml
+++ b/xml/System.Diagnostics.Tracing/EventSource.xml
@@ -2107,7 +2107,7 @@ public void Info1(string arg1)
  You can also create a custom overload by using the <xref:System.Diagnostics.Tracing.EventSource.WriteEventCore%2A> method.  
 
 > [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


gewarren (approved):

xml/System.Diagnostics.Tracing/EventSource.xml
```diff
@@ -2021,6 +2021,9 @@ public void Info1(string arg1)
 ```  
   
  You can also create a custom overload by using the <xref:System.Diagnostics.Tracing.EventSource.WriteEventCore%2A> method.  
+
+> [!IMPORTANT]
+> Event parameters with a type of `string` should not include `<` or `\0` characters. They are unsupported characters and can cause issues for parsers of the event payload.
```
noahfalk, line 2026: The limitation for angle brackets is in the name of the EventSource method, not the string parameters. Both '<' and '>' are problematic.

noahfalk:
This PR aims to fix https://github.com/dotnet/runtime/issues/50892.

noahfalk (approved):
LGTM

PR #8001, opened 2022-04-26
System.Tuple<T1,T2,T3,T4,T5,T6,T7> F# snippets

albert-du:
adds F# snippets for Tuple<T1,T2,T3,T4,T5,T6,T7>


BillWagner (approved):
Thanks @albert-du 

This is great work, as always.

I'll :shipit: now.

PR #8017, opened 2022-04-28
System.TypeInitializationException F# snippets

albert-du:
Adds F# snippets for TypeInitializationException


BillWagner (approved):
Thanks again @albert-du This is great as always.

We'll :shipit: now.

PR #8018, opened 2022-04-28
Clarify casing of escape characters for encoding

daviburg:
WebUtility.UrlEncode(String) Method encodes with upper case characters, unlike HttpUtility.UrlEncode. The difference will cause string comparisons of encoded values to fail, so as for a dictionary key lookup.

## Summary

Describe your changes here.


rzikm (approved):
LGTM! Thanks!

PR #8024, opened 2022-04-29
Update summary for MethodImplOptions.AggressiveOptimization

AndyAyersMS:
## Summary

Update summary for `MethodImplOptions.AggressiveOptimization` to offer a bit more guidance on what it does and when it should be (or shouldn't be) used.

See https://github.com/dotnet/runtime/issues/62148

cc @azhmur @EgorBo

### DIFF
--- a/xml/System.Runtime.CompilerServices/MethodImplOptions.xml
+++ b/xml/System.Runtime.CompilerServices/MethodImplOptions.xml
@@ -143,7 +143,9 @@
       </ReturnValue>
       <MemberValue>512</MemberValue>
       <Docs>
-        <summary>The method contains code that should always be optimized by the just-in-time (JIT) compiler. Use this attribute if running an unoptimized version of the method has undesirable effects, for instance causing too much overhead or extra memory allocation. Methods with this attribute bypass the first tier of <a href="https://docs.microsoft.com/en-us/dotnet/core/runtime-config/compilation#tiered-compilation">Tiered Compilation</a> and so cannot benefit from optimizations that rely on tiering, for example <a href="https://docs.microsoft.com/en-us/dotnet/core/runtime-config/compilation#profile-guided-optimization">Dynamic PGO</a> or optimizations based on initialized classes. Thus they may not have the most optimal code generation.</summary>
+        <summary><para>The method contains code that should always be optimized by the just-in-time (JIT) compiler.</para>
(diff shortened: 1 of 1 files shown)


xml/System.Runtime.CompilerServices/MethodImplOptions.xml
```diff
@@ -143,7 +143,7 @@
       </ReturnValue>
       <MemberValue>512</MemberValue>
       <Docs>
-        <summary>The method contains a hot path and should be optimized.</summary>
+        <summary>The method contains code that should always be optimized by the just-in-time (JIT) compiler. Use this attribute if running an unoptimized version of the method has undesirable effects, for instance causing too much overhead or extra memory allocation. Methods with this attribute bypass the first tier of <a href="https://docs.microsoft.com/en-us/dotnet/core/runtime-config/compilation#tiered-compilation">Tiered Compilation</a> and so cannot benefit from optimizations that rely on tiering, for example <a href="https://docs.microsoft.com/en-us/dotnet/core/runtime-config/compilation#profile-guided-optimization">Dynamic PGO</a> or optimizations based on initialized classes. Thus they may not have the most optimal code generation.</summary>
```
gewarren, line 146:
```suggestion
        <summary><para>The method contains code that should always be optimized by the just-in-time (JIT) compiler.</para>
        <para>Use this attribute if running an unoptimized version of the method has undesirable effects, for instance causing too much overhead or extra memory allocation.</para>
        <para>Methods with this attribute may not have optimal code generation. They bypass the first tier of <a href="/dotnet/core/runtime-config/compilation#tiered-compilation">Tiered Compilation</a> and therefore can't benefit from optimizations that rely on tiering, for example, <a href="/dotnet/core/runtime-config/compilation#profile-guided-optimization">Dynamic PGO</a> or optimizations based on initialized classes.</para></summary>
```

gewarren (approved):
I split it into three paragraphs. What do you think?

xml/System.Runtime.CompilerServices/MethodImplOptions.xml
AndyAyersMS: Thanks! Didn't know that was possible.

gewarren:
Here's the preview if you want to see how it renders: https://review.docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.methodimploptions?view=net-7.0&branch=pr-en-us-8024#fields

azhmur:
Thank you. Looks good.

PR #8038, opened 2022-05-04
Fix escaping of the backslash character

candritzky:
## Summary

The display of the markdown table was crippled for the backslash character. It was shown as the pipe character with the replacement value "_" next to it.


gewarren (approved):
Good catch @candritzky. Thank you!

PR #8043, opened 2022-05-06
itemsInBags will be different from 500 

chenkaigithub:
itemsInBags will be different 500 without atomic operation.

## Summary

Describe your changes here.

Fixes #8043 (if available)

### DIFF
--- a/snippets/csharp/System.Collections.Concurrent/ConcurrentBag/concbag.cs
+++ b/snippets/csharp/System.Collections.Concurrent/ConcurrentBag/concbag.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using System.Threading;
 
 class ConcurrentBagDemo
 {


MihaZupan (approved):

snippets/csharp/System.Collections.Concurrent/ConcurrentBag/concbag.cs
```diff
@@ -36,7 +36,7 @@ static void Main()
                 if (cb.TryTake(out item))
                 {
                     Console.WriteLine(item);
-                    itemsInBag++;
+                    Interlocked.Increment(ref itemsInBag);
```
BillWagner, line 39: This needs the following using statement to compile correctly:

```
using System.Threading;
```

Once you add that, we'll :shipit:

BillWagner (changes requested):
Thanks for submitting this change @chenkaigithub 

Once you make one addition, this is ready merge.

Thanks again!

snippets/csharp/System.Collections.Concurrent/ConcurrentBag/concbag.cs
chenkaigithub: Thank you!  I updated.

BillWagner (approved):

Issue #7420, opened 2021-11-24
Tracking issue for adding F# code snippets for the System namespace


albert-du:
https://github.com/dotnet/docs/issues/26813

to do:
- [x] Action #7352
- [x] Action<'T> #7380
- [x] Action<'T1,'T2> #7403
- [x] Action<'T1,'T2,'T3> #7404
- [x] Action<'T1,'T2,'T3,'T4> #7405
- [x] Activator #7421
- [x] AggregateException #7407
- [x] AppContext #7425
- [x] AppDomain #7914
- [x] AppDomainSetup #7914
- [x] AppDomainUnloadedException #7914
- [x] ApplicationId #7901
- [x] ArgumentException #7428
- [x] ArgumentOutOfRangeException #7429
- [x] Array #7432
- [x] ArraySegment<'T> #7447
- [x] ArrayTypeMismatchException #7514
- [x] AsyncCallback #7518
- [x] Attribute #7522
- [x] AttributeTargets #7525
- [x] AttributeUsageAttribute #7526
- [x] BadImageFormatException #7527
- [x] Base64FormattingOptions #7528
- [x] BitConverter #7530
- [x] Boolean #7430
- [x] Buffer #7531
- [x] Byte #7532
- [x] Char #7535
- [x] CharEnumerator #7536
- [x] Comparison<'T> #7537
- [x] Console #7539
- [x] ConsoleCancelEventArgs #7539
- [x] ConsoleCancelEventHandler #7539
- [x] ConsoleColor #7539
- [x] ConsoleKey #7539
- [x] ConsoleKeyInfo #7539
- [x] ConsoleModifiers #7539
- [x] ConsoleSpecialKey #7539
- [x] Convert #7688
- [x] DateTime #7548
- [x] DateTimeOffset #7596
- [x] DayOfWeek #7597
- [x] DBNull #7598
- [x] Decimal #7605
- [x] Delegate #7606
- [x] DivideByZeroException #7599
- [x] Double #7693
- [x] EntryPointNotFoundException #7694
- [x] Enum #7696
- [x] Environment #7709
- [x] Environment.SpecialFolder #7709 
- [x] EnvironmentVariableTarget #7709 
- [x] EventArgs #7710
- [x] EventHandler #7710
- [x] EventHandler<'TEventArgs> #7710
- [x] Exception #7725 
- [x] FlagsAttribute #7726
- [x] FormatException #7727
- [x] FormattableString #7728
[...]

PR #8046, opened 2022-05-07
System.UInt32 F# snippets

albert-du:
Adds F# snippets for UInt32

been a bit busier recently, finally got a moment to get another one out!

only 10(ish) more to go (for now) #7420


BillWagner (approved):
Thanks again for all the hard work @albert-du !

I've reviewed this, and I'll :shipit: now.

PR #8056, opened 2022-05-08
Fix error in the summary of the Unwrap(Task<Task>)

theodorzoulias:
## Summary

Currently the documentation `<summary>` of the [`Unwrap(Task<Task>)`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskextensions.unwrap) method refers to the unrelated [`TaskScheduler.TryExecuteTaskInline`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskscheduler.tryexecutetaskinline) method, instead of the correct `Task<Task>` class.

This PR fixes this issue.


buyaa-n (approved):
LGTM, thanks!

PR #8068, opened 2022-05-12
Remove Windows Forms code from sample

softwarepronto:
1) Renamed method settings_ValidationEventHandler to ValidationEventHandler. The underscore was in the method name because the Windows Forms designer created event handler methods and it added an underscore to each event handler method (20-year-old knowledge). I used the same method name as in https://docs.microsoft.com/en-us/dotnet/standard/data/xml/reading-and-writing-xml-schemas.
2) The body of ValidationCallback contained Windows Forms MessageBoxes which means any user compiling the documentation would have to reference the Windows Forms assembly (coding style 2002-2006). I replaced this with System.Console (see the link for reading-and-writing-xml-schemas above which has the same behavior).
3) There was variable declared at the end of settings_ValidationEventHandler (in the handling of XmlSeverityType.Error): Type objectType = sender.GetType();. This variable was not used and served no purpose in the sample.

## Summary

Describe your changes here.

### DIFF
--- a/snippets/csharp/System.Xml/XmlDocument/Overview/xmlhelpermethods.cs
+++ b/snippets/csharp/System.Xml/XmlDocument/Overview/xmlhelpermethods.cs
@@ -106,7 +106,7 @@ namespace XMLProcessingApp
 
             settings.Schemas.Add(schema);
 
-            settings.ValidationEventHandler += settings_ValidationEventHandler;
+            settings.ValidationEventHandler += ValidationCallback;
             settings.ValidationFlags =
                 settings.ValidationFlags | XmlSchemaValidationFlags.ReportValidationWarnings;
             settings.ValidationType = ValidationType.Schema;
@@ -216,7 +216,7 @@ namespace XMLProcessingApp
                 doc.Schemas.Add(schema);
             }
 
-            // Use an event handler to validate the XML node against the schema.
+            // Use a callback to validate the XML node against the schema.
             doc.Validate(ValidationCallback);
         }
 


gewarren:
@softwarepronto It looks like there's a compiler error to fix. Can you take a look?

snippets/csharp/System.Xml/XmlDocument/Overview/xmlhelpermethods.cs
```diff
@@ -216,28 +216,27 @@ private void validateXML(bool generateSchema, XmlDocument doc)
                 doc.Schemas.Add(schema);
             }
 
-            // Use an event handler to validate the XML node against the schema.
-            doc.Validate(settings_ValidationEventHandler);
+            // Use a callback to validate the XML node against the schema.
```
MSDN-WhiteKnight, line 219: The "event handler" terminology is baked into the API signature itself: https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmldocument.validate?view=net-6.0 

>public void Validate (System.Xml.Schema.ValidationEventHandler? validationEventHandler);

Is it really needed to rename it into callback?

softwarepronto: @MSDN-WhiteKnight I know you're busy. I replied as to the specific reason why I changed the code. I included a link to other documentation that is coded the same way and I pasted a code sample from the other documentation that matches the change I made. If this is not sufficient to convince, you I can back out renaming the callback and only remove the Windows Forms from the sample.

MSDN-WhiteKnight: I'm not a maintainer in this repo. If callback is more correct term here and maintainers approve it, seems fine to change it.

gewarren (approved):
Looks fine to me. The [project that compiles this file](https://github.com/dotnet/dotnet-api-docs/blob/366ba1f6b880643e2bf111500f84003d49bb40d4/snippets/csharp/System.Xml/XmlDocument/Overview/xmlprocessingapp.csproj) is set to use Windows Forms, but I don't think that's a problem.

krwq (approved):

PR #8076, opened 2022-05-15
System.Version F# snippets

albert-du:
Adds F# snippets for Version


BillWagner (approved):
Thanks again @albert-du 

This looks great as well. I'll :shipit: now.

PR #8077, opened 2022-05-15
System.WeakReference F# snippets

albert-du:
Adds an F# snippet for WeakReference


BillWagner (approved):
This also LGTM @albert-du 

I'll :shipit: now.

PR #8094, opened 2022-05-20
Update Assembly.xml

steveharter:
Fix typo in example.


gewarren (commented):
I'm not sure this change is correct - because then nothing calls M1?

steveharter:
Sorry my change is not correct; the doc assumes someone calling M3 (not M1)

PR #8097, opened 2022-05-23
Update Unsafe and MemoryMarshal docs

GrabYourPitchforks:
## Summary

Significantly fleshes out the docs for `Unsafe`, including:

* Using more accurate terminology, such as using "managed pointer" instead of "reference" in APIs which are performing pointer manipulation. With this change I'm trying to reserve the word "reference" for APIs which are reading \& writing values rather than performing pointer manipulation.
* Writing the API's assumptions in the _Remarks_ section, alongside ECMA sections we expect callers to be familiar with.

The changes for `MemoryMarshal` are a bit tamer by comparison:

* Adding missing "this API can throw `OverflowException`" lines.
* Removing incorrect references to "_T_ can't contain pointers."
* Removing incorrect references to "this method ensures type safety."

### DIFF
--- a/xml/System.Runtime.CompilerServices/Unsafe.xml
+++ b/xml/System.Runtime.CompilerServices/Unsafe.xml
@@ -992,7 +992,7 @@ static void Copy<T>(ref T destination, void* source)
 This API corresponds to the `cpblk` opcode. Both the `destination` and `source` pointers are assumed to be pointer-aligned. See [ECMA-335](https://www.ecma-international.org/publications-and-standards/standards/ecma-335/), Sec. III.3.30 ("cpblk - copy data from memory to memory") for more information.
 
 > [!CAUTION]
-> This API is not intended for copying arbitrary-length runs of memory. Consider instead using <xref:System.Buffer.MemoryCopy> or <xref:System.Span%601.CopyTo%2A> for this scenario.
+> This API is not intended for copying arbitrary-length runs of memory. Consider instead using <xref:System.Buffer.MemoryCopy%2A> or <xref:System.Span%601.CopyTo%2A> for this scenario.
 
 ]]></format>
         </remarks>
@@ -1041,7 +1041,7 @@ This API corresponds to the `cpblk` opcode. Both the `destination` and `source` 
 This API corresponds to the `cpblk` opcode. Both the `destination` and `source` pointers are assumed to be pointer-aligned. See [ECMA-335](https://www.ecma-international.org/publications-and-standards/standards/ecma-335/), Sec. III.3.30 ("cpblk - copy data from memory to memory") for more information.
 
 > [!CAUTION]
-> This API is not intended for copying arbitrary-length runs of memory. Consider instead using <xref:System.Buffer.MemoryCopy> or <xref:System.Span%601.CopyTo%2A> for this scenario.
+> This API is not intended for copying arbitrary-length runs of memory. Consider instead using <xref:System.Buffer.MemoryCopy%2A> or <xref:System.Span%601.CopyTo%2A> for this scenario.
 
 ]]></format>
         </remarks>
@@ -1089,7 +1089,7 @@ This API corresponds to the `cpblk` opcode. Both the `destination` and `source` 
 This API corresponds to the `unaligned.1 cpblk` opcode sequence. No alignment assumptions are made about the `destination` or `source` pointers. See [ECMA-335](https://www.ecma-international.org/publications-and-standards/standards/ecma-335/), Sec. III.3.30 ("cpblk - copy data from memory to memory") and Sec. III.2.5 ("unaligned. (prefix) - pointer instruction might be unaligned") for more information.
 
 > [!CAUTION]
-> This API is not intended for copying arbitrary-length runs of memory. Consider instead using <xref:System.Buffer.MemoryCopy> or <xref:System.Span%601.CopyTo%2A> for this scenario.
+> This API is not intended for copying arbitrary-length runs of memory. Consider instead using <xref:System.Buffer.MemoryCopy%2A> or <xref:System.Span%601.CopyTo%2A> for this scenario.
 
 ]]></format>
         </remarks>
@@ -1136,7 +1136,7 @@ This API corresponds to the `unaligned.1 cpblk` opcode sequence. No alignment as
 This API corresponds to the `unaligned.1 cpblk` opcode sequence. No alignment assumptions are made about the `destination` or `source` pointers. See [ECMA-335](https://www.ecma-international.org/publications-and-standards/standards/ecma-335/), Sec. III.3.30 ("cpblk - copy data from memory to memory") and Sec. III.2.5 ("unaligned. (prefix) - pointer instruction might be unaligned") for more information.
 
 > [!CAUTION]
-> This API is not intended for copying arbitrary-length runs of memory. Consider instead using <xref:System.Buffer.MemoryCopy> or <xref:System.Span%601.CopyTo%2A> for this scenario.
+> This API is not intended for copying arbitrary-length runs of memory. Consider instead using <xref:System.Buffer.MemoryCopy%2A> or <xref:System.Span%601.CopyTo%2A> for this scenario.
 
 ]]></format>
         </remarks>


xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -407,12 +488,33 @@
...
+        <summary>Reinterprets the given managed pointer as a new managed pointer to a value of type <typeparamref name="TTo" />.</summary>
+        <returns>A managed pointer to a value of type <typeparamref name="TTo" />.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+This API is conceptually similar to C++'s `reinterpret_cast<>`. It is the caller's responsibility to ensure that the cast is legal. No runtime check will be performed.
+
+Only the managed pointer is reinterpreted. The referenced value itself will remain unchanged. Consider the example below.
```
gewarren, line 503:
```suggestion
Only the managed pointer is reinterpreted. The referenced value itself will remain unchanged. Consider the following example.
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -407,12 +488,33 @@
...
+
+```cs
+int[] intArray = new int[] { 0x12345678 }; // a 1-element array
+ref int refToInt32 = ref intArray[0]; // managed pointer to first Int32 in array
+ref short refToInt16 = ref Unsafe.As<int, short>(ref refToInt32); // reinterpret as managed pointer to Int16
+Console.WriteLine(refToInt16);
+```
+
+The output of this program depends on the endianness of the current machine. On big-endian architectures this will output `0x1234`, and on little-endian architectures this will output `0x5678`.
```
gewarren, line 512:
```suggestion
The output of this program depends on the endianness of the current machine. On big-endian architectures, this code outputs `0x1234`, and on little-endian architectures, it outputs `0x5678`.
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -584,12 +696,22 @@
...
       <Docs>
-        <typeparam name="T">The type of reference.</typeparam>
-        <param name="origin">The reference to origin.</param>
-        <param name="target">The reference to target.</param>
-        <summary>Determines the byte offset from origin to target from the given references.</summary>
+        <typeparam name="T">The elemental type of the managed pointers.</typeparam>
+        <param name="origin">The managed pointer to origin.</param>
+        <param name="target">The managed pointer to target.</param>
+        <summary>Determines the byte offset from origin to target from the given managed pointers.</summary>
         <returns>Byte offset from origin to target i.e. <paramref name="target" /> - <paramref name="origin" />.</returns>
```
gewarren, line 703:
```suggestion
        <returns>The byte offset from origin to target, that is, <paramref name="target" /> - <paramref name="origin" />.</returns>
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1339,10 +1601,69 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
-        <remarks>To be added.</remarks>
+        <typeparam name="T">The type of the reference.</typeparam>
+        <param name="value">The reference whose initialization should be skipped.</param>
+        <summary>Bypasses definite assignment rules for a given reference.</summary>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+This is typically used to avoid double-assignment when initializing union-type structs. Consider the following example, which produces a C# compiler error.
```
gewarren, line 1617:
```suggestion
This method is typically used to avoid double-assignment when initializing union-type structs. Consider the following example, which produces a C# compiler error.
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1339,10 +1601,69 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
+    public int SomeIntField;
+    
+    [FieldOffset(0)]
+    public byte SomeByteField;
+}
+```
+
+This error occurs because the compiler expects _all_ fields of the struct to be initialized before the struct is used or returned to the caller.
+
+One way to avoid this compiler error is to ensure the entire struct is zero-inited before individual fields are set, as shown in the below example.
```
gewarren, line 1642:
```suggestion
One way to avoid this compiler error is to ensure the entire struct is zero-initialized before individual fields are set, as shown in the following example.
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1339,10 +1601,69 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
+```cs
+// This sample compiles successfully.
+public static MyUnionStruct MyMethod() {
+    MyUnionStruct value = default; // the struct is now guaranteed assigned
+    value.SomeIntField = 42;
+    return value;
+}
+```
+
+For developers who don't want to avoid the initial zero-init assignment, the `SkipInit` method can be used to suppress the compiler warning.
```
gewarren, line 1653:
```suggestion
If you want to avoid the initial zero-init assignment, you can call the `SkipInit` method to suppress the compiler warning.
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1339,10 +1601,69 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
+public static MyUnionStruct MyMethod() {
+    MyUnionStruct value;
+    Unsafe.SkipInit(out value); // compiler believes the struct has been guaranteed assigned
+    value.SomeIntField = 42;
+    return value;
+}
+```
+
+> [!WARNING]
+> The developer must take care to ensure that the struct has been initialized appropriately, otherwise the struct's fields could contain uninitialized data from the stack.
```
gewarren, line 1668:
```suggestion
> Take care to ensure that the struct has been initialized appropriately, otherwise the struct's fields could contain uninitialized data from the stack.
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1376,12 +1697,22 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
         <Parameter Name="elementOffset" Type="System.Int32" Index="1" FrameworkAlternate="dotnet-plat-ext-2.1;dotnet-plat-ext-2.2;net-5.0;net-6.0;netcore-3.0;netcore-3.1" />
       </Parameters>
       <Docs>
-        <typeparam name="T">The type of the void pointer.</typeparam>
-        <param name="source">The void pointer to subtract the offset from.</param>
+        <typeparam name="T">The type whose width will be used as a scale factor for `elementOffset`.</typeparam>
```
gewarren, line 1705:
```suggestion
        <typeparam name="T">The type whose width will be used as a scale factor for <paramref name=elementOffset" />.</typeparam>
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1376,12 +1697,22 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
+        <param name="source">The unmanaged pointer to subtract the offset froms.</param>
```
gewarren, line 1706:
```suggestion
        <param name="source">The unmanaged pointer to subtract the offset from.</param>
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1376,12 +1697,22 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
         <param name="elementOffset">The offset to subtract.</param>
-        <summary>Subtracts an element offset from the given void pointer.</summary>
-        <returns>A new void pointer that reflects the subtraction of offset from the specified pointer.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Subtracts an element offset from the given unmanaged pointer.</summary>
+        <returns>A new unmanaged pointer that reflects the subtraction of offset from the source pointer.</returns>
```
gewarren, line 1709:
```suggestion
        <returns>A new unmanaged pointer that reflects the subtraction of the specified offset from the source pointer.</returns>
```

GrabYourPitchforks:
BTW, I'm totally open to suggestions on how we might flesh out the `Unsafe` documentation to call out when you'd actually want to use it. If nothing else I can add a "here's how you reverse a span with no bounds checks" sample, which would show `GetReference`, `Add`, `Subtract`, and `IsAddressLessThan`. But something with a bit more oomph might be better.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1417,12 +1748,22 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
-        <typeparam name="T">The type of reference.</typeparam>
-        <param name="source">The reference to subtract the offset from.</param>
+        <typeparam name="T">The elemental type of the managed pointer.</typeparam>
+        <param name="source">The managed pointer to subtract the offset from.</param>
         <param name="elementOffset">The offset to subtract.</param>
-        <summary>Subtracts an element offset from the given reference.</summary>
-        <returns>A new reference that reflects the subtraction of offset from pointer.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Subtracts an offset from the given managed pointer.</summary>
+        <returns>A new managed pointer that reflects the subtraction of offset from the source pointer.</returns>
```
gewarren, line 1760:
```suggestion
        <returns>A new managed pointer that reflects the subtraction of the specified offset from the source pointer.</returns>
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1457,12 +1798,22 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
-        <typeparam name="T">The type of reference.</typeparam>
-        <param name="source">The reference to subtract the offset from.</param>
+        <typeparam name="T">The elemental type of the managed pointer.</typeparam>
+        <param name="source">The managed pointer to subtract the offset from.</param>
         <param name="elementOffset">The offset to subtract.</param>
-        <summary>Subtracts an element offset from the given reference.</summary>
-        <returns>A new reference that reflects the subtraction of offset from pointer.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Subtracts an element offset from the given managed pointer.</summary>
+        <returns>A new managed pointer that reflects the subtraction of offset from the source pointer.</returns>
```
gewarren, line 1810:
```suggestion
        <returns>A new managed pointer that reflects the subtraction of the specified offset from the source pointer.</returns>
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1499,12 +1850,22 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
-        <typeparam name="T">The type of reference.</typeparam>
-        <param name="source">The reference to subtract the offset from.</param>
+        <typeparam name="T">The elemental type of the managed pointer.</typeparam>
+        <param name="source">The managed pointer to subtract the offset from.</param>
         <param name="elementOffset">The offset to subtract.</param>
-        <summary>Subtracts an element offset from the given reference.</summary>
-        <returns>A new reference that reflects the subraction of offset from pointer.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Subtracts an element offset from the given managed pointer.</summary>
+        <returns>A new managed pointer that reflects the subtraction of offset from the source pointer.</returns>
```
gewarren, line 1862:
```suggestion
        <returns>A new managed pointer that reflects the subtraction of the specified offset from the source pointer.</returns>
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1499,12 +1850,22 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+
+## Remarks
+
+The `elementOffset` parameter is the number of `T`-sized elements (not bytes) to subtract from the `source` pointer. For example, given a source pointer _ptr_ of type `ref int`, the call `Unsafe.Subtract<int>(ref ptr, (nuint)20)` will return a new pointer whose address points 20 bytes (= 10 elements \* 4 bytes per element) before _ptr_.
```
gewarren, line 1868:
```suggestion
The `elementOffset` parameter is the number of `T`-sized elements (not bytes) to subtract from the `source` pointer. For example, given a source pointer _ptr_ of type `ref int`, the call `Unsafe.Subtract<int>(ref ptr, (nuint)20)` will return a new pointer whose address points 80 bytes (= 20 elements \* 4 bytes per element) before _ptr_.
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1539,12 +1900,22 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
-        <typeparam name="T">The type of reference.</typeparam>
-        <param name="source">The reference to subtract the offset from.</param>
+        <typeparam name="T">The elemental type of the managed pointer.</typeparam>
+        <param name="source">The managed pointer to subtract the offset from.</param>
         <param name="byteOffset">The offset to subtract.</param>
-        <summary>Subtracts a byte offset from the given reference.</summary>
-        <returns>A new reference that reflects the subtraction of byte offset from pointer.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Subtracts a byte offset from the given managed pointer.</summary>
+        <returns>A new managed pointer that reflects the subtraction of byte offset from the source pointer.</returns>
```
gewarren, line 1912:
```suggestion
        <returns>A new managed pointer that reflects the subtraction of the specified byte offset from the source pointer.</returns>
```

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1581,12 +1952,22 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
-        <typeparam name="T">The type of reference.</typeparam>
-        <param name="source">The reference to subtract the offset from.</param>
+        <typeparam name="T">The elemental type of the managed pointer.</typeparam>
+        <param name="source">The managed pointer to subtract the offset from.</param>
         <param name="byteOffset">The offset to subtract.</param>
-        <summary>Subtracts a byte offset from the given reference.</summary>
-        <returns>A new reference that reflects the subraction of byte offset from pointer.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Subtracts a byte offset from the given managed pointer.</summary>
+        <returns>A new managed pointer that reflects the subtraction of byte offset from the source pointer.</returns>
```
gewarren, line 1964:
```suggestion
        <returns>A new managed pointer that reflects the subtraction of the specified byte offset from the source pointer.</returns>
```

xml/System.Runtime.InteropServices/MemoryMarshal.xml
```diff
@@ -565,19 +578,19 @@ This method can be useful if part of a managed object represents a fixed array.
...
+        <param name="value">The pointer to the null-terminated narrow character string.</param>
+        <summary>Creates a new read-only span for a null-terminated narrow character string.</summary>
         <returns>A read-only span representing the specified null-terminated string, or an empty span if the pointer is <see langword="null" />.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
-The returned span does not include the `null` terminator, nor does it validate the well-formedness of the UTF8 data.
+This method is typically used with narrow character strings, such as ANSI or UTF-8 strings. The returned span does not include the `null` terminator, nor does it validate the well-formedness of the narrow character data.
```
gewarren, line 589:
```suggestion
This method is typically used with narrow character strings, such as ANSI or UTF-8 strings. The returned span does not include the `null` terminator, nor does it validate whether the narrow character data is well-formed.
```

gewarren: Might be better for localization.

gewarren (approved):
Should we add a link for the references to ECMA 335? https://www.ecma-international.org/publications-and-standards/standards/ecma-335/

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -956,16 +1184,26 @@
         <ReturnType>System.Void</ReturnType>
       </ReturnValue>
       <Parameters>
-        <Parameter Name="startAddress" Type="System.Void*" Index="0" FrameworkAlternate="dotnet-plat-ext-2.1;dotnet-plat-ext-2.2;net-5.0;net-6.0;netcore-3.0;netcore-3.1" />
+        <Parameter Name="startAddress" Type="System.Byte" RefType="ref" Index="0" FrameworkAlternate="dotnet-plat-ext-2.1;dotnet-plat-ext-2.2;net-5.0;net-6.0;netcore-3.0;netcore-3.1" />
```
gewarren, line 1187: This line should not have changed.

GrabYourPitchforks: Good catch. One of my copy \& paste jobs really went awry. :)

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -25,8 +25,19 @@
...
+    <summary>Contains generic, low-level functionality for manipulating managed and unmanaged pointers.</summary>
+    <remarks>
+      <format type="text/markdown"><![CDATA[  
+
+## Remarks
+
+> [!WARNING]
+> This type is intended for advanced pointer manipulation scenarios. Callers are assumed to be familiar with ECMA-335, Sec. II.14.4 and III.1.1.5, and with the [ECMA-335 CLI Specification Addendum](https://github.com/dotnet/runtime/blob/main/docs/design/specs/Ecma-335-Augments.md).
+>
+> Most APIs on this type do not perform any argument checking or validation. Incorrect use of these APIs could destabilize the .NET runtime.
```
gfoidl, line 37:
```suggestion
> Most APIs on this type do not perform any argument checking or validation. Incorrect use of these APIs could lead to access violations, GC holes or destabilize the .NET runtime.
```
Came into my mind when reading this.
In essence it's similar, but to put some emphasis on the AV, and GC holes.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -223,12 +274,22 @@
...
-        <returns>A new reference that reflects the addition of byte offset to pointer.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Adds a byte offset to the given managed pointer.</summary>
+        <returns>A new managed pointer that reflects the addition of byte offset to the source pointer.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+
+## Remarks
+
+The `byteOffset` parameter is the number of bytes to add to the `source` pointer. For example, given a source pointer _ptr_ of type `ref int`, the call `Unsafe.AddByteOffset<int>(ref ptr, 20)` will return a new pointer whose address points 20 bytes beyond _ptr_.
```
gfoidl, line 287:
```suggestion
The `byteOffset` parameter is the number of bytes to add to the managed `source` pointer. For example, given a managed source pointer _ptr_ of type `ref int`, the call `Unsafe.AddByteOffset<int>(ref ptr, 20)` will return a new managed pointer whose address points 20 bytes beyond _ptr_.
```
Is this overkill to add _managed_ to pointer? 
To distinguish between native- / raw-pointers (`void*`) and managed pointers (`ref T`) here?
Out of the context it's very clear what is meant, and the APIs are targeting advanced users, so this shouldn't be needed.
But as docs they should be very precise, even for people that don't know these differences and to avoid potential confusion.

This would apply to other places as well.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -370,8 +441,18 @@
...
-        <remarks>To be added.</remarks>
+        <returns>The original object, cast to the given type.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+This API is used to cast an object to the given type, suppressing the runtime's normal type safety checks. It is the caller's responsibility to ensure that the cast is legal. No `InvalidCastException` will be thrown.
+
+The behavior of `Unsafe.As<T>(o)` is only well-defined if the typical "safe" casting operation `(T)o` would have succeeded. Use of this API to circumvent casts that would otherwise have failed is unsupported and could result in runtime instability.
```
gfoidl, line 452:
```suggestion
The behavior of `Unsafe.As<T>(o)` is only well-defined if the typical "safe" casting operation `(T)o` would have succeeded. Use of this API to circumvent casts that would otherwise have failed is unsupported and could result in runtime instability.
One could use `Debug.Assert(o is T)` to have an assertion at least in Debug-builds.
```
Or similar.

I see that quite a lot of people aren't aware of the possibilities of `Debug.Assert`, so maybe it's a good idea to give a pointer (😉) to that use in this context.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -407,12 +488,33 @@
...
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+This API is conceptually similar to C++'s `reinterpret_cast<>`. It is the caller's responsibility to ensure that the cast is legal. No runtime check will be performed.
+
+Only the managed pointer is reinterpreted. The referenced value itself will remain unchanged. Consider the example below.
+
+```cs
+int[] intArray = new int[] { 0x12345678 }; // a 1-element array
```
gfoidl, line 506:
```suggestion
int[] intArray = new int[] { 0x_1234_5678 }; // a 1-element array
```
For the example below it's easier to follow then.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -407,12 +488,33 @@
...
+ref int refToInt32 = ref intArray[0]; // managed pointer to first Int32 in array
+ref short refToInt16 = ref Unsafe.As<int, short>(ref refToInt32); // reinterpret as managed pointer to Int16
+Console.WriteLine(refToInt16);
+```
+
+The output of this program depends on the endianness of the current machine. On big-endian architectures this will output `0x1234`, and on little-endian architectures this will output `0x5678`.
+
+When casting a managed pointer from a narrower type to a wider type, the caller must ensure that dereferencing the pointer will not incur an out-of-bounds access. The caller is also responsible for ensuring that the resulting pointer is properly aligned for the referenced type.
```
gfoidl, line 514: > properly aligned for the referenced type

Can we link this to where "properly aligned" is specified?

I'm aware (at least I hope so...) on what rules apply here, but a natural question is here: "What is properly aligned for TTo?".

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -447,11 +549,21 @@
...
+        <summary>Converts a managed pointer into an unmanaged pointer.</summary>
+        <returns>An unmanaged pointer corresponding to the original source pointer.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+Unmanaged pointers are not tracked by the .NET garbage collector. If the original managed pointer points to a GC-managed object, it is the caller's responsibility to ensure that the source object is pinned.
+
+If the pointers reference memory in GC-owned address space, and if the source object is not pinned, the managed and unmanaged pointer values may eventually diverge due to GC relocation operations.
```
gfoidl, line 563:
```suggestion
If the pointers reference memory in GC-owned address space, and if the source object is not pinned, the managed and unmanaged pointer values may eventually diverge due to GC relocation operations. Dereferencing the unmanaged pointer could lead to access violations in such a case.
```
?

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -629,7 +751,25 @@
         <param name="destination">The location to copy to.</param>
         <param name="source">A reference to the value to copy.</param>
         <summary>Copies a value of type <typeparamref name="T" /> to the given location.</summary>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+Both `destination` and `source` are assumed to be properly aligned for pointers to data of type `T`.
```
gfoidl, line 759: Link to "property aligned"?

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -669,7 +809,25 @@
         <param name="destination">The location to copy to.</param>
         <param name="source">A pointer to the value to copy.</param>
         <summary>Copies a value of type <typeparamref name="T" /> to the given location.</summary>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+Both `destination` and `source` are assumed to be properly aligned for pointers to data of type `T`.
```
gfoidl, line 817: Link to "property aligned"?

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -702,11 +860,21 @@
...
         <summary>Copies bytes from the source address to the destination address.</summary>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+This API corresponds to the `cpblk` opcode.
+
+Both `destination` and `source` are assumed to be aligned to the machine's natural word size. See ECMA-335, Sec. III.3.30 ("cpblk - copy data from memory to memory") for more information.
```
gfoidl, line 874: Such a hint for alignment is IMO enough. Something like this could be used for the earlier comments about alignment.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1192,12 +1438,20 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
          </summary>
-        <returns>An object of type <typeparamref name="T" /> read from the given location.</returns>
-        <remarks>To be added.</remarks>
+        <returns>A value of type <typeparamref name="T" /> read from the given location.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+`source` is assumed to be a properly aligned pointer to a value of type `T`.
```
gfoidl, line 1451: Link to "property aligned"?

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1695,11 +2081,19 @@ For more information, see sections III.1.8.1.2.2 ("Controlled-muttability manage
...
         <param name="destination">The location to write to.</param>
         <param name="value">The value to write.</param>
         <summary>Writes a value of type <typeparamref name="T" /> to the given location.</summary>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+`destination` is assumed to be a properly aligned pointer to a value of type `T`.
```
gfoidl, line 2093: Link to "property aligned"?

xml/System.Runtime.InteropServices/MemoryMarshal.xml
```diff
@@ -992,7 +1008,16 @@ This method allows a read-only memory buffer to be used in existing APIs that re
...
         <returns>
           <see langword="true" /> if the method call succeeds; <see langword="false" /> otherwise.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+  
+## Remarks  
+
+> [!CAUTION]
+> <xref:System.ReadOnlyMemory%601> is used to represent immutable data. <xref:System.ArraySegment%601> instances returned by this method should not be written to.
```
gfoidl, line 1017:
```suggestion
> <xref:System.ReadOnlyMemory%601> is used to represent immutable data. <xref:System.ArraySegment%601> instances returned by this method should not be written to, and treated as read-only as well.
```
?
It's redundant, but one can't this express often enough...

xml/System.Runtime.InteropServices/MemoryMarshal.xml
```diff
@@ -565,19 +578,19 @@ This method can be useful if part of a managed object represents a fixed array.
         <Parameter Name="value" Type="System.Byte*" Index="0" FrameworkAlternate="net-6.0;net-7.0" />
       </Parameters>
       <Docs>
-        <param name="value">The pointer to the null-terminated string of bytes.</param>
-        <summary>Creates a new read-only span for a null-terminated UTF8 string.</summary>
+        <param name="value">The pointer to the null-terminated narrow character string.</param>
```
AaronRobinsonMSFT, line 581: I'm not sure talking about characters in describing the arguments is helpful here. I think this should instead be "a sequence of bytes ending with a null value". The remarks can provide potential use cases, but `byte` here is hard for me to accept as "narrow character".

xml/System.Runtime.CompilerServices/Unsafe.xml
GrabYourPitchforks: Neat suggestion!

xml/System.Runtime.InteropServices/MemoryMarshal.xml
GrabYourPitchforks: I think calling out the use cases is merited, since it talks to why we made the API and what "supported" use cases look like. But your comment about moving that to remarks is well taken.

AaronRobinsonMSFT: > I think calling out the use cases is merited

Agree. I should have been clearer about my nit. I don't think the notion of "narrow character" should be on a parameter typed `byte`. It makes perfect sense in remarks though and I agree, our intent on use for these kinds of APIs should be shared clearly.

xml/System.Runtime.CompilerServices/Unsafe.xml
GrabYourPitchforks: Wanted to keep the Warnings section somewhat understandable by all devs, so I'll move the mention of GC holes later in the remarks section.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -60,12 +71,22 @@
         <Parameter Name="elementOffset" Type="System.Int32" Index="1" FrameworkAlternate="dotnet-plat-ext-2.1;dotnet-plat-ext-2.2;net-5.0;net-6.0;netcore-3.0;netcore-3.1" />
       </Parameters>
       <Docs>
-        <typeparam name="T">The type of void pointer.</typeparam>
-        <param name="source">The void pointer to add the offset to.</param>
+        <typeparam name="T">The type whose width will be used as a scale factor for `elementOffset`.</typeparam>
```
tannergooding, line 74: Why `width` and not `size`? This directly corresponds to `sizeof(T)`

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -60,12 +71,22 @@
...
+        <summary>Adds an element offset to the given unmanaged pointer.</summary>
+        <returns>A new unmanaged pointer that reflects the addition of offset to the source pointer.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+
+## Remarks
+
+The `elementOffset` parameter is the number of `T`-sized elements (not bytes) to add to the `source` pointer. For example, given a source pointer _ptr_, the call `Unsafe.Add<int>(ptr, 20)` will return a new pointer whose address points 80 bytes (= 20 elements \* 4 bytes per element) beyond _ptr_.
+
+If `elementOffset` is a calculated value rather than a hardcoded literal, callers should consider the possibility of integer overflow and should take appropriate precautions.
```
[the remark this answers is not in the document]
GrabYourPitchforks: Glad you like it! This was based on a similar warning I once saw regarding `malloc`. Will reword slightly to make it clearer.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -60,12 +71,22 @@
...
       <Docs>
-        <typeparam name="T">The type of void pointer.</typeparam>
-        <param name="source">The void pointer to add the offset to.</param>
+        <typeparam name="T">The type whose width will be used as a scale factor for `elementOffset`.</typeparam>
+        <param name="source">The unmanaged pointer to add the offset to.</param>
         <param name="elementOffset">The offset to add.</param>
-        <summary>Adds an element offset to the given void pointer.</summary>
-        <returns>A new void pointer that reflects the addition of offset to the specified pointer.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Adds an element offset to the given unmanaged pointer.</summary>
```
tannergooding, line 77: if we're touching this anyways, can we order this "properly":
```
<summary></summary>
<typeparam name=""></typeparam>
<param name=""></param>
<returns></returns>
<remarks></remarks>
```

tannergooding: (Noting "properly" is defined here as the order VS declares these by default when you type `///` and matches a good reading order for understanding the method)

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -101,12 +122,22 @@
...
-        <returns>A new reference that reflects the addition of offset to pointer.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Adds an offset to the given managed pointer.</summary>
+        <returns>A new managed pointer that reflects the addition of offset to the source pointer.</returns>
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+
+## Remarks
+
+The `elementOffset` parameter is the number of `T`-sized elements (not bytes) to add to the `source` pointer. For example, given a source pointer _ptr_ of type `ref int`, the call `Unsafe.Add<int>(ref ptr, 20)` will return a new pointer whose address points 80 bytes (= 20 elements \* 4 bytes per element) beyond _ptr_.
```
tannergooding, line 135: Is it intentional we use code quotations rather than `<paramref name="" />`?

xml/System.Runtime.CompilerServices/Unsafe.xml
tannergooding: nit: I much prefer `0x1234_5678`, the `x_` makes it much harder to read for me.

xml/System.Runtime.CompilerServices/Unsafe.xml
tannergooding: This is machine/architecture dependent.

The requirement of ECMA-335 is `byte`/`sbyte` are 1-byte aligned, `short`/`ushort` are 2-byte aligned, and `int`/`uint` are 4-byte aligned.

The alignment of `nint`/`nuint` must be 4-byte aligned on a 32-bit platform and 8-byte aligned on a 64-bit platform.

The alignment of types like `double`, `long`, and `ulong` could be 4-bytes on a 32-bit platform.

The layout and alignment requirements of user-defined types build off the above but can vary from between systems.

Generally its best to just use `Unsafe.Read/WriteUnaligned` if you're unsure and have the JIT optimize it down to a regular read/write on platforms that support it.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -531,10 +643,10 @@
...
       </Parameters>
       <Docs>
-        <typeparam name="T">The type of reference.</typeparam>
+        <typeparam name="T">The underlying type of the reference.</typeparam>
         <param name="source">The read-only reference to reinterpret.</param>
-        <summary>Reinterprets the given read-only reference as a reference.</summary>
-        <returns>A reference to a value of type <typeparamref name="T" />.</returns>
+        <summary>Reinterprets the given read-only reference as a mutable reference.</summary>
+        <returns>A mutable reference to a value of type <typeparamref name="T" />.</returns>
         <remarks>To be added.</remarks>
```
tannergooding, line 650: This deserves remarks that it is the callers responsibility to ensure the data is not mutated. Mutating a readonly reference may cause access violations or other runtime instability.

In some cases, the JIT even optimizes based on assumed immutable data and so mutating it may cause two methods to have differing views of the same data.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -486,10 +598,10 @@
...
       <Docs>
-        <typeparam name="T">The type of the interpreted location.</typeparam>
-        <param name="source">The location of the value to reference.</param>
-        <summary>Reinterprets the given location as a reference to a value of type <typeparamref name="T" />.</summary>
-        <returns>A reference to a value of type <typeparamref name="T" />.</returns>
+        <typeparam name="T">The elemental type of the managed pointer.</typeparam>
+        <param name="source">The unmanaged pointer to convert.</param>
+        <summary>Converts an unmanaged pointer into a managed pointer to a value of type <typeparamref name="T" />.</summary>
+        <returns>A managed pointer to a value of type <typeparamref name="T" />.</returns>
         <remarks>To be added.</remarks>
```
tannergooding, line 605: This deserves a remark that it is the callers responsibility to ensure the pointer is properly aligned for `T`. The JIT assumes byrefs are aligned by default.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -740,11 +908,21 @@
...
         <summary>Copies bytes from the source address to the destination address.</summary>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+This API corresponds to the `cpblk` opcode.
+
+Both `destination` and `source` are assumed to be aligned to the machine's natural word size. See ECMA-335, Sec. III.3.30 ("cpblk - copy data from memory to memory") for more information.
```
tannergooding, line 922: "natural word size" is somewhat confusing/ambiguous even in relation to the spec reference.

Some other wording that clarifies this is just saying the alignment must be `sizeof(nint)` (the same alignment as required for a pointer) would be immensely better.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1154,10 +1400,10 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
       <Parameters />
       <Docs>
-        <typeparam name="T">The type of the reference.</typeparam>
-        <summary>Returns a reference to a value of type <typeparamref name="T" /> that is a null reference.</summary>
-        <returns>A reference to a value of type <typeparamref name="T" /> that is a null reference.</returns>
-        <remarks>To be added.</remarks>
+        <typeparam name="T">The elemental type of the managed pointer.</typeparam>
+        <summary>Returns a null managed pointer to a value of type <typeparamref name="T" />.</summary>
+        <returns>A null managed pointer to a value of type <typeparamref name="T" />.</returns>
+        <remarks>The return value is conceptually similar to <code>ref *((T*)nullptr)</code>.</remarks>
```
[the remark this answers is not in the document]
tannergooding: Would `AsRef<T>(null)` be less confusing?

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1306,10 +1560,23 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+> [!CAUTION]
+> This API returns the size of the _managed_ view of the type. For the size of the _unmanaged_ view of the type, such as for interop purposes, use <xref:System.Runtime.InteropServices.Marshal.SizeOf%2A>.
+
+This API corresponds to the `sizeof` opcode.
+
+If `T` is a reference type, the return value is the size of the reference itself, or `IntPtr.Size`. See ECMA-335, Sec. Sec. III.4.25  ("sizeof - load the size, in bytes, of a type") for more information.
```
tannergooding, line 1576: `, or IntPtr.Size` is going to read as if its one or the other to some people.

It might be better to just specify `sizeof(nint)` or `sizeof(void*)`

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -1581,12 +1957,22 @@ This check is conceptually similar to `(void*)(&left) < (void*)(&right)`.
...
+
+The `byteOffset` parameter is the number of bytes to subtract from the `source` pointer. For example, given a source pointer _ptr_ of type `ref int`, the call `Unsafe.SubtractByteOffset<int>(ref ptr, (nuint)20)` will return a new pointer whose address points 20 bytes before _ptr_.
+
+If `elementOffset` is a calculated value rather than a hardcoded literal, callers should consider the possibility of integer overflow and should take appropriate precautions.
+
+]]></format>
+        </remarks>
       </Docs>
     </Member>
     <Member MemberName="Unbox&lt;T&gt;">
```
tannergooding, line 1978: not changes to `Unbox`?

xml/System.Runtime.CompilerServices/Unsafe.xml
GrabYourPitchforks: @gewarren Thoughts on this? I'm looking through a small sample of other .xml files in the repo, and the consistent pattern I see is "typeparam, param, summary, returns, remarks." If we wanted to reorder this to be more proper I wonder if that should be something that's done repo-wide rather than as a one-off inside this file.

gewarren: Yes, I don't think it's worthwhile to reorder the tags. Also, I have a suspicion that mdoc will just reorder them the next time we update the APIs.

xml/System.Runtime.CompilerServices/Unsafe.xml
GrabYourPitchforks: The pattern I see elsewhere is that we use `<paramref>` for non-markdown sections and backticks for markdown sections. But @gewarren to confirm this?

gewarren: Since this is a Markdown section, the backticks are correct.

xml/System.Runtime.CompilerServices/Unsafe.xml
```diff
@@ -141,12 +172,22 @@
...
-        <typeparam name="T">The type of reference.</typeparam>
-        <param name="source">The reference to add the offset to.</param>
+        <typeparam name="T">The elemental type of the managed pointer.</typeparam>
+        <param name="source">The managed pointer to add the offset to.</param>
         <param name="elementOffset">The offset to add.</param>
-        <summary>Adds an element offset to the given reference.</summary>
-        <returns>A new reference that reflects the addition of offset to pointer.</returns>
-        <remarks>To be added.</remarks>
+        <summary>Adds an element offset to the given managed pointer.</summary>
+        <returns>A new managed pointer that reflects the addition of offset to the source pointer.</returns>
```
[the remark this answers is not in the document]
GrabYourPitchforks: The pattern I've tried to follow throughout the doc is that I reserve _reference_ for instances where the API shape involves `T*` or `T&` and where the operation involves reading or writing a full _T_ at that address. You'll see this in APIs like `Copy`.

For APIs that perform arithmetic or comparisons against the pointer value rather than dereference it, I use _\[un\]managed pointer_, as this terminology matches the ECMA sections listed in the type's Summary section.

For APIs like `ReadUnaligned`, I take a little bit of license here. Nominally the overloads would've been `ReadUnaligned<T>(void*)` and `ReadUnaligned<T>(void&)`, which .NET fully supports but which C# disallows. In both of those cases they're taking void pointers, but they're not dereferencing a `void`. They're instead performing a read or write of some differently-sized data at that address. Since they're not dereferencing a `void` I opted to use the terminology _\[un\]managed pointer_ here rather than _reference_, since IMO it's more accurate.

xml/System.Runtime.CompilerServices/Unsafe.xml
tannergooding: Is it worth a tracking issue for mdoc to be updated? It's definitely much harder to read/review these files because of the difference

xml/System.Runtime.CompilerServices/Unsafe.xml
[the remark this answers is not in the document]
GrabYourPitchforks: I had originally avoided `AsRef` because it increases the concept count of this API's docs.

xml/System.Runtime.CompilerServices/Unsafe.xml
GrabYourPitchforks: We recently updated the `Unbox` docs and I couldn't think of anything new to say. If you have suggestions I can definitely fold them in to this PR!

xml/System.Runtime.InteropServices/MemoryMarshal.xml
GrabYourPitchforks: I reworded this sentence entirely. Hopefully it's more clear now. :)

xml/System.Runtime.InteropServices/MemoryMarshal.xml
```diff
@@ -578,15 +578,15 @@ This method can be useful if part of a managed object represents a fixed array.
         <Parameter Name="value" Type="System.Byte*" Index="0" FrameworkAlternate="net-6.0;net-7.0" />
       </Parameters>
       <Docs>
-        <param name="value">The pointer to the null-terminated narrow character string.</param>
-        <summary>Creates a new read-only span for a null-terminated narrow character string.</summary>
+        <param name="value">The pointer to the null-terminated sequence of bytes.</param>
+        <summary>Creates a new read-only span for a null-terminated sequence of bytes.</summary>
         <returns>A read-only span representing the specified null-terminated string, or an empty span if the pointer is <see langword="null" />.</returns>
```
GrabYourPitchforks, line 583: I missed killing a "string" here.

GrabYourPitchforks:
@gewarren Any idea why the validation system is producing this warning?

```txt
Member InitBlockUnaligned's name and signature is not unique
```

Nothing really jumps out at me. It seems to follow the pattern for _CopyBlockUnaligned_ closely, and that method isn't generating warnings.

gewarren:
> @gewarren Any idea why the validation system is producing this warning?
> 
> ```
> Member InitBlockUnaligned's name and signature is not unique
> ```
> 
> Nothing really jumps out at me. It seems to follow the pattern for _CopyBlockUnaligned_ closely, and that method isn't generating warnings.

It should go away the next time the build completes. I believe it was due to your copy/paste error - https://github.com/dotnet/dotnet-api-docs/pull/8097#discussion_r879966308.

xml/System.Runtime.CompilerServices/Unsafe.xml
gewarren: Yes - I logged https://dev.azure.com/ceapex/Engineering/_workitems/edit/619653.

GrabYourPitchforks:
@gewarren I think we're almost there! But the docs engine is complaining about the xref in this sentence:

```txt
Consider instead using <xref:System.Buffer.MemoryCopy> or <xref:System.Span%601.CopyTo%2A> for this scenario.
```

I'm trying to link to https://docs.microsoft.com/dotnet/api/system.buffer.memorycopy. Am I using the wrong syntax here?

gewarren:
> @gewarren I think we're almost there! But the docs engine is complaining about the xref in this sentence:
> 
> ```
> Consider instead using <xref:System.Buffer.MemoryCopy> or <xref:System.Span%601.CopyTo%2A> for this scenario.
> ```
> 
> I'm trying to link to https://docs.microsoft.com/dotnet/api/system.buffer.memorycopy. Am I using the wrong syntax here?

Yes, since there's more than one overload, you need to add the asterisk, so `<xref:System.Buffer.MemoryCopy%2A>`.

GrabYourPitchforks:
Thanks kindly @gewarren @tannergooding @AaronRobinsonMSFT @gfoidl for your input! 🥳

PR #8098, opened 2022-05-24
Add HttpClient instancing info

### LINKED ISSUE #8082: Some docs pages recommend to instantiate HttpClient once and reuse it for the lifetime of the application, others recommend to instantate HttpClient many times with IHttpClientFactory
Two different docs pages seem to be inconsistent with each other.

[HttpClient Class § Remarks](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-6.0#remarks) says:

> [HttpClient](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-6.0) is intended to be instantiated once and re-used throughout the life of an application. Instantiating an HttpClient class for every request will exhaust the number of sockets available under heavy loads. This will result in SocketException errors. Below is an example using HttpClient correctly.

This page makes no reference to `IHttpClientFactory`.

However, [Make HTTP requests using IHttpClientFactory in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-6.0) recommends to instead use `IHttpClientFactory` to create multiple instances of `HttpClient` and manage them. **Note that this page does not limit itself to the context of ASP.NET Core, in section [§ Use IHttpClientFactory in a console app](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-6.0#use-ihttpclientfactory-in-a-console-app) it also instructs us how to do this outside of ASP.NET Core**.

From this docs page:

> IHttpClientFactory offers the following benefits:
> 
> (…)
> 
> - Manages the pooling and lifetime of underlying HttpClientMessageHandler instances. Automatic management avoids common DNS (Domain Name System) problems that occur when manually managing HttpClient lifetimes.

Section [§ HttpClient and lifetime management](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-6.0#httpclient-and-lifetime-management) elaborates that:

> IHttpClientFactory pools the HttpMessageHandler instances created by the factory to reduce resource consumption. An HttpMessageHandler instance may be reused from the pool when creating a new HttpClient instance if its lifetime hasn't expired.
> 
> Pooling of handlers is desirable as each handler typically manages its own underlying HTTP connections. Creating more handlers than necessary can result in connection delays. Some handlers also keep connections open indefinitely, which can prevent the handler from reacting to DNS (Domain Name System) changes.

It later explicitly hints that creating a single instance of `HttpClient` and reusing it for the lifetime of the application is made obsolete by `IHttpClientFactory`:

> Keeping a single HttpClient instance alive for a long duration is a common pattern used before the inception of IHttpClientFactory. This pattern becomes unnecessary after migrating to IHttpClientFactory.

OTOH in section [§ Cookies](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-6.0#cookies) this page mentions that `IHttpClientFactory` might not be the best option when using cookies:

> The pooled HttpMessageHandler instances results in CookieContainer objects being shared. Unanticipated CookieContainer object sharing often results in incorrect code. For apps that require cookies, consider either:
>
> - Disabling automatic cookie handling
> - Avoiding IHttpClientFactory

Suggestion
===

Given the above I belive that the docs should be made consistent by either:

- Mentioning in HttpClient Class § Remarks](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-6.0#remarks) that, with the possible exception of applications heavily utilizing cookies, `IHttpClientFactory` is the recommended alternative to manually creating one `HttpClient` instance and reusing it for the lifetime of the application;
- Or no longer recommending `IHttpClientFactory` in the ASP.NET Core docs.

gewarren:
Fixes #8082.

### DIFF
--- a/xml/System.Net.Http/HttpClient.xml
+++ b/xml/System.Net.Http/HttpClient.xml
@@ -42,7 +42,7 @@ The <xref:System.Net.Http.HttpClient> class instance acts as a session to send H
 
 ### Instancing
 
-<xref:System.Net.Http.HttpClient> is intended to be instantiated once and reused throughout the life of an application. HttpClient pools connections and reuses a connection across multiple requests. If you instantiate an HttpClient class for every request, the number of sockets available under heavy loads will be exhausted. This exhaustion will result in <xref:System.Net.Sockets.SocketException> errors.
+<xref:System.Net.Http.HttpClient> is intended to be instantiated once and reused throughout the life of an application. In .NET Core and .NET 5+, HttpClient pools connections inside the handler instance and reuses a connection across multiple requests. If you instantiate an HttpClient class for every request, the number of sockets available under heavy loads will be exhausted. This exhaustion will result in <xref:System.Net.Sockets.SocketException> errors.
 
 You can configure additional options by passing in a "handler", such as <xref:System.Net.Http.HttpClientHandler> (or <xref:System.Net.Http.SocketsHttpHandler> in .NET Core 2.1 or later), as part of the constructor. The connection properties on the handler cannot be changed once a request has been submitted, so one reason to create a new HttpClient instance would be if you need to change the connection properties. If different requests require different settings, this may also lead to an application having multiple <xref:System.Net.Http.HttpClient> instances, where each instance is configured appropriately, and then requests are issued on the relevant client.
 
@@ -121,7 +121,7 @@ Connection pool properties can be configured on a <xref:System.Net.Http.HttpClie
 Disposing of the HttpClient instance closes the open connections and cancels any pending requests.
 
 > [!NOTE]
-> If you later send a request to the same server, a new connection is created. If the rate of requests is high, or if there are any firewall limitations, that can exhaust the available sockets because of default TCP cleanup timers.
+> If you concurrently send HTTP/1.1 requests to the same server, new connections can be created. Even if you reuse the `HttpClient` instance, if the rate of requests is high, or if there are any firewall limitations, that can exhaust the available sockets because of default TCP cleanup timers. To limit the number of concurrent connections, you can set the `MaxConnectionsPerServer` property. By default, the number of concurrent HTTP/1.1 connections is unlimited.
 
 ### Buffering and request lifetime
 


xml/System.Net.Http/HttpClient.xml
```diff
@@ -42,38 +42,30 @@ The <xref:System.Net.Http.HttpClient> class instance acts as a session to send H
...
+
+HttpClient only resolves DNS entries when a connection is created. It does not track any time to live (TTL) durations specified by the DNS server. If DNS entries change regularly, which can happen in some container scenarios, the client won't respect those updates. To solve this issue, you can limit the lifetime of the connection by setting the <xref:System.Net.Http.SocketsHttpHandler.PooledConnectionLifetime?displayProperty=nameWithType> property, so that DNS lookup is required when the connection is replaced.
   
 ```csharp
 public class GoodController : ApiController
 {
-    private static readonly HttpClient HttpClient;
+    private static readonly HttpClient httpClient;
```
gfoidl, line 54:
```suggestion
    private static readonly HttpClient s_httpClient;
```
This is the standard naming convention in dotnet/runtime.
I don't know if this should be applied here too or what convention is followed in the docs.

xml/System.Net.Http/HttpClient.xml
```diff
@@ -42,38 +42,30 @@ The <xref:System.Net.Http.HttpClient> class instance acts as a session to send H
 
 ### Instancing
 
-<xref:System.Net.Http.HttpClient> is intended to be instantiated once and reused throughout the life of an application. HttpClient is designed to pool connections and reuse a connection across multiple requests. If you instantiate an HttpClient class for every request, the number of sockets available under heavy loads will be exhausted. This exhaustion will result in <xref:System.Net.Sockets.SocketException> errors. Following is an example that uses HttpClient correctly.  
+<xref:System.Net.Http.HttpClient> is intended to be instantiated once and reused throughout the life of an application. HttpClient pools connections and reuses a connection across multiple requests. If you instantiate an HttpClient class for every request, the number of sockets available under heavy loads will be exhausted. This exhaustion will result in <xref:System.Net.Sockets.SocketException> errors.
```
CarnaViire, line 45: As we are touching this -- maybe not just "HttpClient pools connections and reuses a connection across multiple requests." but "HttpClient pools connections *inside the instance* and reuses a connection across multiple requests." (or "inside the *handler* instance") -- as opposed to static pooling that was in .NET Framework?

xml/System.Net.Http/HttpClient.xml
```diff
@@ -128,6 +120,9 @@ Connection pool properties can be configured on a <xref:System.Net.Http.HttpClie
 
 Disposing of the HttpClient instance closes the open connections and cancels any pending requests.
 
+> [!NOTE]
+> If you later send a request to the same server, a new connection is created. If the rate of requests is high, or if there are any firewall limitations, that can exhaust the available sockets because of default TCP cleanup timers.
```
CarnaViire, line 124: I am not sure this statement is correct. Maybe concurrent requests were meant? Because if you send the request *later*, the pooled connection will be reused...

Maybe this could be something like
```suggestion
> If you concurrently send HTTP/1.1 requests to the same server, new connections can be created. Even if you reuse `HttpClient` instance, if the rate of requests is high, or if there are any firewall limitations, that can exhaust the available sockets because of default TCP cleanup timers. To limit the number of concurrent connections, you can use `MaxConnectionsPerServer` property. By default, the number of concurrent HTTP/1.1 connections is unlimited.
```

CarnaViire:
BTW, a huge thank you @gewarren for helping with HttpClient docs recently (e.g. https://github.com/dotnet/docs/pull/29568) ❤️

xml/System.Net.Http/HttpClient.xml
antonfirsov: I don't see consistent rules being followed in the samples, but the `s_` prefix is not very popular outside of `dotnet/runtime` source, I wouldn't advertise it.

xml/System.Net.Http/HttpClient.xml
gewarren, line 45:
```suggestion
<xref:System.Net.Http.HttpClient> is intended to be instantiated once and reused throughout the life of an application. In .NET Core and .NET 5+, HttpClient pools connections inside the handler instance and reuses a connection across multiple requests. If you instantiate an HttpClient class for every request, the number of sockets available under heavy loads will be exhausted. This exhaustion will result in <xref:System.Net.Sockets.SocketException> errors.
```

CarnaViire (approved):
LGTM, thanks!

PR #8107, opened 2022-05-26
Update Math.xml

wangyoutian:
Per the explanation for each quadrant and boundary, the returned value cannot equal to -π in any case. (it's -π < θ , not  -π  <= θ , in quadrant 2 )

[Preview link](https://review.docs.microsoft.com/en-us/dotnet/api/system.math.atan2?view=net-6.0&branch=pr-en-us-8107#returns).

### DIFF
--- a/xml/System/Math.xml
+++ b/xml/System/Math.xml
@@ -1027,7 +1027,7 @@
   
 -   For (<paramref name="x" />, <paramref name="y" />) in quadrant 2, π/2 &lt; θ ≤ π.
   
--   For (<paramref name="x" />, <paramref name="y" />) in quadrant 3, -π &le; θ &lt; -π/2.  
+-   For (<paramref name="x" />, <paramref name="y" />) in quadrant 3, -π ≤ θ &lt; -π/2.  
   
 -   For (<paramref name="x" />, <paramref name="y" />) in quadrant 4, -π/2 &lt; θ &lt; 0.  
   


xml/System/Math.xml
```diff
@@ -1021,7 +1021,7 @@
         <param name="y">The y coordinate of a point.</param>
         <param name="x">The x coordinate of a point.</param>
         <summary>Returns the angle whose tangent is the quotient of two specified numbers.</summary>
-        <returns>An angle, θ, measured in radians, such that -π ≤ θ ≤ π, and tan(θ) = <paramref name="y" /> / <paramref name="x" />, where (<paramref name="x" />, <paramref name="y" />) is a point in the Cartesian plane. Observe the following:
+        <returns>An angle, θ, measured in radians, such that -π \< θ ≤ π, and tan(θ) = <paramref name="y" /> / <paramref name="x" />, where (<paramref name="x" />, <paramref name="y" />) is a point in the Cartesian plane. Observe the following:
```
gewarren, line 1024:
```suggestion
        <returns>An angle, θ, measured in radians, such that -π &lt; θ ≤ π, and tan(θ) = <paramref name="y" /> / <paramref name="x" />, where (<paramref name="x" />, <paramref name="y" />) is a point in the Cartesian plane. Observe the following:
```

dakersnar (approved):
LGTM. @tannergooding this could use a second review to confirm this is an accurate correction.

xml/System/Math.xml
```diff
@@ -1021,7 +1021,7 @@
         <param name="y">The y coordinate of a point.</param>
         <param name="x">The x coordinate of a point.</param>
         <summary>Returns the angle whose tangent is the quotient of two specified numbers.</summary>
-        <returns>An angle, θ, measured in radians, such that -π ≤ θ ≤ π, and tan(θ) = <paramref name="y" /> / <paramref name="x" />, where (<paramref name="x" />, <paramref name="y" />) is a point in the Cartesian plane. Observe the following:
```
tannergooding, line 1024: This is not a valid change because of requirements put in place by the IEEE 754 (Chapter 9) and C Language (Annex F) Specifications:
* `atan2(±0, −0)` returns `±π`.
* `atan2(±0, x)` returns `±π` for `x < 0`.
* `atan2(±y, −∞)` returns `±π` for `finite y > 0`.

There are all special values that are not in the normal input domain and where the result returned is considered "inexact".

wangyoutian:
This cannot be closed, as the problem is not whether it shall be ">=" or ">" but that the statements in the document are contradictory in that: the overview statement in the document says one thing while the case-by-case statement in the document says another.

dakersnar:
Ahh, understood. It sounds like the case-by-case statements need to be updated instead, is that correct?

tannergooding (approved):

dakersnar (approved):

PR #8110, opened 2022-05-26
Update Thread.Abort docs

gewarren:
Follow up to https://github.com/dotnet/docs/issues/29624.

- Mostly cleans up formatting.
- Removes some .NET Framework 1.1 info.
- Notates some remarks as applicable only to .NET Framework.

### DIFF
--- a/xml/System.Threading/Thread.xml
+++ b/xml/System.Threading/Thread.xml
@@ -94,7 +94,7 @@
   
  The <xref:System.Threading.Thread> constructors can take either of two delegate types, depending on whether you can pass an argument to the method to be executed:  
   
--   If the method has no arguments, you pass a <xref:System.Threading.ThreadStart> delegate to  the constructor. It has the signature:  
+- If the method has no arguments, you pass a <xref:System.Threading.ThreadStart> delegate to the constructor. It has the signature:  
   
     ```csharp  
     public delegate void ThreadStart()  
@@ -104,12 +104,12 @@
     Public Delegate Sub ThreadStart()  
     ```  
   
-     The following example creates and starts a thread that executes the `ExecuteInForeground` method. The method displays  information about some thread properties, then executes a loop in which it pauses for half a second and displays the elapsed number of seconds. When the thread has executed for at least five seconds, the loop ends and the thread terminates execution.  
+     The following example creates and starts a thread that executes the `ExecuteInForeground` method. The method displays information about some thread properties, then executes a loop in which it pauses for half a second and displays the elapsed number of seconds. When the thread has executed for at least five seconds, the loop ends and the thread terminates execution.  
   
      :::code language="csharp" source="~/snippets/csharp/System.Threading/Thread/Overview/ThreadStart1.cs" id="Snippet1":::
      :::code language="vb" source="~/snippets/visualbasic/VS_Snippets_CLR_System/System.Threading.Thread/vb/ThreadStart1.vb" id="Snippet1":::  
   
--   If the method has an argument, you pass a <xref:System.Threading.ParameterizedThreadStart> delegate to the constructor. It has the signature:  
+- If the method has an argument, you pass a <xref:System.Threading.ParameterizedThreadStart> delegate to the constructor. It has the signature:  
   
     ```csharp  
     public delegate void ParameterizedThreadStart(object obj)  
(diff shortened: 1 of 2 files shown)


xml/System.Threading/Thread.xml
```diff
@@ -94,7 +94,7 @@
   
  The <xref:System.Threading.Thread> constructors can take either of two delegate types, depending on whether you can pass an argument to the method to be executed:  
   
--   If the method has no arguments, you pass a <xref:System.Threading.ThreadStart> delegate to  the constructor. It has the signature:  
+- If the method has no arguments, you pass a <xref:System.Threading.ThreadStart> delegate to the constructor. It has the signature:
```
kouvel, line 97: Not relevant to your change, but this statement doesn't make sense to me. "If the method has no arguments" followed by "you pass a <type> delegate to the constructor" are contradictory; if the method has no arguments, nothing could be passed into it. Perhaps it could just say "If the method has no arguments, it has the following signature:".

xml/System.Threading/Thread.xml
```diff
@@ -530,7 +530,7 @@ Main thread: ThreadProc.Join has returned.  Press Enter to end program.
 > [!NOTE]
 > On versions of Microsoft Windows prior to Windows XP and Windows Server 2003, `maxStackSize` is ignored, and the stack size specified in the executable header is used.  
   
- If you specify a very small stack size, you might need to disable stack-overflow probing. When the stack is severely constrained, the probing can itself cause a stack overflow. To disable stack overflow probing, add the following to your application configuration file.  
+ If you specify a very small stack size, you might need to disable stack-overflow probing. When the stack is severely constrained, the probing can itself cause a stack overflow. To disable stack overflow probing, add the following to your application configuration file in a .NET Framework app.
```
kouvel, line 533: I don't think this config section applies to .NET Core. My knowledge about this is limited, but AFAIK this is not configurable in .NET Core in `*.runtimeconfig.json`. Perhaps the previous statement is more accurate.

xml/System.Threading/Thread.xml
```diff
@@ -555,14 +555,14 @@ Main thread: ThreadProc.Join has returned.  Press Enter to end program.
...
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
   
 > [!IMPORTANT]
-> The `Thread.Abort` method should be used with caution. Particularly when you call it to abort a thread other than the current thread, you don't know what code has executed or failed to execute when the <xref:System.Threading.ThreadAbortException> is thrown. You also cannot be certain of the state of your application or any application and user state that it's responsible for preserving. For example, calling `Thread.Abort` may prevent the execution of static constructors or the release of unmanaged resources.
+> Use the `Thread.Abort` method with caution. Particularly when you call it to abort a thread other than the current thread, you don't know what code has executed or failed to execute when the <xref:System.Threading.ThreadAbortException> is thrown. You also cannot be certain of the state of your application or any application and user state that it's responsible for preserving. For example, calling `Thread.Abort` may prevent the execution of static constructors or the release of managed or unmanaged resources.
```
kouvel, line 565: > or the release of managed or unmanaged resources

Managed resources would be released at some point, I think the main issue is with unmanaged resources.

kouvel (approved):

xml/System.Threading/Thread.xml
kouvel: All of the overloads of the constructors have arguments, I guess what it's referring to is whether there are arguments to the callback. Maybe it can say "if the callback has no arguments" and that would be more clear?

kouvel: Perhaps the previous sentence adds enough context to say what "the method" is referring to. Maybe it's fine as is.

xml/System.Threading/Thread.xml
gewarren: @kouvel I've added "in a .NET Framework app" - so doesn't that make the addition more accurate since it excludes .NET Core?

kouvel: Ah yes I think I confused this with something else, this looks more accurate to me, thanks!

xml/System.Threading/Thread.xml
gewarren, line 565:
```suggestion
> Use the `Thread.Abort` method with caution. Particularly when you call it to abort a thread other than the current thread, you don't know what code has executed or failed to execute when the <xref:System.Threading.ThreadAbortException> is thrown. You also cannot be certain of the state of your application or any application and user state that it's responsible for preserving. For example, calling `Thread.Abort` may prevent the execution of static constructors or the release of unmanaged resources. It can even corrupt managed data structures.
```

gewarren: @kouvel I updated it based on https://github.com/dotnet/docs/issues/29624#issuecomment-1137770672. Does this look okay to you?

PR #8116, opened 2022-05-29
Update Regex.xml to clarify behavior of `startat` and `beginning`

danmoseley:
## Summary

Per discussion in https://github.com/dotnet/runtime/issues/69949#issuecomment-1140388943

### DIFF
--- a/xml/System.Text.RegularExpressions/Regex.xml
+++ b/xml/System.Text.RegularExpressions/Regex.xml
@@ -3224,7 +3224,7 @@ The <xref:System.Text.RegularExpressions.Regex.Match(System.String,System.Int32)
 
 ### The `startat` parameter
   
- You can optionally specify a starting position in the string by using the `startat` parameter. Any matches starting before `startat` in the string are ignored. If you don't specify a starting position, the search begins at the default position, which is at the left end of `input` in a left-to-right search, and at the right end of `input` in a right-to-left search. Despite starting at `startat`, the index of any returned match is relative to the start of the string.
+ You can optionally specify a starting position in the string by using the `startat` parameter. Any matches starting before `startat` in the string are ignored. If you don't specify a starting position, the search begins at the default position, which is the left end of `input` in a left-to-right search, and the right end of `input` in a right-to-left search. Despite starting at `startat`, the index of any returned match is relative to the start of the string.
 
  Although the regular expression engine doesn't return any match starting before `startat`, it doesn't ignore the string before `startat`. This means that assertions such as [anchors](/dotnet/standard/base-types/anchors-in-regular-expressions) or [lookbehind assertions](/dotnet/standard/base-types/backtracking-in-regular-expressions#lookbehind-assertions) still apply to the input as a whole. For example, the following code includes a pattern with a lookbehind assertion that's satisfied even though it occurs before the `startat` index of 5 in the input string.
 
@@ -3233,7 +3233,7 @@ The <xref:System.Text.RegularExpressions.Regex.Match(System.String,System.Int32)
 > [!TIP]
 >
 > - If a pattern starts with the `^` anchor but `startat` is greater than 0, no matches will ever be found in a single-line search since they are constrained by `^` to start at index 0.
-> - The [`\G` anchor](/dotnet/standard/base-types/anchors-in-regular-expressions#contiguous-matches-g) is satisfied at `startat`. Because of this, if you want to restrict a match so that it begins exactly at a particular character position in the string, anchor the regular expression with a `\G` at the left for a left-to-right pattern. This restricts the match so it must start exactly at `startat` (or, when multiple matches are desired, so the matches are contiguous).
+> - The [`\G` anchor](/dotnet/standard/base-types/anchors-in-regular-expressions#contiguous-matches-g) is satisfied at `startat`. Because of this, if you want to restrict a match so that it begins exactly at a particular character position in the string, anchor the regular expression with a `\G` on the left for a left-to-right pattern. This restricts the match so it must start exactly at `startat` (or, when multiple matches are desired, so the matches are contiguous).
 
 ### Right-to-left searches
 
@@ -3246,7 +3246,7 @@ A right-to-left search, that is, when the regular expression pattern is construc
 
 For more information about right-to-left searches, see [Right-to-left mode](/dotnet/standard/base-types/regular-expression-options#right-to-left-mode).
 
-### Determine if a match is found
+### Determine whether a match is found
   
  You can determine whether the regular expression pattern has been found in the input string by checking the value of the returned <xref:System.Text.RegularExpressions.Match> object's <xref:System.Text.RegularExpressions.Group.Success%2A> property. If a match is found, the returned <xref:System.Text.RegularExpressions.Match> object's <xref:System.Text.RegularExpressions.Capture.Value%2A> property contains the substring from `input` that matches the regular expression pattern. If no match is found, its value is <xref:System.String.Empty?displayProperty=nameWithType>.
 


xml/System.Text.RegularExpressions/Regex.xml
```diff
@@ -3066,9 +3068,13 @@ Allows an <see cref="T:System.Object" /> to attempt to free resources and perfor
...
+ You can optionally specify a starting position in the string by using the `startat` parameter. When the regular expression engine parses from left to right (the default), any matches starting before `startat` in the string are ignored. When the regular expression engine parses from right to left (when the regular expression pattern is constructed with the <xref:System.Text.RegularExpressions.RegexOptions.RightToLeft?displayProperty=nameWithType> option), the match and scan move in the opposite direction and begin with the character at `startat` - 1 and any matches ending after `startat` are ignored. If you do not specify a starting position, the search begins at the default `startat`. If the regular expression searches from left to right, the default `startat` position is at the left end of `input`; if it searches from right to left, the default `startat` position is at the right end of `input`.
+
+ Despite starting at `startat`, the index of any returned match will be relative to the start of the string.
```
gewarren, line 3073:
```suggestion
 Despite starting at `startat`, the index of any returned match is relative to the start of the string.
```

xml/System.Text.RegularExpressions/Regex.xml
```diff
@@ -2427,7 +2427,9 @@ Allows an <see cref="T:System.Object" /> to attempt to free resources and perfor
...
   
 ## Remarks  
- The <xref:System.Text.RegularExpressions.Regex.IsMatch%2A> method is typically used to validate a string or to ensure that a string conforms to a particular pattern without retrieving that string for subsequent manipulation. If you want to determine whether one or more strings match a regular expression pattern and then retrieve them for subsequent manipulation, call the <xref:System.Text.RegularExpressions.Regex.Match%2A> or <xref:System.Text.RegularExpressions.Regex.Matches%2A> method.  
+ The <xref:System.Text.RegularExpressions.Regex.IsMatch%2A> method is typically used to validate a string or to ensure that a string conforms to a particular pattern without retrieving that string for subsequent manipulation. If you want to determine whether one or more strings match a regular expression pattern and then retrieve them for subsequent manipulation, call the <xref:System.Text.RegularExpressions.Regex.Match%2A> or <xref:System.Text.RegularExpressions.Regex.Matches%2A> method.
+
+ See the Remarks section of <xref:System.Text.RegularExpressions.Regex.Match(System.String,System.Int32)> for more details about `startat`.
```
gewarren, line 2432:
```suggestion
For more information about the `startat` parameter, see the Remarks section of <xref:System.Text.RegularExpressions.Regex.Match(System.String,System.Int32)>.
```

gewarren: Can you change this in the other places too?

xml/System.Text.RegularExpressions/Regex.xml
```diff
@@ -3066,9 +3068,13 @@ Allows an <see cref="T:System.Object" /> to attempt to free resources and perfor
...
+
+ Despite starting at `startat`, the index of any returned match will be relative to the start of the string.
+
+ Although the regular expression engine will not return any match starting before `startat` (or ending after `startat`, for a right-to-left pattern), it does not ignore the string before `startat` (after for a right-to-left pattern). This means that any anchors, such as `^` or `\A`, or zero-width assertions at the start of the pattern may or may not be satisfied depending on the content of the string before `startat` (after for a right-to-left pattern).
```
gewarren, line 3075:
```suggestion
 Although the regular expression engine doesn't return any match starting before `startat` (or ending after `startat`, for a right-to-left search), it doesn't ignore the string before `startat`. This means that any anchors, such as `^` or `\A`, or zero-width assertions at the start of the pattern may or may not be satisfied depending on the content of the string before `startat`.
```

gewarren: I think right-to-left only means that the search starts at the right end instead of the left end. The pattern itself is still matched from left to right.

xml/System.Text.RegularExpressions/Regex.xml
```diff
@@ -3241,7 +3247,11 @@ Allows an <see cref="T:System.Object" /> to attempt to free resources and perfor
...
+ The <xref:System.Text.RegularExpressions.Regex.Match(System.String,System.Int32,System.Int32)> method searches the portion of `input` defined by the `beginning` and `length` parameters for the regular expression pattern. `beginning` always defines the index of the leftmost character to include in the search, and `length` defines the maximum number of characters to search. Together, they define the range of the search. The behavior is exactly as if the `input` was effectively `input.Substring(beginning, length)` except that the index of any match is counted relative to the start of `input`. This means that any anchors or zero-width assertions at the start or end of the pattern behave as if there is no `input` outside of this range. For example the anchors `^`, `\G`, and `\A` will be satisfied at `beginning` and `$` and `\z` will be satisfied at `beginning - length`.
```
gewarren, line 3250:
```suggestion
 The <xref:System.Text.RegularExpressions.Regex.Match(System.String,System.Int32,System.Int32)> method searches the portion of `input` defined by the `beginning` and `length` parameters for the regular expression pattern. `beginning` always defines the index of the leftmost character to include in the search, and `length` defines the maximum number of characters to search. Together, they define the range of the search. The behavior is exactly as if the `input` was effectively `input.Substring(beginning, length)` except that the index of any match is counted relative to the start of `input`. This means that any anchors or zero-width assertions at the start or end of the pattern behave as if there is no `input` outside of this range. For example the anchors `^`, `\G`, and `\A` will be satisfied at `beginning` and `$` and `\z` will be satisfied at `beginning + length - 1`.  
```

gewarren: I think?

gewarren (commented):
I left some comments - they may be incorrect though. But for this code, it matches "building ", not "dub " or "bud ", so the pattern is still matched from left to right even when the search is right-to-left.

```
string pattern = @"\bb\w+\s";
string input = "builder building dub";
Regex regex = new Regex(pattern, RegexOptions.RightToLeft);
Match match = regex.Match(input);
```

danmoseley:
> so the pattern is still matched from left to right even when the search is right-to-left.

That's right. It just means "start on the right". I find it a confusing name, because I think of RTL in the context of scripts, which implies that the pattern is reversed as well and it's not.

We know from the corpus that it's very rarely used. I guess the scenarios are -- 
1. you want the first match on the right, and either don't want to iterate through preceding ones to get there, or the matches overlap so you'd get different matches if you started on the left; or
2. there's only one match, but you know it's near the end of a very large string so you're providing a performance hint (which may well be a deoptimization as we have many optimizations we only did for LTR)

is that about right @joperezr ?

joperezr:
Yup that seems correct to me.

Easiest way to think about RTL is essentially as if the whole pattern was a look behind. In fact, AFAIK when the initial support for lookbehinds was added all those years ago they decided to expose the Right-to-left functionality since all the code for "starting from the right and advancing in the opposite direction" was already implemented, so RTL was just added as an option. @stephentoub can correct me if I'm wrong, but I believe that is mainly why RTL option was added.

stephentoub:
> @stephentoub can correct me if I'm wrong, but I believe that is mainly why RTL option was added.

I don't know for certain, but that's been my hypothesis. :)

> It just means "start on the right". I find it a confusing name, because I think of RTL in the context of scripts, which implies that the pattern is reversed as well and it's not.

The pattern is reversed:
https://github.com/dotnet/runtime/blob/187bb1f22727310c669fb36dee0dfaf8932b4253/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexNode.cs#L116-L117
and the reversed pattern is then evaluated against the input going from right to left.  In the cited example:
```C#
string pattern = @"\bb\w+\s";
string input = "builder building dub";
Regex regex = new Regex(pattern, RegexOptions.RightToLeft);
Match match = regex.Match(input);
```
it searches right-to-left for a space, then right-to-left for word characters, then right-to-left for a b, etc.  That's why it matches `building`.

danmoseley:
> Easiest way to think about RTL is essentially as if the whole pattern was a look behind.

Hmm, can you give a example? I'm having trouble visualizing. Do you mean something like, precede the pattern with `(?<=^.{X})` where X starts as "string length -1" and then backs up by 1 or the match length each time?

danmoseley:
Never mind I see Stephen's reply.

joperezr:
I had to correct that comment a few minutes ago :P, what I meant is that it behaves as a lookbehind, I have edited the comment and added an example of what I meant.

xml/System.Text.RegularExpressions/Regex.xml
gewarren: By "... (after for a RTL pattern)", do you mean it would be to the left of `startat`? I think before/after could be confusing here.

danmoseley: No, to the right. Eg., given text "abcxxx" the pattern "abc(?=def)" and startat of 3 in a RTL match, it will start on the "c" and move left, but the match will fail because the "(?=def)" lookahead does not match the "xxx".

This is further complicated because when you pass "startat" to a RTL pattern, it starts at "startat - 1" (this is already documented) 

I'll think about how to write this more clearly. It's all a bit unfortunate, since 99.9% of people reading these docs don't care about RTL matching.

xml/System.Text.RegularExpressions/Regex.xml
danmoseley: You're right. I don't know where that came from.

danmoseley:
> The pattern is reversed:...and the reversed pattern is then evaluated against the input going from right to left

That's the implementation, but if I understand right, if one describes the LTR maching behavior as "pattern is evaluated against input left to right, looking for matches that start at the index being examined" one could describe RTL as "pattern is evaluated against input right to left, looking for matches that end at the index being examined". It's just easier to implement it by reversing the pattern.

stephentoub:
>>> which implies that the pattern is reversed as well and it's not.

>> The pattern is reversed:...and the reversed pattern is then evaluated against the input going from right to left

> That's the implementation, but if I understand right, if one describes the LTR maching behavior as "pattern is evaluated against input left to right, looking for matches that start at the index being examined" one could describe RTL as "pattern is evaluated against input right to left, looking for matches that end at the index being examined". It's just easier to implement it by reversing the pattern.

I'm not sure what distinction you're trying to draw.  The pattern is evaluated in reverse; whether that's achieved by walking it in reverse or by reversing it and then walking it forward, it's the same thing.  And evaluating it in reverse is critical to the semantics.

stephentoub:
> And evaluating it in reverse is critical to the semantics.

There are many examples of this, but a really simple case is with backreferences. Consider:
```C#
Regex.Match("abcabc", @"\1(\w{3})", RegexOptions.None)
```
and
```C#
Regex.Match("abcabc", @"\1(\w{3})", RegexOptions.RightToLeft)
```
It is not correct to say "pattern is evaluated against input right to left, looking for matches that end at the index being examined", as is evidenced by the fact that the former doesn't match but the latter does.

stephentoub:
Another example:
```C#
Regex.Match("123456789", @"^(\d+)(\d+)$", RegexOptions.None).Groups[1]
```
vs 
```C#
Regex.Match("123456789", @"^(\d+)(\d+)$", RegexOptions.RightToLeft).Groups[1]
```
Here, again, evaluation order matters, because it impacts which capture group captures the text.  If the pattern is evaluated left-to-right, then the first capture group is going to get the bulk of the digits, and if it's evaluated right-to-left, then the second capture group is going to get the bulk of the digits.

And another example:
```C#
Regex.IsMatch("123456789", @"(?>\d*)\d", RegexOptions.None)
```
vs
```C#
Regex.IsMatch("123456789", @"(?>\d*)\d", RegexOptions.RightToLeft)
```
The former is false, because the atomic group will consume the whole input and then there won't be anything left to match the final \d.  But the latter is true, because when evaluated right-to-left, the \d is matched first, and only after it's matched is the \d* evaluated on the remainder of the input.

xml/System.Text.RegularExpressions/Regex.xml
gewarren: Maybe we should just have a separate section for RTL searches and have all the info specific to that kind of search there, instead of cluttering up the other paragraphs where 99.9% of readers don't need to know about it.

danmoseley: That would be my preference. I could even imagine removing any mention of RTL from all of the regex API docs except to link to one doc or section that says "but if you request RTL, these are all the things that change".

I don't have time to work on such a change though. What we have has sufficed for 20 years, I guess.

gewarren:
@danmoseley I created https://github.com/dotnet/docs/pull/29708 to update the conceptual doc about RTL search. If you like I can take over this PR and add links to that section.

danmoseley:
>  If you like, I can take over this PR and add links to that section.

Just saw this. @gewarren yes, that would be great. I wouldn't have time to continue it for a while. appreciate it.

gewarren:
[Preview link](https://review.docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.match?view=net-7.0&branch=pr-en-us-8116#the-startat-parameter) looks good to me.

xml/System.Text.RegularExpressions/Regex.xml
```diff
@@ -3218,18 +3218,43 @@ Allows an <see cref="T:System.Object" /> to attempt to free resources and perfor
...
+A right-to-left search, that is, when the regular expression pattern is constructed with the <xref:System.Text.RegularExpressions.RegexOptions.RightToLeft?displayProperty=nameWithType> option, behaves in the following ways:
+
+- The scan moves in the opposite direction and the pattern is matched from back (right) to front (left).
+- The default starting position is the right end of the input string.
+- If `startat` is specified, the right-to-left scan begins at the character at `startat` - 1 (not `startat`).
+- When the `\G` anchor is specified at the right end of a pattern, it restricts the (first) match to end exactly at `startat` - 1.
+
+For more information about right-to-left searches, see [Right-to-left mode](/dotnet/standard/base-types/regular-expression-options#right-to-left-mode).
+
+### Determine if a match is found
```
danmoseley, line 3249: Determine whether?

xml/System.Text.RegularExpressions/Regex.xml
```diff
@@ -3218,18 +3218,43 @@ Allows an <see cref="T:System.Object" /> to attempt to free resources and perfor
...
+
+ Although the regular expression engine doesn't return any match starting before `startat`, it doesn't ignore the string before `startat`. This means that assertions such as [anchors](/dotnet/standard/base-types/anchors-in-regular-expressions) or [lookbehind assertions](/dotnet/standard/base-types/backtracking-in-regular-expressions#lookbehind-assertions) still apply to the input as a whole. For example, the following code includes a pattern with a lookbehind assertion that's satisfied even though it occurs before the `startat` index of 5 in the input string.
+
+:::code language="csharp" source="~/snippets/csharp/System.Text.RegularExpressions/Regex/Match/startat.cs" interactive="try-dotnet":::
+
+> [!TIP]
+>
+> - If a pattern starts with the `^` anchor but `startat` is greater than 0, no matches will ever be found in a single-line search since they are constrained by `^` to start at index 0.
```
danmoseley, line 3235: “At the left” in some places - should be “to the left”?

danmoseley:
It won't let me approve (because I'm the author?) but I sign off - looks much improved, thank you @gewarren

xml/System.Text.RegularExpressions/Regex.xml
```diff
@@ -3218,18 +3218,43 @@ Allows an <see cref="T:System.Object" /> to attempt to free resources and perfor
...
+> - The [`\G` anchor](/dotnet/standard/base-types/anchors-in-regular-expressions#contiguous-matches-g) is satisfied at `startat`. Because of this, if you want to restrict a match so that it begins exactly at a particular character position in the string, anchor the regular expression with a `\G` at the left for a left-to-right pattern. This restricts the match so it must start exactly at `startat` (or, when multiple matches are desired, so the matches are contiguous).
```
gewarren, line 3236:
```suggestion
> - The [`\G` anchor](/dotnet/standard/base-types/anchors-in-regular-expressions#contiguous-matches-g) is satisfied at `startat`. Because of this, if you want to restrict a match so that it begins exactly at a particular character position in the string, anchor the regular expression with a `\G` to the left for a left-to-right pattern. This restricts the match so it must start exactly at `startat` (or, when multiple matches are desired, so the matches are contiguous).
```

xml/System.Text.RegularExpressions/Regex.xml
```diff
@@ -3218,18 +3218,43 @@ Allows an <see cref="T:System.Object" /> to attempt to free resources and perfor
...
+
+### The `startat` parameter
   
- If you want to restrict a match so that it begins at a particular character position in the string and the regular expression engine does not scan the remainder of the string for a match, anchor the regular expression with a `\G` (at the left for a left-to-right pattern, or at the right for a right-to-left pattern). This restricts the match so it must start exactly at `startat`.  
+ You can optionally specify a starting position in the string by using the `startat` parameter. Any matches starting before `startat` in the string are ignored. If you don't specify a starting position, the search begins at the default position, which is at the left end of `input` in a left-to-right search, and at the right end of `input` in a right-to-left search. Despite starting at `startat`, the index of any returned match is relative to the start of the string.
```
gewarren, line 3227:
```suggestion
 You can optionally specify a starting position in the string by using the `startat` parameter. Any matches starting before `startat` in the string are ignored. If you don't specify a starting position, the search begins at the default position, which is the left end of `input` in a left-to-right search, and the right end of `input` in a right-to-left search. Despite starting at `startat`, the index of any returned match is relative to the start of the string.
```

gewarren (approved):

PR #8123, opened 2022-06-02
System.IO.BinaryReader F# snippets

albert-du:
Adds F# snippets for BinaryReader


BillWagner (approved):
Thanks again @albert-du 

This is excellent work, as always. I'll :shipit: now.

PR #8130, opened 2022-06-03
make used terms more uniform

zp-zpanda:
## Summary

Makes used terms more uniform.

As far as I can see, one of the twos was an Arabic one, while the others are English, so I corrected it.


BillWagner (approved):
Thanks for making this fix @Zhu-Panda We appreciate it.

Thanks for helping us improve.

PR #8131, opened 2022-06-03
Fix transaction doc samples

### LINKED ISSUE #7892: SqlTransaction.Rollback(string) example doesn't actually exemplify the method
Docs at: https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqltransaction.rollback?view=dotnet-plat-ext-6.0

Source at: https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Data.SqlClient/SqlTransaction.xml

The `Rollback(string savePointName)` method is intended to roll back to a save point named as per the given argument to `Rollback`..

..but the example code given in the docs doesn't create any save points or demonstrate rolling back to them. It just gives another example of basic Begin/Commit/Rollback behavior, just like the no arg `Rollback()` above

it would be more useful if the demo code for `Rollback(string)` created a save point and demonstrated rolling back to it

David-Engel:
## Summary

Fixing transaction doc samples

BeginTransaction - no parameters
BeginTransaction2 - named transaction/savepoint
BeginTransaction3 - named transaction/savepoint + isolation level

Fixes #7892

### DIFF
--- a/snippets/csharp/VS_Snippets_ADO.NET/Classic WebData SqlConnection.BeginTransaction Example/CS/Project.csproj
+++ b/snippets/csharp/VS_Snippets_ADO.NET/Classic WebData SqlConnection.BeginTransaction Example/CS/Project.csproj
@@ -1,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>net6.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Data.SqlClient" Version="4.8.3" />
+  </ItemGroup>
+
+</Project>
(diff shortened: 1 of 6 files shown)


DavoudEshtehari (approved):

David-Engel:
@dotnet/docs I'm not sure why the Snippets 5000 check would be failing in this PR. The changes are minor ones in existing files, so the failure doesn't make sense unless there is an existing code/infra problem.

gewarren:
> @dotnet/docs I'm not sure why the Snippets 5000 check would be failing in this PR. The changes are minor ones in existing files, so the failure doesn't make sense unless there is an existing code/infra problem.

We recently made it a requirement to have a project file so any code snippets can be compiled - that's why Snippets 5000 is failing. Can you add a simple C# or VB project file to each folder?

David-Engel:
@dotnet/docs Can this be reviewed and merged?

gewarren (approved):

PR #8132, opened 2022-06-03
Fixing up the IntPtr/UIntPtr docs to clarify they are integer types

tannergooding:
This resolves https://github.com/dotnet/runtime/issues/69614

This updates the `IntPtr` and `UIntPtr` docs to clarify they are integer types and not pointers or handles. Adding clarifying comment that outside specialized scenarios, users should prefer using pointers or SafeHandle.

The docs generation likely needs an update for .NET 7 and later to prefer the `nint` and `nuint` keywords anywhere it shows `IntPtr` or `UIntPtr` today. This will bring it inline with how C# is going to behave for .NET 7 and newer. Prior to .NET 7, there is still a language distinction between `nint` and `IntPtr`.

### DIFF
--- a/xml/System/IntPtr.xml
+++ b/xml/System/IntPtr.xml
@@ -263,7 +263,7 @@
 
  In C# starting from version 9.0, you can use the built-in `nint` type to define native-sized integers. This type is represented by the <xref:System.IntPtr> type internally and provides operations and conversions that are appropriate for integer types. For more information, see [nint and nuint types](/dotnet/csharp/language-reference/builtin-types/nint-nuint).
  
- In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, `nint` is an alias for <xref:System.IntPtr> in the same way that `int` is an alias for <xref:System.Int32>.
+ In C# starting from version 11 and when targeting the .NET 7 or later runtime, `nint` is an alias for <xref:System.IntPtr> in the same way that `int` is an alias for <xref:System.Int32>.
 
 ## Examples
 
@@ -956,7 +956,7 @@ For <xref:System.Numerics.IFloatingPoint%601>, this method matches the IEEE 754:
 
  The addition operation does not throw an exception if the result is too large to represent as a signed integer in the executing process. Instead, it is performed in an unchecked context.
 
- In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The addition operator is directly recognized by the language and will follow the normal language behavior for addition operations, including overflowing in a `checked` context if the result is to large to represent.
+ In C# starting from version 11 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The addition operator is directly recognized by the language and will follow the normal language behavior for addition operations, including overflowing in a `checked` context if the result is too large to represent.
 
  The equivalent method for this operator is <xref:System.IntPtr.Add%28System.IntPtr%2CSystem.Int32%29?displayProperty=nameWithType>]]></format>
         </remarks>
@@ -1549,7 +1549,7 @@ For <xref:System.Numerics.IFloatingPoint%601>, this method matches the IEEE 754:
 
  The subtraction operation does not throw an exception if the result is too small to represent as a signed integer in the executing process. Instead, it is performed in an unchecked context.
 
- In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The subtraction operator is directly recognized by the language and will follow the normal language behavior for subtraction operations, including overflowing in a `checked` context if the result is to small to represent.
+ In C# starting from version 11 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The subtraction operator is directly recognized by the language and will follow the normal language behavior for subtraction operations, including overflowing in a `checked` context if the result is too small to represent.
 
  The equivalent method for this operator is <xref:System.IntPtr.Subtract%28System.IntPtr%2CSystem.Int32%29?displayProperty=nameWithType>]]></format>
         </remarks>
--- a/xml/System/UIntPtr.xml
+++ b/xml/System/UIntPtr.xml
@@ -265,7 +265,7 @@
 
  In C# starting from version 9.0, you can use the built-in `nuint` type to define native-sized integers. This type is represented by the <xref:System.UIntPtr> type internally and provides operations and conversions that are appropriate for integer types. For more information, see [nint and nuint types](/dotnet/csharp/language-reference/builtin-types/nint-nuint).
 
- In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, `nuint` is an alias for <xref:System.UIntPtr> in the same way that `uint` is an alias for <xref:System.UInt32>.
+ In C# starting from version 11 and when targeting the .NET 7 or later runtime, `nuint` is an alias for <xref:System.UIntPtr> in the same way that `uint` is an alias for <xref:System.UInt32>.
 
  ]]></format>
     </remarks>
@@ -893,7 +893,7 @@ For <xref:System.Numerics.IFloatingPoint%601>, this method matches the IEEE 754:
 
  The addition operation does not throw an exception if the result is too large to represent as an unsigned integer in the executing process. Instead, it is performed in an unchecked context.
  
- In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The addition operator is directly recognized by the language and will follow the normal language behavior for addition operations, including overflowing in a `checked` context if the result is to large to represent.
+ In C# starting from version 11 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The addition operator is directly recognized by the language and will follow the normal language behavior for addition operations, including overflowing in a `checked` context if the result is too large to represent.
 
  The equivalent method for this operator is <xref:System.UIntPtr.Add%28System.UIntPtr%2CSystem.Int32%29?displayProperty=nameWithType>. ]]></format>
         </remarks>
@@ -1409,7 +1409,7 @@ For <xref:System.Numerics.IFloatingPoint%601>, this method matches the IEEE 754:
 
  The subtraction operation does not throw an exception if the result is too small to represent as an unsigned integer in the executing process. Instead, it is performed in an unchecked context.
 
- In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The subtraction operator is directly recognized by the language and will follow the normal language behavior for subtraction operations, including overflowing in a `checked` context if the result is to small to represent.
+ In C# starting from version 11 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The subtraction operator is directly recognized by the language and will follow the normal language behavior for subtraction operations, including overflowing in a `checked` context if the result is too small to represent.
 
  ]]></format>
         </remarks>


xml/System/IntPtr.xml
```diff
@@ -244,23 +244,25 @@
...
 
  The <xref:System.IntPtr> type can be used by languages that support pointers and as a common means of referring to data between languages that do and do not support pointers.
 
  <xref:System.IntPtr> objects can also be used to hold handles. For example, instances of <xref:System.IntPtr> are used extensively in the <xref:System.IO.FileStream?displayProperty=nameWithType> class to hold file handles.
 
- The <xref:System.IntPtr> type is CLS-compliant, while the <xref:System.UIntPtr> type is not. Only the <xref:System.IntPtr> type is used in the common language runtime. The <xref:System.UIntPtr> type is provided mostly to maintain architectural symmetry with the <xref:System.IntPtr> type.
+ However, <xref:System.IntPtr> is not itself a pointer or a handle. It is simply an integer type that can be used as an interchange format for pointers and handles due to being the same size. Outside of specific interchange requirements, such as for passing data to a language that doesn't support pointers, a correctly typed pointer should be used to represent pointers and <xref:System.Runtime.InteropServices.SafeHandle> should be used to represent handles.
```
tannergooding, line 259: This is likely `important`. Is there any way we can raise visibility on it and point users towards using `SafeHandle` for handles and the corresponding docs for it?

Likewise, should we elaborate at all why using `IntPtr` as a pointer type is error prone and `unsafe`? For example, converting some `int*` to `IntPtr` loses the context that it is a single indirection and the backing data is `int`. A user getting an `IntPtr` could therefore accidentally pass in an `int**` or a `byte*` and the data may end up incorrectly handled.

xml/System/IntPtr.xml
```diff
@@ -244,23 +244,25 @@
...
 
- This type implements the <xref:System.Runtime.Serialization.ISerializable>, and, in .NET 5 and later versions, the <xref:System.IFormattable> interfaces.
+ This type implements the <xref:System.Runtime.Serialization.ISerializable>. In .NET 5 and later versions, this type also implements the <xref:System.IFormattable> interfaces. In .NET 7 and later versions, this type also implements the <xref:System.Numerics.IBinaryInteger>, <xref:System.Numerics.IMinMaxValue>, and <xref:System.Numerics.ISignedNumber> interfaces.
 
  In C# starting from version 9.0, you can use the built-in `nint` type to define native-sized integers. This type is represented by the <xref:System.IntPtr> type internally and provides operations and conversions that are appropriate for integer types. For more information, see [nint and nuint types](/dotnet/csharp/language-reference/builtin-types/nint-nuint).
+ 
+ In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, <xref:System.IntPtr> and `nint` are fully interchangeable with eachother in the same way that <xref:System.Int32> and `int` are interchangeable.
```
tannergooding, line 265: I'm not sure if there is a better way to describe this change in relationship.

xml/System/IntPtr.xml
```diff
@@ -244,23 +244,25 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>A platform-specific type that is used to represent a pointer or a handle.</summary>
+    <summary>Represents a signed integer where the bit-width is the same as a pointer.</summary>
```
tannergooding, line 247: The format used by the other integer types is:
> Represents a 64-bit signed integer.

This would then correspond, technically speaking, to:
> Represents a ptr-bit signed integer.

However, I don't think that is necessarily as clear. I've found people sometimes understand it instantly and other times do not, when it's described that way.

xml/System/IntPtr.xml
```diff
@@ -537,15 +539,15 @@
...
-        <summary>Adds an offset to the value of a pointer.</summary>
-        <returns>A new pointer that reflects the addition of <paramref name="offset" /> to <paramref name="pointer" />.</returns>
+        <summary>Adds an offset to a signed integer.</summary>
+        <returns>A new signed integer that reflects the addition of <paramref name="offset" /> to <paramref name="pointer" />.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- The <xref:System.IntPtr.Add%2A> method does not throw an exception if the result is too large to represent as a pointer on the specified platform. Instead, the addition operation is performed in an unchecked context.
+ The <xref:System.IntPtr.Add%2A> method does not throw an exception if the result is too large to represent as a signed integer in the executing process. Instead, the addition operation is performed in an unchecked context.
```
tannergooding, line 550: It might be worth a comment calling out that `Add` and `Subtract` aren't recommended for use from C# or F#. Rather devs should prefer using the direct operators (`+` and `-`) instead and to make use of `checked` and `unchecked` contexts depending on their needs.

xml/System/IntPtr.xml
```diff
@@ -951,7 +953,9 @@ For <xref:System.Numerics.IFloatingPoint%601>, this method matches the IEEE 754:
 
  Languages that do not support custom operators can call the <xref:System.IntPtr.Add%2A> method instead.
 
- The addition operation does not throw an exception if the result is too large to represent as a pointer on the specified platform. Instead, it is performed in an unchecked context.
+ The addition operation does not throw an exception if the result is too large to represent as a signed integer in the executing process. Instead, it is performed in an unchecked context.
+
+ In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The addition operator is directly recognized by the language and will follow the normal language behavior for addition operations, including overflowing in a `checked` context if the result is to large to represent.
```
tannergooding, line 958: Not sure if its worth raising visibility on this distinction. For `operator +` this is a breaking change and contradictory to the immediate preceding statement. It will impact users who previously have code such as `IntPtr x = checked(value + int.MaxValue)` when they recompile.

xml/System/IntPtr.xml
```diff
@@ -1017,8 +1021,8 @@ For <xref:System.Numerics.IFloatingPoint%601>, this method matches the IEEE 754:
         <Parameter Name="value2" Type="System.IntPtr" />
       </Parameters>
       <Docs>
-        <param name="value1">The first pointer or handle to compare.</param>
-        <param name="value2">The second pointer or handle to compare.</param>
+        <param name="value1">The first signed integer to compare.</param>
```
tannergooding, line 1024: APIs in the other integer types use `64-bit signed integer`, as an example. 

It's not as easy to describe this the same way without using `ptr-bit signed integer` so I just left it as `signed integer`

xml/System/IntPtr.xml
```diff
@@ -244,23 +244,25 @@
...
   <Docs>
-    <summary>A platform-specific type that is used to represent a pointer or a handle.</summary>
+    <summary>Represents a signed integer where the bit-width is the same as a pointer.</summary>
     <remarks>
       <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
- The <xref:System.IntPtr> type is designed to be an integer whose size is platform-specific. That is, an instance of this type is expected to be 32-bits on 32-bit hardware and operating systems, and 64-bits on 64-bit hardware and operating systems.
+ The <xref:System.IntPtr> type is designed to be an integer whose size is the same as a pointer. That is, an instance of this type is expected to be 32-bits in a 32-bit process and 64-bits in a 64-bit process.
```
tannergooding, line 253: The previous description here was misleading. The underlying hardware/operating system had little overall to do with the size of `IntPtr`/`UIntPtr`.

Instead, it matches the bitness of the executing process. A 32-bit process even on 64-bit hardware/operating system would still have a 32-bit `IntPtr`/`UIntPtr`.

xml/System/IntPtr.xml
```diff
@@ -2889,7 +2895,7 @@ This member is an explicit interface member implementation. It can be used only
         <ReturnType>System.IntPtr</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>A read-only field that represents a pointer or handle that has been initialized to zero.</summary>
+        <summary>A read-only field that represents a signed integer that has been initialized to zero.</summary>
```
tannergooding, line 2898: Similarly to `Add`. Users should likely prefer `(nint)0` or `(nuint)0` instead. It is smaller and more efficient IL and is more closely similar to how you represent `Zero` for other integer types.

xml/System/UIntPtr.xml
```diff
@@ -248,23 +248,24 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>A platform-specific type that is used to represent a pointer or a handle.</summary>
+    <summary>Represents an unsigned integer where the bit-width is the same as a pointer.</summary>
```
tannergooding, line 251: `UIntPtr` has the same annotations I gave for `IntPtr`

xml/System/IntPtr.xml
gfoidl: It's clear, especially with the example of `int` given.

xml/System/IntPtr.xml
MSDN-WhiteKnight: You could use syntax like this:

    > [!NOTE]
    > Using IntPtr as a pointer type is error prone and unsafe.

There are more exceptions to this then only passing to a language that doesn't support pointers, though. For example Marshal.AllocHGlobal/FreeHGlobal also use IntPtr to represent pointers.

xml/System/IntPtr.xml
```diff
@@ -1853,7 +1859,7 @@ For <xref:System.Numerics.IFloatingPoint%601>, this method matches the IEEE 754:
       </ReturnValue>
       <Docs>
         <summary>Gets the size of this instance.</summary>
-        <value>The size of a pointer or handle in this process, measured in bytes. The value of this property is 4 in a 32-bit process, and 8 in a 64-bit process. You can define the process type by setting the <see langword="/platform" /> switch when you compile your code with the C# and Visual Basic compilers.</value>
+        <value>The size of a signed integer in the executing process, measured in bytes. The value of this property is 4 in a 32-bit process, and 8 in a 64-bit process. You can define the process type by setting the <see langword="/platform" /> switch when you compile your code with the C# and Visual Basic compilers.</value>
```
gfoidl, line 1862:
```suggestion
        <value>The size of a signed integer in the executing process, measured in bytes. The value of this property is 4 in a 32-bit process, and 8 in a 64-bit process. You can define the process type by setting the <see langword="/platform" /> switch when you compile your application with the C# and Visual Basic compilers.</value>
```
? To make it clearer that class-libraries aren't affected by this.

xml/System/IntPtr.xml
MSDN-WhiteKnight: It's fine, but you could also say "nint is an alias for System.IntPtr", because this seems to be common term for decribing this: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types

>In all of the table rows except the last two, each C# type keyword from the leftmost column is an alias for the corresponding .NET type

xml/System/IntPtr.xml
MSDN-WhiteKnight, line 253:
```suggestion
 The <xref:System.IntPtr> type is designed to be an integer whose size is the same as a pointer. That is, an instance of this type is expected to be 32 bits in a 32-bit process and 64 bits in a 64-bit process.
```

xml/System/IntPtr.xml
```diff
@@ -244,23 +244,25 @@
...
+ However, <xref:System.IntPtr> is not itself a pointer or a handle. It is simply an integer type that can be used as an interchange format for pointers and handles due to being the same size. Outside of specific interchange requirements, such as for passing data to a language that doesn't support pointers, a correctly typed pointer should be used to represent pointers and <xref:System.Runtime.InteropServices.SafeHandle> should be used to represent handles.
 
- This type implements the <xref:System.Runtime.Serialization.ISerializable>, and, in .NET 5 and later versions, the <xref:System.IFormattable> interfaces.
+ This type implements the <xref:System.Runtime.Serialization.ISerializable>. In .NET 5 and later versions, this type also implements the <xref:System.IFormattable> interfaces. In .NET 7 and later versions, this type also implements the <xref:System.Numerics.IBinaryInteger>, <xref:System.Numerics.IMinMaxValue>, and <xref:System.Numerics.ISignedNumber> interfaces.
```
MSDN-WhiteKnight, line 261: Is there a non-generic IBinaryInteger, or should it be a xref to generic IBinaryInteger: `<xref:System.Numerics.IBinaryInteger%601>`?

xml/System/IntPtr.xml
tannergooding: Class libraries are effected by this. You can compile libraries to be `x86` or `x64` specific and to not support `AnyCPU`.

xml/System/IntPtr.xml
tannergooding: Only generic, I'll update to include the escaped accent.

tannergooding: Fixed.

xml/System/IntPtr.xml
MSDN-WhiteKnight: My understanding is that it's hyphenated in adjectives (64-bit process), but non-hyphenated in number+noun (size is 64 bits). The style guide says "hyphenate if the measurement modifies a noun" (https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/term-collections/bits-bytes-terms), but maybe i'm not interpreting it correctly.

tannergooding:
Any other changes needed here or can this be merged?

xml/System/IntPtr.xml
```diff
@@ -244,23 +244,26 @@
...
 
- This type implements the <xref:System.Runtime.Serialization.ISerializable>, and, in .NET 5 and later versions, the <xref:System.IFormattable> interfaces.
+ This type implements the <xref:System.Runtime.Serialization.ISerializable>. In .NET 5 and later versions, this type also implements the <xref:System.IFormattable> interfaces. In .NET 7 and later versions, this type also implements the <xref:System.Numerics.IBinaryInteger%601>, <xref:System.Numerics.IMinMaxValue%601>, and <xref:System.Numerics.ISignedNumber%601> interfaces.
 
  In C# starting from version 9.0, you can use the built-in `nint` type to define native-sized integers. This type is represented by the <xref:System.IntPtr> type internally and provides operations and conversions that are appropriate for integer types. For more information, see [nint and nuint types](/dotnet/csharp/language-reference/builtin-types/nint-nuint).
+ 
+ In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, `nint` is an alias for <xref:System.IntPtr> in the same way that `int` is an alias for <xref:System.Int32>.
```
gewarren, line 266:
```suggestion
 In C# starting from version 11 and when targeting the .NET 7 or later runtime, `nint` is an alias for <xref:System.IntPtr> in the same way that `int` is an alias for <xref:System.Int32>.
```

xml/System/IntPtr.xml
```diff
@@ -951,7 +954,9 @@ For <xref:System.Numerics.IFloatingPoint%601>, this method matches the IEEE 754:
 
  Languages that do not support custom operators can call the <xref:System.IntPtr.Add%2A> method instead.
 
- The addition operation does not throw an exception if the result is too large to represent as a pointer on the specified platform. Instead, it is performed in an unchecked context.
+ The addition operation does not throw an exception if the result is too large to represent as a signed integer in the executing process. Instead, it is performed in an unchecked context.
+
+ In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The addition operator is directly recognized by the language and will follow the normal language behavior for addition operations, including overflowing in a `checked` context if the result is to large to represent.
```
gewarren, line 959:
```suggestion
 In C# starting from version 11 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The addition operator is directly recognized by the language and will follow the normal language behavior for addition operations, including overflowing in a `checked` context if the result is too large to represent.
```

xml/System/IntPtr.xml
```diff
@@ -1542,7 +1547,9 @@ For <xref:System.Numerics.IFloatingPoint%601>, this method matches the IEEE 754:
 
  Languages that do not support custom operators can call the <xref:System.IntPtr.Subtract%2A> method instead.
 
- The subtraction operation does not throw an exception if the result is too small to represent as a pointer on the specified platform. Instead, it is performed in an unchecked context.
+ The subtraction operation does not throw an exception if the result is too small to represent as a signed integer in the executing process. Instead, it is performed in an unchecked context.
+
+ In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The subtraction operator is directly recognized by the language and will follow the normal language behavior for subtraction operations, including overflowing in a `checked` context if the result is to small to represent.
```
gewarren, line 1552:
```suggestion
 In C# starting from version 11 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The subtraction operator is directly recognized by the language and will follow the normal language behavior for subtraction operations, including overflowing in a `checked` context if the result is too small to represent.
```

xml/System/UIntPtr.xml
```diff
@@ -248,23 +248,25 @@
...
+
+ This type implements the <xref:System.Runtime.Serialization.ISerializable>. In .NET 5 and later versions, this type also implements the <xref:System.IFormattable> interfaces. In .NET 7 and later versions, this type also implements the <xref:System.Numerics.IBinaryInteger%601>, <xref:System.Numerics.IMinMaxValue%601>, and <xref:System.Numerics.IUnsignedNumber%601> interfaces.
 
  In C# starting from version 9.0, you can use the built-in `nuint` type to define native-sized integers. This type is represented by the <xref:System.UIntPtr> type internally and provides operations and conversions that are appropriate for integer types. For more information, see [nint and nuint types](/dotnet/csharp/language-reference/builtin-types/nint-nuint).
 
+ In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, `nuint` is an alias for <xref:System.UIntPtr> in the same way that `uint` is an alias for <xref:System.UInt32>.
```
gewarren, line 268:
```suggestion
 In C# starting from version 11 and when targeting the .NET 7 or later runtime, `nuint` is an alias for <xref:System.UIntPtr> in the same way that `uint` is an alias for <xref:System.UInt32>.
```

xml/System/UIntPtr.xml
```diff
@@ -889,7 +891,9 @@ For <xref:System.Numerics.IFloatingPoint%601>, this method matches the IEEE 754:
 
  Languages that do not support custom operators can call the <xref:System.UIntPtr.Add%2A> method instead.
 
- The addition operation does not throw an exception if the result is too large to represent as a pointer on the specified platform. Instead, it is performed in an unchecked context.
+ The addition operation does not throw an exception if the result is too large to represent as an unsigned integer in the executing process. Instead, it is performed in an unchecked context.
+ 
+ In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The addition operator is directly recognized by the language and will follow the normal language behavior for addition operations, including overflowing in a `checked` context if the result is to large to represent.
```
gewarren, line 896:
```suggestion
 In C# starting from version 11 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The addition operator is directly recognized by the language and will follow the normal language behavior for addition operations, including overflowing in a `checked` context if the result is too large to represent.
```

xml/System/UIntPtr.xml
```diff
@@ -1403,7 +1407,9 @@ For <xref:System.Numerics.IFloatingPoint%601>, this method matches the IEEE 754:
 
  Languages that do not support custom operators can call the <xref:System.UIntPtr.Subtract%2A> method instead.
 
- The subtraction operation does not throw an exception if the result is too small to represent as a pointer on the specified platform. Instead, it is performed in an unchecked context.
+ The subtraction operation does not throw an exception if the result is too small to represent as an unsigned integer in the executing process. Instead, it is performed in an unchecked context.
+
+ In C# starting from version 11.0 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The subtraction operator is directly recognized by the language and will follow the normal language behavior for subtraction operations, including overflowing in a `checked` context if the result is to small to represent.
```
gewarren, line 1412:
```suggestion
 In C# starting from version 11 and when targeting the .NET 7 or later runtime, this API is only accessible via reflection. The subtraction operator is directly recognized by the language and will follow the normal language behavior for subtraction operations, including overflowing in a `checked` context if the result is too small to represent.
```

gewarren (approved):
Approved with some minor suggestions.

tannergooding:
@gewarren, am I good to merge this then or should that be done by someone from the docs team?

gewarren:
@tannergooding Feel free to merge it yourself once you get an approved review (which you have in this case).

tannergooding:
Thanks for the confirmation! I'll keep that in mind for the future :)

PR #8141, opened 2022-06-07
Microsoft.Extensions.Hosting.[Systemd|WindowsServices] are in .NET 5, 6 an 7 as well

CodeCasterNL:
## Summary
The docs for these Platform Extensions:
* Microsoft.Extensions.Hosting.Systemd 
* Microsoft.Extensions.WindowsServices

Were never updated to mention .NET 5, 6 and 7. See the warning banner on this URL: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.windowsservicelifetimehostbuilderextensions.usewindowsservice?view=dotnet-plat-ext-6.0:

> The requested page is not available for .NET Platform Extensions 6. You have been redirected to the newest product version this page is available for.

Their source was moved from the extensions repo to the runtime repo in https://github.com/dotnet/runtime/pull/47686/files. I've checked that the mentioned types and members were present in the dotnet/runtime's main branch and when present (all of them) I assumed they've been there since the merge (so none were, say, removed in 5 and re-added in 6). 

Two questions though: 

1. Why wasn't this change made automatically? Some XML element missing, some directory structure incorrect?
2. Following up from the latter: in which directory should the file `WindowsServiceLifetimeHostBuilderExtensions.xml` be, describing a type in the assembly `Microsoft.Extensions.Hosting.WindowsServices`:
   * `xml/Microsoft.Extensions.Hosting`, where it's now, according to its namespace, or
   * `xml/Microsoft.Extensions.Hosting.WindowsServices`, according to its assembly name and source location?

### DIFF
--- a/xml/Microsoft.Extensions.Hosting.Systemd/ISystemdNotifier.xml
+++ b/xml/Microsoft.Extensions.Hosting.Systemd/ISystemdNotifier.xml
@@ -9,6 +9,9 @@
     <AssemblyName>Microsoft.Extensions.Hosting.Systemd</AssemblyName>
     <AssemblyVersion>3.0.0.0</AssemblyVersion>
     <AssemblyVersion>3.1.0.0</AssemblyVersion>
+    <AssemblyVersion>5.0.0.0</AssemblyVersion>
+    <AssemblyVersion>6.0.0.0</AssemblyVersion>
+    <AssemblyVersion>7.0.0.0</AssemblyVersion>
   </AssemblyInfo>
   <Interfaces />
   <Docs>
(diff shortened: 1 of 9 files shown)


gewarren:
It looks like the packages are missing from [this file](https://github.com/dotnet/api-catalog-infra/blob/main/eng/droplayout.props), which is why the documentation hasn't been updated. @carlossanlop Can you look into this and generate a new drop if necessary (or just wait until the Preview 5 drop is generated)?

We can't accept this PR though, since the XML you updated is autogenerated by the reflection process.

CodeCasterNL:
> We can't accept this PR though, since the XML you updated is autogenerated by the reflection process.

No problem, I get that. Should I close it and move my comments to an issue?

gewarren:
> > We can't accept this PR though, since the XML you updated is autogenerated by the reflection process.
> 
> No problem, I get that. Should I close it and move my comments to an issue?

Yes, that would be great - thanks!

PR #8148, opened 2022-06-09
Fix summary for ImmutableList<T>.Empty

### LINKED ISSUE #8129: Documentation for ImmutableList<T>.Empty is incorrect
The [summary for ImmutableList&lt;T&gt;.Empty](https://github.com/dotnet/dotnet-api-docs/blob/792c515f3b6be9c7b44807644baccb8d9548a6d2/xml/System.Collections.Immutable/ImmutableList%601.xml#L565) says, "Gets an empty set with the default sort comparer."

To match [ImmutableArray&lt;T&gt;.Empty](https://github.com/dotnet/dotnet-api-docs/blob/792c515f3b6be9c7b44807644baccb8d9548a6d2/xml/System.Collections.Immutable/ImmutableArray%601.xml#L744), I think it should say, "Gets an empty immutable list."

mjolka:
## Summary

Fixes the summary for ImmutableList&lt;T&gt;.Empty

Fixes #8129

### DIFF
--- a/xml/System.Collections.Immutable/ImmutableList`1.xml
+++ b/xml/System.Collections.Immutable/ImmutableList`1.xml
@@ -562,7 +562,7 @@
         <ReturnType>System.Collections.Immutable.ImmutableList&lt;T&gt;</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets an empty set with the default sort comparer.</summary>
+        <summary>Gets an empty immutable list.</summary>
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks for spotting this and making the fix @mjolka 

I'll :shipit: now.

You should see the changes on the live site in the next day or two, on our regular publishing cycle.

Thanks again for contributing to docs! 🎆

PR #8149, opened 2022-06-09
Fix Control.Top Property definition

KennyTK:
Top property definition previously referenced bottom of control

## Summary

The Control.Top Property definition was inconsistent. I believe the correct variation is "top edge of the control" rather than bottom.

TopBottomEdge


RussKie (commented):
Thank you

PR #8155, opened 2022-06-11
Removed poor coding style in schema validation

softwarepronto:
getSchema/LoadDocumentWithSchemaValidation
* Returned null on error (versus throwing an exception) -- now just falls back on exceptions
* Used a catch in the success path -- removed try/catch 
* Removed .NET 1.0 code (below UTF8.GetBytes/MemoryStream) and implemented using simpler, .NET 1.1 StringReader
byte[] byteArray = Encoding.UTF8.GetBytes(xml);
            MemoryStream stream = new MemoryStream(byteArray);

## Summary

Describe your changes here.

### DIFF
--- a/snippets/csharp/System.Xml/XmlDocument/Overview/xmlhelpermethods.cs
+++ b/snippets/csharp/System.Xml/XmlDocument/Overview/xmlhelpermethods.cs
@@ -57,7 +57,6 @@ namespace XMLProcessingApp
         #region Validate XML against a Schema
 
         //<Snippet2>
-            
         //************************************************************************************
         //
         //  Helper method that generates an XML string.


snippets/csharp/System.Xml/XmlDocument/Overview/xmlhelpermethods.cs
```diff
@@ -98,35 +98,24 @@ public XmlDocument LoadDocumentWithSchemaValidation(bool generateXML, bool gener
             // Helper method to retrieve schema.
             XmlSchema schema = getSchema(generateSchema);
 
-            if (schema == null)
```
MSDN-WhiteKnight, line 101: Is it OK to remove this validation here? Adding into schema set below would crash on null schema.

snippets/csharp/System.Xml/XmlDocument/Overview/xmlhelpermethods.cs
```diff
@@ -98,35 +98,24 @@ public XmlDocument LoadDocumentWithSchemaValidation(bool generateXML, bool gener
...
-            try
+            if (generateXML)
             {
-                reader = XmlReader.Create("booksData.xml", settings);
+                string xml = generateXMLString();
+                StringReader stringReader = new StringReader(xml);
+
+                reader = XmlReader.Create(stringReader, settings);
             }
-            catch (System.IO.FileNotFoundException)
```
MSDN-WhiteKnight, line 117: Is it okay to remove handling of FileNotFoundException? XmlReader.Create could throw it if file does not exist. The guidelines  say that samples should catch common exceptions that could be thrown by the method: https://docs.microsoft.com/en-us/contribute/dotnet/dotnet-contribute#write-a-new-sample

>Your sample should include appropriate exception handling. It should handle all exceptions that are likely to be thrown in the context of the sample.

softwarepronto: I am open to criticism but I'm not seeing a clean exception handling example coming out of this code.

The original code did not handle the exception. The original code returned null on that one error (FileNotFoundException) and all other errors such as malformed XML threw an exception. On a file access issue threw an exception. It was inconsistent error handling. Returning value errors is 1985 C programming style. The original is an example of incorrect error handling.

generateSchema and generateXML are required parameters but are only used when a FileNotFoundException is thrown. There names appear to me "generate" but in reality they should be generateIfTheHardCodeFileIsNotFound. Not a clear method signature.

A developer who wanted to handle file not found would use File.Exists so as not to generate an FileNotFoundException.

I don't see a clear set of common exceptions. Four of the five documented Exceptions for XmlReader.Create are possible. There are a half dozen other methods that could also generate errors.

I could put a try/catch and a Console.Error.Writeline but I don't see it as a realistic example of error.

Once this change is approved I can do another one and get a "using" to take advantage of XmlReader's IDisposable. That is a lessen that should be taught.

MSDN-WhiteKnight: >I am open to criticism but I'm not seeing a clean exception handling example coming out of this code.

The code currently tries to open xml file for reading, and if it throws exception, instead fallbacks to some hardcoded xml. I'm not sure is it clean or not, but you PR seems to change the meaning of the code, to make it throw where it previously wasn't throwing. It's not just a cleanup. So i wonder whether it's OK to do it in this case.

>On a file access issue threw an exception. It was inconsistent error handling.

You are right. In fact, the guidelines i cited aren't followed consistently, because it's not always easy to determine "likely" exceptions. Maybe this sample needs some improvement, like handle IOException instead to account for more error conditions. I don't see it as an argument to completely remove the exception handling, though.

>Returning value errors is 1985 C programming style. The original is an example of incorrect error handling.

I have no idea what do you mean by this. Such pattern is used all the time in .NET. The production code might need to log the exception additionally when catching it, but for an example, this error handling seems fine.

softwarepronto: @MSDN-WhiteKnight how about this?

if generateXML==true and the XML file is missing, catch the FileNotFoundException and fall back on the hard-coded XML.

if generateSchema ==true and the schema file is missing, catch the FileNotFoundException and fall back on the hard-coded schema.

I think it has a better workflow than the original.

softwarepronto: I can add the using to the XmlReader. I am not sure how Microsoft wants their code written. Do we go old school?

using (Xmlreader reader = XmlReader.Create(blah blah))
{
}

Or new school?
using Xmlreader reader = XmlReader.Create(blah blah);

MSDN-WhiteKnight: Fine to use new language features. I'm not Microsoft, but using declaration syntax is already used, for example, here: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.metadata.metadatareader?view=net-6.0#examples

snippets/csharp/System.Xml/XmlDocument/Overview/xmlhelpermethods.cs
```diff
@@ -91,45 +91,41 @@ private string generateXMLString()
...
+                try
                 {
-                    string xml = generateXMLString();
-                    byte[] byteArray = Encoding.UTF8.GetBytes(xml);
-                    MemoryStream stream = new MemoryStream(byteArray);
-                    reader = XmlReader.Create(stream, settings);
+                    reader = XmlReader.Create("booksData.xml", settings);
                 }
-                else
+
```
MSDN-WhiteKnight, line 110:
```suggestion
```

snippets/csharp/System.Xml/XmlDocument/Overview/xmlhelpermethods.cs
```diff
@@ -178,26 +174,32 @@ private string generateXMLSchema()
...
                 {
-                    string xmlSchemaString = generateXMLSchema();
-                    byte[] byteArray = Encoding.UTF8.GetBytes(xmlSchemaString);
-                    MemoryStream stream = new MemoryStream(byteArray);
-                    XmlReader reader = XmlReader.Create(stream);
-                    schema = xs.Add("http://www.contoso.com/books", reader);
+                    schema = xs.Add("http://www.contoso.com/books", "booksData.xsd");
                 }
-                else
+
```
MSDN-WhiteKnight, line 185:
```suggestion
```

snippets/csharp/System.Xml/XmlDocument/Overview/xmlhelpermethods.cs
```diff
@@ -57,7 +57,7 @@ public void SaveXML(XmlDocument doc)
         #region Validate XML against a Schema
 
         //<Snippet2>
-
+
```
MSDN-WhiteKnight, line 60:
```suggestion

```

MSDN-WhiteKnight (commented):
Updates look good :+1: . There are some extra newlines that need to be deleted (catch statement should not have preceding empty line).

snippets/csharp/System.Xml/XmlDocument/Overview/xmlhelpermethods.cs
krwq: Yes but please add test cases showing old behavior being broken and then fixing the testcase. This code is very old and touching it is always a risk of breaking someone

softwarepronto: The original code did not have broken behavior. It was poorly written and a reader won't know that it is 22 year old. A reader is always looking for the best practices way to write C#. The origin code is poorly written because ignores the generateXML parameter if the XML file was found on disk and it ignores the generateSchema parameter if the schema file was found on disk. Ignoring the parameters passed to a method on success is not proper behavior for well written code. @MSDN-WhiteKnight  and came up with an example where the XML is only read from file if generateXML=false. The schema is only read from a file if generateSchema=false. The parameters are no longer ignored. generateSchema is set to true to work around the XML file not found error by generating the XML programmatically. generateSchema is set to true to work around the schema file not found error by generating the schema programmatically. 

The other changes already committed for this page were to remove the Windows Forms specific code and to add a method to the snippet that cause the code not to compile when pasted into a file.

The method signatures in the change remained the same. The new code just more appropriately uses the exception to handle an error and the new code uses all the parameters passed to the methods (they are not ignored).

The code may be old but we can still clean it up.

snippets/csharp/System.Xml/XmlDocument/Overview/xmlhelpermethods.cs
```diff
@@ -91,45 +90,40 @@ private string generateXMLString()
...
-                    MemoryStream stream = new MemoryStream(byteArray);
-                    reader = XmlReader.Create(stream, settings);
+                    reader = XmlReader.Create("booksData.xml", settings);
                 }
-                else
+                catch (FileNotFoundException ex)
                 {
-                    return null;
+                    Console.WriteLine(
+                        $"XML file not found so generating: {ex.Message}");
```
krwq, line 112: can you check `File.Exists` and let the exception happen instead?

softwarepronto: I am not sure what you are asking. If I check File.Exists then I can avoid the FileNotFoundException all together. I checked File.Exists in my first attempt to clean up and avoided FileNotFoundException . I was told by a reviewer that the example is to show the XmlReader.Create throws a documented exception of FIleNotFoundException. I should handle the documented exception gracefully.  The key to the changes 1) Removing .NET 1.0 code for handling Memory Stream (moved to .NET 1.1 which is cleaner) and 2) the generateXML parameter is always used in the method. In the original example the generateXML parameter was ignored depending on if errors manifested themselves. Setting generateXML=true was simply ignored in the original code at times.

krwq (approved):

krwq:
@softwarepronto this will get auto-merged once all conversations are closed. Please validate if you've fixed and resolve the conversation. I'm ok with as is version.

MSDN-WhiteKnight:
Looks like there are still unresolved review conversations from me that block auto-merge. I don't have Resolve conversation button available, so either the PR author or maintainers need to press it.

PR #8161, opened 2022-06-14
CreateSubdirectory(path) does not throw a IOException in case path already exists

FireEmerald:
## Summary

`System.IO.DirectoryInfo CreateSubdirectory (string path)` does not throw a IOException in case `path` already exists.

### DIFF
--- a/xml/System.IO/DirectoryInfo.xml
+++ b/xml/System.IO/DirectoryInfo.xml
@@ -432,10 +432,10 @@ namespace ConsoleApp
         <exception cref="T:System.ArgumentNullException">
           <paramref name="path" /> is <see langword="null" />.</exception>
         <exception cref="T:System.IO.DirectoryNotFoundException">The specified path is invalid, such as being on an unmapped drive.</exception>
-        <exception cref="T:System.IO.IOException">The subdirectory cannot be created.
-
+        <exception cref="T:System.IO.IOException">The subdirectory cannot be created.  
+  
  -or-  
-
+  
  A file already has the name specified by <paramref name="path" />.</exception>
         <exception cref="T:System.IO.PathTooLongException">The specified path, file name, or both exceed the system-defined maximum length. </exception>
         <exception cref="T:System.Security.SecurityException">The caller does not have code access permission to create the directory.  


xml/System.IO/DirectoryInfo.xml
```diff
@@ -432,11 +432,7 @@ namespace ConsoleApp
         <exception cref="T:System.ArgumentNullException">
           <paramref name="path" /> is <see langword="null" />.</exception>
         <exception cref="T:System.IO.DirectoryNotFoundException">The specified path is invalid, such as being on an unmapped drive.</exception>
-        <exception cref="T:System.IO.IOException">The subdirectory cannot be created.  
-  
- -or-  
-  
- A file or directory already has the name specified by <paramref name="path" />.</exception>
```
BillWagner, line 439: @jeffhandley 

Is this change correct on all platforms?

If so, I'll :shipit:

FireEmerald: I tested it on windows, when you pass a path to a _existing file_ to `CreateSubdirectory(string path)` it does in fact throw a IOException:
> System.IO.IOException: 'Cannot create 'C:\test\parent\child' because a file or directory with the same name already exists.'

But if the path points to a existing _directory_ the method does not throw a `IOException`.

FireEmerald: Tested it now also on Linux, same behavior: **Doesn't** throw if path is a existing directory, throws if a _existing file_:
> Unhandled exception. System.IO.IOException: The file '/home/testuser/play/parent/child' already exists.
>   at System.IO.FileSystem.CreateDirectory(String fullPath)
>   at System.IO.DirectoryInfo.CreateSubdirectory(String path)

FireEmerald: Here is also already written that the method does not throw a exception if the subdirectory already exists:
https://github.com/dotnet/dotnet-api-docs/blob/d532baa9c9f8a759696337a1f6fd61e9723c47a3/xml/System.IO/DirectoryInfo.xml#L415

BillWagner (approved):
Thanks for doing the extra research @FireEmerald 

I'll :shipit: now.

PR #8182, opened 2022-06-29
Fix example of Unsafe.As<T>(Object)

Flash0ver:
## Summary

The example of `Unsafe.As<T>(Object)` did not compile (Compiler Error `CS0165` - Use of unassigned local variable 's').
Changed the example to use the parameter correctly like the example above.


BillWagner (approved):
Thanks for fixing this @Flash0ver 

I've reviewed this, and I'll :shipit: now.

PR #8197, opened 2022-07-04
System.IO.DirectoryInfo F# snippets

albert-du:
Adds F# snippets for DirectoryInfo


BillWagner (approved):
Thanks again @albert-du 

This looks great. I'll :shipit: now.

PR #8201, opened 2022-07-07
Fix potential IndexOutOfRangeException

guneyizol:
## Summary

The expression ```Count > array.Length - arrayIndex + 1``` causes a System.IndexOutOfRangeException to be thrown when BoxCollection.CopyTo method is called for copying the elements to the end of the given array with BoxCollection.Count is 1 more than the available space at the end of the given array.

For example, when BoxCollection.Count equals 4 and the given array's length is 10 and the starting index is 7, last if statement in BoxCollection.CopyTo evaluates to false, although it should evaluate to true.

### DIFF
--- a/snippets/csharp/System.Collections.Generic/ICollectionT/Overview/program.cs
+++ b/snippets/csharp/System.Collections.Generic/ICollectionT/Overview/program.cs
@@ -198,7 +198,7 @@ public class BoxCollection : ICollection<Box>
            throw new ArgumentNullException("The array cannot be null.");
         if (arrayIndex < 0)
            throw new ArgumentOutOfRangeException("The starting array index cannot be negative.");
-        if (Count > array.Length - arrayIndex + 1)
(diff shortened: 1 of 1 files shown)


snippets/csharp/System.Collections.Generic/ICollectionT/Overview/program.cs
```diff
@@ -198,7 +198,7 @@ public void CopyTo(Box[] array, int arrayIndex)
            throw new ArgumentNullException("The array cannot be null.");
         if (arrayIndex < 0)
            throw new ArgumentOutOfRangeException("The starting array index cannot be negative.");
-        if (Count > array.Length - arrayIndex + 1)
+        if (Count > array.Length - arrayIndex)
```
krwq, line 201: please add a test case and make sure it fails before this change and passes after

krwq (approved):

PR #8207, opened 2022-07-11
EventLogReader.ReadEvent - documenting null return

### LINKED ISSUE #8206: EventLogReader.ReadEvent does not document null-returning behaviour or give usage example
(Creating an issue for a PR)

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.eventing.reader.eventlogreader.readevent

None of these behaviours are documented:

- `ReadEvent()` and `ReadEvent(TimeSpan)` will return `null`...
  -  on the first attempt to read past the end of the query results (if running a query)
  -  on the first attempt to read past the last entry in the log-file on disk (if reading from a log directly).
  -  subsequent attempts to read will throw an `InvalidOperationException` - so reading must be stopped at the first `null`.
- The `ReadEvent` method will return `EventLogRecord` objects instead of `EventRecord` - should this be documented too?
- The `EventRecord` type is `IDisposable` because it wraps an `EventLogHandle`, but the documentation doesn't say if the `EventRecord` or `EventLogRecord` should always be disposed inside the read-loop or not, and if it's safe to pass the `EventLogRecord` around or emit it in a `yield break`.
   - I assume it _should_ be eagerly-disposed and not exposed to callers, so the example in my PR copies scalar data from the `EventLogRecord` properties into a `ValueTuple` and emits those instead.

daiplusplus:
## Summary

The documentation for `EventLogReader.ReadEvent` does not indicate how to detect being at the last result (it returns `null`, but this isn't documented at all) - so this PR adds that detail.

Fixes #8206

### DIFF
--- a/xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
+++ b/xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
@@ -467,6 +467,16 @@
         <summary>Reads and returns the next event from the event query results - or null if the end of the results has been reached.</summary>
         <returns>Returns an <see cref="T:System.Diagnostics.Eventing.Reader.EventRecord" /> object - or a null reference.</returns>
         <remarks>
+          <para>
+            This overload is equivalent to calling <see cref="M:System.Diagnostics.Eventing.Reader.EventLogReader.ReadEvent(System.TimeSpan)" /> with an <see cref="M:System.Threading.Timeout.InfiniteTimeSpan" /> timeout parameter argument.
+          </para>
+          <para>
+            Each time this method is called the reader will advance to the next event record in its query results or the log file being read.
+            When the reader reaches the end of the query results or log file the method will return null on its next call, but subsequent calls will throw an InvalidOperationException.
+          </para>
+          <para>
+            The returned EventRecord objects implement IDisposable and should be disposed of appropriately. To avoid leaking resources consider copying desired data from the EventRecord object into a custom class or tuple and then immediately disposing of the returned EventRecord before advancing to the next logged event.
+          </para>
           <format type="text/markdown"><![CDATA[
 public static IEnumerable<(String logPath, DateTime dateTime, String level, String description)> GetEvents(String logPath)
 {
@@ -474,7 +484,7 @@ public static IEnumerable<(String logPath, DateTime dateTime, String level, Stri
 
     using (EventLogReader rdr = new EventLogReader(path: logPath))
     {
-        for (EventRecord? e = rdr.ReadEvent(); e != null; e = rdr.ReadEvent())
+        while (rdr.ReadEvent() is EventRecord e)
         {
             try
             {
@@ -495,6 +505,11 @@ public static IEnumerable<(String logPath, DateTime dateTime, String level, Stri
 }
  ]]></format>
         </remarks>
+          <!--<exception cref="T:System.Diagnostics.Eventing.Reader.EventLogException"></exception>-->
+          <exception cref="T:System.Diagnostics.Eventing.Reader.EventLogReadingException"></exception>
+          <exception cref="T:System.Diagnostics.Eventing.Reader.EventLogInvalidDataException"></exception>
+          <exception cref="T:System.InvalidOperationException">Thrown when ReadEvent is called again after a previous call returned null.</exception>
+          <exception cref="T:System.UnauthorizedAccessException"></exception>
         <related type="Article" href="https://docs.microsoft.com/previous-versions/bb671205(v=vs.90)">Technology Summary for Reading and Managing Event Logs</related>
         <related type="Article" href="https://docs.microsoft.com/previous-versions/bb671204(v=vs.90)">Event Log Scenarios</related>
       </Docs>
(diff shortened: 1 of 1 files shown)


xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
```diff
@@ -433,7 +433,7 @@
         <AssemblyVersion>4.0.0.0</AssemblyVersion>
       </AssemblyInfo>
       <Docs>
-        <summary>Reads the next event that is returned from the event query in this object.</summary>
+        <summary>Reads and returns the next event from the event query results - or null if the end of the results has been reached.</summary>
```
gewarren, line 436:
```suggestion
        <summary>Reads and returns the next event from the event query results, or null if the end of the results has been reached.</summary>
```

xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
```diff
@@ -464,16 +464,52 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>Reads the next event that is returned from the event query in this object.</summary>
-        <returns>Returns an <see cref="T:System.Diagnostics.Eventing.Reader.EventRecord" /> object.</returns>
+        <summary>Reads and returns the next event from the event query results - or null if the end of the results has been reached.</summary>
```
gewarren, line 467:
```suggestion
        <summary>Reads and returns the next event from the event query results.</summary>
```

xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
```diff
@@ -464,16 +464,52 @@
...
+        <returns>Returns an <see cref="T:System.Diagnostics.Eventing.Reader.EventRecord" /> object - or a null reference.</returns>
```
gewarren, line 468:
```suggestion
        <returns>Returns an <see cref="T:System.Diagnostics.Eventing.Reader.EventRecord" /> object, or a null reference if the end of the results has been reached.</returns>
```

xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
```diff
@@ -464,16 +464,52 @@
...
-          <format type="text/markdown"><![CDATA[  
-  
-## Examples  
- For example code using this class, see [How To: Query for Events](https://docs.microsoft.com/previous-versions/bb671200(v=vs.90)).  
-  
+          <para>
+            This overload is equivalent to calling <see cref="M:System.Diagnostics.Eventing.Reader.EventLogReader.ReadEvent(System.TimeSpan)" /> with an <see cref="M:System.Threading.Timeout.InfiniteTimeSpan" /> timeout parameter argument.
+          </para>
+          <para>
+            Each time this method is called the reader will advance to the next event record in its query results or the log file being read.
```
gewarren, line 474:
```suggestion
            Each time this method is called, the reader advances to the next event record in the query results or the log file.
```

xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
```diff
@@ -464,16 +464,52 @@
...
+            When the reader reaches the end of the query results or log file the method will return null on its next call, but subsequent calls will throw an InvalidOperationException.
```
gewarren, line 475:
```suggestion
            When the reader reaches the end of the query results or log file, the method returns <c>null</c> on its next call, and subsequent calls will throw an <see cref="T:System.InvalidOperationException" />.
```

xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
```diff
@@ -464,16 +464,52 @@
...
+          </para>
+          <para>
+            The returned EventRecord objects implement IDisposable and should be disposed of appropriately. To avoid leaking resources consider copying desired data from the EventRecord object into a custom class or tuple and then immediately disposing of the returned EventRecord before advancing to the next logged event.
```
gewarren, line 478:
```suggestion
            The returned EventRecord objects implement IDisposable and should be disposed of appropriately. To avoid leaking resources, consider copying desired data from the EventRecord object into a custom class or tuple and then immediately disposing of the returned EventRecord before advancing to the next logged event. The following code shows an example.
```

xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
```diff
@@ -464,16 +464,52 @@
...
+          </para>
+          <format type="text/markdown"><![CDATA[
```
gewarren, line 480: The example should be code-fenced.

xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
```diff
@@ -464,16 +464,52 @@
...
+            }
+        }
+    }
+}
  ]]></format>
         </remarks>
+          <!--<exception cref="T:System.Diagnostics.Eventing.Reader.EventLogException"></exception>-->
+          <exception cref="T:System.Diagnostics.Eventing.Reader.EventLogReadingException"></exception>
+          <exception cref="T:System.Diagnostics.Eventing.Reader.EventLogInvalidDataException"></exception>
+          <exception cref="T:System.InvalidOperationException">Thrown when ReadEvent is called again after a previous call returned null.</exception>
```
gewarren, line 511:
```suggestion
          <exception cref="T:System.InvalidOperationException">ReadEvent was called again after a previous call returned null.</exception>
```

xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
```diff
@@ -515,7 +551,48 @@
         <param name="timeout">The maximum time to allow the read operation to run before canceling the operation.</param>
         <summary>Reads the next event that is returned from the event query in this object.</summary>
         <returns>Returns an <see cref="T:System.Diagnostics.Eventing.Reader.EventRecord" /> object.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
```
gewarren, line 554: Please make all the same changes below.

gewarren (changes requested):

xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
Jehoel: Pardon my ignorance, but what is code-fencing? Also, what about the build validation warnings about `<format>` being invalid HTML? (…and I’m not even writing HTML, weird)

xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
Jehoel: I’m unsure if I should retain the `<exception>` reference to `EventLogReaderException` even though it’s redundant as the other exception types are subclasses thereof) - is it possible to render a hierarchical exception type list here?

xml/System.Diagnostics.Eventing.Reader/EventLogReader.xml
Jehoel: Methinks I should add the `<returns>` element back…

…but why remove the important details about `null` here?

PR #8215, opened 2022-07-14
Clean up for BigInteger

GitHubPang:
- There were 2 occurrences of "subtraction" in the same sentence, so I removed one.
- The sentence mentioned "negation" and "unary negation" which are equivalent (since negation is always unary), so I removed one.


BillWagner (approved):
Thanks again @GitHubPang 

I'll :shipit: now.

PR #8223, opened 2022-07-15
Fix docs for CompressionLevel.Optimal

stephentoub:
The docs were ambiguous as to what "optimal" was referring to... it was meant to refer to the balance between speed and size, whereas Fastest is best speed regardless of compression ratio, and SmallestSize is best compression ratio regardless of speed.

### DIFF
--- a/xml/System.IO.Compression/CompressionLevel.xml
+++ b/xml/System.IO.Compression/CompressionLevel.xml
@@ -83,7 +83,7 @@
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>The compression operation should optimally balance compression speed and output size.</summary>
+        <summary>The compression operation should complete as quickly as possible, even if the resulting file is not optimally compressed.</summary>
       </Docs>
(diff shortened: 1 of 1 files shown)


gewarren (approved):

xml/System.IO.Compression/CompressionLevel.xml
```diff
@@ -83,7 +83,7 @@
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>The compression operation should complete as quickly as possible, even if the resulting file is not optimally compressed.</summary>
```
adamsitnik, line 86: In the PR description you write:

> whereas Fastest is best speed regardless of compression ratio

and here you update `Fastest` description (`MemberValue=1`) to be the same as `Optimal` (`MemberValue=0`):

> optimally balance compression speed and output size.

I don't understand?

stephentoub: This line shouldn't have been edited.

PR #8224, opened 2022-07-15
Add remark discouraging HttpRequestMessage reuse

MihaZupan:
## Summary

Document that modifying and/or reusing an `HttpRequestMesssage` instance after it has been sent is discouraged (things may blow up in interesting ways at runtime).

Contributes to https://github.com/dotnet/runtime/issues/66989


ManickaP (approved):
Thanks.

PR #8226, opened 2022-07-15
Fix DateTime ISO 8601 example

### LINKED ISSUE #7848: DateTime ISO 8601 example violates this standard slightly
Link: <https://docs.microsoft.com/en-us/dotnet/api/system.datetime?view=net-6.0#formatting-05>

The example's line of code that contains the format string:

```
Console.WriteLine(date1.ToString("yyyyMMddTHH:mm:ssZ"));
```

Said format string uses basic format for date and extended format for time. (_Basic format_ means not using separators. _Extended format_ means using separators. Separators are `-` for dates and `:` for times.) The issue is that both date and time halves must have the same format.

Source: <https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations>

Sorry for using Wikipedia as a source. I can neither read nor refer to ISO itself because it's pay-walled. And I can't think of a better alternative than to give up and trust Wikipedia.

dakersnar:
Fixes #7848

### DIFF
--- a/snippets/csharp/System.DateTime/StringFormat.cs
+++ b/snippets/csharp/System.DateTime/StringFormat.cs
@@ -53,7 +53,7 @@ namespace SystemDateTimeReference
         {
             // <Snippet5>
             var date1 = new DateTime(2008, 3, 1, 7, 0, 0, DateTimeKind.Utc);
-            Console.WriteLine(date1.ToString("yyyy-MM-ddTHH:mm:sszzz", CultureInfo.InvariantCulture));
+            Console.WriteLine(date1.ToString("yyyy-MM-ddTHH:mm:sszzz", System.Globalization.CultureInfo.InvariantCulture));
             // Displays 2008-03-01T07:00:00+00:00
             // </Snippet5>
         }
--- a/snippets/visualbasic/System.DateTime/StringFormat.vb
+++ b/snippets/visualbasic/System.DateTime/StringFormat.vb
@@ -42,7 +42,7 @@
     Private Sub ShowIsoDateTime()
         ' <Snippet5>
         Dim date1 As DateTime = New DateTime(2008, 3, 1, 7, 0, 0, DateTimeKind.Utc)
-        Console.WriteLine(date1.ToString("yyyy-MM-ddTHH:mm:sszzz", CultureInfo.InvariantCulture))
+        Console.WriteLine(date1.ToString("yyyy-MM-ddTHH:mm:sszzz", System.Globalization.CultureInfo.InvariantCulture))
         ' Displays 2008-03-01T07:00:00+00:00
         ' </Snippet5>
     End Sub
--- a/snippets/fsharp/System.DateTime/StringFormat.fs
+++ b/snippets/fsharp/System.DateTime/StringFormat.fs
@@ -33,7 +33,7 @@ let showCultureSpecificFullDateAndTime () =
 let showIso8601Format () =
     // <Snippet5>
     let date1 = DateTime(2008, 3, 1, 7, 0, 0, DateTimeKind.Utc)
-    printfn $"""{date1.ToString("yyyy-MM-ddTHH:mm:sszzz", CultureInfo.InvariantCulture)}"""
+    printfn $"""{date1.ToString("yyyy-MM-ddTHH:mm:sszzz", System.Globalization.CultureInfo.InvariantCulture)}"""
     // Displays 2008-03-01T07:00:00+00:00
     // </Snippet5>
 


snippets/csharp/System.DateTime/StringFormat.cs
```diff
@@ -53,8 +53,8 @@ private static void ShowIso8601Format()
         {
             // <Snippet5>
             var date1 = new DateTime(2008, 3, 1, 7, 0, 0);
-            Console.WriteLine(date1.ToString("yyyyMMddTHH:mm:ssZ"));
-            // Displays 20080301T07:00:00Z
+            Console.WriteLine(date1.ToString("yyyy-MM-ddTHH:mm:ssZ"));
```
tarekgh, line 56: >. [](http://example.com/codeflow?start=35&length=1)

will be good to use the Invariant culture when formatting such data. or use single quotes around `:`. I prefer the former than the later.

dakersnar: Can you elaborate on what "Invariant culture" and "zzz" mean in this context? I'm not super familiar with DateTime formatting, I was just going off of the standard pointed out in the original issue.

tarekgh: > Can you elaborate on what "Invariant culture" mean

Calling `date1.ToString("yyyy-MM-ddTHH:mm:ssZ")` without passing any culture will cause the formatting using the CurrentCulture configured by the user. Some cultures can give a time separator different than `:` which can cause the output of the formatting not matching the ISO 8601 formats. Here is some example:

```C#
Console.WriteLine(dt.ToString("yyyy-MM-ddTHH:mm:sszzz", CultureInfo.GetCultureInfo("fi-FI")));
```

This will produce output like the following:

```
2022-07-18T10.09.16-07:00
```

To ensure this will not happen, we recommend using the invariant culture, like the following:

```C#
Console.WriteLine(dt.ToString("yyyy-MM-ddTHH:mm:sszzz", CultureInfo.InvariantCulture));
```

> Can you elaborate on what "zzz" mean

Please have a look at the [Custom date and time format strings doc](https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings), you can see the following there:

| Format specifier  | Description | Examples |
| ------------- | ------------- |  ------------- |
| "z"	| Hours offset from UTC, with no leading zeros. | More information: The "z" Custom Format Specifier.	2009-06-15T13:45:30-07:00 -> -7 |
| "zz"	| Hours offset from UTC, with a leading zero for a single-digit value. | More information: The "zz" Custom Format Specifier.	2009-06-15T13:45:30-07:00 -> -07 |
| "zzz" | Hours and minutes offset from UTC.  | More information: The "zzz" Custom Format Specifier.	2009-06-15T13:45:30-07:00 -> -07:00 |
 
By the way, using uppercase `Z` is wrong anyway.

dakersnar:
@tarekgh Thanks for the detailed suggestions. I pushed some edits, let me know if they look correct.

snippets/csharp/System.DateTime/StringFormat.cs
```diff
@@ -53,8 +53,8 @@ private static void ShowIso8601Format()
         {
             // <Snippet5>
             var date1 = new DateTime(2008, 3, 1, 7, 0, 0);
```
tarekgh, line 55: Would be nice of you can change this line to 

```C#
var dt1 = new DateTime(2008, 3, 1, 7, 0, 0, DateTimeKind.Utc);
```

then change the comment to 

```
// Displays 2008-03-01T07:00:00+00:00
```

tarekgh (approved):
Thanks @dakersnar for fixing this. I added a little suggestion. LGTM otherwise.

snippets/visualbasic/System.DateTime/StringFormat.vb
```diff
@@ -42,8 +42,8 @@
     Private Sub ShowIsoDateTime()
         ' <Snippet5>
         Dim date1 As Date = #3/1/2008 7:00AM#
-        Console.WriteLine(date1.ToString("yyyyMMddTHH:mm:ssZ"))
-        ' Displays 20080301T07:00:00Z
```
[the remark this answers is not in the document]
tarekgh: Try 

```VB
Dim date1 As DateTime = New DateTime(2008, 3, 1, 7, 0, 0, DateTimeKind.Utc)
```

snippets/visualbasic/System.DateTime/StringFormat.vb
```diff
@@ -42,8 +42,8 @@
     Private Sub ShowIsoDateTime()
         ' <Snippet5>
         Dim date1 As Date = #3/1/2008 7:00AM#
```
dakersnar, line 44:
```suggestion
        Dim date1 As DateTime = New DateTime(2008, 3, 1, 7, 0, 0, DateTimeKind.Utc)
```

PR #8249, opened 2022-07-22
Document AllocateUninitializedArray security and performance traps

jkotas:
Based on feedback from https://github.com/dotnet/roslyn/issues/62864#issuecomment-1192655037

### DIFF
--- a/xml/System/GC.xml
+++ b/xml/System/GC.xml
@@ -274,7 +274,7 @@ If pinned is set to `true`, `T` must not be a reference type or a type that cont
 Skipping zero-initialization is a security risk. The unitialized array can contain invalid valuetype instances or sensitive information created by other parts of the application.
 The code operating on unitialized arrays should be heavily scrutinized to ensure that the unitialized data is never read.
 
-Skipping zero-initialization is only profitable for large arrays, such as buffers of several kilobytes or more.
+Skipping zero-initialization using this API only has a material performance benefit for large arrays, such as buffers of several kilobytes or more.
 
           ]]></format>
         </remarks>


xml/System/GC.xml
```diff
@@ -263,14 +263,19 @@ If pinned is set to `true`, `T` must not be a reference type or a type that cont
...
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
 
 If pinned is set to `true`, `T` must not be a reference type or a type that contains object references.
 
+Skipping zero-initialization is a security risk. The unitialized array can contain invalid valuetype instances or sensitive information created by other parts of the application.
+The code operating on unitialized arrays should be heavily scrutinized to ensure that the unitialized data is never read.
+
+Skipping zero-initialization is only profitable for large arrays, such as buffers of several kilobytes or more.
```
333fred, line 277: IMO, profitable isn't particularly clear. Consider something more specific, like "only has a performance benefit".

tannergooding: I'm not sure "profitable" is the right word either.

Skipping zero initialization, even for small data, is almost always "better for performance" and it certainly is measurable. Even skipping zero initialization of locals (or optimizing it to use SIMD zeroing) is impactful enough to move the needle on many of our perf benchmarks.

The real issue is that the relative cost for small data is typically minimal, to the point that it doesn't matter in most contexts. If you're spending 10ns zeroing memory and 1us doing other things, then the zeroing isn't going to be impactful and its not what you should be looking at.

jkotas: Updated

tannergooding (approved):

333fred (approved):

PR #8259, opened 2022-07-27
Mention that HttpClient's proxy should not be modified after first use.

rzikm:
Closes https://github.com/dotnet/runtime/issues/65138.

### DIFF
--- a/xml/System.Net.Http/HttpClient.xml
+++ b/xml/System.Net.Http/HttpClient.xml
@@ -160,7 +160,7 @@ By default, HttpClient reads proxy configuration from environment variables or u
 
 You can disable the proxy using <xref:System.Net.Http.HttpClientHandler.UseProxy>. The default configuration for Windows users is to try and detect a proxy using network discovery, which can be slow. For high throughput applications where it's known that a proxy isn't required, you should disable the proxy.
 
-Proxy settings (like <xref:System.Net.IWebProxy.Credentials) should be changed only before the first request is made using the HttpClient. Changes made after using the HttpClient for the first time may not be reflected in subsequent requests.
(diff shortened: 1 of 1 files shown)


CarnaViire (commented):
I believe this doesn't only apply to proxy settings -- I think we should document that for all HttpClient properties that bear the same behavior.

Also, I think it's better to use "**should** only be changed" instead of "**can** only be changed" -- because, well, you *can* change those settings, meaning it wouldn't throw, but it will be useless 😄

rzikm:
> I think we should document that for all HttpClient properties that bear the same behavior.

The properties themselves throw when set after first use. The problem with proxy is that you are not setting the reference, but changing something inside the object.

xml/System.Net.Http/HttpClient.xml
```diff
@@ -160,6 +160,8 @@ By default, HttpClient reads proxy configuration from environment variables or u
 
 You can disable the proxy using <xref:System.Net.Http.HttpClientHandler.UseProxy>. The default configuration for Windows users is to try and detect a proxy using network discovery, which can be slow. For high throughput applications where it's known that a proxy isn't required, you should disable the proxy.
 
+Proxy settings (like <xref:System.Net.IWebProxy.Credentials) should be changed only before the first request is made using the HttpClient. Changes made after using the HttpClient for the first time may not be reflected in subsequent requests.
```
CarnaViire, line 163:
```suggestion
Proxy settings (like <xref:System.Net.IWebProxy.Credentials>) should be changed only before the first request is made using the HttpClient. Changes made after using the HttpClient for the first time may not be reflected in subsequent requests.
```

CarnaViire (approved):
LGTM modulo nit, thanks!

PR #8271, opened 2022-08-01
Fix stream parameter description on GZipStream, BrotliStream, and DeflateStream

stephentoub:
For the ctors that take a mode, the stream parameter is written to when CompressionMode.Compress is used or is read from when CompressionMode.Decompress is used.  These comments, in particular the ones on GZipStream, suggest the stream is written to regardless of the CompressionMode.


adamsitnik (approved):
LGTM, thank you @stephentoub !

PR #8274, opened 2022-08-03
SerialPort: remove discussion of encoding from `byte[]` overload

Porges:
This text seems to have been copied from the `string`/`char[]` overloads and doesn't make sense here.

This has been confusing people for [over 10 years](https://stackoverflow.com/a/10797494/10311).


gewarren (approved):
Thank you @Porges

PR #8278, opened 2022-08-03
Fix decimal represenation for 2-byte integer

klaza:
## Summary

Fix decimal represenation for 2-byte integers in BlobBuilder.WriteCompressedInteger(Int32) Method


buyaa-n (approved):
LGTM, thanks!

PR #8282, opened 2022-08-04
Change F# System.Random.Shared signature to static member

bisen2:
## Summary

Both C# and VB signatures correctly denote that `System.Random.Shared` is a static member, but the F# signature showed it as an instance member. This PR updates the F# signature to be in line with C#, VB, and the actual API.


gewarren:
I've opened internal work item https://dev.azure.com/ceapex/Engineering/_workitems/edit/686206 to address this bug. Unfortunately your change will just be overwritten the next time the tool reflects over the APIs (unless the tool is fixed by then).

gewarren:
This is fixed in https://github.com/dotnet/dotnet-api-docs/pull/8308/files#diff-5affc3a429d2127448ef55158a731a42672b6724504e9160bb0a6c0d1b5d2097. Thanks for letting us know about this issue!

PR #8303, opened 2022-08-12
Update Enum.xml

code4clouds:
Added a small clarification of what is the "out" default behavior when a TryParse fails using enums.

## Summary

Added a small clarification about the TryParse out behavior when a string fails to parse to the tenum.


buyaa-n:
The default value not have to be the first, look https://stackoverflow.com/questions/4967656/what-is-the-default-value-for-enum-variable for more info

code4clouds (commented):
Update to mention the default value could be at any position but most have a 0 value assigned.

xml/System/Enum.xml
```diff
@@ -4132,7 +4132,7 @@ thisInstance And flag = flag
       <Docs>
         <typeparam name="TEnum">The enumeration type to which to convert <paramref name="value" />.</typeparam>
         <param name="value">The case-sensitive string representation of the enumeration name or underlying value to convert.</param>
-        <param name="result">When this method returns, contains an object of type <c>TEnum</c> whose value is represented by <paramref name="value" /> if the parse operation succeeds. If the parse operation fails, contains the default value of the underlying type of <c>TEnum</c>. This value need not be a member of the <c>TEnum</c> enumeration. This parameter is passed uninitialized.</param>
+        <param name="result">When this method returns, contains an object of type <c>TEnum</c> whose value is represented by <paramref name="value" /> if the parse operation succeeds. If the parse operation fails, contains the default value of the underlying type of <c>TEnum</c>. The default value is the first element on the <c>TEnum</c> or a member in the enum with a value assigned of 0. This value need not be a member of the <c>TEnum</c> enumeration. This parameter is passed uninitialized.</param>
```
gewarren, line 4135:
```suggestion
        <param name="result">When this method returns, contains an object of type <c>TEnum</c> whose value is represented by <paramref name="value" /> if the parse operation succeeds. If the parse operation fails, contains the default value of the underlying type of <c>TEnum</c>. This parameter is passed uninitialized.</param>
```

gewarren: I'd suggest adding this info into the Remarks section below instead. Thoughts?

> If the parse operation fails, the `result` parameter contains the default value of the underlying type of `TEnum`. The default value is the first element on the `TEnum` or a member in the enumeration with an assigned value of 0. This value need not be a member of the `TEnum` enumeration.

But I also have a question - if the default value is the first element or the member with a value of 0, why does it say "this value need not be a member of the enum"?

buyaa-n: >  if the default value is the first element or the member with a value of 0, why does it say "this value need not be a member of the enum"?

The default value is 0 (numerically), which is by default equal to the first element of the Enum, if the dev assigned custom numbers then default Enum not have to be the 1st element or can be even not exist within the enum members. Example:
```cs
public enum DefaultValueEnum
{
    First, // defaulted to 0
    Second, // 1
    Third // 2
}

// The case where default 0 not exist:
public enum EnumWithCustomValue
{
    First = 1, 
    Second = 2, 
    Third  = 3
}
```
So not sure if we really need this update here, or the wording needs to be improved

code4clouds:
From the sample in the link provided an assignment to zero is used if the dev doesn't want to use / modify the current order of the enum members.  I don't know what will happen when all numbers are assigned and there is no zero.   Will it pick the first member as default?

buyaa-n:
> Will it pick the first member as default?

No, the result will have value of 0, that is why it has the sentence `This value need not be a member of the <c>TEnum</c> enumeration.`

xml/System/Enum.xml
code4clouds: From the sample in the SO link provided an assignment to zero is used if the dev doesn't want to use/modify the current order of the enum members. If there is no zero assigned on any member, but all members have a value return then tryparse returns a 0.
``` C#
        enum SubscriptionStatusEnumExtension
        {
            Unrecognized = 3,
            Check = 2,
            OK = 1,
        }
```
This return the value 0 as default and not any of the enum members.

code4clouds:
``` c#
        enum SubscriptionStatusEnumExtension
        {
            UnRecognized = 3,
            CHECK = 2,
            OK = 0,  // OK will be the default
        }
```

code4clouds:
The doc needs to be improved to make clear the behavior.  Maybe provide these examples.

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=8303) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=8303) before we can accept your contribution.<br/><br/>:x: code4clouds [sign now](https://cla.dotnetfoundation.org/dotnet/dotnet-api-docs?pullRequest=8303)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/dotnet-api-docs?pullRequest=8303) it.</sub>

xml/System/Enum.xml
```diff
@@ -4132,19 +4132,22 @@ thisInstance And flag = flag
...
+
+If `value` is a name that does not correspond to a named constant of `TEnum`, the method returns `false`. If `value` is the string representation of an integer that does not represent an underlying value of the `TEnum` enumeration, the method returns an enumeration member whose underlying value is `value` converted to an integral type. If this behavior is undesirable, call the <xref:System.Enum.IsDefined%2A> method to ensure that a particular string representation of an integer is actually a member of `TEnum`.
+
+If the parse operation fails, the `result` parameter contains the default value of the underlying type of `TEnum`. The default value is the member in the enumeration with an assigned value of 0, or the first member of the `TEnum` if no values are assigned, or `0` if values are assigned to each member but none of them is 0.
```
gewarren, line 4150: This is the new paragraph.

gewarren:
Do we need to update the other two enum-related overloads of TryParse? https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=net-7.0#system-enum-tryparse-1(system-readonlyspan((system-char))-0@)

xml/System/Enum.xml
buyaa-n: Thanks, I think the wording still needs update, because the result will be defaulted to 0; which may or may not match with the underlying value of the Enum.
```suggestion
If the parse operation fails, the `result` parameter contains the default value of 0 which may not be a member of the underlying <c>TEnum</c> enumeration. If no values are assigned to named constant of `TEnum` default is equal to the first member of the `TEnum` else it is equal to the member in the enumeration with an assigned value of 0.
```

code4clouds (commented):
I just noticed there is no explanation to what happens when there are non-zero values and zero is missing on the elements.

xml/System/Enum.xml
```diff
@@ -4277,19 +4280,22 @@ thisInstance And flag = flag
...
 
- If `value` is a name that does not correspond to a named constant of `TEnum`, the method returns `false`. If `value` is the string representation of an integer that does not represent an underlying value of the `TEnum` enumeration, the method returns an enumeration member whose underlying value is `value` converted to an integral type. If this behavior is undesirable, call the <xref:System.Enum.IsDefined%2A> method to ensure that a particular string representation of an integer is actually a member of `TEnum`.
+If the parse operation fails, the `result` parameter contains the default value of the underlying type of `TEnum`. The default value is the member in the enumeration with an assigned value of 0, or the first member of the `TEnum` if no values are assigned, or `0` if values are assigned to each member but none of them is 0.
```
buyaa-n, line 4298:
```suggestion
If the parse operation fails, the `result` parameter contains the default value of 0, which may not be a member of the underlying `TEnum` enumeration in case no member in the enumeration has an assigned value of 0. If no numeric values are assigned to the named constants in `TEnum`, the default is the first member of the `TEnum`.
```

code4clouds: This doesn't sound like the way it behaves... Is this related to a new version released?

buyaa-n:
> Do we need to update the other two enum-related overloads of TryParse? https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=net-7.0#system-enum-tryparse-1(system-readonlyspan((system-char))-0@)

It seems they do not have a remarks section that explains this behavior, so nothing to update, but this part can be added  

> @buyaa-n @gewarren - I just noticed there is no explanation to what happens when all elements have non-zero values and zero is missing. In that specific case no member from the enum is return and a zero is returned.

Well, the `which may not be a member of the underlying 'TEnum' enumeration.` part is covering that: https://github.com/dotnet/dotnet-api-docs/blob/9adc850f451375f7be1915961550387ddcc94a91/xml/System/Enum.xml#L4150 

If the above version is not understandable what about this:
`If the parse operation fails, the 'result' parameter contains the default value of 0, which may not be a member of the underlying 'TEnum' enumeration in case no member in the enumeration has an assigned value of 0. If no any numeric values are assigned to the named constants in 'TEnum', the default is the first member of the 'TEnum'.
`

xml/System/Enum.xml
buyaa-n: Nope its same, just trying to make the wording better, sounds like it did not improve 😛, which part sounds so different?

dakersnar: To clarify, what exactly is the current behavior? 

For example, what would a failed parse operation return if it was trying to parse to this Enum?
image

buyaa-n: if try to parse an enum member name that does not exist in the enum list the parse would fail and result would have default value of 0.

For your example `testEnum` if you try `Enum.TryParse<testEnum>("test3", out var result);` parse fails and the result will be defaulted to 0 (non-existing value in the enum members); and as you know if try parse `Enum.TryParse<testEnum>("test2", out var result);` parse succeeds and result will be `test2` (2 numerically)

If the `testEnum` were not numbered:
```cs
    public enum testEnum
    {
        test1,
        test2 
    }
```
then by default, the members would be numbered with 0 and 1 therefore `Enum.TryParse<testEnum>("test3", out var result);` parse fails and the result value will be defaulted to `test1` (still 0 numerically)

dakersnar: Got it. Given that, I actually think the original text is clear.

buyaa-n: Well from the original text the the `'result' parameter contains the default value of the underlying type of TEnum'` section does not sound correct because the default value (0) might not be part of the underlying type of `TEnum`. We have updated the same remarks above like this: 
```suggestion
If the parse operation fails, the `result` parameter contains the default value of 0, which may not be a member of the underlying `TEnum` enumeration. If no values are assigned to the named constants in `TEnum`, the default is equal to the first member of the `TEnum`. Otherwise, the default is equal to the member in the enumeration with an assigned value of 0.
```

PR #8328, opened 2022-08-22
Spelling.

daverayment:
Fix spelling issue.

### DIFF
--- a/xml/System.Runtime.Intrinsics.X86/Popcnt+X64.xml
+++ b/xml/System.Runtime.Intrinsics.X86/Popcnt+X64.xml
@@ -87,7 +87,7 @@
 
 ## Remarks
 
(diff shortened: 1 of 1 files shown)


xml/System.Runtime.Intrinsics.X86/Popcnt+X64.xml
```diff
@@ -87,7 +87,7 @@
 
 ## Remarks
 
-This intrinsic is only available on 64-bit processes.
```
tannergooding, line 90: This is actually not a typo. The requirement is that you are running in a 64-bit process. Running a 32-bit process on a 64-bit processor will still have this as `unsupported`.

BillWagner:
Thanks for adding this @daverayment 

As @tannergooding points out, the original doc is correct. I'll close this.

PR #8331, opened 2022-08-24
Fix remark on NetworkStream.BeginRead

rzikm:
Fixes https://github.com/dotnet/runtime/issues/69269.


antonfirsov:
This is better covered by #8332

PR #8335, opened 2022-08-25
Port inheritdoc tag for runtime APIs that are missing docs

carlossanlop:
MS Docs is smart enough to automatically show inherited documentation if the `<inheritdoc/>` tag is present. If an API contains such tag in the intellisense xml, we can port that tag instead of copying the parent's documentation directly. But if the API contains documentation (partial or total) in the intellisense xml, we want to port that, because we should assume the owner wants to override some or all of the parent's documentation.

The docs porting tool is now able to port documentation following that logic.

The undoc APIs report we generate from MS Docs is also able to assume an API to be documented if it has the inheritdoc tag, and the parent is documented.

This PR ports the inheritdoc tag only for APIs that are partially or fully undocumented, so that they stop showing in the report.

I submitted another PR https://github.com/dotnet/dotnet-api-docs/pull/8336 that includes the rest of the inheritdoc tags that were detected in APIs that have already been fully documented before (by copying the base type or the interface's documentation).

PR #8336, opened 2022-08-25
Port inheritdoc tags for APIs that are already fully documented

carlossanlop:
This is a continuation of https://github.com/dotnet/dotnet-api-docs/pull/8335 . Full explanation can be found there.
This PR includes the tags that were found in APIs that were already documented.

### DIFF
--- a/xml/System/TimeOnly.xml
+++ b/xml/System/TimeOnly.xml
@@ -1012,7 +1012,7 @@ If `start` and `end` are equal, this method returns `false`, meaning there is ze
         <summary>Parses a span of characters into a value.</summary>
         <returns>The result of parsing <paramref name="s" />.</returns>
         <remarks>To be added.</remarks>
-        <inheritdoc cref="M:System.ISpanParsable`1.Parse(System.ReadOnlySpan{{System.Char}},System.IFormatProvider)" />
+        <inheritdoc cref="M:System.ISpanParsable`1.Parse(System.ReadOnlySpan{System.Char},System.IFormatProvider)" />
       </Docs>
     </Member>
     <Member MemberName="Parse">
@@ -1075,7 +1075,7 @@ If `start` and `end` are equal, this method returns `false`, meaning there is ze
         <remarks>To be added.</remarks>
         <exception cref="T:System.FormatException">
           <paramref name="s" /> does not contain a valid string representation of a time.</exception>
-        <inheritdoc cref="M:System.ISpanParsable`1.Parse(System.ReadOnlySpan{{System.Char}},System.IFormatProvider)" />
+        <inheritdoc cref="M:System.ISpanParsable`1.Parse(System.ReadOnlySpan{System.Char},System.IFormatProvider)" />
       </Docs>
     </Member>
     <Member MemberName="Parse">
@@ -1789,7 +1789,7 @@ The accepted standard formats are 'r', 'R', 'o', 'O', 't', and 'T'.
         <returns>
           <see langword="true" /> if <paramref name="s" /> was successfully parsed; otherwise, <see langword="false" />.</returns>
         <remarks>To be added.</remarks>
-        <inheritdoc cref="M:System.ISpanParsable`1.TryParse(System.ReadOnlySpan{{System.Char}},System.IFormatProvider,`0@)" />
+        <inheritdoc cref="M:System.ISpanParsable`1.TryParse(System.ReadOnlySpan{System.Char},System.IFormatProvider,`0@)" />
       </Docs>
     </Member>
     <Member MemberName="TryParse">
@@ -1861,7 +1861,7 @@ The accepted standard formats are 'r', 'R', 'o', 'O', 't', and 'T'.
         <returns>
           <see langword="true" /> if<paramref name="s" /> was converted successfully; otherwise, <see langword="false" />.</returns>
         <remarks>To be added.</remarks>
-        <inheritdoc cref="M:System.ISpanParsable`1.TryParse(System.ReadOnlySpan{{System.Char}},System.IFormatProvider,`0@)" />
+        <inheritdoc cref="M:System.ISpanParsable`1.TryParse(System.ReadOnlySpan{System.Char},System.IFormatProvider,`0@)" />
       </Docs>
     </Member>
     <Member MemberName="TryParse">


carlossanlop:
@dotnet/area-extensions-caching @dotnet/area-extensions-configuration @dotnet/area-extensions-dependencyinjection @dotnet/area-extensions-filesystem @dotnet/area-extensions-hosting @dotnet/area-extensions-logging @dotnet/area-extensions-primitives @dotnet/area-system-buffers @dotnet/area-system-codedom @dotnet/ncl @HongGit @StephenMolloy @dotnet/area-system-runtime @dotnet/area-system-security @dotnet/area-system-text-json @dotnet/area-system-text-encoding @dotnet/area-system-xml @dotnet/area-system-globalization

tarekgh:
@carlossanlop does this change include all inherit docs or just a subset of it? I am asking because I recall there were some of them in DateOnly and component model which not showing up here. 

MS extensions and System changes LGTM.

carlossanlop:
@tarekgh the main PR is this one: https://github.com/dotnet/dotnet-api-docs/pull/8335, which contains the subset of APIs that are currently showing as undocumented in the report.

The rest of them are the ones you're seeing in this PR. They are the APIs that have already been fully documented in the past, by copying the strings from the base type or the interface.

carlossanlop:
I'm investigating the warnings. I see the tool is porting some crefs with characters escaped, and that should not happen.

xml/System.Security.Cryptography/CryptoStream.xml
```diff
@@ -591,6 +591,7 @@ Copying begins at the current position in the current stream, and does not reset
         <exception cref="T:System.ObjectDisposedException">Either the current stream or <paramref name="destination" /> were closed before the <see cref="M:System.IO.Stream.CopyTo(System.IO.Stream)" /> method was called.</exception>
         <exception cref="T:System.IO.IOException">An I/O error occurred.</exception>
         <exception cref="T:System.Security.Cryptography.CryptographicException">An error occurred during a cryptographic operation.</exception>
+        <inheritdoc />
```
bartonjs, line 594: This method has full documentation... what does inheritdoc mean in context?

tannergooding (approved):

BillWagner (approved):
This LGTM @carlossanlop 

Once the open conversations are resolved, you can :shipit:

xml/System.Security.Cryptography/CryptoStream.xml
carlossanlop: This is what we should've ported from the beginning, instead of the individual texts. I'm adding these since the tool now detects them. We can later decide if we want to clear the summary, returns, params, typeparams of these so that we use the inherited documentation.

gewarren:
> The undoc APIs report we generate from MS Docs is also able to assume an API to be documented if it has the inheritdoc tag, and the parent is documented.
> This PR ports the inheritdoc tag only for APIs that are partially or fully undocumented, so that they stop showing in the report.

Why do the APIs show up in the report if they are already documented?

xml/System.Runtime.Serialization/ImportOptions.xml
```diff
@@ -107,7 +107,7 @@
       <Docs>
         <summary>Gets or sets a data contract surrogate that can be used to modify the code generated during an import operation.</summary>
         <value>An implementation of the <see cref="T:System.Runtime.Serialization.IDataContractSurrogate" /> interface that handles schema import.</value>
-        <remarks>To be added.</remarks>
+        <remarks>The interface type for this option is ISerializationSurrogateProvider, but to take full advantage of the imported code modification abilities, using an <see cref="T:System.Runtime.Serialization.ISerializationSurrogateProvider2" />ISerializationSurrogateProvider2 that also implements <see cref="T:System.Runtime.Serialization.ISerializationCodeDomSurrogateProvider" /> is recommended.</remarks>
```
carlossanlop, line 110:
```suggestion
        <remarks>To be added.</remarks>
```

carlossanlop:
> > The undoc APIs report we generate from MS Docs is also able to assume an API to be documented if it has the inheritdoc tag, and the parent is documented.
> > This PR ports the inheritdoc tag only for APIs that are partially or fully undocumented, so that they stop showing in the report.
> 
> Why do the APIs show up in the report if they are already documented?

That's the description of the other PR, in which I am porting the inheritdoc tag of those APIs that were still partially or fully undocumented. So those are the ones showing up in the report.

This PR contains APIs that are already fully documented, but the tool also found they had the inheritdoc tag in triple slash. These aren't showing up in the report, I'm just porting them for completeness.

carlossanlop (commented):
The rest of the warnings are being checked by the Docs dev team, they seem to be a bug.

xml/System.Runtime.Serialization/ImportOptions.xml
carlossanlop: This is a remark that mentions APIs from 8.0. Removing.

gewarren:
> This PR contains APIs that are already fully documented, but the tool also found they had the inheritdoc tag in triple slash. These aren't showing up in the report, I'm just porting them for completeness.

I wonder what mdoc would do in this case. Perhaps we should test.

xml/System/TimeOnly.xml
```diff
@@ -1006,6 +1012,7 @@ If `start` and `end` are equal, this method returns `false`, meaning there is ze
         <summary>Parses a span of characters into a value.</summary>
         <returns>The result of parsing <paramref name="s" />.</returns>
         <remarks>To be added.</remarks>
+        <inheritdoc cref="M:System.ISpanParsable`1.Parse(System.ReadOnlySpan{{System.Char}},System.IFormatProvider)" />
```
carlossanlop, line 1015:
```suggestion
        <inheritdoc cref="M:System.ISpanParsable`1.Parse(System.ReadOnlySpan{System.Char},System.IFormatProvider)" />
```

xml/System/TimeOnly.xml
```diff
@@ -1067,6 +1075,7 @@ If `start` and `end` are equal, this method returns `false`, meaning there is ze
         <remarks>To be added.</remarks>
         <exception cref="T:System.FormatException">
           <paramref name="s" /> does not contain a valid string representation of a time.</exception>
+        <inheritdoc cref="M:System.ISpanParsable`1.Parse(System.ReadOnlySpan{{System.Char}},System.IFormatProvider)" />
```
carlossanlop, line 1078:
```suggestion
        <inheritdoc cref="M:System.ISpanParsable`1.Parse(System.ReadOnlySpan{System.Char},System.IFormatProvider)" />
```

xml/System/TimeOnly.xml
```diff
@@ -1780,6 +1789,7 @@ The accepted standard formats are 'r', 'R', 'o', 'O', 't', and 'T'.
         <returns>
           <see langword="true" /> if <paramref name="s" /> was successfully parsed; otherwise, <see langword="false" />.</returns>
         <remarks>To be added.</remarks>
+        <inheritdoc cref="M:System.ISpanParsable`1.TryParse(System.ReadOnlySpan{{System.Char}},System.IFormatProvider,`0@)" />
```
carlossanlop, line 1792:
```suggestion
        <inheritdoc cref="M:System.ISpanParsable`1.TryParse(System.ReadOnlySpan{System.Char},System.IFormatProvider,`0@)" />
```

xml/System/TimeOnly.xml
```diff
@@ -1850,6 +1861,7 @@ The accepted standard formats are 'r', 'R', 'o', 'O', 't', and 'T'.
         <returns>
           <see langword="true" /> if<paramref name="s" /> was converted successfully; otherwise, <see langword="false" />.</returns>
         <remarks>To be added.</remarks>
+        <inheritdoc cref="M:System.ISpanParsable`1.TryParse(System.ReadOnlySpan{{System.Char}},System.IFormatProvider,`0@)" />
```
carlossanlop, line 1864:
```suggestion
        <inheritdoc cref="M:System.ISpanParsable`1.TryParse(System.ReadOnlySpan{System.Char},System.IFormatProvider,`0@)" />
```

carlossanlop:
> I wonder what mdoc would do in this case. Perhaps we should test.

You were right to raise the concern. If we add inheritdoc to an API with partial or full documentation, inheritdoc is ignored.

Example: The GUID operator "<=":

Here is the preview:
https://review.docs.microsoft.com/en-us/dotnet/api/system.guid.op_greaterthanorequal?view=net-7.0&branch=pr-en-us-8345

Here is the source file:
https://github.com/dotnet/runtime/blob/fe9a5a4685f1314bab48804ae9b9716890b331e1/src/libraries/System.Private.CoreLib/src/System/Guid.cs#L1302

Here are the partially documented docs:
https://github.com/dotnet/dotnet-api-docs/blob/83a7d69c083dbb5fa02fb991777edcadc0ee01a4/xml/System/Guid.xml#L1149-L1156

Here is the interface from which it should inherit automatically:
https://github.com/dotnet/dotnet-api-docs/blob/83a7d69c083dbb5fa02fb991777edcadc0ee01a4/xml/System.Numerics/IComparisonOperators%603.xml#L82-L83

So I guess the Docs porting tool should strip all the existing documentation (except remarks) if it is going to add an inheritdoc. What do you think, @gewarren?

rzikm (approved):
System.Net.* changes LGTM

gewarren:
> > I wonder what mdoc would do in this case. Perhaps we should test.
> 
> You were right to raise the concern. If we add inheritdoc to an API with partial or full documentation, inheritdoc is ignored.

But isn't there a bug with inheritdoc on operators specifically?

carlossanlop:
> But isn't there a bug with inheritdoc on operators specifically?

Ah, you're right. It's basically the same problem as the bug we opened: the parent API to inherit from is not found, so it's still considered undocumented, wrongly.

> So I guess this PR is okay.

Thank you for checking. I'll merge.

PR #8351, opened 2022-08-31
Add missing namespace docs for System.IO.Hashing

bartonjs:
Contributes to https://github.com/dotnet/runtime/issues/72406.

### DIFF
--- a/xml/ns-System.IO.Hashing.xml
+++ b/xml/ns-System.IO.Hashing.xml
@@ -1,6 +1,6 @@
 <Namespace Name="System.IO.Hashing">
   <Docs>
-    <summary>To be added.</summary>
+    <summary>Contains types used in computing non-cryptographic hash values.</summary>
     <remarks>To be added.</remarks>
   </Docs>
 </Namespace>


xml/ns-System.IO.Hashing.xml
```diff
@@ -1,6 +1,6 @@
 <Namespace Name="System.IO.Hashing">
   <Docs>
-    <summary>To be added.</summary>
+    <summary>Contains types used in computing non-cryptographic hash values.</summary>
```
Jozkee, line 3: Shouldn't namespaces summaries align with NuGet package descriptions?
> Provides non-cryptographic hash algorithms, such as CRC-32.

Jozkee (approved):

carlossanlop (approved):
LGTM, pending addressing the @Jozkee question.

xml/ns-System.IO.Hashing.xml
bartonjs: Namespaces and NuGet packages have different purposes and show up in docs in different places.  A single NuGet package can span multiple namespaces, and a single namespace can be populated from multiple NuGet packages.

I don't really have objections (other than "Contains types" seems to be the predominant namespace style and "Provides [something]" is the predominant NuGet package description style), but I don't think that there's necessarily any suggestion that they should be the same.

gewarren (approved):

PR #8356, opened 2022-09-02
Update descriptions with alternate type params

gewarren:
[Internal preview](https://review.docs.microsoft.com/en-us/dotnet/api/system.runtime.intrinsics.vector128.as?view=net-7.0&branch=pr-en-us-8356) - Not sure if it's too busy with both versions of the name.


BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

tannergooding:
I don't think this has ever been known as `T`/`U` and so I think it would be better if we only used `TFrom`/`TTo` as those are semantically clear/relevant (the `from type` and the `to type`).

PR #8361, opened 2022-09-06
GetTempFileName

danmoseley:
Fix https://github.com/dotnet/runtime/issues/73793

1. Note that the limit of 65,535 temp files on Windows does not exist in .NET 8 or later.
2. Clean up/clarify GetTempPath as well.

cc @eerhardt 

BTW, I know it's a bit early for .NET 8. I don't know what the protocol is -- do you branch?

### DIFF
--- a/xml/System.IO/Path.xml
+++ b/xml/System.IO/Path.xml
@@ -1949,7 +1949,7 @@ Paths are resolved by calling the <xref:System.IO.Path.GetFullPath%2A> method be
 
 This method creates a temporary file with a *.tmp* file extension and returns the path to it. The temporary file is created within the user's temporary folder, which is the path returned by the <xref:System.IO.Path.GetTempPath%2A> method.  
   
- Previous to .NET 8.0, when using this method on Windows the <xref:System.IO.Path.GetTempFileName%2A> method will raise an <xref:System.IO.IOException> if it is used to create more than 65535 files without deleting previous temporary files. This limitation does not exist on other operating systems, nor does it exist on any operating system in .NET 8.0 and later versions.
+On .NET 7 and earlier versions, when using this method on Windows, the <xref:System.IO.Path.GetTempFileName%2A> method raises an <xref:System.IO.IOException> if it's used to create more than 65535 files without deleting previous temporary files. This limitation does not exist on operating systems other than Windows. Starting in .NET 8, the limitation does not exist on *any* operating system.
   
  The <xref:System.IO.Path.GetTempFileName%2A> method will raise an <xref:System.IO.IOException> if no unique temporary file name is available. To resolve this error, delete all unneeded temporary files.  
   


xml/System.IO/Path.xml
```diff
@@ -1946,9 +1946,9 @@ Paths are resolved by calling the <xref:System.IO.Path.GetFullPath%2A> method be
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- This method creates a temporary file with a .TMP file extension. The temporary file is created within the user's temporary folder, which is the path returned by the <xref:System.IO.Path.GetTempPath%2A> method.  
+ This method creates a temporary file with a .tmp file extension and returns the path to it. The temporary file is created within the user's temporary folder, which is the path returned by the <xref:System.IO.Path.GetTempPath%2A> method.  
   
- The <xref:System.IO.Path.GetTempFileName%2A> method will raise an <xref:System.IO.IOException> if it is used to create more than 65535 files without deleting previous temporary files.  
+ Previous to .NET 8.0, when using this method on Windows the <xref:System.IO.Path.GetTempFileName%2A> method will raise an <xref:System.IO.IOException> if it is used to create more than 65535 files without deleting previous temporary files. This limitation does not exist on other operating systems, nor does it exist on any operating system in .NET 8.0 and later versions.
```
danmoseley, line 1951: do we prefer to use something like ".NET 7.0 and earlier" (with the complications of .NET Framework) rather than caveating this way? i

xml/System.IO/Path.xml
```diff
@@ -2027,11 +2029,15 @@ Paths are resolved by calling the <xref:System.IO.Path.GetFullPath%2A> method be
   
 3.  The path specified by the USERPROFILE environment variable.  
   
-4.  The Windows directory.  
+4.  The Windows directory.
+
+In .NET 7.0 and later, when the process calling the method is running as SYSTEM, the method skips the sequence above and returns C:\Windows\SystemTemp. This location is inaccessible to processes not running as SYSTEM.
```
gewarren, line 2034:
```suggestion
In .NET 7 and later, when the process calling the method is running as `SYSTEM`, the method skips the preceding sequence and returns `C:\Windows\SystemTemp`. This location is inaccessible to processes not running as `SYSTEM`.
```

xml/System.IO/Path.xml
```diff
@@ -2016,8 +2016,10 @@ Paths are resolved by calling the <xref:System.IO.Path.GetFullPath%2A> method be
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
-## Remarks  
- This method checks for the existence of environment variables in the following order and uses the first path found:  
+## Remarks
+ The method does not verify that the path exists, nor does it test to see if the current process can access the path.
```
gewarren, line 2020:
```suggestion

The method does not verify that the path exists or test to see if the current process can access the path. 
```

xml/System.IO/Path.xml
```diff
@@ -1946,9 +1946,9 @@ Paths are resolved by calling the <xref:System.IO.Path.GetFullPath%2A> method be
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- This method creates a temporary file with a .TMP file extension. The temporary file is created within the user's temporary folder, which is the path returned by the <xref:System.IO.Path.GetTempPath%2A> method.  
+ This method creates a temporary file with a .tmp file extension and returns the path to it. The temporary file is created within the user's temporary folder, which is the path returned by the <xref:System.IO.Path.GetTempPath%2A> method.
```
gewarren, line 1949:
```suggestion

This method creates a temporary file with a *.tmp* file extension and returns the path to it. The temporary file is created within the user's temporary folder, which is the path returned by the <xref:System.IO.Path.GetTempPath%2A> method.  
```

xml/System.IO/Path.xml
gewarren, line 1951:
```suggestion
On .NET 7 and earlier versions, when using this method on Windows, the <xref:System.IO.Path.GetTempFileName%2A> method raises an <xref:System.IO.IOException> if it's used to create more than 65535 files without deleting previous temporary files. This limitation does not exist on operating systems other than Windows. Starting in .NET 8, the limitation does not exist on *any* operating system.
```

gewarren (approved):
We might want to create a .NET 8 branch going forward, but I think this one is fine to merge now.

xml/System.IO/Path.xml
gewarren: Yes, I like that better especially since .NET 8 hasn't released yet.

PR #8371, opened 2022-09-08
Fix regex typo

rccarlson:
Regex was invalid. Added escape character to minus symbol.
Example of new regex in use: https://regex101.com/r/Xd6EH9/1


gewarren (approved):
Thanks @rccarlson!

PR #8377, opened 2022-09-11
Fixes #7669 by updating doc of Enumerable.Sum as suggested

### LINKED ISSUE #7669: Documentation on Enumerable.Sum(IEnumerable<Nullable<double>>) should be clearer on whether/when it returns null
The current documentation says
~~~
This method returns zero if source contains no elements.

The result does not include values that are null.
~~~
You could deduce from this that the method never returns null, but I think 
* `the result does not include` could be put more precisely
* whether the method returns null should be stated explicitly, because users need to know whether they need to handle a null return value. 
~~~
This method returns zero if source contains no elements.

Items in source that are null are excluded from the computation of the sum.
This method never returns null.
~~~

The same goes for all the other `Enumerable.Sum(IEnumerable<Nullable<TNumber>>)` overloads.

eriawan:
## Summary

This PR fixes #7669 by updating `Enumerable.Sum` docs page with suggested change from the original issue.

cc @eiriktsarpalis @layomia

### DIFF
--- a/xml/System.Linq/Enumerable.xml
+++ b/xml/System.Linq/Enumerable.xml
@@ -11050,9 +11050,10 @@ If `count` is not a positive number, this method returns an identical copy of th
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- The <xref:System.Linq.Enumerable.Sum%28System.Collections.Generic.IEnumerable%7BSystem.Decimal%7D%29> method returns zero if `source` contains no elements.
 
- In Visual Basic query expression syntax, an `Aggregate Into Sum()` clause translates to an invocation of <xref:System.Linq.Enumerable.Sum%2A>.
+The <xref:System.Linq.Enumerable.Sum%28System.Collections.Generic.IEnumerable%7BSystem.Decimal%7D%29> method returns zero if `source` contains no elements.
+
+In Visual Basic query expression syntax, an `Aggregate Into Sum()` clause translates to an invocation of <xref:System.Linq.Enumerable.Sum%2A>.
 
   ]]></format>
         </remarks>
@@ -11106,9 +11107,10 @@ If `count` is not a positive number, this method returns an identical copy of th
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- This method returns zero if `source` contains no elements.
 
- In Visual Basic query expression syntax, an `Aggregate Into Sum()` clause translates to an invocation of <xref:System.Linq.Enumerable.Sum%2A>.
+This method returns zero if `source` contains no elements.
+
+In Visual Basic query expression syntax, an `Aggregate Into Sum()` clause translates to an invocation of <xref:System.Linq.Enumerable.Sum%2A>.
 
  ]]></format>
         </remarks>
@@ -11161,9 +11163,10 @@ If `count` is not a positive number, this method returns an identical copy of th
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- This method returns zero if `source` contains no elements.
 
- In Visual Basic query expression syntax, an `Aggregate Into Sum()` clause translates to an invocation of <xref:System.Linq.Enumerable.Sum%2A>.
+This method returns zero if `source` contains no elements.
+
+In Visual Basic query expression syntax, an `Aggregate Into Sum()` clause translates to an invocation of <xref:System.Linq.Enumerable.Sum%2A>.
 
  ]]></format>
         </remarks>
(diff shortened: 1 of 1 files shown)


xml/System.Linq/Enumerable.xml
```diff
@@ -11275,7 +11275,7 @@ If `count` is not a positive number, this method returns an identical copy of th
 ## Remarks
  This method returns zero if `source` contains no elements.
 
- The result does not include values that are `null`.
+ Items in `source` that are null are excluded from the computation of the sum. This method never returns `null`.
```
eiriktsarpalis, line 11278: I would probably combine this with the sentence above to say something like:

> Items in `source` that are `null` are excluded from the computation of the sum. This method returns zero if `source` contains no elements or all elements are `null`.

eriawan: @eiriktsarpalis 
Thanks for the feedback!
Done update the wording based on feedback above. Please review 🙂

xml/System.Linq/Enumerable.xml
```diff
@@ -11993,9 +11983,7 @@ If `count` is not a positive number, this method returns an identical copy of th
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- This method returns zero if `source` contains no elements.
-
- The result does not include values that are `null`.
+ Items in `source` that are null are excluded from the computation of the sum. This method never returns `null`.
```
eiriktsarpalis, line 11986: Any particular reason why this remark uses different wording than the ones above?

eiriktsarpalis: Right, my point is that it's not applied to all methods consistently. Shouldn't it be the same everywhere?

eriawan: I see, I understand now,  thanks! 
I was thinking that the wordings only applicable to Sum that takes nullable parameters such as `int?` 
Ok, I'm going to apply to all methods as well today. 🙂

eiriktsarpalis: Are some of these overloads not taking nullable parameters? I haven't checked, but I would assume they do because the remarks already make reference to `null` values.

eriawan: > Are some of these overloads not taking nullable parameters? I haven't checked, but I would assume they do because the remarks already make reference to `null` values.

AFAIK yes. For example:
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sum?view=net-6.0#system-linq-enumerable-sum(system-collections-generic-ienumerable((system-decimal)))

and looking at the remark, I think it doesn't have reference to `null` values:

image

CMIIW, I think I need to ask your confirmation about this, because going back to my original understanding before, the update that I have applied should only matter for sum that takes IEnumerable<T> that takes T nullable such as `int?` or `double?`, like below overloads:

image

eiriktsarpalis: All remark sections that this PR updates originally contain this comment:

> The result does not include values that are `null`.

Presumably they're all nullable variants? If so then the same change should be applied everywhere.

eriawan: > Presumably they're all nullable variants? If so then the same change should be applied everywhere.

Yes, I've only updated all of the remarks that have that sentence and as long as the IEnumerable parameter has with T of nullable primitive number types such as `int?`, `long?`, `Double?`.

Ok, let me check again and ensure that I'm going to update all of the remarks section of any Sum that has met the above requirements.

xml/System.Linq/Enumerable.xml
```diff
@@ -11273,9 +11273,7 @@ If `count` is not a positive number, this method returns an identical copy of th
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- This method returns zero if `source` contains no elements.
-
- The result does not include values that are `null`.
+ Items in `source` that are `null` are excluded from the computation of the sum. This method returns zero if `source` contains no elements or all elements are `null`.
```
gewarren, line 11276: Since you are changing all these remarks, could you also add a blank line after the heading:

```suggestion

Items in `source` that are `null` are excluded from the computation of the sum. This method returns zero if `source` contains no elements or all elements are `null`.
```

eiriktsarpalis:
@eriawan would it be possible to address the remaining feedback? Thanks!

eriawan:
@eiriktsarpalis 
Yes, I'm addressing this. Apologize, I've been busy for the last 10 days.
I'll update this Saturday.

xml/System.Linq/Enumerable.xml
eriawan: updared

eriawan:
I'm still going thru all of the usage of Nullable structs. Going to finish this today.

xml/System.Linq/Enumerable.xml
eriawan: @gewarren 
done applying feedback on `Sum()`. Please review.

eriawan:
@eiriktsarpalis 

As far as I know I have addressed all feedback, including feedback from @gewarren , focusing on `Sum()` mehods only.

@gewarren 

There are some whitespaces on other methods of Enumerable. In this PR I'm focusing on `Sum()` method and its overloads first. then I'll update on separate PR to ensure that this PR review should be focusing on Sum() first, IMHO.

Thanks in advance! 🙂

eiriktsarpalis (approved):
Thanks!

gewarren (approved):

PR #8378, opened 2022-09-12
Add explicit variable names

drewnoakes:
## Summary

In the `Semaphore` class's documentation, these particular APIs are the focus within this snippet. It's not obvious what the integer literal arguments mean.

Adding parameter names makes this clearer to the reader.

Describe your changes here.


BillWagner (approved):
Thanks @drewnoakes 

This is much better. I'll :shipit:

PR #8379, opened 2022-09-12
Clarify Process.HasExited and that it raises Exited

kierenj:
## Summary

The existing description implied that the Exited event could be raised "before" the user calls HasExited. This wasn't very clear in my opinion - so I've suggested a tweak to make it more explicit by revealing that the HasExited property accessor itself has a side-effect; invoking HasExited if the process has exited.

### DIFF
--- a/xml/System.Diagnostics/Process.xml
+++ b/xml/System.Diagnostics/Process.xml
@@ -927,7 +927,7 @@ process.BeginOutputReadLine();
       <Docs>
         <summary>Gets or sets whether the <see cref="E:System.Diagnostics.Process.Exited" /> event should be raised when the process terminates.</summary>
         <value>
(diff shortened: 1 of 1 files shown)


xml/System.Diagnostics/Process.xml
```diff
@@ -927,7 +927,7 @@ process.BeginOutputReadLine();
...
+          <see langword="true" /> if the <see cref="E:System.Diagnostics.Process.Exited" /> event should be raised when the associated process is terminated (through either an exit or a call to <see cref="M:System.Diagnostics.Process.Kill" />); otherwise, <see langword="false" />. The default is <see langword="false" />. Note that even if the value of <see cref="P:System.Diagnostics.Process.EnableRaisingEvents" /> is <see langword="false" />, the <see cref="E:System.Diagnostics.Process.Exited" /> event will be raised by the <see cref="P:System.Diagnostics.Process.HasExited" /> property accessor itself, if it determines that the process has exited.</value>
```
Jozkee, line 930:
```suggestion
          <see langword="true" /> if the <see cref="E:System.Diagnostics.Process.Exited" /> event should be raised when the associated process is terminated (through either an exit or a call to <see cref="M:System.Diagnostics.Process.Kill" />); otherwise, <see langword="false" />. The default is <see langword="false" />. Note that even if the value of <see cref="P:System.Diagnostics.Process.EnableRaisingEvents" /> is <see langword="false" />, the <see cref="E:System.Diagnostics.Process.Exited" /> event will be raised by the <see cref="P:System.Diagnostics.Process.HasExited" /> property accessor, if it determines that the process has exited.</value>
```

Jozkee (approved):
Otherwise, LGTM. Thanks!

PR #8384, opened 2022-09-13
Update [Alg].Create() samples for 3DES/RC2.

### LINKED ISSUE #8140: Improve or remove code examples for TripleDES, DES, and RC2.
We have a number of examples for using these symmetric ciphers that I think either need to be significantly improved, or removed.

1. They demonstrate incorrect use of `CryptoStream`. Ex: https://github.com/dotnet/dotnet-api-docs/blob/13d97c7e536ce822c000883dba873a79d5d6e841/snippets/csharp/System.Security.Cryptography/DES/Create/memoryexample.cs#L99
2. They are incomplete examples. They, for example, do not demonstrate proper nonce management, they do not demonstrate proper authentication (authentication in cryptographic terms). It would be rather hard to demonstrate how to do these things properly in short examples, and nor do I think it's something that should be in API documentation.

The `CryptoStream` issue doesn't apply to `Aes` because we don't have that example for `Aes`, but we do have them for the other symmetric ciphers.

If we want to keep the examples, then they need to be re-written to 1. be correct 2. use modern APIs and 3. be heavily caveated that they are incomplete examples and do not demonstrate safe examples of using symmetric encryption.

/cc @bartonjs @GrabYourPitchforks

bartonjs:
This updates TripleDES/3DES and RC2 to match the changes done for DES in #8227, namely:

* Removes the samples from the Create(string) overload
  * The method is being marked as Obsolete in .NET 7
  * It had a different sample than Create() for C#, but reused the Create() sample in VB and C++/CLI
* Fixes the bug in the sample that assumed CryptoStream.Read aggressively fills the destination.
* Adds using statements (or the language-specific equivalent) for all IDisposable values.
* Uses names for parameters and locals that are consistent with .NET guidelines and best practices.
* Changes C++/CLI to use function prototypes so the functions are presented in the same order across all three languages.
* Changes the method declarations in VB to a more canonical style.

The samples were all copied from their DES versions, but with the following transformations:

3DES:

* s/des/tripleDes/g
* s/DES/TripleDES/g

RC2:

* s/des/rc2/g
* s/a DES/an RC2/g
* s/DES/RC2/g

Fixes #8140.


gewarren (approved):

bartonjs:
@gewarren https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.tripledescryptoserviceprovider?view=net-7.0 is still showing examples from before this change.  Would you happen to know why? :smile:

gewarren:
> @gewarren https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.tripledescryptoserviceprovider?view=net-7.0 is still showing examples from before this change. Would you happen to know why? 😄

@bartonjs It doesn't look like you updated those examples:

snippets/cpp/VS_Snippets_CLR/Cryptography.3DESCSP.CreateEncryptor.File/CPP/fileexample.cpp
snippets/csharp/System.Security.Cryptography/TripleDESCryptoServiceProvider/Overview/fileexample.cs
snippets/visualbasic/VS_Snippets_CLR/Cryptography.3DESCSP.CreateEncryptor.File/VB/fileexample.vb 
  
snippets/cpp/VS_Snippets_CLR/Cryptography.3DESCSP.CreateEncryptor.Memory/CPP/memoryexample.cpp
snippets/csharp/System.Security.Cryptography/TripleDESCryptoServiceProvider/Overview/memoryexample.cs
snippets/visualbasic/VS_Snippets_CLR/Cryptography.3DESCSP.CreateEncryptor.Memory/VB/memoryexample.vb

bartonjs:
Ah.  That would, of course, do it.

PR #8387, opened 2022-09-14
GetFullPath does not throw NotSupportedException in netcore

### LINKED ISSUE #2411: Path.GetFullPath implementation/documentation discrepancy
I encountered an issue this morning with a unit test revolving around `Path.GetFullPath`. I was upgrading unit tests from a .NET Framework 4.6.2 project to a .NET Core 2.2 project, which caused this test to go from passing to failing. I am unsure if this is an implementation discrepancy or a document discrepancy.

According to the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getfullpath?view=netframework-4.6.2) which is labelled for both Framework 4.6.2 and Core 2.2, a call to `Path.GetFullPath` will throw an `NotSupportedException` if the path contains an extra `:` beyond the one used to specify the volume identifier. However, in Core 2.2 this exception does not throw for the unit test in question. 

I attempted to find the source of this issue in the [source code](https://referencesource.microsoft.com/#mscorlib/system/io/path.cs,9d69b1f37ab42ca1,references) for Framework 4.8 and found a possibly pertinent comment stating `:` could be a valid character in an object. In either case I have attached an XUnit snippet with a unit test demonstrating this issue. It passes when run on a 4.6.2 project and fails on a Core 2.2 project since no exception is thrown. 

```cs
public class PathHelperTests
    {
[Theory]
        [InlineData(@"file:\\NetworkShare\NetworkFolder\NetworkFile.avi", @"C:\MyFolder\MyProject.project")]
        public void WhenRelativePathContainsAbsolutePathThrowsNotSupportedException(string relativePath, string relativeTo)
        {

            Exception ex = Assert.Throws<NotSupportedException>(() =>
                Path.GetFullPath(Path.Combine(Path.GetDirectoryName(relativeTo) ?? "", relativePath)));
            Assert.Equal("The given path's format is not supported.", ex.Message);
        }
    }
``` #

plade:
## Summary

GetFullPath does not throw NotSupportedException for invalid paths since Net Core 2.1.
More details are in [this blog post](https://docs.microsoft.com/en-us/archive/blogs/jeremykuhne/system-io-in-net-core-2-1-sneak-peek) and in the issue discussion.

Fixes #2411

### DIFF
--- a/xml/System.IO/Path.xml
+++ b/xml/System.IO/Path.xml
@@ -1404,8 +1404,6 @@ For more information on file path formats on Windows, see [File path formats on 
         <exception cref="T:System.Security.SecurityException">The caller does not have the required permissions.</exception>
         <exception cref="T:System.ArgumentNullException">
           <paramref name="path" /> is <see langword="null" />.</exception>
-        <exception cref="T:System.NotSupportedException">
-          <paramref name="path" /> contains a colon (":") that is not part of a volume identifier (for example, "c:\\").</exception>
         <exception cref="T:System.IO.PathTooLongException">The specified path, file name, or both exceed the system-defined maximum length. </exception>
         <related type="Article" href="/dotnet/standard/io/file-path-formats">File path formats on Windows systems</related>
         <related type="Article" href="/dotnet/standard/io/">File and Stream I/O</related>


gewarren:
Related breaking change doc: https://docs.microsoft.com/en-us/dotnet/core/compatibility/2.1#path-apis-dont-throw-an-exception-for-invalid-characters. Does that need to be updated too @carlossanlop?

gewarren (changes requested):
We do need to leave the exception there for .NET Framework, so instead of removing it, just prefix it with: ".NET Framework only: ...". Or if we still need to mention the .NET Core versions, ".NET Framework and .NET Core 1.0 - 2.0 only: ...".

plade:
I'm not sure what the intended behavior is like. Closing so that someone else can take over.

PR #8408, opened 2022-09-19
Clarify docs for the `Volatile` class with some examples of how it can be used for synchronization

kouvel:
Fixes https://github.com/dotnet/runtime/issues/67330

### DIFF
--- a/xml/System.Threading/Volatile.xml
+++ b/xml/System.Threading/Volatile.xml
@@ -63,7 +63,7 @@
  | 3 | `...` | `int y2 = Volatile.Read(ref y);` |
  | 4 | `...` | `int x2 = x;` |
   
- Even though the volatile write to `y` on thread 1 occurred before the volatile read of `y` on thread 2, thread 2 may still see `y2 == 0` because a volatile operation only restricts reordering in the same thread.
+ Even though the volatile write to `y` on thread 1 occurred before the volatile read of `y` on thread 2, thread 2 may still see `y2 == 0`. The volatile write to `y` does not guarantee that a following volatile read of `y` on a different processor will see the updated value.
   
 > [!NOTE]
 > - Volatile reads and writes ensure that a value is read or written to memory and not cached (for example, in a processor register). Thus, you can use these operations to synchronize access to a field that can be updated by another thread or by hardware.  


xml/System.Threading/Volatile.xml
```diff
@@ -39,14 +39,35 @@
...
+ | 4 | `...` | `int x2 = x;` |
+  
+ Even though the volatile write to `y` on thread 1 occurred before the volatile read of `y` on thread 2, thread 2 may still see `y2 == 0`. The volatile write to `y` does not guarantee that a following volatile read of `y` on a different processor will see the updated value.
   
 > [!NOTE]
-> - On a multiprocessor system, a volatile read operation is not guaranteed to obtain the latest value written to that memory location by any processor. Similarly, a volatile write operation does not guarantee that the value written would be immediately visible to other processors.
-> - On a uniprocessor system, volatile reads and writes ensure that a value is read or written to memory and not cached (for example, in a processor register). Thus, you can use these operations to synchronize access to a field that can be updated by another thread or by hardware.  
+> - Volatile reads and writes ensure that a value is read or written to memory and not cached (for example, in a processor register). Thus, you can use these operations to synchronize access to a field that can be updated by another thread or by hardware.
```
VSadov, line 69: A volatile read may not happen at all if the result is not used, but it may not be worth mentioning as there is no way to take dependency on such reads to happen or not.

VSadov (approved):
LGTM. Thanks!

xml/System.Threading/Volatile.xml
kouvel: It may be observable if the memory is mapped to a device, but perhaps unlikely that an unused read would be used in an observable way. I'll keep it the way it is for now.

andredasilvapinto: Isn't this inconsistent with https://github.com/dotnet/dotnet-api-docs/pull/8408/files#diff-1cd5ee87b602b1035daff1e5a6f8fbe323864ce39dc6d61f13c1d5726f5f5e5dR66 ?

If volatile reads and writes are done to memory and not cached, how is it not guaranteed that a volatile write of y by thread 1 is seen by a subsequent volatile read of y by thread 2? And if the latter is not guaranteed how is it recommended that we use these operations to synchronize access to a field that can be updated by another thread or by hardware?

kouvel: > If volatile reads and writes are done to memory and not cached, how is it not guaranteed that a volatile write of y by thread 1 is seen by a subsequent volatile read of y by thread 2?

A processor's core may still cache a volatile write, so the write may not be immediately visible to another core. Processors typically have some cache consistency that would guarantee that the write would eventually become visible to other cores. So volatile-reading a field in a loop should be sufficient to see changes to the field from elsewhere, it just may not see the changes immediately.

andredasilvapinto: sorry if I am sounding pedantic but this is an important feature of the language, so I think it would be beneficial to have precise documentation:

docs:
> written to memory and **not cached**

your comment:
> A processor's core **may still cache** a volatile write

This seems to be inconsistent.

I also think that if there is a need to read volatile fields in a loop this should be documented and it can't be recommended "to synchronize access to a field that can be updated by another thread". The "loop till the value changes" is a valid strategy, but that approach only works in specific scenarios. It does not work when you simply want to guarantee the use of the latest value for a field, i.e. you don't want to wait till the value changes (as it might even already had changed a few ms before you reach that code).

PR #8416, opened 2022-09-22
Fix JSON ignore conditions

gewarren:


xml/System.Text.Json.Serialization/JsonIgnoreCondition.xml
```diff
@@ -84,7 +84,7 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>Property will only be ignored if it is <see langword="null" />.</summary>
+        <summary>Property is ignored only if it has the default value for its type.</summary>
```
eiriktsarpalis, line 87: Suggested change in wording to better emphasize semantics in case of structs:
```suggestion
        <summary>Property is ignored only if it equals the <see langword="default" /> value for its type.</summary>
```

Might be worth referencing https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values as well, but not sure if we can link to that in the summary section.

eiriktsarpalis (approved):

xml/System.Text.Json.Serialization/JsonIgnoreCondition.xml
gewarren:
```suggestion
        <summary>Property is ignored only if it has the <see href=" https://learn.microsoft.com/dotnet/csharp/language-reference/builtin-types/default-values">default value</see> for its type.</summary>
```

PR #8445, opened 2022-09-30
Add remarks about GC memory info

### LINKED ISSUE #6756: Better explanation of GCMemoryInfo properties
The new GCMemoryInfo added in .NET 5.0 seem very promising, but many of them are not well documented on https://docs.microsoft.com/en-us/dotnet/api/system.gcmemoryinfo?view=net-5.0

For example:

- **MemoryLoadBytes**: `Gets the memory load when the last garbage collection occurred.` That's great, but what *is* the memory load? That's not a common, well-understood term, to my knowledge, and a web search for ".NET memory load" does not explain it.
- **PauseDurations** : `Gets the pause durations. For blocking GCs, there is one pause; for background GC, there are two.` OK, that's something, but why are there 2 for BGCs? What do I do with the 2 TimeSpans? Just add them together? Or do they mean something different? Is this the pause duration only for the one GC at the returned `Index` or all GCs up to and including that one? If it's for all GCs, and I pass in a GCKind is it then the total pause duration for all GCs of that kind? But what about GCKind.Any? (`PauseTimePercentage` is explained well, but I'm not sure if I can assume the same applies to `PauseDurations`.)
- **PromotedBytes**: `Gets the promoted bytes for this GC.` - promoted from which generation to which? When I request GCKind.FullBlocking I get non-zero values, so I guess it must be *to* gen 2? Or does it include bytes promoted from gen 0 to gen 1 as well?
- **TotalCommittedBytes**: `Gets the total committed bytes of the managed heap.` It would be useful to include here what OS metrics this should correspond to on Windows, Linux and macOS, i.e. if I was trying to find the same info from the OS, how would I do that?
- **HighMemoryLoadThresholdBytes**: `Gets the high memory load threshold when the last garbage collection occurred.` What is the "high memory load threshold" - what is the significance of this?
- **Compacted** and **Concurrent**: links to explanations of what these are would be useful, though these are easily findable

gewarren:
Fixes #6756.

### DIFF
--- a/xml/System/GCMemoryInfo.xml
+++ b/xml/System/GCMemoryInfo.xml
@@ -57,7 +57,7 @@ A garbage collection (GC) is identified by its <xref:System.GCMemoryInfo.Index>,
         <summary>Gets a value that indicates if this is a compacting GC or not.</summary>
         <value>
           <see langword="true" /> if this is a compacting GC; <see langword="false" /> otherwise.</value>
-        <remarks>A compacting GC is when space occupied by dead objects is reclaimed and surviving objects are compacted. The compacting phase moves objects that have survived a garbage collection toward the older end of the segment.</remarks>
+        <remarks>A compacting GC is when space occupied by dead objects is reclaimed and surviving objects are compacted.</remarks>
       </Docs>
     </Member>
     <Member MemberName="Concurrent">
@@ -244,7 +244,7 @@ The memory after `OBJ_D` is not considered part of the `FragmentedBytes` but is 
       <Docs>
         <summary>Gets the high memory load threshold when the last garbage collection occurred.</summary>
         <value>The high memory load threshold, in bytes, when the last garbage collection occurred.</value>
-        <remarks>Most of the GC performance heuristics are based on per-process measurements. However, to avoid paging, GC is aware of the global physical memory load on the machine. GC recognizes a certain memory load percentage as a "high memory load situation". When the memory load percentage is over that threshold, GC becomes more aggressive and will do more full blocking GCs if necessary to reduce the heap size.</remarks>
+        <remarks>Most of the GC performance heuristics are based on per-process measurements. However, to avoid paging, GC is aware of the global physical memory load on the machine or VM or in the container. GC recognizes a certain memory load percentage as a "high memory load situation". When the memory load percentage is over that threshold, GC becomes more aggressive and will do more full blocking GCs if necessary to reduce the heap size.</remarks>
       </Docs>
     </Member>
     <Member MemberName="Index">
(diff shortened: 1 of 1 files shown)


xml/System/GCMemoryInfo.xml
```diff
@@ -57,7 +57,7 @@ A garbage collection (GC) is identified by its <xref:System.GCMemoryInfo.Index>,
         <summary>Gets a value that indicates if this is a compacting GC or not.</summary>
         <value>
           <see langword="true" /> if this is a compacting GC; <see langword="false" /> otherwise.</value>
-        <remarks>To be added.</remarks>
+        <remarks>A compacting GC is when space occupied by dead objects is reclaimed and surviving objects are compacted. The compacting phase moves objects that have survived a garbage collection toward the older end of the segment.</remarks>
```
Maoni0, line 60:
```suggestion
        <remarks>A compacting GC is when space occupied by dead objects is reclaimed and surviving objects are compacted.</remarks>
```

Maoni0: I would not describe the detail here as it may not be the same segment

xml/System/GCMemoryInfo.xml
```diff
@@ -243,7 +244,7 @@ The memory after `OBJ_D` is not considered part of the `FragmentedBytes` but is
       <Docs>
         <summary>Gets the high memory load threshold when the last garbage collection occurred.</summary>
         <value>The high memory load threshold, in bytes, when the last garbage collection occurred.</value>
-        <remarks>To be added.</remarks>
+        <remarks>Most of the GC performance heuristics are based on per-process measurements. However, to avoid paging, GC is aware of the global physical memory load on the machine. GC recognizes a certain memory load percentage as a "high memory load situation". When the memory load percentage is over that threshold, GC becomes more aggressive and will do more full blocking GCs if necessary to reduce the heap size.</remarks>
```
Maoni0, line 247:
```suggestion
        <remarks>Most of the GC performance heuristics are based on per-process measurements. However, to avoid paging, GC is aware of the global physical memory load on the machine or VM or in the container. GC recognizes a certain memory load percentage as a "high memory load situation". When the memory load percentage is over that threshold, GC becomes more aggressive and will do more full blocking GCs if necessary to reduce the heap size.</remarks>
```

xml/System/GCMemoryInfo.xml
```diff
@@ -292,7 +293,8 @@ The memory after `OBJ_D` is not considered part of the `FragmentedBytes` but is
       <Docs>
         <summary>Gets the physical memory load when the last garbage collection occurred.</summary>
         <value>The physical memory load, in bytes, when the last garbage collection occurred.</value>
-        <remarks>Data is only brought into physical memory on first touch. If you allocated a big object but haven't actually used it, most of its memory isn't in physical memory. In this case, the allocation won't affect the memory load significantly.</remarks>
+        <remarks><para>Memory load is a field in the [MEMORYSTATUS structure](/windows/win32/api/winbase/ns-winbase-memorystatus). It's a number between 0 and 100 that specifies the approximate percentage of physical memory that's in use (0 indicates no memory use and 100 indicates full memory use).</para>
```
Maoni0, line 296:
```suggestion
        <remarks><para>On Windows memory load is a field in the [MEMORYSTATUS structure](/windows/win32/api/winbase/ns-winbase-memorystatus). It's a number between 0 and 100 that specifies the approximate percentage of physical memory that's in use (0 indicates no memory use and 100 indicates full memory use). Corresponding info is given on other OSs.</para>
```

xml/System/GCMemoryInfo.xml
```diff
@@ -313,9 +315,9 @@ The memory after `OBJ_D` is not considered part of the `FragmentedBytes` but is
         <ReturnType>System.ReadOnlySpan&lt;System.TimeSpan&gt;</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets the pause durations. For blocking GCs, there is one pause; for background GC, there are two.</summary>
+        <summary>Gets the durations of the pauses when all managed threads are suspended.</summary>
         <value>A timespan representing the pause durations.</value>
-        <remarks>To be added.</remarks>
+        <remarks> For blocking GCs, there is one pause; for background GC, there are two (they occur when the ephemeral generations 0 and 1 are collected).</remarks>
```
Maoni0, line 320: it actually doesn't include the ephemeral pauses at all. we just do an initial pause to figure out a bit of info like what gen to collect and etc. and later another pause.

Maoni0, line 320:
```suggestion
        <remarks> For blocking GCs, there is one pause; for background GC, the pauses are described here: for background GC it's explained here: https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/background-gc#background-workstation-vs-server-gc
.</remarks>
```

Maoni0 (approved):
thanks!

PR #8450, opened 2022-10-03
Add remark about MaxConnectionsPerServer clarifying word 'server'

### LINKED ISSUE #8166: Can we clarify what is a 'server'?
When speaking about the max connections per server, can we clarify what is a server to the SocketsHttpHandler? Is it an IP address, domain, TLD? As an example, if you are using this to connect to an api at api.contoso.com, and api.contoso.com is served with DNS-RR and has two IPs, would the MaxConnectionsPerServer setting be applied to api.contoso.com, each individual IP, or contoso.com?

https://github.com/dotnet/dotnet-api-docs/blob/d532baa9c9f8a759696337a1f6fd61e9723c47a3/xml/System.Net.Http/SocketsHttpHandler.xml#L589-L613

rzikm:
Closes #8166.

I added information about what we consider a unique Server based on the HttpConnectionPoolManager.GetConnectionKey

https://github.com/dotnet/runtime/blob/a08d30012999806f2ee4b0c346bdaca0002d5993/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionPoolManager.cs#L259

### DIFF
--- a/xml/System.Net.Http/SocketsHttpHandler.xml
+++ b/xml/System.Net.Http/SocketsHttpHandler.xml
@@ -608,7 +608,18 @@ Timeout must be greater than or equal to 1 second. Set to <xref:System.Threading
       <Docs>
         <summary>Gets or sets the maximum number of simultaneous TCP connections allowed to a single server.</summary>
         <value>The maximum number of simultaneous TCP connections allowed to a single server.</value>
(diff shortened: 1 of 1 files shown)


ManickaP (approved):
Thanks!

PR #8451, opened 2022-10-03
Add annotations for UriFormatException in HttpClient methods

### LINKED ISSUE #4201: HttpClient.GetAsync can throw UriFormatException
There are 4 overloads of GetAsync that take a string and convert it to a URI in the implementation. All 4 of these should have UriFormatException documented as a possible exception:
```
GetAsync(String)
GetAsync(String, HttpCompletionOption)
GetAsync(String, CancellationToken)
GetAsync(String, HttpCompletionOption, CancellationToken)
```

rzikm:
Fixes #4201.

I fixed it in (hopefully all) methods which accept `string requestUri` parameter.


ManickaP (approved):
Thanks!

PR #8455, opened 2022-10-04
Mention that RemoteCertificate is not disposed when accessed

### LINKED ISSUE #4872: SslStream.Dispose incorrectly states that unmanaged resources are released
If the remote certificate has been accessed from the SslStream.RemoteCertificate property, SslStream does not cleanup the X509Certificate2 instance when it is disposed. There should be language to say that the caller of this property is responsible for disposing the certificate, either on the Dispose documentation or the RemoteCertificate documentation. It needs to be carefully worded as you only get the one instance, if you Dispose it prematurely, the certificate returned from RemoteCertificate will be the same disposed instance.

rzikm:
Fixes #4872.

I did not find a section for (parameterless) SslStream.Dispose method, so I added the mention only to the property.

### DIFF
--- a/xml/System.Net.Security/SslStream.xml
+++ b/xml/System.Net.Security/SslStream.xml
@@ -4318,7 +4318,7 @@ This property gets the cipher suite that is going to be used in the communicatio
         <value>An X509Certificate object that represents the certificate supplied for authentication or <see langword="null" /> if no certificate was supplied.</value>
         <remarks>
           <format type="text/markdown"><![CDATA[  
-If the property is accessed, the remote certificate will not be disposed when the SslStream instance is disposed. The caller of the property is responsible for disposing the returned <xref:System.Security.Cryptography.X509Certificate> object.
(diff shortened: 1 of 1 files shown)


xml/System.Net.Security/SslStream.xml
```diff
@@ -4318,6 +4318,7 @@ This property gets the cipher suite that is going to be used in the communicatio
         <value>An X509Certificate object that represents the certificate supplied for authentication or <see langword="null" /> if no certificate was supplied.</value>
         <remarks>
           <format type="text/markdown"><![CDATA[  
+If the property is accessed, the remote certificate will not be disposed when the SslStream instance is disposed. The caller of the property is responsible for disposing the returned X509Certificate object.
```
ManickaP, line 4321: Should the reference to X509Certificate be either a reference or at least formatted as code?

ManickaP (approved):
Small NIT, otherwise LGTM, thanks!

PR #8456, opened 2022-10-04
Update Timer.xml

tohoit02:
When providing the period value as a TimeSpan in ticks, the value must represent at least 1 millisecond. Any value below will be rounded down to zero (0) and the callback is called only once.

## Summary

Added one line for periode parameter as TimeSpan in ticks, which is rounded down if the value is calculated to be less 1ms.


tohoit02:
Sorry that this merge took so long :)

PR #8460, opened 2022-10-05
Make sure TcpClient is disposed in connect examples

### LINKED ISSUE #2023: TcpClient not disposing
The example in this page: https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.tcpclient?view=netframework-4.7.2

Is calling `stream.Close` and `client.Close` inside the `try`. Between the client being opened and the `Close` method being called, an exception can occur, leaving the socket open. There should be a `using` directive to ensure correct disposal of the client and stream.

The same thing happens in the `TcpListener` example in https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.tcplistener?view=netframework-4.7.2. Neither `TcpClient` nor `NetworkStream` are being correctly disposed of.

rzikm:
Fixes #2023

### DIFF
--- a/snippets/cpp/VS_Snippets_Remoting/System.Net.Sockets.TcpClient/CPP/tcpclient.cpp
+++ b/snippets/cpp/VS_Snippets_Remoting/System.Net.Sockets.TcpClient/CPP/tcpclient.cpp
@@ -40,7 +40,7 @@ void Connect( String^ server, String^ message )
       // connected to the same address as specified by the server, port
       // combination.
       Int32 port = 13000;
-      client = gcnew TcpClient( server,port );
+      client = gcnew TcpClient(server, port);
       
       // Translate the passed message into ASCII and store it as a Byte array.
       array<Byte>^data = Text::Encoding::ASCII->GetBytes( message );


snippets/cpp/VS_Snippets_Remoting/System.Net.Sockets.TcpClient/CPP/tcpclient.cpp
```diff
@@ -32,29 +32,28 @@ using namespace System::Security::Permissions;
...
+   TcpClient^ client = nullptr;
    try
    {
       // Create a TcpClient.
       // Note, for this client to work you need to have a TcpServer 
       // connected to the same address as specified by the server, port
       // combination.
       Int32 port = 13000;
-      TcpClient^ client = gcnew TcpClient( server,port );
+      client = gcnew TcpClient( server,port );
```
karelz, line 43:
```suggestion
      client = gcnew TcpClient(server, port);
```

karelz (approved):

gewarren (approved):

rzikm:
@gewarren Looks like there is one more project file missing, can you please have a look?

gewarren:
> @gewarren Looks like there is one more project file missing, can you please have a look?

I'm not sure how to satisfy Snippets 5000 by creating a project file for the C++ files, so I expected that error and it's okay.

snippets/csharp/System.Net.Sockets/TcpClient/Overview/tcpclient.cs
```diff
@@ -65,22 +65,22 @@ static void Connect(String server, String message)
       // connected to the same address as specified by the server, port
       // combination.
       Int32 port = 13000;
-      TcpClient client = new TcpClient(server, port);
+
+      // Prefer using declaration to ensure the instance is Disposed later.
```
stephentoub, line 69: > Prefer using declaration

"Prefer a using declaration"... otherwise it sounds like we should prefer to use a declaration rather than we should prefer to use a using declaration.

PR #8462, opened 2022-10-05
Fix note on TcpClient constructor about IPv4

### LINKED ISSUE #8075: `TcpClient` constructor docs misleading
[TcpClient Constructors docs](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.tcpclient.-ctor?f1url=%3FappId%3DDev16IDEF1%26l%3DZH-CN%26k%3Dk(System.Net.Sockets.TcpClient.%2523ctor)%3Bk(DevLang-csharp)%26rd%3Dtrue&view=net-5.0#system-net-sockets-tcpclient-ctor)

>  Note
>  This constructor works only with IPv4 address types.

## But in source code, it's not only with IPv4 address types.

Let's look at its implementation in .NET 5.

https://github.com/dotnet/runtime/blob/6a984143635bde23e728abaaccbde52f5ea8fa3e/src/libraries/System.Net.Sockets/src/System/Net/Sockets/TCPClient.cs#L26-L28

https://github.com/dotnet/runtime/blob/6a984143635bde23e728abaaccbde52f5ea8fa3e/src/libraries/System.Net.Sockets/src/System/Net/Sockets/TCPClient.cs#L31-L45

https://github.com/dotnet/runtime/blob/6a984143635bde23e728abaaccbde52f5ea8fa3e/src/libraries/System.Net.Sockets/src/System/Net/Sockets/TCPClient.cs#L412-L428

> // **If AF was not explicitly set try to initialize dual mode socket or fall-back to IPv4.**
> _clientSocket = new Socket(SocketType.Stream, ProtocolType.Tcp);

https://github.com/dotnet/runtime/blob/6a984143635bde23e728abaaccbde52f5ea8fa3e/src/libraries/System.Net.Sockets/src/System/Net/Sockets/Socket.cs#L77-L84

rzikm:
Fixes #8075.

Behavior changed by https://github.com/dotnet/corefx/pull/30036. So I assume it is enough to specify that the note is only valid for .NET Framework.

### DIFF
--- a/xml/System.Net.Sockets/TcpClient.xml
+++ b/xml/System.Net.Sockets/TcpClient.xml
@@ -131,7 +131,7 @@
  This constructor creates a new <xref:System.Net.Sockets.TcpClient> and allows the underlying service provider to assign the most appropriate local IP address and port number. You must first call the <xref:System.Net.Sockets.TcpClient.Connect%2A> method before sending and receiving data.  
   
 > [!NOTE]
->  This constructor works only with IPv4 address types.  
+>  On .NET Framework, this constructor works only with IPv4 address types.  
   
 > [!NOTE]
 >  This member outputs trace information when you enable network tracing in your application. For more information, see [Network Tracing in the .NET Framework](/dotnet/framework/network-programming/network-tracing).  


liveans (approved):
LGTM

PR #8463, opened 2022-10-05
Clarify remark on DirectoryInfo.Create (#8212)

### LINKED ISSUE #8212: DirectoryInfo.Create() can create intermediate directories
I think it'd be nice to mention that you can call `DirectoryInfo.Create()` to create multiple nested directories in this block:

https://github.com/dotnet/dotnet-api-docs/blob/289fe0b3d1974b15f5522b12588599912a650d25/xml/System.IO/DirectoryInfo.xml#L248-L254

amd-9:
Fixes #8212

### DIFF
--- a/xml/System.IO/DirectoryInfo.xml
+++ b/xml/System.IO/DirectoryInfo.xml
@@ -247,11 +247,7 @@
   
 ## Remarks  
 
- If the directory already exists, this method does nothing.  
-
- If the directory did not exist before calling this method, then any cached attribute information about the directory will be flushed if the creation is successful.  
-
- If the directory path contains non-existent intermediate directories they will be created also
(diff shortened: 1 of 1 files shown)


xml/System.IO/DirectoryInfo.xml
```diff
@@ -250,6 +250,8 @@
  If the directory already exists, this method does nothing.  
 
  If the directory did not exist before calling this method, then any cached attribute information about the directory will be flushed if the creation is successful.  
+
+ If the directory path contains non-existent intermediate directories they will be created also
```
Jozkee, line 254: We already have a wording for this in DirectoryInfo.CreateSubdirectory(String) 
> Any and all directories specified in path are created, unless some part of path is invalid. The path parameter specifies a directory path, not a file path. If the subdirectory already exists, this method does nothing.

and Directory.CreateDirectory(String)
> Any and all directories specified in path are created, unless they already exist or unless some part of path is invalid. If the directory already exists, this method does not create a new directory, but it returns a [DirectoryInfo](https://learn.microsoft.com/en-us/dotnet/api/system.io.directoryinfo?view=net-6.0) object for the existing directory.


This description should be aligned to those two descriptions of the same.

Jozkee (approved):

PR #8464, opened 2022-10-05
Fix api documentation for System.Runtime.Intrinsics

dakersnar:
Fixes https://github.com/dotnet/runtime/issues/72402

### DIFF
--- a/xml/System.Numerics/Vector`1.xml
+++ b/xml/System.Numerics/Vector`1.xml
@@ -67,21 +67,8 @@
   
  The <xref:System.Numerics.Vector%601> structure provides support for hardware acceleration.
  
- The term *primitive numeric data type* in this article refers to numeric data types that are directly supported by the CPU and have instructions that can manipulate those data types. The following table shows which primitive numeric data type and operation combination uses intrinsic instructions for faster execution:
+ The term *primitive numeric data type* in this article refers to numeric data types that are directly supported by the CPU and have instructions that can manipulate those data types.
  
-| Primitive type | `+` | `-` | `*` | `/` |
-| --- | :---: | :---: | :---: | :---: |
-|`sbyte` | Yes | Yes | No | No |
-|`byte` | Yes | Yes | No | No |
-|`short` | Yes | Yes | Yes | No |
-|`ushort` | Yes | Yes | No | No |
-| `int` | Yes | Yes | Yes | No |
-| `uint` | Yes | Yes | No | No |
-| `long` | Yes | Yes | No | No |
-| `ulong` | Yes | Yes | No | No |
-| `float` | Yes | Yes | Yes | Yes |
-| `double` | Yes | Yes | Yes | Yes |
-
  ]]></format>
     </remarks>
   </Docs>


xml/System.Runtime.Intrinsics/Vector128`1.xml
```diff
@@ -37,8 +37,8 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <typeparam name="T">To be added.</typeparam>
-    <summary>To be added.</summary>
+    <typeparam name="T">The vector type. <c>T</c> can be any primitive numeric type.</typeparam>
```
gewarren, line 40:
```suggestion
    <typeparam name="T">The vector type. <typeparamref name="T"> can be any primitive numeric type.</typeparam>
```

xml/System.Runtime.Intrinsics/Vector256`1.xml
```diff
@@ -37,8 +37,8 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <typeparam name="T">To be added.</typeparam>
-    <summary>To be added.</summary>
+    <typeparam name="T">The vector type. <c>T</c> can be any primitive numeric type.</typeparam>
```
gewarren, line 40:
```suggestion
    <typeparam name="T">The vector type. <typeparam name="T"> can be any primitive numeric type.</typeparam>
```

xml/System.Runtime.Intrinsics/Vector64`1.xml
```diff
@@ -37,8 +37,8 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <typeparam name="T">To be added.</typeparam>
-    <summary>To be added.</summary>
+    <typeparam name="T">The vector type. <c>T</c> can be any primitive numeric type.</typeparam>
```
gewarren, line 40:
```suggestion
    <typeparam name="T">The vector type. <typeparam name="T"> can be any primitive numeric type.</typeparam>
```

gewarren (approved):

xml/System.Runtime.Intrinsics/Vector256`1.xml
dakersnar: I assume you mean "typeparamref" here too, right? And below.

dakersnar: Also, these should have a slash at the end, right? Like `<typeparamref name="T" />`

gewarren: Oh yes, sorry about that.

gewarren: I updated the suggestions.

gewarren: > Also, these should have a slash at the end, right? Like `<typeparamref name="T" />`

Yes, good catch.

dakersnar:
@gewarren I added a commit with your suggestions (with an extra backslash that my IDE was complaining about the lack of). I also changed Vector_1.xml in the same way for consistency.

gewarren:
> @gewarren I added a commit with your suggestions (with an extra slash that my IDE was complaining about the lack of). I also changed Vector_1.xml in the same way for consistency.

Cool! Which IDE are you using for that functionality?

dakersnar:
VSCode. I'm not an XML expert, but I think it is complaining that `</typeparam>`was closing `<typeparamref name="T">`. It couldn't quite connect the dots to what the real problem was, but adding the slash in `<typeparamref name="T">` fixed it.

image

xml/System.Runtime.Intrinsics/Vector128`1.xml
```diff
@@ -37,8 +37,8 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <typeparam name="T">To be added.</typeparam>
-    <summary>To be added.</summary>
+    <<typeparam name="T">>The vector type. <typeparamref name="T" /> can be any primitive numeric type.</typeparam>
```
carlossanlop, line 40:
```suggestion
    <typeparam name="T">The vector type. <typeparamref name="T" /> can be any primitive numeric type.</typeparam>
```

gewarren:
@dakersnar I see you were updating the /// in the source code. I found an extension (XML by Red Hat) to do similar validation in the XML file itself. Thanks for the idea!

image

xml/System.Runtime.Intrinsics/Vector128.xml
```diff
@@ -18,7 +18,7 @@
   </Base>
   <Interfaces />
   <Docs>
-    <summary>To be added.</summary>
+    <summary>Provides a collection of static convenience methods for creating, manipulating, combining, and converting generic 128-bit vectors.</summary>
```
tannergooding, line 21: Same as https://github.com/dotnet/runtime/pull/76687

Combining and converting are a small subset of the total operations supported. Maybe generalizing this as the following would be better:
> Provides a collection of static methods for creating, manipulating, and otherwise operating on 128-bit vectors.

xml/System.Numerics/Vector`1.xml
```diff
@@ -57,7 +57,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <typeparam name="T">The vector type. <typeparamref name="T" /> can be any primitive numeric type.</typeparam>
+    <typeparam name="T">The type of the elements in the vector. <typeparamref name="T" /> can be any primitive numeric type.</typeparam>
```
dakersnar, line 60: @tannergooding FYI the `remarks` for this specific case defines what they mean by "primitive numeric type". I removed the term from the other files that don't have that context.

tannergooding: Would probably be good to update the remarks to cover `nint` and `nuint`.

The general table there is also out of date since its really only covering the semantics around `x86/x64` and not other platforms such as `Arm64`.

dakersnar: Would removing the table be the best course of action in that case (see my new commit)? The rest of the remarks section is general enough to apply to both architectures. I suppose it depends on how easy it would be for users to figure out those types with other sources.

carlossanlop: You can remove the remarks table in this PR and resubmit it without urgency on a later date.

Remarks do not show up in intellisense, so they aren't urgent for getting them merged today.

dakersnar: Sounds good. I'll merge this now.

PR #8470, opened 2022-10-06
Document when async HttpListener send completes

### LINKED ISSUE #2068: Document when async HttpListener send completes
When you asynchronously send a response to a request to a `HttpListener`, the docs don't seem to indicate how to tell when the response has been sent. Part of the problem seems that async was only partially implemented for this class. That makes it all the more important that the docs do their best to help you navigate through.

Here are some key questions that are unanswered:
1. Does `HttpListener.Stop` block to flush any response being sent?
2. How can I be notified when `HttpListenerResponse.Close(Byte[], Boolean)` completes when `willBlock` is false?
3. If there's no answer to 2, how can I asynchronously close `HttpListener` such that any async response gets sent?

rzikm:
Fixes #2068

### DIFF
--- a/xml/System.Net/HttpListenerResponse.xml
+++ b/xml/System.Net/HttpListenerResponse.xml
@@ -432,10 +432,10 @@
   
  This method closes the response stream and the <xref:System.Net.HttpListenerRequest> associated with the response.
 
- If `false` is specified as the `willBlock` parameter is specified, the method returns immediately and the connection is closed asynchronously once the data is sent.
+ If `false` is specified as the `willBlock`, the method returns immediately and the connection is closed asynchronously once the data is sent.
 
 > [!NOTE]
->  If closing asynchronously, it is not possible for the caller to determine when all data have been sent. In addition to that, disposing the <xref:System.Net.HttpListenerResponse>, <xref:System.Net.HttpListenerContext> or <xref:System.Net.HttpListener> may close the connection before the final data is sent.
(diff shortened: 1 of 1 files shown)


xml/System.Net/HttpListenerResponse.xml
```diff
@@ -430,9 +430,12 @@
...
   
- This method closes the response stream and the <xref:System.Net.HttpListenerRequest> associated with the response.  
-  
-   
+ This method closes the response stream and the <xref:System.Net.HttpListenerRequest> associated with the response.
+
+ If `false` is specified as the `willBlock` parameter is specified, the method returns immediately and the connection is closed asynchronously once the data is sent.
+
+> [!NOTE]
+>  If closing asynchronously, it is not possible for the caller to determine when all data have been sent. In addition to that, disposing the <xref:System.Net.HttpListenerResponse>, <xref:System.Net.HttpListenerContext> or <xref:System.Net.HttpListener> may close the connection before the final data is sent.
```
ManickaP, line 438: I'm not a native speaker, so this suggestion might be completely wrong :smile: 
```suggestion
>  If closing asynchronously, it is not possible for the caller to determine when the data have been sent. In addition to that, disposing the <xref:System.Net.HttpListenerResponse>, <xref:System.Net.HttpListenerContext> or <xref:System.Net.HttpListener> may close the connection before all the data are sent.
```

ManickaP (approved):
Small wording suggestions.

PR #8474, opened 2022-10-07
Fix up boilerplate text for Dispose/Finalize

### LINKED ISSUE #8333: HttpClient.Dispose documentation says it is invoked by Finalize but the code does not support this statement
https://github.com/dotnet/dotnet-api-docs/blame/main/xml/System.Net.Http/HttpClient.xml#L922
image

None of the implementations I can find in Framework or Core support this statement.  As far as I can see there are no explicit finalizers in the code at all.

https://github.com/microsoft/referencesource/blob/master/System/net/System/Net/Http/HttpClient.cs

The statement may well have once been true (I can't find out when that might have been) but it doesn't look like it's true today.

gewarren:
Fixes #8333.

Notes that `Dispose(Boolean)` is only called by `Finalize` if it's been overridden.

### DIFF
--- a/xml/System.CodeDom.Compiler/TempFileCollection.xml
+++ b/xml/System.CodeDom.Compiler/TempFileCollection.xml
@@ -553,7 +553,7 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks
- This method is called by the public `Dispose()` method and the <xref:System.Object.Finalize> method, if it has been overridden. `Dispose()` invokes the protected `Dispose(Boolean)` method with the `disposing` parameter set to `true`. `Finalize` invokes `Dispose` with `disposing` set to `false`.
+ This method is called by the public `Dispose()` method and the <xref:System.Object.Finalize> method, if it has been overridden. `Dispose()` invokes this method with the `disposing` parameter set to `true`. `Finalize` invokes this method with `disposing` set to `false`.
   
  When the `disposing` parameter is `true`, this method releases all resources held by any managed objects that this <xref:System.CodeDom.Compiler.TempFileCollection> references. This method invokes the `Dispose()` method of each referenced object.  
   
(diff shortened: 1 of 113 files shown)


antonfirsov (approved):

xml/System.CodeDom.Compiler/TempFileCollection.xml
```diff
@@ -552,8 +552,8 @@
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
-## Remarks  
- This method is called by the public `Dispose()` method and the <xref:System.Object.Finalize%2A> method. `Dispose()` invokes the protected `Dispose(Boolean)` method with the `disposing` parameter set to `true`. <xref:System.Object.Finalize%2A> invokes `Dispose` with `disposing` set to `false`.  
+## Remarks
+ This method is called by the public `Dispose()` method and the <xref:System.Object.Finalize> method, if it has been overridden. `Dispose()` invokes this method with the `disposing` parameter set to `true`. `Finalize` invokes this method with `disposing` set to `false`.
```
stephentoub, line 556: This remark isn't generic boilerplate but is actually the text on the TempFileCollection page.  We know whether TempFileCollection implements a finalizer (it does).  Seems like it would make sense to specialize this with that knowledge?  Same for the rest of the occurrences here where we know whether it does or doesn't implement a finalizer.

antonfirsov: Do you mean specializing this general text in both cases? (when finalizer is overridden VS when it's missing)

What if we change the implementation (delete or add a finalizer)? Are we sure we have the processes, the awareness and the capacity to always notice such a change and change the documentation accordingly? Also, I would find a specialized text confusing, in case the finalizer is *missing*. `Dispose(false)` is never invoked from the built-in hierarchy, that case only makes sense if a subclass overrides the finalizer. We would need to properly explain this in the remarks section.

IMO the boilerplate text is a safer option if we are not sure if we can address these concerns.

PR #8483, opened 2022-10-10
Dispose of WebClient and Stream in examples

### LINKED ISSUE #2232: Miss using to dispose stream
@lixiaoqiang commented on [Tue Apr 02 2019](https://github.com/dotnet/dotnet-api-docs.zh-cn/issues/24)

文档
https://github.com/dotnet/dotnet-api-docs.zh-cn/blob/live/xml/System.Net/WebClient.xml
中的示例
~/samples/snippets/csharp/VS_Snippets_Remoting/NCLWebClientUserAgent/CS/useragent.cs#1

没有dispose掉WebClient OpenRead 生成的Stream以及WebClient本身。

rzikm:
Fixes #2232

### DIFF
--- a/snippets/cpp/VS_Snippets_Remoting/NCLWebClientUserAgent/CPP/useragent.cpp
+++ b/snippets/cpp/VS_Snippets_Remoting/NCLWebClientUserAgent/CPP/useragent.cpp
@@ -25,6 +25,7 @@ int main()
    Console::WriteLine( s );
    data->Close();
    reader->Close();
+   delete client;
 }
 
 //</Snippet1>
--- a/snippets/csharp/System.Net/WebClient/Overview/useragent.cs
+++ b/snippets/csharp/System.Net/WebClient/Overview/useragent.cs
@@ -5,25 +5,25 @@ using System.IO;
 
 public class Test
 {
-    public static void Main (string[] args)
+    public static void Main(string[] args)
     {
         if (args == null || args.Length == 0)
         {
-            throw new ApplicationException ("Specify the URI of the resource to retrieve.");
+            throw new ApplicationException("Specify the URI of the resource to retrieve.");
         }
-        WebClient client = new WebClient ();
+        using WebClient client = new WebClient();
 
         // Add a user agent header in case the
         // requested URI contains a query.
 
-        client.Headers.Add ("user-agent", "Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.2; .NET CLR 1.0.3705;)");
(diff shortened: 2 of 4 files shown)


snippets/cpp/VS_Snippets_Remoting/NCLWebClientUserAgent/CPP/useragent.cpp
```diff
@@ -25,6 +25,7 @@ int main()
    Console::WriteLine( s );
    data->Close();
    reader->Close();
+   delete client;
```
liveans, line 28: I would expect that, if we're creating an object instance via `gcnew`, garbage collector should be responsible for deleting this object and it should cause call on `Dispose` member function as well. Since, I'm not really aware of C++CLI, I have doubts on it.

liveans (approved):

snippets/csharp/System.Net/WebClient/Overview/useragent.cs
```diff
@@ -5,25 +5,25 @@
...
-        StreamReader reader = new StreamReader (data);
-        string s = reader.ReadToEnd ();
-        Console.WriteLine (s);
-        data.Close ();
-        reader.Close ();
+        using Stream data = client.OpenRead(args[0]);
+        StreamReader reader = new StreamReader(data);
+        string s = reader.ReadToEnd();
+        Console.WriteLine(s);
+        data.Close();
```
stephentoub, line 25: You added a `using` for data... why is the data.Close() still needed?  Also, why add a using for the Stream but not for the StreamReader?  The StreamReader will itself by default close the underlying stream, so this could instead just be:
```C#
using StreamReader reader = new StreamReader(client.OpenRead(args[0]));
```
and then you don't need the line above or either of the .Close calls.

snippets/visualbasic/VS_Snippets_Remoting/NCLWebClientUserAgent/VB/useragent.vb
```diff
@@ -10,18 +10,19 @@ Public Class Test
...
+            ' Add a user agent header in case the 
+            ' requested URI contains a query.
+            client.Headers.Add("user-agent", "Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.2; .NET CLR 1.0.3705;)")
+            
+            Dim data As Stream = client.OpenRead(args(0))
+            Dim reader As New StreamReader(data)
+            Dim s As String = reader.ReadToEnd()
+            Console.WriteLine(s)
+            data.Close()
+            reader.Close()
```
stephentoub, line 24: Same questions here.

PR #8486, opened 2022-10-10
Add using to NetworkStream/WriteTimeout examples

### LINKED ISSUE #1897: Both TcpClient and NetworkStream implement IDisposable, so why not use using block in documentation examples?
I was wondering why dont you use `using` block in the example considering that both `TcpClient` and `NetworkStream` implement `IDisposable`.  None of your documentation on this subject is using "using" blocks which should be used if IDisposable is implemented.  Here is just one example [https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.networkstream.writetimeout?view=netframework-4.7.2#System_Net_Sockets_NetworkStream_WriteTimeout](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.networkstream.writetimeout?view=netframework-4.7.2#System_Net_Sockets_NetworkStream_WriteTimeout), but there are many more.  **The documentation seam to be very basic and full of miss-guiding information.**

The other question I have  is that your documentation shows that you read data from the stream **right after** writing data to stream.  Is this data going to be available for reading right after you write data to stream though?

I have below code sample with my question capitalized, it would be nice someone can provide some explanation.  

```
using (TcpClient client = new TcpClient())  // WHY NONE OF YOUR EXAMPLES USES USING BLOCK LIKE THIS?
{
	client.Connect(ip, port);

	using (NetworkStream stream = client.GetStream())  // WHY NONE OF YOUR EXAMPLES USES USING BLOCK LIKE THIS?
	{
		byte[] messageBytes = ...;
		if (stream.CanWrite)  //WHAT IS THE POINT OF THIS IF CanWrite IS SET BY DEFAULT TO TRUE?
		{
			await stream.WriteAsync(messageBytes, 0, messageBytes.Length);
			stream.FlushAsync();
		}

		// IS THIS DATA GOING TO BE AVAILABLE AT THIS POINT, RIGHT AFTER BEING WRITTEN?  
		// IF NOT, THEN YOU WILL HAVE NO OPPORTUNITY TO READ DATA 
		// SINCE NOTHING IS GOING TO TELL YOU WHEN IT IS GOING TO BE AVAILABLE.  
		// BY THE TIME IT IS AVAILABLE, THE EXECUTION MAY PASS THIS POINT, WHAT THEN???
                // IN FACT, BOTH CanWrite AND CanRead ARE SET TO DEFAULT VALUE OF TRUE ON
                // NETWORKSTREAM INITIALIZATION, SO WHAT IS EVEN POINT OF HAVING THESE
                // TWO PROPERTIES?
		if (stream.CanRead)  //WHAT IS THE POINT OF THIS IF CanRead IS SET BY DEFAULT TO TRUE?
		{
			byte[] buffer = newe byte[1024];
			int readSoFar = 0;
			StringBuilder builder = new StringBuilder();

			while(stream.DataAwailable)
			{
				readSoFar = await stream.ReadAsync(buffer, 0, buffer.Length);
				builder.AppendFormat("{0}", Encoding.ASCII.GetString(buffer, 0, readSoFar));
			}

			string msg = builder.ToString();
		}
	}
}
```

liveans:
Fixes #1897

### DIFF
--- a/snippets/cpp/VS_Snippets_Remoting/NclTcpServerSync/cpp/tcplistener.cpp
+++ b/snippets/cpp/VS_Snippets_Remoting/NclTcpServerSync/cpp/tcplistener.cpp
@@ -35,6 +35,5 @@ int main()
     // Dispose allocated resources.
     delete networkStream;
     delete tcpClient;
-    delete tcpListener;
 }
 //</snippet0>
--- a/snippets/csharp/System.Net.Sockets/NetworkStream/WriteTimeout/tcplistener.cs
+++ b/snippets/csharp/System.Net.Sockets/NetworkStream/WriteTimeout/tcplistener.cs
@@ -12,7 +12,7 @@ namespace Examples.System.Net
         {
             // Create the server side connection and
             // start listening for clients.
-            using TcpListener tcpListener = new TcpListener(IPAddress.Any,11000);
+            TcpListener tcpListener = new TcpListener(IPAddress.Any,11000);
             tcpListener.Start();
             Console.WriteLine("Waiting for a connection....");
 


rzikm (approved):
LGTM

liveans:
I'll merge it since failure is unrelated.

PR #8490, opened 2022-10-10
Update parameter type

gewarren:
Workaround for https://dev.azure.com/ceapex/Engineering/_workitems/edit/608073, which is unlikely to be fixed.

[Preview link](https://review.learn.microsoft.com/en-us/dotnet/api/system.runtime.intrinsics.vector128.as?view=net-7.0&branch=pr-en-us-8490#system-runtime-intrinsics-vector128-as-2(system-runtime-intrinsics-vector128((tfrom)))) (both the heading and parameter type in the Parameters subsection are fixed)

### DIFF
--- a/xml/System.Runtime.Intrinsics/Vector128.xml
+++ b/xml/System.Runtime.Intrinsics/Vector128.xml
@@ -134,7 +134,7 @@
       </Docs>
     </Member>
     <Member MemberName="As&lt;TFrom,TTo&gt;">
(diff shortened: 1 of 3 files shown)


dakersnar (approved):
I think this makes sense. Could use a second review from @tannergooding

BillWagner (approved):
This also LGTM @gewarren 

Let's :shipit:

PR #8494, opened 2022-10-11
Document default headers for web socket requests

greenEkatherine:
## Summary

Add info abut default headers for web socket requests

Fixes https://github.com/dotnet/runtime/issues/73564

### DIFF
--- a/xml/System.Net.WebSockets/ClientWebSocket.xml
+++ b/xml/System.Net.WebSockets/ClientWebSocket.xml
@@ -381,17 +381,17 @@
 
  This operation will not block. The returned <xref:System.Threading.Tasks.Task> object will complete after the connect request on the <xref:System.Net.WebSockets.ClientWebSocket> instance has completed.
  
- The request has default headers for HTTP/2 request:
+ The WebSocket handshake request has default headers for HTTP/1.1 request:
  
- :protocol: websocket
+ `Connection: Upgrade`
  
- For HTTP/1.1 request:
+ `Upgrade: websocket`
  
- Connection: Upgrade
+ `Sec-WebSocket-Key: [generated key]`
  
- Upgrade: websocket
+ For HTTP/2 request:
  
- Sec-WebSocket-Key: [generated key]
+ `:protocol: websocket`
  
  By default, HTTP/1.1 will be used. To change HTTP version used or enable HTTP version upgrade/downgrade, see <xref:System.Net.WebSockets.ClientWebSocketOptions.HttpVersion> option and <xref:System.Net.WebSockets.ClientWebSocketOptions.HttpVersionPolicy> option.
 


xml/System.Net.WebSockets/ClientWebSocket.xml
```diff
@@ -369,7 +369,17 @@
       <Docs>
         <param name="uri">The URI of the WebSocket server to connect to.</param>
         <param name="cancellationToken">A cancellation token used to propagate notification that the operation should be canceled.</param>
-        <summary>Connects to a WebSocket server as an asynchronous operation.</summary>
+        <summary>
+			Connects to a WebSocket server as an asynchronous operation.
```
CarnaViire, line 373: The resulted page will not retain this formatting, you can see it on the preview link generated https://review.learn.microsoft.com/en-us/dotnet/api/system.net.websockets.clientwebsocket.connectasync?view=net-7.0&branch=pr-en-us-8494#system-net-websockets-clientwebsocket-connectasync(system-uri-system-threading-cancellationtoken). I think this information might look better in the `Remarks` section, which supports formatting.

xml/System.Net.WebSockets/ClientWebSocket.xml
```diff
@@ -369,7 +369,17 @@
...
+			The request has default headers:
+			VersionPolicy = RequestVersionExact
```
CarnaViire, line 375: VersionPolicy is not an HTTP header. It is an application-level setting for `ClientWebSocket`. But we might want to still mention the behavior with words, like "By default, HTTP/1.1 will be used. To change HTTP version used or enable HTTP version upgrade/downgrade, see (xref to HttpVersion option) and (xref to VersionPolicy option)"

xml/System.Net.WebSockets/ClientWebSocket.xml
```diff
@@ -369,7 +369,17 @@
...
+			For HTTP/2 request:
+			:protocol = websocket
+			For HTTP/1.1 request:
+			Connection = Upgrade
```
CarnaViire, line 379: I think we should follow conventional headers format, like `Connection: Upgrade`

xml/System.Net.WebSockets/ClientWebSocket.xml
greenEkatherine: Moved to remarks

xml/System.Net.WebSockets/ClientWebSocket.xml
greenEkatherine: Added

xml/System.Net.WebSockets/ClientWebSocket.xml
```diff
@@ -380,6 +380,20 @@
...
+ 
+ :protocol: websocket
+ 
+ For HTTP/1.1 request:
+ 
+ Connection: Upgrade
+ 
+ Upgrade: websocket
+ 
+ Sec-WebSocket-Key: [generated key]
```
CarnaViire, line 394: I believe it's better to start with 1.1 as it's the default one so it's more interesting to the users, and add a bit of clarification on what request we are talking about. I also think it might be better to use `code formatting`.

```suggestion
 The WebSocket handshake request has default headers for HTTP/1.1 request:
 
 `Connection: Upgrade`
 
 `Upgrade: websocket`
 
 `Sec-WebSocket-Key: [generated key]`
 
 For HTTP/2 request:
 
 `:protocol: websocket`
```

CarnaViire (approved):
LGTM, thanks!

Issue #1640, opened 2019-01-18
Smtp Exceptions using SendMailAsync


saysmaster:
The documentation of `SmtpClient.SendMailAsync` specifies `ArgumentNullException` as the only exception thrown by this method.
On the contrary, `SmtpClient.Send` also specifies `SmtpException`, `SmtpFailedRecipientException`, `SmtpFailedRecipientsException`.

How can I handle failed recipient exceptions in `SmtpClient.SendMailAsync`?

PR #8495, opened 2022-10-11
Copy exception docs from SendAsync to SendMailAsync

### LINKED ISSUE #860: Clarify how SendMailAsync handles multiple messages
So, the older function, SendAsync, specifies "After calling SendAsync, you must wait for the email transmission to complete before attempting to send another email message using Send or SendAsync." I've been using Send synchronously instead to avoid this problem.

SendMailAsync doesn't specify in the documentation how it handles multiple SendMailAsyncs, it simply states that it sends as an asychronous operation. But my understanding is that your program is not going to be able to talk to an SMTP server multiple times simultaneously, so either SendMailAsync queues mail in order for processing, needs to be queued for processing by code that I have to write (in which some sample code would be nice), or should have a similar warning to SendAsync about a potential crash if I don't wait before calling two of them.

I don't really know what this thing does behaviorally based on the documentation, so I don't really know what the solution is.

---
#### Document Details

* ID: fc951970-adbe-37d9-9a35-c9911992df93
* Version Independent ID: 8e04b05f-4579-6d9e-a093-ba492fba5ff5
* Content: [SmtpClient.SendMailAsync Method (System.Net.Mail)](https://docs.microsoft.com/en-us/dotnet/api/system.net.mail.smtpclient.sendmailasync?view=netframework-4.7.2)
* Content Source: [xml/System.Net.Mail/SmtpClient.xml](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System.Net.Mail/SmtpClient.xml)
* Product: **dotnet-api**
* GitHub Login: @dotnet-bot

rzikm:
Fixes #860.
Fixes https://github.com/dotnet/dotnet-api-docs/issues/1640

### DIFF
--- a/xml/System.Net.Mail/SmtpClient.xml
+++ b/xml/System.Net.Mail/SmtpClient.xml
@@ -1372,6 +1372,8 @@ The classes shown in the following table are used to construct email messages th
  -or-  
   
  The <paramref name="message" /> could not be delivered to one or more of the recipients in <see cref="P:System.Net.Mail.MailMessage.To" />, <see cref="P:System.Net.Mail.MailMessage.CC" />, or <see cref="P:System.Net.Mail.MailMessage.Bcc" />.</exception>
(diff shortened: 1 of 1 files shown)


rzikm:
bump @dotnet/ncl, please take a look.

liveans (approved):
Well, it is LGTM, but it'd be nice if somebody else take a look as well.

PR #8519, opened 2022-10-17
Remove code example for SmtpClient.ServicePoint

### LINKED ISSUE #5050: Code example for SmtpClient.ServicePoint Property is misleading
https://docs.microsoft.com/en-us/dotnet/api/system.net.mail.smtpclient.servicepoint?view=netcore-3.1

The code example claims to show how to access the ServicePoint property, but the code actually doesn't access the property at any point in the code  :)

image

rzikm:
Fixes #5050.

The C# example was missing use of the property and since the snippet is reused multiple times it seemed better to remove it from the C++ example as well.


liveans (approved):
just a question, otherwise LGTM.

Should we consider add a simple example for this as a separate code snippet?

rzikm:
> Should we consider add a simple example for this as a separate code snippet?

I don't think it is necessary, as ServicePoint is obsolete, the original code just accessed the property and dumped some property from it, which does not seem too useful to me.

PR #8524, opened 2022-10-17
Replaced class keyword with struct keyword

vcrobe:
GCHandle is a struct not a class

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Runtime.InteropServices/GCHandleType.xml
+++ b/xml/System.Runtime.InteropServices/GCHandleType.xml
@@ -60,7 +60,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>Represents the types of handles the <see cref="T:System.Runtime.InteropServices.GCHandle" /> struct can allocate.</summary>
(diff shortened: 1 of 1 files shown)


xml/System.Runtime.InteropServices/GCHandleType.xml
```diff
@@ -60,7 +60,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>Represents the types of handles the <see cref="T:System.Runtime.InteropServices.GCHandle" /> class can allocate.</summary>
+    <summary>Represents the types of handles the <see cref="T:System.Runtime.InteropServices.GCHandle" /> struct can allocate.</summary>
```
AaronRobinsonMSFT, line 63:
```suggestion
    <summary>Represents the types of handles the <see cref="T:System.Runtime.InteropServices.GCHandle" /> type can allocate.</summary>
```

I accept this can be confusing, but the `struct` keyword is a C#-ism. The official term is actually "value class", so in this case "class" is broadly accurate. However, I accept the terminology here is worth updating. Thoughts on splitting the difference and we go with "type"?

AaronRobinsonMSFT (approved):

PR #8532, opened 2022-10-19
Mention supported HTTP versions in WinHttpHandler

### LINKED ISSUE #885: HTTP version support?
The [HttpClient](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient) docs say that WinHttpHandler is HTTP 2.0 capable. Unfortunately, the WinHttpHandler docs don't say anything about HTTP 2.0. Instead, they should go into detail about its HTTP 2.0 support, especially regarding to operating system version requirements.

---
#### Document Details

* ID: 10a449ca-bae2-357d-c22e-03c3779187ca
* Version Independent ID: eb88a9b2-c382-2fe3-df46-e70b44cf4818
* Content: [WinHttpHandler Class (System.Net.Http)](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.winhttphandler?view=netcore-2.0&viewFallbackFrom=netcore-2.1)
* Content Source: [xml/System.Net.Http/WinHttpHandler.xml](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System.Net.Http/WinHttpHandler.xml)
* Product: **dotnet-api**
* GitHub Login: @dotnet-bot

rzikm:
Fixes #885.

I did not find when HTTP2 was introduced but my tests and search suggest that the WinHTTP present on supported Windows versions supports it.

### DIFF
--- a/xml/System.Net.Http/WinHttpHandler.xml
+++ b/xml/System.Net.Http/WinHttpHandler.xml
@@ -30,7 +30,7 @@
   
 ## Remarks
 
- <xref:System.Net.Http.WinHttpHandler> is similar to other existing classes such as <xref:System.Net.Http.HttpClientHandler>. <xref:System.Net.Http.WinHttpHandler> provides a handler underneath an <xref:System.Net.Http.HttpClient> instance and is used to send HTTP requests out to a server and receive server responses.  
+ <xref:System.Net.Http.WinHttpHandler> is similar to other existing classes such as <xref:System.Net.Http.HttpClientHandler>. <xref:System.Net.Http.WinHttpHandler> provides a handler underneath an <xref:System.Net.Http.HttpClient> instance and is used to send HTTP requests out to a server and receive server responses. The <xref:System.Net.Http.WinHttpHandler> implementation supports HTTP versions up to HTTP/2.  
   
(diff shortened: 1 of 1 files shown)


xml/System.Net.Http/WinHttpHandler.xml
```diff
@@ -30,7 +30,7 @@
   
 ## Remarks
 
- <xref:System.Net.Http.WinHttpHandler> is similar to other existing classes such as <xref:System.Net.Http.HttpClientHandler>. <xref:System.Net.Http.WinHttpHandler> provides a handler underneath an <xref:System.Net.Http.HttpClient> instance and is used to send HTTP requests out to a server and receive server responses.  
+ <xref:System.Net.Http.WinHttpHandler> is similar to other existing classes such as <xref:System.Net.Http.HttpClientHandler>. <xref:System.Net.Http.WinHttpHandler> provides a handler underneath an <xref:System.Net.Http.HttpClient> instance and is used to send HTTP requests out to a server and receive server responses. The <xref:System.Net.Http.WinHttpHandler> implementation supports HTTP versions up to HTTP/2.
```
ManickaP, line 33: Would we throw for H/3? Or any different version from H/1(.1) and H/2? Just wondering what the actual behavior is.

ManickaP (approved):
LGTM, thanks!

xml/System.Net.Http/WinHttpHandler.xml
antonfirsov: I don't see any validation logic. I think we would silently treat HTTP/3 (or any other unknown) requests as if they were HTTP/1(.1).

rzikm: Anton is right, we pass null to WinHTTP and it uses H/1.1 as the default.

PR #8539, opened 2022-10-19
Fix the name of the LocationTracker object in the documentation examp…

dnovatchev:
…le the currently used TrackLocation name is a method-name, not a classname

## Summary

Describe your changes here.


gewarren (approved):
Thank you @dnovatchev

PR #8541, opened 2022-10-21
Update CancellationTokenSource.xml

brettshearer:
Should be disposed even if there is no exception.

## Summary

Describe your changes here.


kouvel (approved):
LGTM, thanks!

PR #8543, opened 2022-10-21
Add Warning for Full Cache Unexpected Behavior

Nick-Stanton:
## Summary

As noted in dotnet/runtime#36122, unexpected behavior occurs when MemoryCache is at its SizeLimit due to transient states. Since over-optimizing for this case would ultimately block application threads and slow down the app, we have opted to instead document a warning and not push a fix.

I'm crossing my fingers that I got the syntax right here.

### DIFF
--- a/xml/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.xml
+++ b/xml/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.xml
@@ -238,7 +238,7 @@
         <remarks>
           <format type="text/markdown"><![CDATA[
 > [!CAUTION]
-> There may be transient states while the cache is full but eviction is occuring in a background thread. Due to this, unexpected behavior may occur if the cache is being overutilized. 
+> There may be transient states while the cache is full but blocks are being released on a background thread. In this state, unexpected behavior might occur if the cache is being overutilized. 
           ]]></format>
         </remarks>
       </Docs>


xml/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.xml
```diff
@@ -235,7 +235,12 @@
       <Docs>
         <summary>Gets or sets the maximum size of the cache.</summary>
         <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+> [!CAUTION]
+> There may be transient states while the cache is full but eviction is occuring in a background thread. Due to this, unexpected behavior may occur if the cache is being overutilized.
```
gewarren, line 241:
```suggestion
> There may be transient states while the cache is full but blocks are being released on a background thread. In this state, unexpected behavior might occur if the cache is being overutilized. 
```

gewarren: I reworded it so "occur" wasn't used consecutively.

gewarren (approved):

maryamariyan (approved):

tarekgh (approved):
After applying @gewarren suggestion, LGTM.

gewarren:
Preview looks good - https://review.learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycacheoptions.sizelimit?view=dotnet-plat-ext-7.0&branch=pr-en-us-8543#remarks.

Nick-Stanton:
Nailed it!

gavinltomra:
I looked on the following pages and didn't find any warning.

* https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycacheoptions.sizelimit?view=net-8.0-pp
* https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.memorycacheoptions?view=net-8.0-pp
* https://learn.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-10.0#use-setsize-size-and-sizelimit-to-limit-cache-size

The current behaviour is more weird than just transient states, it means you can't use this cache at all for "every item is Size 1 and SizeLimit is max number of items" cases (which I would think would be common usage) because when you try to add items at full capacity it will fail to add the new item instead of evicting the LRU as documented.

adegeo:
@gavinltomra I did some investigation and it looks like this was dropped in some sort of transition from documenting it in this file to documenting it in the source code. I'll look at getting it fixed. Thanks!

Regarding the behavior, you might want to have that discussion in https://github.com/dotnet/runtime/issues

PR #8547, opened 2022-10-21
Fix documentation for INumberBase IsComplex and IsImaginary

dakersnar:
Matches changes here https://github.com/dotnet/runtime/pull/77330

### DIFF
--- a/xml/System.Numerics/INumberBase`1.xml
+++ b/xml/System.Numerics/INumberBase`1.xml
@@ -259,7 +259,7 @@
 
 ## Remarks
 
-This function returns `false` for a complex number `a + bi` where `a` or `b` are zero. In other words, it excludes real numbers and pure imaginary numbers.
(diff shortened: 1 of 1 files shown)


xml/System.Numerics/INumberBase`1.xml
```diff
@@ -259,7 +259,7 @@
 
 ## Remarks
 
-This function returns `false` for a complex number `a + bi` where `b` is zero.
+This function returns `false` for a complex number `a + bi` where `a` or `b` are zero. In other words, it excludes real numbers and pure imaginary numbers.
```
gewarren, line 262:
```suggestion
This function returns `false` for a complex number `a + bi` where `a` or `b` is zero. In other words, it excludes real numbers and pure imaginary numbers.
```

gewarren (approved):

PR #8553, opened 2022-10-22
OpenReadCompleted no event handler method

softwarepronto:
## Summary

An example of subscribe event is provided for the WebClient events. Some events have the event handler method in the documentation (DownloadFileCompleted, DownloadProgressChanged, and UploadProgressChanged). Some events are missing the method specified when the event is subscribed. This includes OpenReadCompleted  which is missing the method to when the event was subscribed. OpenReadCompleted was fixed to include the missing event handler method using the same pattern as UploadProgressChanged (a code section for the subscribe and a second code section for the event handler method).


rzikm (approved):
LGTM, thanks!

PR #8558, opened 2022-10-22
Missing using for StreamReader variable

softwarepronto:
## Summary

The example includes:
using WebClient client = new WebClient();
using Stream data = client.OpenRead(args[0]);

But was missing:
StreamReader reader = new StreamReader(data);

### DIFF
--- a/snippets/csharp/System.Net/WebClient/Overview/useragent.cs
+++ b/snippets/csharp/System.Net/WebClient/Overview/useragent.cs
@@ -22,8 +22,6 @@ public class Test
         using StreamReader reader = new StreamReader(data);
         string s = reader.ReadToEnd();
         Console.WriteLine(s);
-        data.Close();
-        reader.Close();
     }
 }
 //</Snippet1>


snippets/csharp/System.Net/WebClient/Overview/useragent.cs
```diff
@@ -19,11 +19,11 @@ public static void Main(string[] args)
         client.Headers.Add("user-agent", "Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.2; .NET CLR 1.0.3705;)");
 
         using Stream data = client.OpenRead(args[0]);
-        StreamReader reader = new StreamReader(data);
+        using StreamReader reader = new StreamReader(data);
         string s = reader.ReadToEnd();
         Console.WriteLine(s);
         data.Close();
         reader.Close();
```
antonfirsov, line 26: Do we still need the explicit `Close()` calls?

softwarepronto:
My goal when changing the documentation is to change as little as possible.
The Close methods were there to start but there was inconsistent use of the
using keyword. Also, the Close methods were ordered incorrectly.

When I personally code, I explicitly include the Close method such as with
Streams. I do not explicitly call Dispose for a class like WebClient which
does not implement a Close.

antonfirsov:
I think it would be still better to delete the explicit `Close()` calls. Since they are functionally equivalent to `Dispose()` for these types, they are redundant. Leaving the calls could result in confusion, and false conclusions: "Why is the official sample also closing? So it means `Dispose()` and `Close()` are not equivalent here?"

snippets/csharp/System.Net/WebClient/Overview/useragent.cs
softwarepronto: Removed

BillWagner (approved):
Thanks for making this change @softwarepronto 

I like where this landed. I'll :shipit: now.

PR #8560, opened 2022-10-22
Fix snippet buffered read logic

### LINKED ISSUE #4141: FileStream example bug
Hi,

Not sure if this is considered a bug but the first example on the FileStream docs only works if you have a file with less than 1024 bytes or bytes%1024 == 0. This is due to the buffer byte array being reused. 
If you use a 4 byte array and try read the string "hello" you end up with the following in the buffer
loop iteration 0:  hell 
loop iteration 1: oell
> hell
> oell
This is because the byte array holds data from the first loop
https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream?view=netframework-4.8

Kissaki:
The snippet generates a file with 215 bytes.  
It reads the file with a 1024 byte buffer.  
While this works with a new 0-initialized buffer, the buffered reading misses using the returned read bytes count.  
A file with > 1024 bytes will reuse the now non-0 buffer and print unintended values.

Fix buffered reading to be technically correct and work for arbitrary file contents.

Fixes #4141

### DIFF
--- a/snippets/csharp/System.IO/FileStream/Overview/fstream class.cs
+++ b/snippets/csharp/System.IO/FileStream/Overview/fstream class.cs
@@ -35,9 +35,10 @@ class Test
         {
             byte[] b = new byte[1024];
             UTF8Encoding temp = new UTF8Encoding(true);
-            while (fs.Read(b,0,b.Length) > 0)
+            int readLen;
+            while ((readLen = fs.Read(b,0,b.Length)) > 0)
             {
-                Console.WriteLine(temp.GetString(b));
+                Console.WriteLine(temp.GetString(b,0,readLen));
             }
         }
     }


adamsitnik (approved):
LGTM, thank you @Kissaki !

adamsitnik:
The CI leg is red due to missing solution file. I've verified manually that the provided code builds fine:

```log
PS D:\projects\repros\docs8560> dotnet build -c Release
MSBuild version 17.4.0-preview-22428-01+14c24b2d3 for .NET
  Determining projects to restore...
  All projects are up-to-date for restore.
C:\Program Files\dotnet\sdk\7.0.100-rc.1.22431.12\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInferen
ce.targets(219,5): message NETSDK1057: You are using a preview version of .NET. See: https://aka.ms/dotnet-support-poli
cy [D:\projects\repros\docs8560\docs8560.csproj]
  docs8560 -> D:\projects\repros\docs8560\bin\Release\net7.0\docs8560.dll

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:02.40
```

PR #8561, opened 2022-10-22
Improve and fix Regex example 1

### LINKED ISSUE #5043: Incorrect Regex class example with pattern "\d{2}?"
(At least as of current commit c9445d2) Line 195 of `System.Text.RegularExpressions/Regex.xml` states the following:
`\d{2}?        Match two decimal digits zero or one time. `

The description of this behavior does not match .NET's execution of this behavior. This will not match a two digit string zero times. I suspect that the actual behavior of `?` in quantifier expressions is to match a number of characters non-greedily (as might be useful in `{n,}` or `{n,m}` expressions) rather than being interpreted as making the quantifier optional.

The following is a minimal example of this issue. In this example `r1` is the example expression mentioned above, which according to the Regex class documentation should match an empty string as a zero times match. `r2` pulls the operator out to act on a non-capturing group to perform the correct behavior of matching two decimal digits zero or one times.

```
using System;
using System.Text.RegularExpressions;
					
public class Program
{
	public static void Main()
	{
		Regex r1 = new Regex("\\d{2}?");
		Regex r2 = new Regex("(?:\\d{2})?");

		Console.WriteLine(string.Format("Regex 1: {0}", r1.IsMatch("")));
		Console.WriteLine(string.Format("Regex 2: {0}", r2.IsMatch("")));
	}
}
```

The output of this example (on .NET Framework 4.7.2) is the following (with the expected behavior of both lines showing true if the documentation on the Regex class page is correct):
```
Regex 1: False
Regex 2: True
```

----

Furthermore (though it probably could be argued this is a separate issue and should be filed as such), the related example which uses the `{n}?` expression in question seems to be incorrect:
`
(\d*\.?\d{2}?){1}        Match the pattern of integral and fractional digits separated by a decimal point symbol at least one time.
`

This issue can be seen in the context of the encompassing expression example `^\s*[\+-]?\s?\$?\s?(\d*\.?\d{2}?){1}$` and the example currency program that uses it. The example program can state that "$42" represents a currency value, but "$4" does not represent a currency value. This is because the expression actually requires there to be at least two digits (because the `?` character does not make the digits from `\d{2}` optional as the Regex class documentation states, which is the issue stated above). I think the example rather should be `(\d*(?:\.\d{2})?){1}` and `^\s*[\+-]?\s?\$?\s?(\d*(?:\.\d{2})?){1}$` respectively.

Kissaki:
The produced pattern is mentioned in the Regex documentation for 'en-US'. Making the sample snippet use an explicit culture format info of 'en-US' makes it comparable and consistent for users across cultures. *(Commit 1)*

Change decimals pattern to match documented expectation. *(Commit 2)*

`"\\d{2}?"` does not match `""`.

`"(\\d{2})?"` matches `""`.

Fixes #5043

### DIFF
--- a/snippets/csharp/System.Text.RegularExpressions/Regex/Overview/regex_example1.cs
+++ b/snippets/csharp/System.Text.RegularExpressions/Regex/Overview/regex_example1.cs
@@ -5,47 +5,46 @@ using System.Text.RegularExpressions;
 
 public class Example
 {
-   public static void Main()
-   {
-      // Get the en-US NumberFormatInfo object to build the regular 
-      // expression pattern dynamically.
-      NumberFormatInfo nfi = CultureInfo.GetCultureInfo("en-US").NumberFormat;
+    public static void Main()
+    {
+        // Get the en-US NumberFormatInfo object to build the regular 
+        // expression pattern dynamically.
+        NumberFormatInfo nfi = CultureInfo.GetCultureInfo("en-US").NumberFormat;
 
-      // Define the regular expression pattern.
-      string pattern; 
-      pattern = @"^\s*[";
-      // Get the positive and negative sign symbols.
-      pattern += Regex.Escape(nfi.PositiveSign + nfi.NegativeSign) + @"]?\s?";
-      // Get the currency symbol.
-      pattern += Regex.Escape(nfi.CurrencySymbol) + @"?\s?";
-      // Add integral digits to the pattern.
-      pattern += @"(\d*";
-      // Add the decimal separator.
-      pattern += Regex.Escape(nfi.CurrencyDecimalSeparator) + "?";
-      // Add the fractional digits.
-      pattern += @"(\d{";
-      // Determine the number of fractional digits in currency values.
-      pattern += nfi.CurrencyDecimalDigits.ToString() + "})?){1}$";
-      
-      Console.WriteLine("Pattern is {0}", pattern);
-      Console.WriteLine();
-      
-      Regex rgx = new Regex(pattern);
+        // Define the regular expression pattern.
+        string pattern;
+        pattern = @"^\s*[";
+        // Get the positive and negative sign symbols.
+        pattern += Regex.Escape(nfi.PositiveSign + nfi.NegativeSign) + @"]?\s?";
+        // Get the currency symbol.
+        pattern += Regex.Escape(nfi.CurrencySymbol) + @"?\s?";
+        // Add integral digits to the pattern.
+        pattern += @"(\d*";
+        // Add the decimal separator.
+        pattern += Regex.Escape(nfi.CurrencyDecimalSeparator) + "?";
+        // Add the fractional digits.
+        pattern += @"(\d{";
+        // Determine the number of fractional digits in currency values.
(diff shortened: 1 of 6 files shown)


xml/System.Text.RegularExpressions/Regex.xml
```diff
@@ -196,7 +196,7 @@
 |`\s?`|Match zero or one white-space character.|  
 |`\d*`|Match zero or more decimal digits.|  
 |`\.?`|Match zero or one decimal point symbol.|  
-|`\d{2}?`|Match two decimal digits zero or one time.|  
+|`(\d{2})?`|Match two decimal digits zero or one time.|  
 |`(\d*\.?\d{2}?){1}`|Match the pattern of integral and fractional digits separated by a decimal point symbol at least one time.|  
 |`$`|Match the end of the string.|
```
gewarren, line 202: In the next sentence, "defined by the current culture's" should be changed to "defined by the en-US culture's".

gewarren (commented):
Thanks @Kissaki. Can you also update the VB snippet at snippets/visualbasic/VS_Snippets_CLR_System/system.text.regularexpressions.regex.class/vb/regex_example1.vb?

snippets/csharp/System.Text.RegularExpressions/Regex/Overview/regex_example1.cs
```diff
@@ -9,7 +9,7 @@
```
joperezr: The comment above would need to be adjusted as well, as it mentions that it is getting the **current** NumberFormatInfo.

snippets/csharp/System.Text.RegularExpressions/Regex/Overview/regex_example1.cs
```diff
@@ -23,9 +23,12 @@
       // Add the decimal separator.
       pattern += Regex.Escape(nfi.CurrencyDecimalSeparator) + "?";
       // Add the fractional digits.
-      pattern += @"\d{";
+      pattern += @"(\d{";
```
joperezr, line 26: NIT: Indenting is slightly off on these two lines

Kissaki: Was on the Console lines, not this marked line. Fixed.

snippets/visualbasic/VS_Snippets_CLR_System/system.text.regularexpressions.regex.class/vb/regex_example1.vb
```diff
@@ -23,9 +23,12 @@
```
joperezr: NIT: You are correctly fixing the issue so that `?` now has the intended meaning (optional, instead of lazily), but you are also introducing a capturing group by doing so. We could opt either to document that in the description saying that the piece `(\d{2})?` is a `Capturing group 1: Match two decimal digits zero or one time.` or alternatively we can change this line to be `pattern += "(?:\d{"` which would turn this group into a non-capturing group which is only used to be able to wrap the `?`

gewarren (approved):

PR #8564, opened 2022-10-23
Remove wrong path case sensitive claims

### LINKED ISSUE #7989: Directory.Exists is case sensitive on linux
**Issue description**

(https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.exists?view=net-6.0)

The docs state that the "path" parameter is not case sensitive.
This is not correct for .Net running in Linux.
I'd say the behavior is what one needs but then it should be documented properly.

**Target framework**

- [x] .NET Core
- [x] .NET Framework
- [x] .NET Standard

Kissaki:
#7989 describes that the Directory.Exists path parameter is case sensitive on linux despite the documentation stating it is case insensitive.

Presumably, this documentation comes from the Windows-only .NET framework days where the file systems are expected to be case insensitive.
This does not hold for newer multi-platform .NET versions which use other case sensitive platforms.

As such, presumably, this holds for all Directory methods with a `path` parameter. The statement is removed for every such method.

Fixes #7989

### DIFF
--- a/xml/System.IO/Directory.xml
+++ b/xml/System.IO/Directory.xml
@@ -550,6 +550,8 @@ An I/O error occurred.</exception>
   
  This method throws an <xref:System.IO.IOException> if the directory specified in the `path` parameter contains files or subdirectories.  
   
+ The case-sensitivity of the `path` parameter corresponds to that of the file system on which the code is running. For example, it's case-insensitive on NTFS (the default Windows file system) and case-sensitive on Linux file systems.  
+  
  In some cases, if you have the specified directory open in File Explorer, the <xref:System.IO.Directory.Delete%2A> method may not be able to delete it.  
   
    
@@ -656,6 +658,8 @@ An I/O error occurred.</exception>
  The `path` parameter may specify relative or absolute path information. Relative path information is interpreted as relative to the current working directory. To obtain the current working directory, see <xref:System.IO.Directory.GetCurrentDirectory%2A>.  
   
  Trailing spaces are removed from the end of the `path` parameter before deleting the directory.  
+  
+ The case-sensitivity of the `path` parameter corresponds to that of the file system on which the code is running. For example, it's case-insensitive on NTFS (the default Windows file system) and case-sensitive on Linux file systems.  
   
  If the `recursive` parameter is `true`, the user must have write permission for the current directory as well as for all subdirectories.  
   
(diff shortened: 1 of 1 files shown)


xml/System.IO/Directory.xml
```diff
@@ -550,8 +550,6 @@ An I/O error occurred.</exception>
   
  This method throws an <xref:System.IO.IOException> if the directory specified in the `path` parameter contains files or subdirectories.
```
gewarren, line 552:
```suggestion
  
The case-sensitivity of the `path` parameter corresponds to that of the file system on which the code is running. For example, it's case-insensitive on Windows and case-sensitive on Linux.
  
```

gewarren (commented):
I think it's better to be explicit rather than just to remove it.

xml/System.IO/Directory.xml
Kissaki: If we refer to filesystem I feel like naming Windows and Linux is inprecise if not misleading.

Shouldn’t it mention NTFS and e.g. ext3?

xml/System.IO/Directory.xml
```diff
@@ -550,7 +550,7 @@ An I/O error occurred.</exception>
   
  This method throws an <xref:System.IO.IOException> if the directory specified in the `path` parameter contains files or subdirectories.  
   
- The `path` parameter is not case-sensitive.  
+ The case-sensitivity of the `path` parameter corresponds to that of the file system on which the code is running. For example, it’s case-insensitive on NTFS (Windows default filesystem) and case-sensitive on ext3 (a Linux filesystem).
```
gewarren, line 553:
```suggestion
 The case-sensitivity of the `path` parameter corresponds to that of the file system on which the code is running. For example, it's case-insensitive on NTFS (the default Windows file system) and case-sensitive on Linux file systems.
```

gewarren (commented):
What do you think about this suggestion?

xml/System.IO/Directory.xml
Kissaki: The wording in the first parentheses is definitely better. I will apply it.

I don’t know if *all* “Linux file systems” (if you can even say that) are case sensitive, but it seems like a reasonable assumption to make here; at least most will surely be case sensitive.

gewarren (approved):
Thanks @Kissaki!

PR #8565, opened 2022-10-23
Fix escaped XML characters

### LINKED ISSUE #7776: Incorrect output for XmlElement.InnerText property in example code
In https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmlelement.innertext the output in the comment of the example code is incorrect. (It appears that it's necessary to double-escape the & character.)

Existing code:
`// <elem>Text containing <markup/> will have char(<) and char(>) escaped.</elem>`

Correct:
`// <elem>Text containing &lt;markup/&gt; will have char(&lt;) and char(&gt;) escaped.</elem>`

Kissaki:
Fixes #7776


gewarren (approved):
Thanks @Kissaki. I checked the preview and it now matches the actual output you see when you run the program.

PR #8567, opened 2022-10-23
Mention enumeration order of generic collections

### LINKED ISSUE #8083: Is the enumeration order of the Stack<T> collection well defined?
Hi! I used the [`Stack<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.stack-1) collection recently to implement an algorithm, whose correctness now depends on the `Stack<T>` having a LIFO enumeration order. I checked the documentation to see if this enumeration order is well defined, but I can't find an explicit statement about this. I found though an [old StackOverflow question](https://stackoverflow.com/questions/37725449/stack-and-queue-enumeration-order "Stack and Queue enumeration order"), where [the accepted answer](https://stackoverflow.com/a/37725773/11178549) (by Matthew Watson) includes this excerpt:

> **The Stack<T>.GetEnumerator() *strongly implies* that LIFO order is used.**
> 
> If you look at [the example for `Stack<T>.GetEnumerator()` in Microsoft's documentation](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.stack-1.getenumerator) and inspect the stated output, you can see that it is in LIFO order.
> 
> This strongly suggests that Microsoft fully intend a Stack to be enumerated in LIFO order - but they forgot (or didn't bother to) to explicitly document this!

I would like to ask if the current LIFO behavior is just an implementation detail, or if it's a publicly documented guarantee on which algorithms can rely on.

Thanks!

Kissaki:
Fixes #8083

### DIFF
--- a/xml/System.Collections.Generic/List`1+Enumerator.xml
+++ b/xml/System.Collections.Generic/List`1+Enumerator.xml
@@ -72,6 +72,8 @@
   
  <xref:System.Collections.Generic.List%601.Enumerator.Current%2A> returns the same object until <xref:System.Collections.Generic.List%601.Enumerator.MoveNext%2A> is called. <xref:System.Collections.Generic.List%601.Enumerator.MoveNext%2A> sets <xref:System.Collections.Generic.List%601.Enumerator.Current%2A> to the next element.  
   
+ The order of enumeration of list items proceeds from index zero to the last item of the list.
+  
  If <xref:System.Collections.Generic.List%601.Enumerator.MoveNext%2A> passes the end of the collection, the enumerator is positioned after the last element in the collection and <xref:System.Collections.Generic.List%601.Enumerator.MoveNext%2A> returns `false`. When the enumerator is at this position, subsequent calls to <xref:System.Collections.Generic.List%601.Enumerator.MoveNext%2A> also return `false`. If the last call to <xref:System.Collections.Generic.List%601.Enumerator.MoveNext%2A> returned `false`, <xref:System.Collections.Generic.List%601.Enumerator.Current%2A> is undefined. You cannot set <xref:System.Collections.Generic.List%601.Enumerator.Current%2A> to the first element of the collection again; you must create a new enumerator instance instead.  
   
  An enumerator remains valid as long as the collection remains unchanged. If changes are made to the collection, such as adding, modifying, or deleting elements, the enumerator is irrecoverably invalidated and the next call to <xref:System.Collections.Generic.List%601.Enumerator.MoveNext%2A> or <xref:System.Collections.Generic.List%601.Enumerator.System%23Collections%23IEnumerator%23Reset%2A> throws an <xref:System.InvalidOperationException>.  
--- a/xml/System.Collections.Generic/List`1.xml
+++ b/xml/System.Collections.Generic/List`1.xml
@@ -98,7 +98,7 @@
 ## Remarks  
  The <xref:System.Collections.Generic.List%601> class is the generic equivalent of the <xref:System.Collections.ArrayList> class. It implements the <xref:System.Collections.Generic.IList%601> generic interface by using an array whose size is dynamically increased as required.  
   
- You can add items to a <xref:System.Collections.Generic.List%601> by using the <xref:System.Collections.Generic.List%601.Add%2A> or <xref:System.Collections.Generic.List%601.AddRange%2A> methods. Enumerating over the elements goes first to last added (FIFO - first-in-first-out).  
+ You can add items to a <xref:System.Collections.Generic.List%601> by using the <xref:System.Collections.Generic.List%601.Add%2A> or <xref:System.Collections.Generic.List%601.AddRange%2A> methods. The order of enumeration of list items proceeds from index zero to the last item of the list.  
   
  The <xref:System.Collections.Generic.List%601> class uses both an equality comparer and an ordering comparer.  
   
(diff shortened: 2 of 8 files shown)


gewarren:
I'm not sure about this PR, because it's not clear from the original issue where the information is missing. The summaries for these types are clear about the implementation. @GSPP wrote:

> I believe that this question has been answered, but the answer does not follow from the documentation. It should be documented.
> List, Queue and SortedSet are also not documented in this way.

@GSPP can you clarify where you believe information is missing?

GSPP:
@gewarren I don't find anything in the documentation (https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.stack-1.getenumerator?view=net-6.0) that states the enumeration order. Randomized order, for example, would be consistent with this documentation. For that reason, the sample output does not specify the order. It just provides one particular order that might be seen.

I can understand that one might argue that it's not worth the time to address this. But the guarantee we are speaking of is simply not documented (unless I misunderstand).

xml/System.Collections.Generic/List`1.xml
```diff
@@ -98,7 +98,7 @@
 ## Remarks  
  The <xref:System.Collections.Generic.List%601> class is the generic equivalent of the <xref:System.Collections.ArrayList> class. It implements the <xref:System.Collections.Generic.IList%601> generic interface by using an array whose size is dynamically increased as required.  
   
- You can add items to a <xref:System.Collections.Generic.List%601> by using the <xref:System.Collections.Generic.List%601.Add%2A> or <xref:System.Collections.Generic.List%601.AddRange%2A> methods.  
+ You can add items to a <xref:System.Collections.Generic.List%601> by using the <xref:System.Collections.Generic.List%601.Add%2A> or <xref:System.Collections.Generic.List%601.AddRange%2A> methods. Enumerating over the elements goes first to last added (FIFO - first-in-first-out).
```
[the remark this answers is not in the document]
gewarren: I also think, based on @GSPP's comment, that the clarification is missing specifically from the GetEnumerator remarks on each type. Not from the general type remarks.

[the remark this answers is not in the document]
Kissaki: I was wondering whether to add it to the GetEnumerator() methods, in addition, or only. I decided to put it on the type where adding elements is also documented. Users may be using foreach for iteration without being aware of the use of GetEnumerator(), and to look there.

So I guess it should be added in both places - like the adding items docs.

xml/System.Collections.Generic/List`1+Enumerator.xml
```diff
@@ -72,6 +72,8 @@
   
  <xref:System.Collections.Generic.List%601.Enumerator.Current%2A> returns the same object until <xref:System.Collections.Generic.List%601.Enumerator.MoveNext%2A> is called. <xref:System.Collections.Generic.List%601.Enumerator.MoveNext%2A> sets <xref:System.Collections.Generic.List%601.Enumerator.Current%2A> to the next element.  
   
+ The order of enumeration of list items proceeds from index zero to the last item of the list.
```
eiriktsarpalis, line 75: Looks correct, even though it's probably not the best wording to describe this. Perhaps something along the lines of

> Lists are enumerated ordered by index, starting from the element with index zero up to the last item in the list.

I defer to @gewarren for choice of words.

xml/System.Collections.Generic/Queue`1+Enumerator.xml
```diff
@@ -72,6 +72,8 @@
   
  <xref:System.Collections.Generic.Queue%601.Enumerator.Current%2A> returns the same object until <xref:System.Collections.Generic.Queue%601.Enumerator.MoveNext%2A> is called. <xref:System.Collections.Generic.Queue%601.Enumerator.MoveNext%2A> sets <xref:System.Collections.Generic.Queue%601.Enumerator.Current%2A> to the next element.  
   
+ The order of enumeration of queue items proceeds from oldest to newest (FIFO - first-in-first-out).
```
eiriktsarpalis, line 75:
```suggestion
Queues are enumerated using insertion order, starting from the earliest inserted up to the latest inserted (FIFO - first-in-first-out order).  
```

xml/System.Collections.Generic/Stack`1+Enumerator.xml
```diff
@@ -72,6 +72,8 @@
   
  <xref:System.Collections.Generic.Stack%601.Enumerator.Current%2A> returns the same object until <xref:System.Collections.Generic.Stack%601.Enumerator.MoveNext%2A> is called. <xref:System.Collections.Generic.Stack%601.Enumerator.MoveNext%2A> sets <xref:System.Collections.Generic.Stack%601.Enumerator.Current%2A> to the next element.  
   
+ The order of enumeration of items proceeds from top to bottom (newest pushed to oldest pushed).
```
eiriktsarpalis, line 75: Are "top" and "bottom" well-defined/understood terms in the context of stacks? Also, "newest pushed" to "oldest pushed" is technically correct but it doesn't describe how the intermediate elements are enumerated. We might want to use terminology that is accepted in the literature for stacks.

eiriktsarpalis (changes requested):
I think the wording in the additions needs a few improvements.

xml/System.Collections.Generic/Stack`1.xml
```diff
@@ -103,6 +103,8 @@
   
 -   <xref:System.Collections.Generic.Stack%601.Peek%2A> returns an element that is at the top of the <xref:System.Collections.Generic.Stack%601> but does not remove it from the <xref:System.Collections.Generic.Stack%601>.  
   
+ The order of enumeration of items proceeds from top to bottom (newest pushed to oldest pushed).
```
gewarren, line 106:
```suggestion
Stack elements are enumerated from the top to the bottom of the stack.
```

xml/System.Collections.Generic/Stack`1.xml
```diff
@@ -703,6 +705,8 @@
   
  <xref:System.Collections.Generic.Stack%601.Enumerator.Current%2A> returns the same object until <xref:System.Collections.Generic.Stack%601.Enumerator.MoveNext%2A> is called. <xref:System.Collections.Generic.Stack%601.Enumerator.MoveNext%2A> sets <xref:System.Collections.Generic.Stack%601.Enumerator.Current%2A> to the next element.  
   
+ The order of enumeration of items proceeds from top to bottom (newest pushed to oldest pushed).
```
gewarren, line 708:
```suggestion
Stack elements are enumerated from the top to the bottom of the stack.
```

gewarren: Also this should be moved prior to line 704.

xml/System.Collections.Generic/Stack`1+Enumerator.xml
gewarren: I believe top and bottom are commonly used and we use them elsewhere in the Stack documentation.

gewarren:
```suggestion
Stack elements are enumerated from the top to the bottom of the stack.
```

gewarren: And move this to line 69 (add it to the existing paragraph about enumerators).

xml/System.Collections.Generic/List`1+Enumerator.xml
gewarren: And move it before line 71.

gewarren (changes requested):

PR #8569, opened 2022-10-24
Fx handler methods for event UploadProgressChanged

softwarepronto:
## Summary

The documentation for WebClient.UploadProgressChanged event is as follows: https://learn.microsoft.com/en-us/dotnet/api/system.net.webclient.uploadprogresschanged?view=net-6.0

This documentation for UploadProgressChanged uses <Snippet3>that originally included:
            client.UploadFileCompleted += new UploadFileCompletedEventHandler(UploadFileCallback2);
            client.UploadProgressChanged += new UploadProgressChangedEventHandler(UploadProgressCallback);

The documentation page for UploadProgressChanged then contains a reference to Snippet42 to document the above methods but Snippet42 contains the following methods: UploadProgressCallback
DownloadProgressCallback: this is a download method not used in Snippet3

Snippet4 contains the documentation needed by Snippet3 -- the definition of the UploadFileCallback2 method. Snippet4 has no UploadProgressChangedEventHandler. I made a copy of the original UploadProgressCallback and called it UploadProgressCallback2. I placed this in Snippet4 and updated Snippet3 to reference UploadProgressCallback2 instead of UploadProgressCallback.

Once asyncmethods.cs is updated, I will change the source to this link: https://learn.microsoft.com/en-us/dotnet/api/system.net.webclient.uploadprogresschanged?view=net-6.0

I will swap out the reference to Snippet42 and replace it with Snippet4.

I am pretty sure Snippet41 and Snippet42 are orphaned but this is obsolete documentation. I'd rather fix what is obviously broken (Snippet4) than purge unused code.

### DIFF
--- a/xml/System.Net/WebClient.xml
+++ b/xml/System.Net/WebClient.xml
@@ -7163,7 +7163,7 @@ internal class MyWebClient : WebClientProtocol
  The following code example shows an implementation of a handler for this event.
 
  [!code-cpp[NCLWebClientAsync#42](~/snippets/cpp/VS_Snippets_Remoting/NCLWebClientAsync/CPP/asyncmethods.cpp#42)]
- [!code-csharp[NCLWebClientAsync#42](~/snippets/csharp/System.Net/DownloadDataCompletedEventArgs/Overview/asyncmethods.cs#42)]
+ [!code-csharp[NCLWebClientAsync#44](~/snippets/csharp/System.Net/DownloadDataCompletedEventArgs/Overview/asyncmethods.cs#44)]
  [!code-vb[NCLWebClientAsync#42](~/snippets/visualbasic/VS_Snippets_Remoting/NCLWebClientAsync/VB/asyncmethods.vb#42)]
 
  ]]></format>


softwarepronto:
> > I am pretty sure Snippet41 and Snippet42 are orphaned but this is obsolete documentation.
> 
> Snippet42 is used by DownloadProgressChangedEvengArgs documentation
> 
> Since Snippet5 is already used by documentation for UploadFileCompleted (and UploadFileCompletedEventArgs), I don't think we should add the UploadProgressCallback2 to it. It should be a separate snippet to be referenced from UploadProgressChanged documentation

I fixed it for C#. If you agree with the change for C# then I will fix CPP and VB.NET in the same manner.

snippets/csharp/System.Net/DownloadDataCompletedEventArgs/Overview/asyncmethods.cs
```diff
@@ -60,19 +60,31 @@ public static void UploadFileInBackground2(string address, string fileName)
...
 
             // Specify a progress notification handler.
-            client.UploadProgressChanged += new UploadProgressChangedEventHandler(UploadProgressCallback);
+            client.UploadProgressChanged += new UploadProgressChangedEventHandler(UploadProgressCallback2);
             client.UploadFileAsync(uri, "POST", fileName);
             Console.WriteLine("File upload started.");
         }
         //</Snippet4>
 
+        //<Snippet44>
```
rzikm, line 69: I think the snippet should not include the UploadFileCallback2 method to be consistent with examples for other event handlers

rzikm:
> I fixed it for C#. If you agree with the change for C# then I will fix CPP and VB.NET in the same manner.

Yes, consistency across languages would be nice

snippets/csharp/System.Net/DownloadDataCompletedEventArgs/Overview/asyncmethods.cs
softwarepronto: UploadProgressChanged has two events subscribed. 
UploadFileCompleted has two event subscribed.
UploadValuesCompleted has zero events subscribed. This is on my list to fix.

If we want to make it consistent
UploadProgressChanged  should have one event (remove one)
UploadFileCompleted  should have one event (remove one)
UploadValuesCompleted  should have one event (add one)

Is this how we should handle these last three events?

rzikm: It can have multiple events subscribed to, IMO it does not have to show the definition of the other handlers. I.e. show only UploadFileCompleted handler definition in docs for UploadFileCompleted event, as in https://learn.microsoft.com/en-us/dotnet/api/system.net.webclient.uploadfilecompleted?view=net-6.0&branch=main

gewarren: @softwarepronto Do you plan to make any more changes here?

BillWagner:
closing as abandoned

PR #8575, opened 2022-10-25
Update AES encryption snippet to include latest C# standards

Zekiah-A:
## Summary

Update the example in ASP.NET AES encryption to be using latest C# language features and conventions, making the example more usable in newer projects, some unnecessary code, such as null checking on non-nullable  method arguments was also cut out.

### DIFF
--- a/snippets/csharp/System.Security.Cryptography/Aes/Overview/program.cs
+++ b/snippets/csharp/System.Security.Cryptography/Aes/Overview/program.cs
@@ -14,9 +14,9 @@ public static class AesExample
         // Create a new instance of the Aes
         // class.  This generates a new key and initialization
         // vector (IV).
-        using var myAes = Aes.Create();
+        using Aes myAes = Aes.Create();
         // Encrypt the string to an array of bytes.
-        var encrypted = EncryptStringToBytes(original, myAes.Key, myAes.IV);
+        byte[] encrypted = EncryptStringToBytes(original, myAes.Key, myAes.IV);
 
         // Decrypt the bytes to a string.
         var roundtrip = DecryptStringFromBytes(encrypted, myAes.Key, myAes.IV);


snippets/csharp/System.Security.Cryptography/Aes/Overview/program.cs
```diff
@@ -1,119 +1,80 @@
-﻿//<Snippet1>
+//<Snippet1>
 using System;
```
BillWagner, line 2: The using System.IO should stay:

```suggestion
using System;
using System.IO;
```

BillWagner (commented):
There's one build issue. I've made a suggestion for it. Once that's fixed, this is ready to merge.

snippets/csharp/System.Security.Cryptography/Aes/Overview/program.cs
```diff
@@ -1,119 +1,81 @@
...
+
+        // Create the streams used for encryption.
+        using var msEncrypt = new MemoryStream();
+        using var csEncrypt = new CryptoStream(msEncrypt, encryptor, CryptoStreamMode.Write);
+        using var swEncrypt = new StreamWriter(csEncrypt);
+        //Write all data to the stream.
+        swEncrypt.Write(plainText);
+        
+        
+        var encrypted = msEncrypt.ToArray();
```
vcsjones, line 49: I don't think this is functionally the same code as it was before. Previously, `swEncrypt.Dispose()` was called before `msEncrypt.ToArray()` was called because of the explicit using block. This may have an impact on the flushing behavior of `swEncrypt`.

With the `using` as it is now,

Zekiah-A: Latest commit should resolve this.

bartonjs:
Do we have a style guide for samples?  Because as the area owner:

* I don't like file-scoped namespaces
* BCL style rules prohibit `var` for lines like `Aes.Create()`
  * And mostly prefer avoiding var altogether.
* I don't like statement-using
  * In fact, this change to use them changes the lifetime of some objects.  It was previously intentional that the using closed before `return plaintext;` to show that there's no lifetime associated.

Basically... the file (which I updated a month or so ago) looks exactly how I wanted it to, so I'm opposed to all the changes.  But if there's a consistent style for examples in our docs then I'll cave.  Although, the DES/TripleDES/RC2 samples should all be updated at the same time.

gewarren:
@bartonjs In the [.NET docs contributor guide](https://learn.microsoft.com/en-us/contribute/dotnet/dotnet-contribute#write-a-new-sample), we link to https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md for coding style.

snippets/csharp/System.Security.Cryptography/Aes/Overview/program.cs
```diff
@@ -1,119 +1,83 @@
...
-            // Return the encrypted bytes from the memory stream.
-            return encrypted;
-        }
-        //</Snippet2>
+        // Create a new instance of the Aes
+        // class.  This generates a new key and initialization
+        // vector (IV).
+        using var myAes = Aes.Create();
+        // Encrypt the string to an array of bytes.
+        var encrypted = EncryptStringToBytes(original, myAes.Key, myAes.IV);
```
bartonjs, line 19:
```suggestion
        using Aes myAes = Aes.Create();
        // Encrypt the string to an array of bytes.
        byte[] encrypted = EncryptStringToBytes(original, myAes.Key, myAes.IV);
```

These uses of `var` would violate the style guide (`var` is only permissable with an as-cast, hard-cast, or `new`).

MSDN-WhiteKnight:
>I don't like statement-using

But it is useful in code like this with multiple disposable objects, because it reduces the amount of braces and nested blocks, making the code more readable. You don't like it because you write framework code where having explicit lifetimes is good, but regular application developers are going to use this feature because it really helps. It needs to be adopted in examples somehow, like with any new features so users could discover them. It is already used in some examples i've submitted. like https://github.com/dotnet/dotnet-api-docs/blob/main/snippets/csharp/System.Reflection.Metadata/MetadataReader/MetadataReaderSnippets.cs

>In fact, this change to use them changes the lifetime of some objects. It was previously intentional that the using closed before return plaintext; to show that there's no lifetime associated.

Is it really important to highlight lifetimes here? I think it's pretty intuitive that lifetime of string variable is not tied to streams it was read from. It only gets more complicated when disposable objects are passed around. If they are only implemenation detail of the method, it's less important to have lifetime indicated explicitly.

vcsjones:
> Is it really important to highlight lifetimes here?

In my opinion, yes. I don't think this code is more clear and nor is `using var` a replacement for using statements. The lifetime of the streams _do_ matter in this case, and a using declaration resulted in manual flushes being added.

> lifetime of string variable is not tied to streams it was read from

Perhaps, but when you call `ToArray` does matter.

> It needs to be adopted in examples somehow, like with any new features so users could discover them. It is already used in some examples i've submitted.

That's fine. I like using using declarations where the at-time disposal does not matter. But in this case it did. I don't think using declarations are upgrades over using statements entirely, either. If the lifetime of the resource matters, use the thing that lets you control the lifetime of the resource.

PR #8581, opened 2022-10-26
Update `Memory<T>`: Link for "ref struct" is wrong

colejohnson66:
The link to "ref struct" linked to the normal struct page, not the ref struct page.

## Summary

Updated URL.


BillWagner (approved):
Thanks for catching this and making the update @colejohnson66 

I'll :shipit: now.

PR #8584, opened 2022-10-26
Update Instr(Int32, String, String)

KathleenDollard:
We got a bug that InStr did not work because of confusion with this example. I could not work out adding a comment in the code, and this text change seemed more explanatory anyway.

(The example switches both the start position and the `CompareMethod` and the person who had a problem only noticed that the start position is changed.

## Summary

Describe your changes here.

### DIFF
--- a/xml/Microsoft.VisualBasic/Strings.xml
+++ b/xml/Microsoft.VisualBasic/Strings.xml
@@ -1295,7 +1295,7 @@
 |`Text`|1|Performs a text comparison|
   
 ## Examples  
- This example uses the `InStr` function to return the position of the first occurrence of one string within another. In the first example, the search starts from the fourth character and returns the next lower case p because `CompareMethod` is `Text` and is therefore case insensitive. The position is always relative to the beginning of the string, regardless of the start position.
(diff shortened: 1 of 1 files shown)


cston (approved):

xml/Microsoft.VisualBasic/Strings.xml
```diff
@@ -1295,7 +1295,7 @@
 |`Text`|1|Performs a text comparison|
   
 ## Examples  
- This example uses the `InStr` function to return the position of the first occurrence of one string within another.  
+ This example uses the `InStr` function to return the position of the first occurrence of one string within another. In the first example, the search starts from the fourth character and returns the next lower case p because `CompareMethod` is `Text` and is therefore case insensitive. The position is always relative to the beginning of the string, regardless of the start position.
```
gewarren, line 1298:
```suggestion
 This example uses the `InStr` function to return the position of the first occurrence of one string within another. In the first example, the search starts from the fourth character and returns the next lower case "p" because `CompareMethod` is `Text` and therefore case insensitive. The position is always relative to the beginning of the string, regardless of the start position.
```

cston (approved):

PR #8596, opened 2022-10-30
Modernize remarks for System.Threading.Task

### LINKED ISSUE #8133: [System.String.Split] StringSpiltOptions missing examples for .NET 5 and later
The [StringSplitOptions page](https://docs.microsoft.com/en-us/dotnet/api/system.stringsplitoptions?view=net-6.0) is missing an example of how to use `StringSplitOptions.RemoveEmptyEntries` and `StringSplitOptionsTrimEntries` in the same method.

An example should be crafted that uses the `|` or `'` characters to separate values, so there can be leading spaces, trailing spaces, and blank values in the string being split. It should also demonstrate using the correct syntax of:

```
var value = somestring.Split('|', StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries);
```

I'm bringing this up because I used `StringSplitOptions.TrimEntries & StringSplitOptions.RemoveEmptyEntries` first and that didn't work.

### LINKED ISSUE #8138: Remarks for Task should be modernized
[The remarks for `System.Threading.Tasks.Task`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task?view=net-6.0#remarks) seem to largely originate from the .Net Framework 4.0 era and could use some modernization. Some of the issues:

* The first thing mentioned is the `Task` constructor, the second is `StartNew()` and the fourth is `RunSynchronously()`. Those methods should be used only rarely, so they either should not be mentioned here at all, or at a much less prominent place.
* Most of the examples use `Wait()` and not `await`. I think this proportion should be reversed.
* When the example does use `await`, it does not explain it. For something so important to `Task`, I think a sentence with a link to the documentation for the keyword is the minimum.

Less important notes:

* There's quite a few `Main`s in the examples, so they could benefit from using top-level statements, assuming that's acceptable.
* The section *For debugger developers* documents implementation details and has incredibly niche audience. That might have been useful when .Net was not open source, but now debugger developers can read the source code, which [contains the same information](https://github.com/dotnet/runtime/blob/1f15373/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs#L101-L110). So I think this section should be deleted here.

poke:
This pull request fixes #8138 which asked for a modernization of the `Task` remarks section.

I hope that all the references I used are working properly. I wasn’t able to run docfx on this repo to verify it myself (is there a guide somewhere that I didn’t see?). – Okay, the bot was very helpful, all references are correct now 😊

## Summary

Rewrite the remarks section for `Task` to focus more on modern usages of async/await with asynchronous methods, instead of showing older approaches and synchronous consumption using blocking methods.

- The section on task instantiation is replaced by a new section on general usage which demonstrates common interactions with tasks by the utilization of `async` and `await` where the creation of a task is driven by calling asynchronous methods, or using `Task.Run`.
- The section on creating a task is mostly left as is to show that `Task.Run` and `Task.Factory.CreateNew` are still used to manually create tasks. The snippets however were updated to use `await` and non-blocking code.
- The section on waiting for one or more tasks was rewritten with new snippets to demonstrate the asynchronous approaches with `await` instead, while keeping the original structure and content. So this uses `WaitAsync` to enforce a timeout, and `WhenAny` and `WhenAll` to interact with multiple tasks.
- The debugging section is removed since the information there matches the open source implementation and might be too implementation- specific to mention in the API documentation.

The following snippet changes are included:

- Add new snippet `usage` that demonstrates the common usage of `Task` when used with async/await and `Task.Run`.
- Add new snippet `await` that demonstrates how `await` is used to asynchronously wait for a single task, how `WaitAsync` can be used to enforce a timeout while awaiting, and how `WhenAny` and `WhenAll` can be used to await multiple tasks.
- Update `run1` and `startnew1` snippets to use async/await.
- Remove unused and outdated snippets that focused on a synchronous consumption of `Task`.


BillWagner:
closing as abandoned

PR #8597, opened 2022-10-30
Fix minor issues in String.Substring documentation

### LINKED ISSUE #4040: Possible typo on https://docs.microsoft.com/en-us/dotnet/api/system.string.substring?view=netframework-4.8
https://docs.microsoft.com/en-us/dotnet/api/system.string.substring?view=netframework-4.8 

Should the line towards the end be

If startIndex is equal to zero and **results** equals the length of the current string, the method returns the original string unchanged.

poke:
Fixes #4040 which pointed out a missing reference to the `length` parameter and also some other minor issues I found while going through the text.

## Summary

- Fix minor spelling and wording issues.
- Fix references to `LastIndexOf` (instead of mentioning `IndexOf` twice).
- Fix missing mention of `length` when the original string is returned.

### DIFF
--- a/xml/System/String.xml
+++ b/xml/System/String.xml
@@ -13153,12 +13153,12 @@ The following example defines a `StripStartTags` method that uses the <xref:Syst
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- You call the <xref:System.String.Substring(System.Int32)> method to extract a substring from a string that begins at a specified character position and ends at the end of the string. The starting character position is a zero-based; in other words, the first character in the string is at index 0, not index 1. To extract a substring that begins at a specified character position and ends before the end of the string, call the <xref:System.String.Substring(System.Int32%2CSystem.Int32)> method.
+ You call the <xref:System.String.Substring(System.Int32)> method to extract a substring from a string that begins at a specified character position and ends at the end of the string. The starting character position is zero-based; in other words, the first character in the string is at index 0, not index 1. To extract a substring that begins at a specified character position and ends before the end of the string, call the <xref:System.String.Substring(System.Int32%2CSystem.Int32)> method.
 
 > [!NOTE]
 > This method does not modify the value of the current instance. Instead, it returns a new string that begins at the `startIndex` position in the current string.
 
- To extract a substring that begins with a particular character or character sequence, call a method such as  <xref:System.String.IndexOf%2A> or <xref:System.String.IndexOf%2A> to get the value of `startIndex`. The second example illustrates this; it extracts a key value that begins one character position after the "=" character.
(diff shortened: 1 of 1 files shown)


xml/System/String.xml
```diff
@@ -13296,13 +13296,13 @@ The following example defines a `StripStartTags` method that uses the <xref:Syst
...
 
- The <xref:System.String.IndexOf%2A> method is used to get the position of the equals character in the string. The call to the <xref:System.String.Substring(System.Int32%2CSystem.Int32)> method extracts the key name, which starts from the first character in the string and extends for the number of characters returned by the call to the <xref:System.String.IndexOf%2A> method. The call to the <xref:System.String.Substring(System.Int32)> method then extracts the value assigned to the key. It starts at one character position beyond the equals character and extends to the end of the string.
+ The <xref:System.String.IndexOf%2A> method is used to get the position of the equals character in the string. The call to the <xref:System.String.Substring(System.Int32%2CSystem.Int32)> method extracts the key name, which starts from the first character in the string and extends for the number of characters returned by the call to the <xref:System.String.IndexOf%2A> method. The call to the <xref:System.String.Substring(System.Int32)> method then extracts the value assigned to the value. It starts at one character position beyond the equals character and extends to the end of the string.
```
gewarren, line 13305: I think this was correct as initially written - the key (before the =) has a value (after the =).

xml/System/String.xml
```diff
@@ -13296,13 +13296,13 @@ The following example defines a `StripStartTags` method that uses the <xref:Syst
  :::code language="fsharp" source="~/snippets/fsharp/System/String/Substring/source.fs" id="Snippet1":::
  :::code language="vb" source="~/snippets/visualbasic/VS_Snippets_CLR_Classic/classic String.Substring1 Example/VB/source.vb" id="Snippet1":::
 
- The following example uses the <xref:System.String.Substring%2A> method to separate key/value pairs that are delimited by an equals ("=") character.
+ The following example uses the <xref:System.String.Substring%2A> method to separate key/value pairs that are delimited by an equals (`"="`) character.
```
gewarren, line 13299:
```suggestion
 The following example uses the <xref:System.String.Substring%2A> method to separate key/value pairs that are delimited by an equals (`=`) character.
```

gewarren (commented):
Thanks @poke. Overall this looks great!

xml/System/String.xml
poke: Oh, you’re right. I must have missed the part where it said _“extracts **the value assigned** to the key”_. I’ll revert that.

xml/System/String.xml
poke: I’ve also replaced two more instances of `"="`.

gewarren (approved):

PR #8598, opened 2022-10-30
Update samples for StringSplitOptions.TrimEntries

### LINKED ISSUE #8133: [System.String.Split] StringSpiltOptions missing examples for .NET 5 and later

poke:
Since I accidentally mentioned the wrong issue in another pull request, I felt obligated to fix that issue now too 😂

This PR fixes #8133 which requested additional samples for `StringSplitOptions.TrimEntries`. I’ve built upon the existing samples to add additional cases for both `StringSplitOptions.TrimEntries` and `StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries`. I also took the liberty to restructure the output a tiny little bit in an attempt to make the logic a bit easier to follow.

I’ve updated the snippets for all of C#, F# and VB, to continue to keep them all in sync, making them all produce exactly the same output. I wasn’t able to update the C++ sample though since I don’t have a C++/CLI environment around to test it. But the C++ sample appears to be lagging behind on the functionality anyway, so I guess it’s not that much of a problem.

### DIFF
--- a/snippets/visualbasic/System/string.split/Program.vb
+++ b/snippets/visualbasic/System/string.split/Program.vb
@@ -2,8 +2,6 @@
 
 Module Program
     Sub Main(args As String())
-        Intro1()
-        Intro2()
-        Intro3()
+        Sample.StringSplitOptionsExamples()
     End Sub
 End Module
--- a/snippets/visualbasic/System/string.split/limit.vb
+++ b/snippets/visualbasic/System/string.split/limit.vb
@@ -4,13 +4,16 @@
         Console.WriteLine("What is your name?")
         Dim name As String = Console.ReadLine()
 
-        Dim substrings = name.Split(Nothing, 2)
+        Dim substrings = name.Split(" "c, count:=2)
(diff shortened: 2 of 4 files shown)


gewarren (approved):
Thanks @poke - and sorry for the very long delay.

PR #8599, opened 2022-10-31
Add NotSupportedException on IEnumerator.Reset

carlreinke:
`IEnumerator.Reset()` can throw `NotSupportedException` (as mentioned in the remarks).


eiriktsarpalis (approved):
I'm not sure if we're meant to document particular implementation behaviors in interface documentation, but other than that this looks good to me.

PR #8607, opened 2022-11-02
Replaced 'T' type with 'TResult'

vcrobe:
I think the statement on line 3059 

"Starting in .NET 6, for some `T` types and some result values, this method may return a cached singleton object rather than allocating a new object." 

is a bit confusing. I think it should be removed or enclosed in a box with a title like "Internal implementation".

Usually I/we as developer are not interested on the way a library is implemented when reading the documentation. If someone wants to see the implementation details then he/she can browses the source code.


buyaa-n:
> I think the statement on line 3059
> 
> "Starting in .NET 6, for some T types and some result values, this method may return a cached singleton object rather than allocating a new object."
> 
> is a bit confusing. I think it should be removed or enclosed in a box with a title like "Internal implementation".

I think that info would he be helpful for those who cares about, though I agree it is confusing because it did not say when. Anyway, this is not related to your change so better to create a doc issue.

buyaa-n (approved):

PR #8608, opened 2022-11-02
Update Task.xml

vcrobe:
Added a general tip to the Task.WhenAny() method.


xml/System.Threading.Tasks/Task.xml
```diff
@@ -6234,6 +6234,9 @@ The returned task will complete when any of the supplied tasks has completed.  T
 
 The returned task will complete when any of the supplied tasks has completed.  The returned task will always end in the `RanToCompletion` state with its `Result` set to the first task to complete.  The result value is `true` even if the first task to complete ended in the `Canceled` or `Faulted` state.
 
+> [!TIP]
+> In every overloaded versions of Task.WhenAny() when this method return the first completed task the other tasks will run until completion too even when some of them throw an exception. So once the first task complete you may want to cancel all the remaining tasks.
```
buyaa-n, line 6238:
```suggestion
> In every overloaded version of Task.WhenAny() when this method returns the first completed task, the other tasks will continue running until completion, even any of them completed in the `Canceled` or `Faulted` state. So, if that behavior is not desired you may want to cancel all the remaining tasks once the first task complete.
```

buyaa-n (approved):

PR #8611, opened 2022-11-04
Refer to the correct parameters

JeroenOortwijn:
Refer to the correct `path#` parameters.


JeroenOortwijn (commented):
I'm not 100% sure if commit https://github.com/dotnet/dotnet-api-docs/pull/8611/commits/99319b2100af976495dce7b21ad24fc140842591 is correct, as I haven't tested it yet.
But I would find it strange if the `Join` method will discard the previous paths if the last one is an absolute path, because it also doesn't do that in all the other `Join` overloads.

adamsitnik (approved):
LGTM, thank you for your contribution @JeroenOortwijn !

PR #8616, opened 2022-11-07
Changes for socket option overloads

### LINKED ISSUE #7843: Socket.SetSocketOption docs are misleading/confusing
https://github.com/dotnet/dotnet-api-docs/blob/13425c2d339c9f517577e57ff51e0aa727915541/xml/System.Net.Sockets/Socket.xml

The doc for at least _some_ overloads of Socket.SetSocketOption (specifically, the one that takes a System.Object as the last argument) are highly misleading.
A casual reading of them might lead one to assume that _any_ option is supported through this overload. However, in reality the code actually is only willing to accept a few possible options.
Other options require using other overloads. This should be clarified in the docs.
For example, trying to set the TCP option TcpKeepAliveTime with a TimeSpan value won’t work – it throws an ArgumentException (note that this is not a documented exception BTW). One has to use the overload that takes an int and “magically know” that the int is the intended number of msec. (It’s also possible that the overload taking a byte array works, but I didn’t test that.)

liveans:
## Summary

Updates on snippets and formatting example files.

Fixes #7843

### DIFF
--- a/snippets/cpp/VS_Snippets_Remoting/Socket_Socket_Options/CPP/source.cpp
+++ b/snippets/cpp/VS_Snippets_Remoting/Socket_Socket_Options/CPP/source.cpp
@@ -1,4 +1,4 @@
-#using < System.dll>
+#using <System.dll>
 
 using namespace System;
 using namespace System::Text;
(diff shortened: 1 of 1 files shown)


liveans:
You can ignore snippet build fail, because it says `Project file is missing for C++`.

antonfirsov (approved):
LGTM

PR #8617, opened 2022-11-07
Remove IPAddress.Parse example from TryParse docs

### LINKED ISSUE #8615: TryParse() example still uses Parse() example
#6913 fixed that the documentation was misleading, but the example for `ipaddress.tryparse` is still identical to the example for `ipaddress.parse`.

rzikm:
Fixes #8615.


liveans (approved):
Should we change it with an example for `TryParse` at least for C#?

PR #8618, opened 2022-11-10
add remarks for Memory<T>.Span property

ssg:
The cost of accessing `.Span` property of `Memory<T>` wasn't specified in the documentation.

### DIFF
--- a/xml/System/Memory`1.xml
+++ b/xml/System/Memory`1.xml
@@ -731,9 +731,8 @@ The garbage collector will not move the memory until the returned <xref:System.B
       <Docs>
         <summary>Returns a span from the current instance.</summary>
         <value>A span created from the current <see cref="T:System.Memory`1" /> object.</value>
-        <remarks>Although it involves no copying, accessing Span property is still costlier 
-          than reusing its return value. Try to assign its result to a variable, and use the variable
-          in your functions instead of repetitively calling Span.</remarks>
+        <remarks>Although it doesn't involve copying, it's costly to access this property.
+          Instead of accessing this property multiple times, it's better to assign its result to a variable (if possible).</remarks>
       </Docs>
     </Member>
     <Member MemberName="ToArray">


BillWagner:
ping @stephentoub

xml/System/Memory`1.xml
```diff
@@ -731,7 +731,9 @@ The garbage collector will not move the memory until the returned <xref:System.B
       <Docs>
         <summary>Returns a span from the current instance.</summary>
         <value>A span created from the current <see cref="T:System.Memory`1" /> object.</value>
-        <remarks>To be added.</remarks>
+        <remarks>Although it involves no copying, accessing Span property is still costlier 
+          than reusing its return value. Try to assign its result to a variable, and use the variable
+          in your functions instead of repetitively calling Span.</remarks>
```
stephentoub, line 736: This is true today, though it's also true of many properties in .NET in general (anything that does anything more than just read a field), and the perf has both improved over time and is likely to improve further in the future.  We can include this remark if we think it's valuable, I just get a little nervous about docs that try too much to document things based on implementation detail.  This is probably ok, though.

ssg: I think this is more helpful than letting people write code assuming it's zero-cost.

The question is: is caching the Span result an undesirable outcome? If Span eventually becomes zero-cost, would we regret what we've ended up with? I think not.

ssg: Yes, and the compiler quickly clarifies that situation. That's why I also added "try to" in the text instead of making it sound like a rule. Do you think this needs to be clarified even further in the text with a note like "Reusing .Span's result may not be always possible, such as in async methods"?

xml/System/Memory`1.xml
```diff
@@ -731,7 +731,9 @@ The garbage collector will not move the memory until the returned <xref:System.B
       <Docs>
         <summary>Returns a span from the current instance.</summary>
         <value>A span created from the current <see cref="T:System.Memory`1" /> object.</value>
-        <remarks>To be added.</remarks>
+        <remarks>Although it involves no copying, accessing Span property is still costlier
```
gewarren, line 734:
```suggestion
        <remarks>Although it doesn't involve copying, it's costly to access this property.
```

xml/System/Memory`1.xml
```diff
@@ -731,7 +731,9 @@ The garbage collector will not move the memory until the returned <xref:System.B
       <Docs>
         <summary>Returns a span from the current instance.</summary>
         <value>A span created from the current <see cref="T:System.Memory`1" /> object.</value>
-        <remarks>To be added.</remarks>
+        <remarks>Although it involves no copying, accessing Span property is still costlier 
+          than reusing its return value. Try to assign its result to a variable, and use the variable
```
gewarren, line 735:
```suggestion
          Instead of accessing this property multiple times, it's better to assign its result to a variable (if possible).</remarks>
```

xml/System/Memory`1.xml
gewarren, line 736:
```suggestion
```

gewarren (approved):

PR #8641, opened 2022-11-16
Fix the DateTime.Add methods docs to reflect the new behavior in .NET 7.0

### LINKED ISSUE #8292: Fix the DatetIme.Add methods docs to reflect the new behavior in .NET 7.0
In .NET 7.0 we have improved the calculation precision for most of Add methods in DateTime. This is done in the PR https://github.com/dotnet/runtime/pull/73198. The docs of these APIs need to get updated to reflect the new behavior. 
For example the doc https://docs.microsoft.com/en-us/dotnet/api/system.datetime.addseconds?view=net-6.0#remarks mention `The value parameter is rounded to the nearest millisecond.` which is not true any more.

pirrmann:
## Summary

Update DateTime and DateTimeOffset documentation to match the .NET 7 changes

Fixes #8292

### DIFF
--- a/xml/System/DateTime.xml
+++ b/xml/System/DateTime.xml
@@ -2092,11 +2092,9 @@ For applications in which portability of date and time data or a limited degree 
 
  The fractional part of `value` is the fractional part of a day. For example, 4.5 is equivalent to 4 days, 12 hours, 0 minutes, 0 seconds, 0 milliseconds, and 0 ticks.
 
- The `value` parameter is rounded to the nearest millisecond.
+ In .NET 6 and earlier versions, the `value` parameter is rounded to the nearest millisecond. In .NET 7 and later versions, the full <xref:System.Double> precision of the `value` parameter is used. However, due to the inherent imprecision of floating point math, the resulting precision will vary.
 
  The <xref:System.DateTime.AddDays%2A> method takes into account leap years and the number of days in a month when performing date arithmetic.
-
-
 
 ## Examples
  The following example uses the <xref:System.DateTime.AddDays%2A> method to determine the day of the week 36 days after the current date.
@@ -2164,11 +2162,9 @@ For applications in which portability of date and time data or a limited degree 
 
  The fractional part of `value` is the fractional part of an hour. For example, 4.5 is equivalent to 4 hours, 30 minutes, 0 seconds, 0 milliseconds, and 0 ticks.
 
- The `value` parameter is rounded to the nearest millisecond.
+ In .NET 6 and earlier versions, the `value` parameter is rounded to the nearest millisecond. In .NET 7 and later versions, the full <xref:System.Double> precision of the `value` parameter is used. However, due to the inherent imprecision of floating point math, the resulting precision will vary.
 
  Converting time intervals of less than an hour to a fraction can involve a loss of precision if the result is a non-terminating repeating decimal. (For example, one minute is 0.016667 of an hour.) If this is problematic, you can use the <xref:System.DateTime.Add%2A> method, which enables you to specify more than one kind of time interval in a single method call and eliminates the need to convert time intervals to fractional parts of an hour.
-
-
 
 ## Examples
  The following example uses the <xref:System.DateTime.AddHours%2A> method to add a number of whole and fractional values to a date and time. It also illustrates the loss of precision caused by passing the method a value that includes a fractional component.
@@ -2285,9 +2281,7 @@ The value parameter is rounded to the nearest integer.
 
  The fractional part of `value` is the fractional part of a millisecond. For example, 4.5 is equivalent to 4 milliseconds and 5000 ticks, where one millisecond = 10000 ticks.
 
- The `value` parameter is rounded to the nearest integer.
-
-
+ In .NET 6 and earlier versions, the `value` parameter is rounded to the nearest millisecond. In .NET 7 and later versions, the full <xref:System.Double> precision of the `value` parameter is used. However, due to the inherent imprecision of floating point math, the resulting precision will vary.
 
 ## Examples
  The following example uses the <xref:System.DateTime.AddMilliseconds%2A> method to add one millisecond and 1.5 milliseconds to a <xref:System.DateTime> value. It then displays each new value and displays the difference between it and the original value. The difference is displayed both as a time span and as a number of ticks. The example makes it clear that one millisecond equals 10,000 ticks. It also shows that fractional milliseconds are rounded before performing the addition; the <xref:System.DateTime> value that results from adding 1.5 milliseconds to the original date is 2 milliseconds greater than the original date.
@@ -2355,9 +2349,7 @@ The value parameter is rounded to the nearest integer.
 
  The fractional part of `value` is the fractional part of a minute. For example, 4.5 is equivalent to 4 minutes, 30 seconds, 0 milliseconds, and 0 ticks.
 
- The `value` parameter is rounded to the nearest millisecond.
-
-
+ In .NET 6 and earlier versions, the `value` parameter is rounded to the nearest millisecond. In .NET 7 and later versions, the full <xref:System.Double> precision of the `value` parameter is used. However, due to the inherent imprecision of floating point math, the resulting precision will vary.
 
 ## Examples
  The following example uses the <xref:System.DateTime.AddMinutes%2A> method to add a number of whole and fractional values to a date and time.
@@ -2498,9 +2490,7 @@ The value parameter is rounded to the nearest integer.
 
  The fractional part of `value` is the fractional part of a second. For example, 4.5 is equivalent to 4 seconds, 500 milliseconds, and 0 ticks.
 
- The `value` parameter is rounded to the nearest millisecond.
-
-
+ In .NET 6 and earlier versions, the `value` parameter is rounded to the nearest millisecond. In .NET 7 and later versions, the full <xref:System.Double> precision of the `value` parameter is used. However, due to the inherent imprecision of floating point math, the resulting precision will vary.
 
 ## Examples
  The following example uses the <xref:System.DateTime.AddSeconds%2A> method to add 30 seconds and the number of seconds in one day to a <xref:System.DateTime> value. It then displays each new value and displays the difference between it and the original value. The difference is displayed both as a time span and as a number of ticks.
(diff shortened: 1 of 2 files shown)


xml/System/DateTime.xml
```diff
@@ -2033,7 +2033,7 @@ For applications in which portability of date and time data or a limited degree
 
  The fractional part of `value` is the fractional part of a day. For example, 4.5 is equivalent to 4 days, 12 hours, 0 minutes, 0 seconds, 0 milliseconds, and 0 ticks.
 
- The `value` parameter is rounded to the nearest millisecond.
+ The `value` parameter is no longer rounded to the nearest millisecond (as it used to be before .NET 7.0).
```
gewarren, line 2036:
```suggestion
In .NET 6 and earlier versions only, the `value` parameter is rounded to the nearest millisecond.
```

xml/System/DateTimeOffset.xml
```diff
@@ -1020,7 +1020,7 @@ The <see cref="P:System.DateTimeOffset.UtcDateTime" /> property is earlier than
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- The fractional part of the `days` parameter is the fractional part of a day. For example, 4.5 is equivalent to 4 days, 12 hours, 0 minutes, 0 seconds, 0 milliseconds. The `days` parameter is rounded to the nearest millisecond.
+ The fractional part of the `days` parameter is the fractional part of a day. For example, 4.5 is equivalent to 4 days, 12 hours, 0 minutes, 0 seconds, 0 milliseconds. The `days` parameter is no longer rounded to the nearest millisecond (as it used to be before .NET 7.0).
```
gewarren, line 1023:
```suggestion
 The fractional part of the `days` parameter is the fractional part of a day. For example, 4.5 is equivalent to 4 days, 12 hours, 0 minutes, 0 seconds, 0 milliseconds. In .NET 6 and earlier versions only, the `days` parameter is rounded to the nearest millisecond.
```

xml/System/DateTimeOffset.xml
```diff
@@ -1422,10 +1422,10 @@ The resulting <see cref="T:System.DateTimeOffset" /> value is greater than <see
 |.01 second|10 milliseconds|
 |.001 second|1 millisecond|
 
- The `seconds` parameter is rounded to the nearest millisecond.
+ The `seconds` parameter is no longer rounded to the nearest millisecond (as it used to be before .NET 7.0).
```
gewarren, line 1425:
```suggestion
In .NET 6 and earlier versions only, the `seconds` parameter is rounded to the nearest millisecond.
```

xml/System/DateTimeOffset.xml
```diff
@@ -5119,7 +5119,7 @@ Strings that do not specify a UTC offset are assumed to have the offset of the l
  You can use the <xref:System.DateTimeOffset.Subtract%2A> method to subtract more than one kind of time interval (days, hours, minutes, seconds, or milliseconds) in a single operation. Its behavior is identical to the <xref:System.DateTimeOffset.op_Subtraction%28System.DateTimeOffset%2CSystem.TimeSpan%29> method, which defines the subtraction operator. The <xref:System.DateTimeOffset> structure also supports specialized addition methods (such as <xref:System.DateTimeOffset.AddDays%2A>, <xref:System.DateTimeOffset.AddHours%2A>, and <xref:System.DateTimeOffset.AddMinutes%2A>) that allow you to perform subtraction by assigning a negative value to the method parameter.
 
 > [!NOTE]
->  This method returns a new <xref:System.DateTimeOffset> object. It does not modify the value of the current object by adding `timeSpan` to its date and time.
+>  This method returns a new <xref:System.DateTimeOffset> object. It does not modify the value of the current object by adding `value` to its date and time.
```
gewarren, line 5122:
```suggestion
>  This method returns a new <xref:System.DateTimeOffset> object. It does not modify the value of the current object by subtracting `value` from its date and time.
```

gewarren (commented):
The docs are shared amongst all versions, so have to be written in a way to reflect that.

xml/System/DateTime.xml
tarekgh, line 2036: Can we use what @tannergooding suggested in the comment https://github.com/dotnet/dotnet-api-docs/issues/8292#issuecomment-1211390405?

`The value parameter is rounded to at least the nearest millisecond. More accuracy may be available for certain inputs provided they represent less than 2^53 ticks.`

pirrmann: I'm not sure that this comment makes it clearer the the documentation reader. There is no rounding of the `value` parameter in the code anymore. I think that @tannergooding's comment applies to the precision of the value, but can be misanderstood as if it implied that some rounding was performed.

pirrmann: @gewarren Thank you, I'll update the pattern on all cases

tarekgh: What about using something like:

 `The Add result is rounded to at least the nearest millisecond. In .NET 7.0 and up more accuracy may be available for certain inputs provided they represent less than 2^53 ticks.`

tarekgh: To be clear, I want the doc at least to give some hints on how the result is calculated to set the expectation.

pirrmann: How about `In .NET 7 and up, the double precision of the value parameter is used to round to the nearest tick`?

tarekgh: Looks good to me.

gewarren: "In .NET 7 and later versions..."

tarekgh (approved):

xml/System/DateTime.xml
tannergooding: > I think that @tannergooding's comment applies to the precision of the value, but can be misanderstood as if it implied that some rounding was performed.

You must always assume implicit rounding may exist when dealing with `float`/`double`, even if no explicit rounding happens in code.

Simply doing `double x = someInt64;` is a potentially lossy cast and its one of the few places where `implicit casts` can result in a "loss of data". The same goes for `float x = someInt32;` (`int32` and `uint32` are lossless for `double`; as are `int16` and `uint16` for `float`; and `byte`/`sbyte` for `Half`).

xml/System/DateTime.xml
```diff
@@ -2033,7 +2033,7 @@ For applications in which portability of date and time data or a limited degree
 
  The fractional part of `value` is the fractional part of a day. For example, 4.5 is equivalent to 4 days, 12 hours, 0 minutes, 0 seconds, 0 milliseconds, and 0 ticks.
 
- The `value` parameter is rounded to the nearest millisecond.
+ In .NET 6 and earlier versions only, the `value` parameter is rounded to the nearest millisecond. In .NET 7 and and later versions, the <xref:System.Double> precision of the `value` parameter is used to round to the nearest tick.
```
tannergooding, line 2036: Nearest tick isn't really accurate.

Inaccuracies can creep in due to `double` being a `binary floating-point value`. For example, the `double` value `9.2` is actually representing `9.199999999999999289457264239899814128875732421875` and so `AddHours(9.2)` will end up computing a tick count using this actual represented value.

Some precision of this "actual represented value" will get lost since `double` can only accurately represent whole integers up to `2^53`.

As I explained in https://github.com/dotnet/dotnet-api-docs/issues/8292#issuecomment-1211385532, this ultimately only guarantees `millisecond` accuracy still. We may include additional precision in the computed result and there may be non zero data for sub millisecond ranges. However, it is not guaranteed to be "exact" for such scenarios and so we cannot guarantee "microsecond", "tick", or "nanosecond" level accuracy.

tarekgh: @tannergooding do you suggest not changing the doc then? or we need to write it some other way?

tannergooding: There is probably some good way to describe this more succinctly, but I'm not sure quite how we'd do that atm as understanding the limitations here require more in depth technical explanations.

Maybe we could describe this as in some terminology to describe that in .NET 6 it was `Millisecond/Millisecond` but in .NET 7 it is `Millisecond/Nanosecond`.

That is, in .NET 6 we explicitly rounded everything to the nearest millisecond and dropped any additional precision.

In .NET 7, we no longer explicitly round and so we can have nanosecond level precision, but may still only persist user inputs to the millisecond level due to how binary-floating point numbers work.

pirrmann: Anything that would mention nanosecond as the precision would be extremely wrong, as DateTime is only storing the value as a number of ticks, and 1 tick = 100 ns. I don't understand what's wrong in what I've writen: the code uses the double precision (as much as it can , and yes, double can be inaccurate), and the result in expressed in ticks. In .NET 7, we *also* drop any additional precision after the tick, and that's why I suggest that we should *round* to the nearest tick.

pirrmann:
Please note that I'm not talking about accuracy, but about rounding. The
topic of this documentation PR was never about accuracy. Even if a value is
only accurate to the millisecond, you can still round to the nearest tick.

Le jeu. 17 nov. 2022 à 16:31, Tanner Gooding ***@***.***> a
écrit :

GSPP:
How about merely mentioning that there are no more guarantees?

> In .NET 6 and earlier versions only, the `value` parameter is rounded to the nearest millisecond. In .NET 7 and later versions, the full <xref:System.Double> precision of the `value` parameter is used. **However, due to the inherent nature of floating point math, the resulting precision will vary. No particular rounding mode is guaranteed.**

tarekgh:
@pirrmann @tannergooding I am fine with what @GSPP suggested. are you good with it too?

pirrmann:
> @pirrmann @tannergooding I am fine with what @GSPP suggested. Are you good with the suggested text too?

I appreciate the attempt to find a wording that would reach a consensus. I particularly like the "However, due to the inherent imprecision of floating point math, the resulting precision will vary".

I'm not sure why we'd want to write "No particular rounding mode is guaranteed.". Could we maybe just keep that out?

tarekgh:
@tannergooding could you please help advising the final text we should use here? personally, I am fine with what @pirrmann is proposing.

tannergooding:
I think what was suggested, minus the rounding mode comment is fine:
> In .NET 6 and earlier versions only, the value parameter is rounded to the nearest millisecond. In .NET 7 and later versions, the full xref:System.Double precision of the value parameter is used. However, due to the inherent imprecision of floating point math, the resulting precision will vary

tarekgh:
@pirrmann could you please update the PR according to the last feedback so we can proceed merging it? Thanks!

PR #8643, opened 2022-11-17
Clarify docs for Leading/TrailingZeroCount

### LINKED ISSUE #8640: TrailingZeroCount / LeadingZeroCount
The base used for this count (base 2) should be specified in the docs.

I made tests for `Int64.TrailingZeroCount` with decimal numbers like 10, 1_000, 10_000 ... and it appears that by coincidence the binary equivalence has the same number of trailing zeroes for all powers of 10. But this is misleading, as e.g.  for 64 it returns 6.

dakersnar:
Fixes #8640

Any other similar cases that should be bundled in with this change?

### DIFF
--- a/xml/System.Numerics/IBinaryInteger`1.xml
+++ b/xml/System.Numerics/IBinaryInteger`1.xml
@@ -198,9 +198,9 @@
         <Parameter Name="value" Type="TSelf" />
       </Parameters>
       <Docs>
-        <param name="value">The value whose leading zeroes are to be counted.</param>
-        <summary>Computes the number of leading zeros in a value.</summary>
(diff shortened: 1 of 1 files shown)


gfoidl:
> Any other similar cases that should be bundled in with this change?

[PopCount](https://learn.microsoft.com/en-us/dotnet/api/system.byte.popcount?view=net-7.0) already has _bits_ mentioned.

dakersnar:
cc @stephentoub this PR should match the PR you reviewed here: https://github.com/dotnet/runtime/pull/78510

stephentoub (approved):

PR #8648, opened 2022-11-18
Add new ArgumentExceptions for BrotliStream

gewarren:
Contributes to dotnet/docs#32620.

@Jozkee Does this apply to the other two constructors as well?


Jozkee (approved):
> @Jozkee Does this apply to the other two constructors as well?

No, just the ones taking a `compressionLevel`.

PR #8653, opened 2022-11-21
Introduce a System.Formats.Tar area

jeffhandley:
Related to https://github.com/dotnet/fabricbot-config/pull/64

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -5079,6 +5079,17 @@
                   {
                     "name": "hasLabel",
                     "parameters": {
+                      "label": "area-System.Formats.Tar"
+                    }
+                  }
+                ]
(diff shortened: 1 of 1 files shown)


carlossanlop (approved):
Ok I think I understand what every section does. Let me try to describe it:
- There are different triggers: "Needs triage", "Moved to another area", "Needs further triage", etc.
- Each one of them seems to have the same conditions for triggering: When the label is added for the firs time, or when it gets removed, or when it gets substituted with another one, or when the label is already there and there are changes in the issue/PR.

Is this explanation good enough?

jeffhandley:
You got it, @carlossanlop. And it's all generated by the scripts in dotnet/fabricbot-config.

PR #8663, opened 2022-11-26
Current node is returned if it is a root, not null

arviedelgado:
Docs is updated to align with the Source.

## Summary
The [Docs](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.nodes.jsonnode.root?view=net-7.0#system-text-json-nodes-jsonnode-root) says:
Summary: Gets the root JsonNode. If the current JsonNode is a root, null is returned.
Remarks: The current node is returned if it is a root.

The [Source](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonNode.cs#L141-L147) says:
Summary: Gets the root JsonNode.
Remarks: The current node is returned if it is a root.


BillWagner:

gewarren (approved):

gewarren:
Thanks @arviedelgado

PR #8669, opened 2022-11-29
Fix strange Console.WriteLine

ychikazawa:
## Summary

The "\t" in "\table" becomes an escape sequence, resulting in the following strange output.
(Also, these are rows, not tables.)

image

The VB version outputs only values, so I fix it to the same style.

https://github.com/dotnet/dotnet-api-docs/blob/main/snippets/visualbasic/VS_Snippets_ADO.NET/Classic%20WebData%20DataView%20Example/VB/source.vb#L67


gewarren (approved):
Thank you @ychikazawa!

PR #8672, opened 2022-11-29
Minor update in HttpClientHandler.xml

MageFroh:
## Summary

As far as I understand the `DangerousAcceptAnyServerCertificateValidator` mechanism (thanks to the little code snippet), the framework checks whether `ServerCertificateCustomValidationCallback` is set to that value. It does not make sense to check the value of `DangerousAcceptAnyServerCertificateValidator`, since it's a read-only property anyway.


wfurt (approved):
LGTM

PR #8684, opened 2022-12-04
add ImageSharp as an alternative to System.Drawing

SimonCropp:
## Summary

Add ImageSharp as an alternative to System.Drawing with a caveat about licence restrictions


MSDN-WhiteKnight:
It was intentionally removed by maintainers: https://github.com/dotnet/dotnet-api-docs/pull/8577 We need to define a [policy on external library recomendations](https://github.com/dotnet/docs/issues/32861) rather then just reinstating it.

PR #8701, opened 2022-12-08
Fix incorrectly escaped code block

timbussmann:
## Summary

One of the notes in the `Type.GetInterface` documentation contains an incorrectly escaped code block, making it render incorrect:

image

when it should look like this

image

fixing this using double-backticks to make the inner backtick not incorrectly terminate the code block.


gewarren (approved):
Thanks @timbussmann

PR #8714, opened 2022-12-15
Fix errors in codeowners file

gewarren:
These users don't have write access and break the file.


BillWagner:
I'm fine with this change, but I'll defer to @jeffhandley 

I'd also be fine with adding those folks to the `api-docs-write` team.  @gewarren I added you as a maintainer of that team if that's the right change here, or in the future.

BillWagner (approved):

PR #8726, opened 2022-12-19
Update File.Move exception details

adamsitnik:
fixes https://github.com/dotnet/runtime/issues/79213

### DIFF
--- a/xml/System.IO/File.xml
+++ b/xml/System.IO/File.xml
@@ -3198,13 +3198,11 @@ The following example moves a file.
         <exception cref="T:System.ArgumentNullException">
           <paramref name="sourceFileName" /> or <paramref name="destFileName" /> is <see langword="null" />.</exception>
         <exception cref="T:System.ArgumentException">.NET Framework and .NET Core versions older than 2.1: <paramref name="sourceFileName" /> or <paramref name="destFileName" /> is a zero-length string, contains only white space, or contains invalid characters. You can query for invalid characters by using the <see cref="M:System.IO.Path.GetInvalidPathChars" /> method.</exception>
-        <exception cref="T:System.UnauthorizedAccessException">
-        The caller does not have the required permission.
+        <exception cref="T:System.UnauthorizedAccessException">The caller does not have the required permission.
       
-        -or-
+ -or-
 
-        The Operating System has failed to acquire an exclusive access to the destination file.
-        </exception>
+ The Operating System has failed to acquire an exclusive access to the destination file.</exception>
         <exception cref="T:System.IO.PathTooLongException">The specified path, file name, or both exceed the system-defined maximum length. </exception>
         <exception cref="T:System.IO.DirectoryNotFoundException">The path specified in <paramref name="sourceFileName" /> or <paramref name="destFileName" /> is invalid, (for example, it is on an unmapped drive).</exception>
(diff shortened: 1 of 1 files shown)


xml/System.IO/File.xml
```diff
@@ -3198,7 +3198,13 @@ The following example moves a file.
...
           <paramref name="sourceFileName" /> or <paramref name="destFileName" /> is <see langword="null" />.</exception>
         <exception cref="T:System.ArgumentException">.NET Framework and .NET Core versions older than 2.1: <paramref name="sourceFileName" /> or <paramref name="destFileName" /> is a zero-length string, contains only white space, or contains invalid characters. You can query for invalid characters by using the <see cref="M:System.IO.Path.GetInvalidPathChars" /> method.</exception>
-        <exception cref="T:System.UnauthorizedAccessException">The caller does not have the required permission.</exception>
+        <exception cref="T:System.UnauthorizedAccessException">
+        The caller does not have the required permission.
+      
+        -or-
+
+        The Operating System has failed to acquire an exclusive access to the destination file.
+        </exception>
```
Jozkee, line 3207: Subsequent lines need to be indented with one space only. See [this exception ](https://github.com/dotnet/dotnet-api-docs/blob/4c6279f88d8411618554ff3641cfe38fedddc630/xml/System.IO/File.xml#L3567-L3580)as an example.

```suggestion
        <exception cref="T:System.UnauthorizedAccessException">The caller does not have the required permission.
 
 -or-
 
 The Operating System has failed to acquire an exclusive access to the destination file.</exception>
```

xml/System.IO/File.xml
```diff
@@ -3198,7 +3198,13 @@ The following example moves a file.
         <exception cref="T:System.ArgumentNullException">
           <paramref name="sourceFileName" /> or <paramref name="destFileName" /> is <see langword="null" />.</exception>
         <exception cref="T:System.ArgumentException">.NET Framework and .NET Core versions older than 2.1: <paramref name="sourceFileName" /> or <paramref name="destFileName" /> is a zero-length string, contains only white space, or contains invalid characters. You can query for invalid characters by using the <see cref="M:System.IO.Path.GetInvalidPathChars" /> method.</exception>
-        <exception cref="T:System.UnauthorizedAccessException">The caller does not have the required permission.</exception>
+        <exception cref="T:System.UnauthorizedAccessException">
+        The caller does not have the required permission.
+      
+        -or-
+
+        The Operating System has failed to acquire an exclusive access to the destination file.
```
Jozkee, line 3206: Do you know why this is not an issue on Copy?

xml/System.IO/File.xml
adamsitnik, line 3207:
```suggestion
        <exception cref="T:System.UnauthorizedAccessException">The caller does not have the required permission.
      
 -or-

 The Operating System has failed to acquire an exclusive access to the destination file.</exception>
```

xml/System.IO/File.xml
adamsitnik: > Do you know why this is not an issue on Copy?

I don't know for sure, but I suspect that Move is atomic and it requires the exclusive access (the source file gets "deleted" only when the file is successfully moved to the destination) while Copy is not (at the end you have two files)

Jozkee (approved):

PR #8731, opened 2022-12-20
Fix assumed typo about Execution Time from 0(n) to O(n)

sapph42:
## Summary

This include, which is, at minimum, part of /xml/System.Text/StringBuilder.xml, says that lookups of individual characters by index occur in 0(n) time (that is a zero).  On the assumption this is a typo, and O(n) was intended, that change has been made in this branch.


BillWagner (approved):
You have sharp eyes @sapph42 

I'll :shipit: this PR now.

PR #8739, opened 2022-12-25
Document `Exception.StackTrace` nullability

### LINKED ISSUE #8728: `Exception.StackTrace` does not explain how it could be `null`
As the title describes, [Exception.StackTrace](https://learn.microsoft.com/en-us/dotnet/api/system.exception.stacktrace?view=net-7.0) does not explain when it could be `null`.

[Exception.TargetSite](https://learn.microsoft.com/en-us/dotnet/api/system.exception.targetsite?view=net-7.0) mentions that, if `StackTrace` is `null`, it will be `null`.

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System/Exception.xml

colejohnson66:
Fixes #8728.

## Summary

`Exception.StackTrace` returns a nullable `string?`, but it is not documented *when* it will be `null`. Testing confirms that it's `null` before being thrown.

```csharp
// prints `True`
Console.WriteLine((new Exception()).StackTrace is null);

// prints `False`
try
{
    throw new Exception();
}
catch (Exception ex)
{
    Console.WriteLine(ex.StackTrace is null);
}
```


xml/System/Exception.xml
```diff
@@ -1364,7 +1364,7 @@ Stack Trace:
       </ReturnValue>
       <Docs>
         <summary>Gets a string representation of the immediate frames on the call stack.</summary>
-        <value>A string that describes the immediate frames of the call stack.</value>
+        <value>A string that describes the immediate frames of the call stack. If no stack trace is available (such as prior to being thrown), this will be `null`.</value>
```
gfoidl, line 1367: Is "being thrown" correct here or it should it rather be 
```suggestion
        <value>A string that describes the immediate frames of the call stack. If no stack trace is available (such as prior to stack unwinding, this will be `null`.</value>
```
or "stack walking", etc. Or a combination of these?

colejohnson66: Maybe a combination? "…such as prior to stack unwinding from a `throw` statement"?

gfoidl: Sounds good.

Can the stack be unwound w/o throwing too? (I don't know, but docs should be accurate)

colejohnson66: Made a new commit with the wording tweak.

xml/System/Exception.xml
```diff
@@ -1364,7 +1364,7 @@ Stack Trace:
       </ReturnValue>
       <Docs>
         <summary>Gets a string representation of the immediate frames on the call stack.</summary>
-        <value>A string that describes the immediate frames of the call stack.</value>
+        <value>A string that describes the immediate frames of the call stack. If no stack trace is available (such as prior to stack unwinding from a `throw` statement), this will be `null`.</value>
```
gewarren, line 1367:
```suggestion
        <value>A string that describes the immediate frames of the call stack. If no stack trace is available (such as prior to stack unwinding from a <see langword="throw" /> statement), the value is <see langword="null" />.</value>
```

gewarren (approved):
Thanks @colejohnson66

PR #8746, opened 2022-12-28
Fix xref warnings

gewarren:
I don't think the CloudAppendBlob changes will be overwritten even though they're not in the docs section, because the last version they applied to was .NET Core 2.1


BillWagner (approved):
Thanks @gewarren 
This LGTM. Let's :shipit:

gewarren:
Thanks @BillWagner, and welcome back!

Issue #7840, opened 2022-03-17
Document stored exceptions


gewarren:
This issue is to discuss if and how to document "stored exceptions", that is, exceptions that are stored in the returned object by Task-returning methods.

With async methods, and more generally Task-returning methods, a majority of exceptions don't emerge synchronously from the call.  Rather, any exception gets stored into the returned Task, and then later asking that Task for the exception (via await'ing it, .Wait()'ing it, accessing its .Exception), etc., will surface the exception.  If you have a method:

```C#
public static async Task MethodAsync()
{
    await Task.Yield();
    throw new FormatException();
}
```
and you call it as:
```C#
Task t = MethodAsync();
```
the call to MethodAsync will never throw a `FormatException`.  Rather, such an exception may emerge when you later await the task, e.g.

```C#
await t; // throws FormatException
t.GetResult().GetAwaiter(); // throws FormatException
t.Wait(); // throws an AggregateException wrapping FormatException
Exception e = Task.Exception; // returns an AggregateException wrapping FormatException
...
```

Historically, then, we haven't documented FormatException as an exception thrown out of MethodAsync, because technically it's not.  Of course, in the 95% use case where a developer does:

```C#
await MethodAsync();
```

the distinction is irrelevant, as it's not directly observable whether the exception was thrown from MethodAsync or by awaiting the task it returned.

_Originally posted by @stephentoub in https://github.com/dotnet/dotnet-api-docs/issues/7692#issuecomment-1034151896_

gewarren:
Asynchronous methods can throw argument validation exceptions synchronously, so it seems like we'd need to differentiate stored exceptions from synchronous exceptions.

We'd probably need to use a different XML tag in the /// comments (for example, `<storedexception cref="FormatException">The format is incorrect.</storedexception>`. These exceptions could then be added to a new "Stored exceptions" heading/section in the API reference docs, which would exist in addition to the existing [Exceptions](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.comparer-1.compare?view=net-6.0#exceptions) heading. Or we (mdoc) could automatically add boilerplate text to stored exceptions, saying that they are thrown asynchronously, but still list them in the existing Exceptions section.

Thoughts?

stephentoub:
> Asynchronous methods can throw argument validation exceptions synchronously, so it seems like we'd need to differentiate stored exceptions from synchronous exceptions.

[...]

PR #8751, opened 2022-12-30
Add remarks about stored exceptions

gewarren:
Adds boilerplate text that asynchronous methods store the same exceptions that their synchronous counterpart can throw, with a link to that counterpart.

[Preview of remarks](https://review.learn.microsoft.com/en-us/dotnet/api/system.io.pipes.pipestream.flushasync?view=net-7.0&branch=pr-en-us-8751).

I used a utility to add these remarks to \*Async methods *that have a matching synchronous method* (571 of 1136 \*Async methods).

Contributes to #7840.

cc @stephentoub


ManickaP:
Just an idea, but could the link to the sync method point directly to the Exception dub-heading (`#exceptions`)? E.g.: https://review.learn.microsoft.com/en-us/dotnet/api/system.io.pipes.pipestream.flush?view=net-7.0&branch=pr-en-us-8751#exceptions

gewarren:
> Just an idea, but could the link to the sync method point directly to the Exception dub-heading (`#exceptions`)? E.g.: https://review.learn.microsoft.com/en-us/dotnet/api/system.io.pipes.pipestream.flush?view=net-7.0&branch=pr-en-us-8751#exceptions

That's a good idea, except it's only effective for methods with a single overload. That's because all overloads are on a single page, and an xref link to a specific overload is accomplished using a bookmark, for example, https://learn.microsoft.com/en-us/dotnet/api/system.string.substring?view=net-7.0#system-string-substring(system-int32). By adding `#exceptions` to the xref link, the bookmark to the specific overload is superseded by the bookmark to the Exceptions heading (and since there's more than one Exceptions heading on a page with multiple overloads, it just points to the top of the page).

xml/Microsoft.Extensions.Caching.Distributed/DistributedCacheExtensions.xml
```diff
@@ -89,7 +89,7 @@
         <param name="key">The key to get the stored data for.</param>
         <summary>Asynchronously gets a string from the specified cache with the specified key.</summary>
         <returns>A task that gets the string value from the stored cache key.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>This method stores all the exceptions that its synchronous counterpart can throw in the task it returns. If an exception is stored during execution, it will be thrown when you await either this method or the task it returns. For the stored exceptions, see the exceptions thrown by <see cref="M:Microsoft.Extensions.Caching.Distributed.DistributedCacheExtensions.GetString(Microsoft.Extensions.Caching.Distributed.IDistributedCache,System.String)" />.</remarks>
```
stephentoub, line 92: > This method stores all the exceptions that its synchronous counterpart can throw in the task it returns.

I think we should caveat the "all" slightly in the boilerplate... typically we allow/encourage usage exceptions (in particular ArgumentException and its derived types) to escape synchronously from asynchronous methods.  So we might want to make this say something more like "This method stores all non-usage exceptions...".

> If an exception is stored during execution, it will be thrown when you await either this method or the task it returns.

"awaiting a method" isn't actually a thing: you're always using await with some instance, and if you write "await Foo()", that instance is whatever Foo() returns.
https://devblogs.microsoft.com/pfxteam/invoke-the-method-with-await-ugh/
so this should just say something like "it will be thrown when you await the task the method returns".

stephentoub:
Thanks for working on this.

gewarren:
@stephentoub Can you take another look? Is this good to merge?

xml/Microsoft.Extensions.Caching.Distributed/DistributedCacheExtensions.xml
```diff
@@ -89,7 +89,7 @@
         <param name="key">The key to get the stored data for.</param>
         <summary>Asynchronously gets a string from the specified cache with the specified key.</summary>
         <returns>A task that gets the string value from the stored cache key.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>This method stores all non-usage exceptions that its synchronous counterpart can throw in the task it returns. If an exception is stored during execution, it will be thrown when you await that task. Usage exceptions, such as <see cref="T:System.ArgumentException" />, are still thrown synchronously. For the stored exceptions, see the exceptions thrown by <see cref="M:Microsoft.Extensions.Caching.Distributed.DistributedCacheExtensions.GetString(Microsoft.Extensions.Caching.Distributed.IDistributedCache,System.String)" />.</remarks>
```
stephentoub, line 92: Maybe a small tweak to make it read a little less ambiguous:
"This method stores all non-usage exceptions that its synchronous counterpart can throw in the task it returns." =>
"This method stores in the task it returns all non-usage exceptions that the method's synchronous counterpart can throw."

stephentoub, line 92: "If an exception is stored during execution, it will be thrown when you await that task." =>
"If an exception is stored into the returned task, that exception will be thrown when the task is awaited."

stephentoub (approved):

PR #8765, opened 2023-01-05
Update FileShare.xml

danmoseley:
Clarify the doc for System.IO.FileShare to make clear it affects any other access to the file -- not just access through FileStream. The flag is passed through to the OS. On Windows it maps directly to `dwShareMode` and on Unix it leads to a `LOCK_SH` (`LOCK_SH` is purely advisory, but that's not relevant to this edit, the point remains that it's a request to the OS, and whether the "other" access is via a FileStream or something else is not relevant.)

@ericstj I used "other code" here but in docs I sometimes see "other processes". Do you know what's correct? I would have expected the share mode to follow the handle, not the process.

### DIFF
--- a/xml/System.IO/FileShare.xml
+++ b/xml/System.IO/FileShare.xml
@@ -61,7 +61,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>Contains constants for controlling the kind of access other code can have to the same file.</summary>
+    <summary>Contains constants for controlling the kind of access other operations can have to the same file.</summary>
     <remarks>
       <format type="text/markdown"><![CDATA[  
   


xml/System.IO/FileShare.xml
```diff
@@ -61,7 +61,7 @@
     </Attribute>
   </Attributes>
   <Docs>
-    <summary>Contains constants for controlling the kind of access other <see cref="T:System.IO.FileStream" /> objects can have to the same file.</summary>
+    <summary>Contains constants for controlling the kind of access other code can have to the same file.</summary>
```
ericstj, line 64: Maybe "clients", "callers", "consumers"?  It's any other caller to the CreateFile API, either in process or out.  Saying "code" might imply too limited a scope.  cc @dotnet/area-system-io

danmoseley: Perhaps this is overly pedantic, but I suspect it's not just CreateFile, but NtXXXXX or ZwXXXX or whatever. @JeremyKuhne 

Open to any.

danmoseley: Here's an example of where the CreateFileW docs use "process", which while correct, is not complete, as it's surely any other handle, except perhaps a duplicated one (on Unix, flock definitely applies to a file descriptor)
```
Enables subsequent open operations on a file or device to request read access.
Otherwise, no process can open the file or device if it requests read access.
```

danmoseley: perhaps ".. kind of access other code not using this FileStream can have .." is accurate and complete enough.

JeremyKuhne: "users" maybe? I'm fine with whatever- it helps encourage broader thinking that end users will be affected using "normal" OS interaction. At least on Windows it's a low-level construct- there is no way around it.

danmoseley: right, although of course it could be the very next line of code, running as the same user on the same thread, that tries to open the file to write..

bartonjs: I'd go with `calls`, `callers`, or `operations`, personally.  The last one (`operations`) has the benefit of matching the terminology of the underlying documentation of our original platform.

BillWagner (approved):
Based on the discussion, I'll merge this.

It may not be the best language, but it's an improvement.

PR #8745, opened 2022-12-28
Add OperationCanceledException exceptions

gewarren:
Used a utility I wrote to automatically add `OperationCanceledException` exceptions with reason "The cancellation token was canceled" to methods that accept a `CancellationToken` parameter.

[Example preview](https://review.learn.microsoft.com/en-us/dotnet/api/system.io.textreader.readasync?view=net-7.0&branch=pr-en-us-8745#system-io-textreader-readasync(system-memory((system-char))-system-threading-cancellationtoken)).

Contributes to #7840 

cc @stephentoub

PR #8776, opened 2023-01-10
Add OperationCanceledException exceptions automatically

gewarren:
Used a utility I wrote to automatically add `OperationCanceledException` exceptions with reason "The cancellation token was canceled" to methods that accept a `CancellationToken` parameter.

[Example preview](https://review.learn.microsoft.com/en-us/dotnet/api/system.io.textreader.readasync?view=net-7.0&branch=pr-en-us-8776#system-io-textreader-readasync(system-memory((system-char))-system-threading-cancellationtoken)).

Contributes to #7840.
Replaces #8745.

cc @stephentoub


stephentoub:
Thanks, @gewarren, but I'm confused.  Typically `<exception/>` indicates exceptions that are thrown out of a method, but in the case of OperationCanceledException, with an async / Task-returning method, it should always be stored into the resulting task.

gewarren:
> Thanks, @gewarren, but I'm confused. Typically `<exception/>` indicates exceptions that are thrown out of a method, but in the case of OperationCanceledException, with an async / Task-returning method, it should always be stored into the resulting task.

@stephentoub Yes, I agree with that, but in my notes from our meeting in November this was one of the things we decided to do as a stop-gap measure. I'm happy to close this, or perhaps we can still use the `<exception>` tag but note that it's a stored exception? Something like:

`<exception cref="T:System.OperationCanceledException">The cancellation token was canceled. This exception is stored into the returned <see cref="T:System.Threading.Tasks.Task`1" />.</exception>`

stephentoub:
> perhaps we can still use the <exception> tag but note that it's a stored exception? Something like:

Did we discuss just doing that for all the exceptions, e.g. copyimg from the sync counterpart and adding that verbiage in the exception tag itself?

gewarren:
> > perhaps we can still use the  tag but note that it's a stored exception? Something like:
> 
> Did we discuss just doing that for all the exceptions, e.g. copying from the sync counterpart and adding that verbiage in the exception tag itself? I wonder if that'd be sufficient to just call this whole thing done.

@stephentoub We didn't discuss that, but it seems like a good idea to me. Thoughts @antonfirsov @BillWagner? We could run the tool periodically to catch any new exceptions that were added to the synchronous method.

BillWagner:
> We could run the tool periodically to catch any new exceptions that were added to the synchronous method.

I like this plan. It keeps us moving forward.

BillWagner (approved):
This is a great improvement. Let's :shipit:

gewarren:
Converted to draft PR until I add in the extra text on the end.

gewarren:
Closing and replacing due to too many conflicts.

PR #8778, opened 2023-01-11
What is a bind added

surfmuggle:
What is a bind added from https://learn.microsoft.com/en-us/previous-versions/windows/desktop/ldap/binding-to-an-ldap-server

## Summary

Currently it is not explained what a bind is. Therefore i added an explanation. Since i copied it from another msft documentation it might need to be adjusted: the part 

> If a connection was created using [ldap_connect](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/api/Winldap/nf-winldap-ldap_connect), and if no binding function is called

 might need to be changed to a method that is relevant for .Net. 

### DIFF
--- a/xml/System.DirectoryServices.Protocols/LdapConnection.xml
+++ b/xml/System.DirectoryServices.Protocols/LdapConnection.xml
@@ -332,11 +332,16 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>The <see cref="M:System.DirectoryServices.Protocols.LdapConnection.Bind" /> method sends an LDAP bind using the current credentials.
-        Binding is the step where the LDAP server authenticates the client and, if the client is successfully authenticated, allows the client access to the LDAP server based on that client's privileges.
-        If a connection was created using ldap_connect, and if no binding function is called, on a LDAP v3 server, you run as anonymous.
-        </summary>
-        <remarks>To be added.</remarks>
+        <summary>The <see cref="M:System.DirectoryServices.Protocols.LdapConnection.Bind" /> method sends an LDAP bind using the current credentials.</summary>
+        <remarks>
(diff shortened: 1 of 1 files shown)


xml/System.DirectoryServices.Protocols/LdapConnection.xml
```diff
@@ -332,7 +332,10 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>The <see cref="M:System.DirectoryServices.Protocols.LdapConnection.Bind" /> method sends an LDAP bind using the current credentials.</summary>
+        <summary>The <see cref="M:System.DirectoryServices.Protocols.LdapConnection.Bind" /> method sends an LDAP bind using the current credentials.
+        Binding is the step where the LDAP server authenticates the client and, if the client is successfully authenticated, allows the client access to the LDAP server based on that client's privileges.
+        If a connection was created using ldap_connect, and if no binding function is called, on a LDAP v3 server, you run as anonymous.
+        </summary>
         <remarks>To be added.</remarks>
```
MSDN-WhiteKnight, line 339:
```suggestion
        <summary>The <see cref="M:System.DirectoryServices.Protocols.LdapConnection.Bind" /> method sends an LDAP bind using the current credentials.</summary>
        <remarks>
          <format type="text/markdown"><![CDATA[
## Remarks

Binding is the step where the LDAP server authenticates the client and, if the client is successfully authenticated, allows the client access to the LDAP server based on that client's privileges.
If a connection was created using `ldap_connect`, and if no binding function is called, on a LDAP v3 server, you run as anonymous.

   ]]></format>
    </remarks>
```

Long details should be put in remarks. Consider also adding link to the relevant section of LDAP documentation.

buyaa-n (approved):
LGTM, thanks you!

PR #8811, opened 2023-01-27
Add IOException to FileStream.Write methods

gewarren:
Contributes to dotnet/docs#32224

### DIFF
--- a/xml/System.IO/FileStream.xml
+++ b/xml/System.IO/FileStream.xml
@@ -4076,6 +4076,7 @@ If the write operation is successful, the position within the file stream advanc
 
           ]]></format>
         </remarks>
+        <exception cref="T:System.IO.IOException">.NET 8 and later versions: The underlying pipe is closed or disconnected.</exception>
       </Docs>
     </Member>
     <Member MemberName="Write">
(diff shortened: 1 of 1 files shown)


xml/System.IO/FileStream.xml
```diff
@@ -4319,9 +4321,6 @@ If the operation is canceled before it completes, the returned task contains the
         <exception cref="T:System.ArgumentOutOfRangeException">
           <paramref name="offset" /> or <paramref name="count" /> is negative.</exception>
         <exception cref="T:System.ArgumentException">The sum of <paramref name="offset" /> and <paramref name="count" /> is larger than the buffer length.</exception>
-        <exception cref="T:System.NotSupportedException">The stream does not support writing.</exception>
```
gewarren, line 4322: I deleted these exceptions because the guidance is that stored exceptions shouldn't be documented as regular (synchronous) exceptions, since the async method still completes and returns something.

adamsitnik (approved):
LGTM, thank you @gewarren !

BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

PR #8813, opened 2023-01-27
Move stored exceptions blurb to remarks

gewarren:
Exceptions paragraph was incorrectly placed under "Examples" instead of "Remarks".
(Also replaces a fwlink with its target.)

Found these using multi-line regex: `Examples([^\]]|\n)+?This method stores`


gewarren:
Thanks @BillWagner

PR #8852, opened 2023-02-20
Update List<T> assembly

joeltankam:
## Summary

`List<T>` is documented to be in `System.Collections.dll` assembly, while it is in `System.Private.CoreLib`.

```csharp
Console.WriteLine(typeof(List<>).Assembly.FullName); // System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e
```


MSDN-WhiteKnight:
Documentation specifies reference assembly where the public contract for the type is provided. System.Private.CoreLib contains implementation, but it's "private" for a reason. It should not be publicly documented. Also, XML metadata is autogenerated and should not be edited manually (only contents of `Docs` should).

jeffhandley:
@BillWagner I overlooked this notification. To answer your question, we had a service interruption in the issue-labeler that week that has since been resolved.

PR #8863, opened 2023-02-23
Property CacheMemoryLimit represents megabytes

marshallr12:
## Change description of CacheMemoryLimit from bytes to megabytes.

If you look at the source, the configuration key for the config collection passed in the constructor MemoryCache( string name, NameValueCollection config ) is "cacheMemoryLimitMegabytes."


marshallr12:
Actually, after tracing through the code, the MemoryCache is always configured in megabytes, but when it is stored internally, it is widened to a long. So even though the value can never be larger than int.MaxValue, for some reason when you call this property, it's returned as a long representing bytes. So the property is correct, but its usage is inconsistent with how you configure a MemoryCache.

PR #8866, opened 2023-02-24
System.Text.StringBuilder F# snippets

albert-du:
Adds F# snippets for StringBuilder


BillWagner (approved):
Thanks again @albert-du These are great!

I'll :shipit: now.

/cc @dsyme

PR #8883, opened 2023-03-01
Remove erroneous static reference

333fred:
The docs say that only static methods can have `GeneratedRegexAttribute` applied, but this is [not true](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.RegularExpressions/gen/RegexGenerator.Parser.cs#L120-L124).


stephentoub (approved):
Thanks

PR #6786, opened 2021-06-02
Update Ordering of members in DayOfWeek.xml

ChrisSchaller:
The ordering of the enum values is significant, both in terms of real-world calendar systems and the ordinality of the values. The previous document was not sorted by any obvious means, alphabetical might be acceptable at a stretch, but enums in C# have significant values and those values are expected to define their _natural_ order.

## Summary

Sorted the Enum members according to their value. Previous implementation was ordered in an ambiguous manner:

- Friday: 5
- Monday: 1
- Saturday: 6
- Sunday: 0
- Thursday: 4
- Tuesday: 2
- Wedneday: 3

PR #6749, opened 2021-05-13
Reorder DayOfWeek members

darrenkopp:
## Summary

The [documentation for the DayOfWeek enumeration](https://docs.microsoft.com/en-us/dotnet/api/system.dayofweek?view=net-5.0#fields) currently has a bit of a weird order. It's not based on the underlying value, nor is it based on common order of days as it would appear on a calendar, nor is it alphabetical. 

I believe the changes in this pull request will put the values in the order Sunday, Monday, Tuesday, Wednesday, Thursday, Friday, Saturday. Not sure if that's controversial or not, but an alternate could be done with the week starting on Monday.

PR #8892, opened 2023-03-07
Order days of of week in `DayOfWeek` Enum

xtqqczze:
Order the days of the week in their actual order rather than alphabetically.


gewarren:
Thank you @xtqqczze. We get this request often - see #3852, #4352, #4558, #4725, #6749, #6786. Unfortunately the order is autogenerated and will be overwritten soon even if we merge this :(

gewarren:
This has been fixed!

PR #8893, opened 2023-03-08
Add documentation for ConsoleKey.None and ConsoleModifiers.None

terrajobst:
## Summary

This adds reference documentation for https://github.com/dotnet/runtime/pull/83114.

/cc @stephentoub @carlossanlop

### DIFF
--- a/xml/System/ConsoleKey.xml
+++ b/xml/System/ConsoleKey.xml
@@ -3311,6 +3311,26 @@
         <summary>A constant reserved for future use.</summary>
       </Docs>
     </Member>
+    <Member MemberName="None">
+      <MemberSignature Language="C#" Value="None" />
+      <MemberSignature Language="ILAsm" Value=".field public static literal valuetype System.ConsoleKey None = int32(0)" />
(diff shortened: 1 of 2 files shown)


gewarren:
I think there was a misunderstanding - Stephen was referring to XML comments in the ///, not here. I think we should close this PR since we don't want to publish it before Preview 3.

terrajobst:
@gewarren 

> I think there was a misunderstanding - Stephen was referring to XML comments in the ///, not here. I think we should close this PR since we don't want to publish it before Preview 3.

For some assemblies we currently don't do ///, which includes these two types. How are those being documented then? Or is this just a function of waiting until automation populates the API entry in this repo?

gewarren:
Yes, wait until automation populates the XML files in this docs repo. Even if you're not building documentation files when you compile, Carlos can use his tool to port the comments over from ///. However, he said that he's not going to do that until RC, so if you wanted to add descriptions before RC, you can come back here when Preview 3 ships and update the `<summary>` manually.

terrajobst:
Sounds good, will do. Thanks!

PR #8897, opened 2023-03-10
Improve wording about when exactly an NRE is thrown

biancaghiurutan:
The previous wording was not accurate, inferring that an exception is thrown when the method is called, as opposed to when a member of a null object is referenced. I have also removed the documentation around `ArgumentNullException` as it was confusing and also did not appear in the code it was referencing.

## Summary

Describe your changes here.

### DIFF
--- a/xml/System/NullReferenceException.xml
+++ b/xml/System/NullReferenceException.xml
@@ -135,7 +135,7 @@
      :::code language="fsharp" source="~/snippets/fsharp/System/NullReferenceException/Overview/Array2.fs" id="Snippet9":::
      :::code language="vb" source="~/snippets/visualbasic/VS_Snippets_CLR_System/system.nullreferenceexception.class/vb/Array2.vb" id="Snippet9":::  
   
--   A <xref:System.NullReferenceException> exception can be thrown by a method that is passed `null` in one of its arguments. The example below illustrates this scenario. The method below will throw when a member is called on the `null` object: `names.Add(arrName);`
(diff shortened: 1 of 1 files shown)


biancaghiurutan:
@dotnet-policy-service agree

xml/System/NullReferenceException.xml
```diff
@@ -135,7 +135,7 @@
      :::code language="fsharp" source="~/snippets/fsharp/System/NullReferenceException/Overview/Array2.fs" id="Snippet9":::
      :::code language="vb" source="~/snippets/visualbasic/VS_Snippets_CLR_System/system.nullreferenceexception.class/vb/Array2.vb" id="Snippet9":::  
   
--   A <xref:System.NullReferenceException> exception is thrown by a method that is passed `null`. Some methods validate the arguments that are passed to them. If they do and one of the arguments is `null`, the method throws an <xref:System.ArgumentNullException?displayProperty=nameWithType> exception. Otherwise, it throws a <xref:System.NullReferenceException> exception. The following example illustrates this scenario.  
+-   A <xref:System.NullReferenceException> exception can be thrown by a method that is passed `null` in one of its arguments. The example below illustrates this scenario. The method below will throw when a member is called on the `null` object: `names.Add(arrName);`
```
gewarren, line 138:
```suggestion
-   A <xref:System.NullReferenceException> exception can be thrown by a method when it accesses a member of one of its arguments, but that argument is `null`. The `PopulateNames` method in the following example throws the exception at the line `names.Add(arrName);`.
```

gewarren (approved):
Thanks @biancaghiurutan. I updated the wording slightly - what do you think?

PR #8908, opened 2023-03-16
make the return value comment a bit simpler

kung-foo:
## Summary

Documentation has unnecessary indirection. This PR removes it.


kung-foo:
@dotnet-policy-service agree

gewarren (approved):

PR #8911, opened 2023-03-18
fix typo in Examples

ABA2396:
## Summary

Description:
Fixed an error in the WPF documentation where the event "PreviewMouseLeftButtonDownEvent" was incorrectly identified as the event that fires when the right mouse button is pressed. The correct event name is "PreviewMouseRightButtonDownEvent", and I have corrected this throughout the affected documentation.

Reason for change:
The error in the documentation could cause confusion for developers using WPF controls and trying to identify the correct event to handle when the right mouse button is pressed. This change ensures that the documentation is accurate and helpful for developers.

References:
documentation: https://learn.microsoft.com/en-us/dotnet/api/system.windows.eventmanager.registerclasshandler

### DIFF
--- a/xml/System.Windows/EventManager.xml
+++ b/xml/System.Windows/EventManager.xml
@@ -166,7 +166,7 @@
    
   
 ## Examples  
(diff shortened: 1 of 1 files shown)


ABA2396:
@dotnet-policy-service agree

PR #8915, opened 2023-03-20
Correct doc comment for IMultiplyOperators returns

IDisposable:
## Summary

The copy-pasted and inherited documentation for the multiplication operation was incorrect. There were several instances of either _divided-by_ or _divided by_ that should have been _multiplied-by_ or _multiplied by_ (respectively).

Replaced text in XML documentation for System.Half, System.Int128, System.UInt128, System.Numerics.IMultiplyOperators, and System.Runtime.InteropServices.NFloat.

This reflects the matching changes in runtime PR https://github.com/dotnet/runtime/pull/83693 to the `IMultiplyOperators.cs` and `NFloat.cs`.

Fixes https://github.com/dotnet/runtime/issues/80521

### DIFF
--- a/xml/System.Numerics/IMultiplyOperators`3.xml
+++ b/xml/System.Numerics/IMultiplyOperators`3.xml
@@ -52,7 +52,7 @@
         <param name="left">The value which <paramref name="right" /> multiplies.</param>
         <param name="right">The value which multiplies <paramref name="left" />.</param>
         <summary>Multiplies two values together to compute their product.</summary>
-        <returns>The product of <paramref name="left" /> divided by <paramref name="right" />.</returns>
+        <returns>The product of <paramref name="left" /> multiplied by <paramref name="right" />.</returns>
         <remarks>To be added.</remarks>
         <exception cref="T:System.OverflowException">The product of <paramref name="left" /> multiplied by <paramref name="right" /> is not representable by <typeparamref name="TResult" />.</exception>
       </Docs>
--- a/xml/System.Runtime.InteropServices/NFloat.xml
+++ b/xml/System.Runtime.InteropServices/NFloat.xml
@@ -4290,7 +4290,7 @@ For <xref:System.Numerics.IFloatingPoint%601> this method matches the IEEE 754:2
         <param name="left">The value that <paramref name="right" /> multiplies.</param>
         <param name="right">The value that multiplies <paramref name="left" />.</param>
         <summary>Multiplies two values together to compute their product.</summary>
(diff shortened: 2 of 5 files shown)


IDisposable:
I noticed while scanning the documentation that there's a **WHOLE lot** of stuff missing the content that should have been automatically pulled from the source documentation but currently reads `To be added`. Is there a plan in place to sync this stuff up via an automated tool, or would there be value in cleaning some of that up by hand?

AaronRobinsonMSFT (approved):

IDisposable:
Hey @AaronRobinsonMSFT, any comment about https://github.com/dotnet/dotnet-api-docs/pull/8915#issuecomment-1476948541

AaronRobinsonMSFT:
> Thanks for merging! @AaronRobinsonMSFT, any comment about [#8915 (comment)](https://github.com/dotnet/dotnet-api-docs/pull/8915#issuecomment-1476948541)

Oops. Very sorry to have missed that. Yes, there was tooling, unsure if it still exists. @carlossanlop used to be the contact for that. Nowadays we get sync PRs that attempt to synchronize then. Unsure if they are still active.

IDisposable:
Maybe I'll do a sweep then ;)

carlossanlop:
Thanks for your contribution, @IDisposable . Note that dotnet-api-docs is considered our source of truth for documentation, not triple slash. It currently depends on area owners if they want to keep the docs synced in both places, but the default behavior is to have the correct documentation here.

If you'd like to continue contributing with similar changes, my suggestion is that you only do it here, so you don't work double.

We will eventually convert runtime triple slash into the source of truth (after we make the correct adaptations in our infra) and contributors won't have to do what you did of correcting both places.

**Regarding your question**: we run a tool during Preview7 and RC1 to sync the two repos, to ensure we ship the new .NET version with the latest docs. Do you know when these APIs were introduced to source? If they were introduced in .NET 7, then the tool might have a bug that prevented them from getting ported from triple slash and into dotnet-api-docs. But if they were recently introduced in .NET 8, then it's expected that the dotnet-api-docs docs are empty, we will port everything in Preview7, and my recommendation is that you don't do all this manual work yet. We will eventually get it all ported.

IDisposable:
Maybe I won't do a sweep then ;)

As to the specific change, I literally just picked up the first easy "help wanted" that I saw and started down the merry path.

The PR that added this incorrect /// comment was https://github.com/dotnet/runtime/pull/54650 and was merged July 2, 2021.
https://github.com/dotnet/runtime/pull/54650/files#diff-5bef0c49804fbe50148c64a5d94d9aab05e8fd3ea4929080c1b97d7cb6b24b5bR23

PR #8918, opened 2023-03-21
Fix HMAC docs to use hash output size, not block size, for key adjustment

vcsjones:
Some feedback from a mailing list that the HMAC documentation was incorrect. It uses the block size of the HMAC to describe the output of the hash function. Additionally, MD5 incorrectly noted the use of SHA-1 when it meant MD5.

/cc @bartonjs

### DIFF
--- a/xml/System.Security.Cryptography/HMACMD5.xml
+++ b/xml/System.Security.Cryptography/HMACMD5.xml
@@ -191,7 +191,7 @@
         <Parameter Name="key" Type="System.Byte[]" />
       </Parameters>
       <Docs>
-        <param name="key">The secret key HMAC computation. The key can be any length, but if it is more than 64 bytes long it will be hashed (using MD5) to derive a 16-byte key. Therefore, the recommended size of the secret key is 64 bytes.</param>
+        <param name="key">The secret key for HMAC computation. The key can be any length, but if it is more than 64 bytes long it will be hashed (using MD5) to derive a 16-byte key. Therefore, the recommended size of the secret key is 64 bytes.</param>
         <summary>Initializes a new instance of the <see cref="T:System.Security.Cryptography.HMACMD5" /> class by using the specified key.</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[  


xml/System.Security.Cryptography/HMACMD5.xml
```diff
@@ -191,7 +191,7 @@
         <Parameter Name="key" Type="System.Byte[]" />
       </Parameters>
       <Docs>
-        <param name="key">The secret key for <see cref="T:System.Security.Cryptography.HMACMD5" /> encryption. The key can be any length, but if it is more than 64 bytes long it will be hashed (using SHA-1) to derive a 64-byte key. Therefore, the recommended size of the secret key is 64 bytes.</param>
+        <param name="key">The secret key for <see cref="T:System.Security.Cryptography.HMACMD5" /> encryption. The key can be any length, but if it is more than 64 bytes long it will be hashed (using MD5) to derive a 16-byte key. Therefore, the recommended size of the secret key is 64 bytes.</param>
```
bartonjs, line 194: While we're editing these, can we change all of the `<see cref=T:THETYPEIMON"/> encryption" to "HMAC computation".  Since MAC != encryption, and the self-reference doesn't seem worthwhile here.

vcsjones: Yeah. Good idea.

xml/System.Security.Cryptography/HMACMD5.xml
```diff
@@ -191,7 +191,7 @@
         <Parameter Name="key" Type="System.Byte[]" />
       </Parameters>
       <Docs>
-        <param name="key">The secret key for <see cref="T:System.Security.Cryptography.HMACMD5" /> encryption. The key can be any length, but if it is more than 64 bytes long it will be hashed (using SHA-1) to derive a 64-byte key. Therefore, the recommended size of the secret key is 64 bytes.</param>
+        <param name="key">The secret key HMAC computation. The key can be any length, but if it is more than 64 bytes long it will be hashed (using MD5) to derive a 16-byte key. Therefore, the recommended size of the secret key is 64 bytes.</param>
```
bartonjs, line 194:
```suggestion
        <param name="key">The secret key for HMAC computation. The key can be any length, but if it is more than 64 bytes long it will be hashed (using MD5) to derive a 16-byte key. Therefore, the recommended size of the secret key is 64 bytes.</param>
```

bartonjs (approved):

xml/System.Security.Cryptography/HMACMD5.xml
vcsjones: [sigh] oops. Thanks!

PR #8922, opened 2023-03-23
Fix SystemParameters.MenuDropAlignment's description

### LINKED ISSUE #7920: Description of SystemParameters.MenuDropAlignment is backwards
The documentation states the return value means "true when left aligned, otherwise false"

https://docs.microsoft.com/en-us/dotnet/api/system.windows.systemparameters.menudropalignment?view=windowsdesktop-6.0

This is backwards.    True means that the menus are _right_ aligned, as described in the Win32 API system parameter SPI_GETMENUDROPALIGNMENT to which this parameter corresponds

https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-systemparametersinfoa

m2jean:
## Summary
The current description indicates that when the return value is true, it means the menus are left-aligned. However, it's the opposite for SPI_GETMENUDROPALIGNMENT for the Win32 API [SystemParametersInfoA](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-systemparametersinfoa), and the [source code](https://source.dot.net/#PresentationFramework/System/Windows/SystemParameters.cs,2e88725c97fc3540) doesn't do any negation.

Fixes #7920

### DIFF
--- a/xml/System.Windows/SystemParameters.xml
+++ b/xml/System.Windows/SystemParameters.xml
@@ -5087,7 +5087,7 @@
       <Docs>
         <summary>Gets a value indicating whether pop-up menus are left-aligned or right-aligned, relative to the corresponding menu item.</summary>
         <value>
-          <see langword="true" /> if left-aligned; otherwise, <see langword="false" />.</value>
+          <see langword="true" /> if right-aligned; otherwise, <see langword="false" />.</value>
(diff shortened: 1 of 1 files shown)


m2jean:
@dotnet-policy-service agree

PR #8957, opened 2023-04-12
Update strmreader read2.cpp

jontho-microsoft-com:
.NET source code clearly shows Read() may not return a fixed number of characters, but only what's currently available on the stream. ReadBlock returns the stated number of characters and only returns fewer if end of stream

## Summary

Describe your changes here.


BillWagner:
closing as abandoned

PR #8959, opened 2023-04-12
Update strmreader read2.vb

jontho-microsoft-com:
.NET source code clearly shows Read() may not return a fixed number of characters, but only what's currently available on the stream. ReadBlock returns the stated number of characters and only returns fewer if end of stream

## Summary

Describe your changes here.

### DIFF
--- a/snippets/visualbasic/VS_Snippets_CLR/StrmReader Read2/VB/strmreader read2.vb
+++ b/snippets/visualbasic/VS_Snippets_CLR/StrmReader Read2/VB/strmreader read2.vb
@@ -24,7 +24,7 @@ Public Class Test
             Do While sr.Peek() >= 0
                 'This is an arbitrary size for this example.
                 Dim c(5) As Char
-                sr.Read(c, 0, c.Length)
+                sr.ReadBlock(c, 0, c.Length)
                 'The output will look odd, because
                 'only five characters are read at a time.
                 Console.WriteLine(c)


jontho-microsoft-com:
Hi Adam,

Actually, I have several changes I mean to do in a pull, and frankly, I find the explanation of how to work with multiple files as a single file set using the online tool with drafts very confusing. I've got some other related changes I need to make to the docs to make things holistic. My objective is to correct the docs such that they can be used with correct information that doesn't result in more support cases. The fascinating thing is these documentation conflicts have been present since the earliest .NET docs, and it appears I got the purple squirrel scenario of someone working with the ZipArchive classes and their forms of streams.

I've been working with a local git repository for changes, since that's more clear-cut in my mind how that is supposed to work.

Any guidance you can provide me for how to do this in the most efficient/fastest process is greatly welcomed, as I'm certain that we'll have the occasional developer support case in the future where we hit this.

In particular, after I've made the final changes I wish to make in the local git repository, I'd like to:


  1.  Drop the drafts already there
  2.  Send over the repo I have (I forgot what the requirements are there)
  3.  Do the PR and get confirmation ASAP, as I have a requirement to do this while recording labor, and get the associated case closed (internal detail).
Thanks!







Jonathan Thompson
Escalation Engineer
Windows SDK Team
Phone: +1 (425) 722-6428

adamsitnik:
@jontho-microsoft-com If I were you I would push my changes, create a new PR with all fixes and close the existing draft PRs.

PR #8965, opened 2023-04-17
SortedSet.Overlaps is O (n log m) in the worst case

gurustron:
[Current implementation](https://github.com/dotnet/runtime/blob/add51b82baf268c3678bdd124480a46b89d72ce4/src/libraries/System.Collections/src/System/Collections/Generic/SortedSet.cs#L1319) performs iteration over `other` which calls `Contains`:

 ```c#
foreach (T item in other)
{
	if (Contains(item))
	{
		return true;
	}
}
```

And Contains is `O(log m)` ([docs](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.sortedset-1.contains?view=net-6.0#remarks), [implementation](https://github.com/dotnet/runtime/blob/add51b82baf268c3678bdd124480a46b89d72ce4/src/libraries/System.Collections/src/System/Collections/Generic/SortedSet.cs#L693))

## Summary

Fixed docs to current implementation

### DIFF
--- a/xml/System.Collections.Generic/SortedSet`1.xml
+++ b/xml/System.Collections.Generic/SortedSet`1.xml
@@ -1707,7 +1707,7 @@
 ## Remarks  
  Any duplicate elements in `other` are ignored.  
   
- This method is an `O(n)` operation, where `n` is the number of elements in `other`.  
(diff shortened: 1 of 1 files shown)


gurustron:
JIC from [this SO question](https://stackoverflow.com/q/76038138/2501279)

PR #8976, opened 2023-04-21
Fix typo, `callee` -> `caller`

Dynesshely:
## Summary

In `CallingConvention Enum`, `callee` should be `caller`


jkoritzinsky:
In Stdcall, the callee does clean the stack. That's the primary difference between it and Cdecl. This is not a typo.

Dynesshely:
Thanks, I made a mistake.

PR #8977, opened 2023-04-21
Add note that Uri.Query and Fragment include the leading delimiter

MihaZupan:
Contributes to https://github.com/dotnet/runtime/issues/84955.

It's useful if this info shows up in IntelliSense to help clear up any confusion.


MattMcL4475:
@MihaZupan thank you for adding this

gewarren (approved):

PR #8983, opened 2023-04-25
Delete ns-System.Numerics.Tensors.xml

carlossanlop:
@dotnet/area-system-numerics-tensors

These APIs are not customer facing. We don't have xmls for the APIs, so we can remove the namespace file (assuming the CI does not complain).


gewarren (approved):
Thanks @carlossanlop

PR #8985, opened 2023-04-26
Add RFC 3986 non-compliance message to Query and Fragment

MattMcL4475:
## Summary

Adds additional context to the developer to be aware that these properties do not comply with the industry standard RFC 3986. 
 More information here: https://github.com/dotnet/runtime/issues/84955


MihaZupan (changes requested):
As discussed in https://github.com/dotnet/runtime/issues/84955, we feel that including this information in the summary would lead to more confusion, but we would accept a clarification in the remarks section.

MihaZupan:
Replaced by #12002

PR #8989, opened 2023-04-27
Emphasise further the possibly-unexpected default behaviour of StringComparison

Smaug123:
## Summary

Add a couple of words around the API functions which take `StringComparison`.

### Rationale

The current documentation could be read (very much incorrectly!) as:

> We recommend that you avoid calling comparison methods that substitute default values, and instead call methods that require parameters to be explicitly specified. Here is an example of an explicit specification where you override the default behaviour: to compare strings using the string comparison rules of the current culture, call the `System.String.EndsWith(System.String, System.StringComparison)` overload.

This is almost exactly the opposite of what the documentation intends to say, which is:

> The default behaviour is to compare strings using the string comparison rules of the current culture. Do not use this default behaviour. Instead, explicitly signal your intent.

I believe my proposed wording uses five additional words to rule out the incorrect interpretation. (I consider it crucial for readers to read _Best Practices for Using Strings_; I believe more tentatively that this change indicates "here be dragons" and thereby makes it more likely that a reader clicks through to it.)

### DIFF
--- a/xml/System/String.xml
+++ b/xml/System/String.xml
@@ -5074,7 +5074,7 @@ This method performs an ordinal (case-sensitive and culture-insensitive) compari
         <exception cref="T:System.ArgumentNullException">
           <paramref name="value" /> is <see langword="null" />.</exception>
         <block subset="none" type="usage">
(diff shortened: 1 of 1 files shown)


tarekgh:
@gewarren @MSDN-WhiteKnight this change is good to me if you think the wording is good.

gewarren (approved):

PR #8997, opened 2023-05-05
[ConstructorArgumentAttribute] fix incorrect attribution Property vs. Object

glenn-slayden:
As written, the summary line indicated the wrong target entity for the `System.Windows.Markup.ConstructorArgumentAttribute`. The attribute must be placed on the Property that represents the constructor argument, not the owning object itself.


glenn-slayden:
@dotnet-policy-service agree

glenn-slayden:
Note that the patently wrong wording in the original is "...that a property of the specified name supplies construction information."

No, the name that must be specified indicates the name of a constructor argument, not the name of a property.

PR #9008, opened 2023-05-13
Add note about Monitor thread affinity

ericmutta:
The [remarks section](https://learn.microsoft.com/en-us/dotnet/api/system.threading.monitor?view=net-7.0#remarks) of the Monitor class does not mention the fact that the Monitor class has thread affinity.

That important information is found in a [separate article on synchronization primitives](https://learn.microsoft.com/en-us/dotnet/standard/threading/overview-of-synchronization-primitives#monitor-class) that is easy to miss.

I copied the relevant sentence and just tacked it onto the end of the paragraph in the Remarks section (it can be moved anywhere in the article, I just feel it's important to have it appear early on in the Remarks section because that's where I came looking for it but couldn't find it).


ericmutta:
Any chance this could be reviewed?

kouvel (approved):
LGTM, thanks!

ericmutta:
Many thanks @kouvel for following up 👍

PR #9009, opened 2023-05-15
Fix snippet - ConsoleKey1.cs

MichaelDeutschCoding:
Fix the sample output for the snippet in ConsoleKey1.cs to match the code.

## Summary
The sample output for the snippet did not match the described code (the code would print "together with" while the sample output said "along with").
- Fixed the sample output to match the snippet code
- Changed to use `Enum.HasFlag()` instead of bit masking for easier readability


BillWagner:
Hi @MichaelDeutschCoding 

I looked at the changes, and these look great to me. Once you accept the Contributor License Agreement (CLA), I'll :shipit:

MichaelDeutschCoding:
@dotnet-policy-service agree

PR #9015, opened 2023-05-17
Update global ms.date

gewarren:
While ms.date isn't shown on the rendered page, it is in the metadata and is also shown in offline books, so I think we should update it each time we refresh the API docs.


BillWagner (approved):
This is a really good idea @gewarren 

Let's :shipit:

PR #9017, opened 2023-05-18
Fix documentation for FileInfo.IsReadOnly property

### LINKED ISSUE #8979: FileInfo.IsReadOnly is true when file does not exist
### Overview
The state of the `FileInfo.IsReadOnly` property is not what I would expect for a file that doesn't exist. I'm assuming that for some reason this is the desired behavior and is therefore just an issue with the documentation. Or it could also be an actual bug. But either way, the behavior is definitely not accurately reflected by the current documentation.

### Current Documentation
Current documentation reads:
> `true` if the current file is read only; otherwise, `false`.

And under exceptions it also states that it throws a FileNotFoundException:
> The file described by the current [FileInfo](https://learn.microsoft.com/en-us/dotnet/api/system.io.fileinfo?view=net-7.0) object could not be found.

### Demonstration
Yet, I've found this to not be the case. 

A very simple repro:
```c#
var fileInfo = new FileInfo("some path to a file that does not exist");
Console.WriteLine($"Is ReadOnly: {fileInfo.IsReadOnly}"); // prints True
``` 
I would expect it to either print false (the default value for any boolean property) or throw a `FileNotFoundException` (as it does if one tries to access other properties of the FileInfo object, such as `Length`).

### Suggested Fix
Either amend the actual code to set the default to `false` or amend the documentation to clearly indicate that the default for this property is `true`.

MichaelDeutschCoding:
## Summary
`FileInfo.IsReadOnly` returns `true` for a file that does not exist. This is unusual behavior (which may be a bug) that is not documented accordingly.

Fixes #8979

### DIFF
--- a/snippets/cpp/VS_Snippets_CLR/IO.FileInfo.isReadOnly/cpp/sample.cpp
+++ b/snippets/cpp/VS_Snippets_CLR/IO.FileInfo.isReadOnly/cpp/sample.cpp
@@ -32,6 +32,8 @@ int main()
 
     // Delete the temporary file.
     fInfo->Delete();
+
+    return 0;
 };
 
 // This code produces output similar to the following,


snippets/csharp/System.IO/FileInfo/IsReadOnly/sample.cs
```diff
@@ -2,60 +2,55 @@
...
+        string fileName = @"c:\test.xml";
 
-            string FileName = @"c:\test.xml";
+        // Get the read-only value for a file.
+        bool isReadOnly = IsFileReadOnly(fileName);
 
-            // Get the read-only value for a file.
-            bool isReadOnly = IsFileReadOnly(FileName);
+        // Display whether the file is read-only.
+         Console.WriteLine("The file read-only value for " + fileName + " is: " + isReadOnly);
```
adamsitnik, line 15: nit: extra space

```suggestion
        // Display whether the file is read-only.
        Console.WriteLine("The file read-only value for " + fileName + " is: " + isReadOnly);
```

snippets/csharp/System.IO/FileInfo/IsReadOnly/sample.cs
```diff
@@ -2,60 +2,55 @@
...
-//The file read-only value for c:\test.xml is: True
-//Changing the read-only value for c:\test.xml to true.
-//The file read-only value for c:\test.xml is: True
+// The file read-only value for c:\test.xml is: True
+// Changing the read-only value for c:\test.xml to false.
+// The file read-only value for c:\test.xml is: False
 //
-//</SNIPPET1>
\ No newline at end of file
+//</SNIPPET1>
```
adamsitnik, line 56: thank you for updating the example! While looking at it now I got to a conclusion that we should improve it further. We should make it cross-platform (not Windows-specific path) and working (create the file before trying to change the readonly property).

Here is my proposal:

```cs
//<SNIPPET1>
using System;
using System.IO;

public class FileExample
{
    public static void Main()
    {
        // combine a path to temporary file
        string filePath = Path.Combine(Path.GetTempPath(), "readonly_example.txt");

        // create the file
        File.WriteAllText(filePath, "contents");

        // Create a new FileInfo object.
        FileInfo fInfo = new FileInfo(filePath);

        // Get the read-only value for a file.
        bool isReadOnly = fInfo.IsReadOnly;

        // Display whether the file is read-only.
        Console.WriteLine($"The file read-only value for '{fInfo.Name}' is {isReadOnly}");

        // Set the file to read-only.        
        Console.WriteLine($"Changing the read-only value for '{fInfo.Name}' to true.");
        fInfo.IsReadOnly = true;

        // Get the read-only value for a file.
        isReadOnly = fInfo.IsReadOnly;

        // Display that the file is now read-only.
        Console.WriteLine($"The file read-only value for '{fInfo.Name}' is {isReadOnly}");

        // make the file mutable again so it can be deleted
        fInfo.IsReadOnly = false;
        // delete the temporary file
        File.Delete(filePath);
    }
}

// This code produces output similar to the following,
// though results may vary based on the computer, file structure, etc.:
//
// The file read-only value for 'readonly_example.txt' is False
// Changing the read-only value for 'readonly_example.txt' to true.
// The file read-only value for 'readonly_example.txt' is True
//
//</SNIPPET1>
```

Thoughts?

MichaelDeutschCoding: Personally, I like the more comprehensive (and robust) approach. I was initially hesitant to make too many changes for a few reasons:
- Didn't want the various languages' samples to diverge significantly (unless I update all the snippets)
- Wasn't sure if this snippet works in tandem with any other tutorial or documentation that assumes a certain state
- Ensuring the sample is as terse and relevant to the current property as possible

In terms of the string interpolation, I really wanted to change that! But I noticed that every sample uses this (ugly!) string concatenation when writing to the console. I thought maybe it runs in some unique context or required backwards compatibility that precludes the common `$` interpolated strings.

My one comment about the snippet you shared: would it be possible to remove the `File.WriteAllText()` line? It just feels out of scope for this sample (if it's feasible without it). Otherwise, that looks great!

MichaelDeutschCoding: Do we need to additionally:
- Update the other language snippets to match this process?
- Update the documentation's summary of what the snippet does? (currently states that `The following example uses the <xref:System.IO.FileInfo.IsReadOnly> property to mark a file as read only and then mark it as read-write.`)

MichaelDeutschCoding: > My one comment about the snippet you shared: would it be possible to remove the File.WriteAllText() line? It just feels out of scope for this sample (if it's feasible without it). Otherwise, that looks great!

Though the first alternative I thought of (using `File.Create(filePath)`) would require immediately calling `.Dispose();`; otherwise the stream would still be left open. So I don't know if that's really any better...

adamsitnik: > My one comment about the snippet you shared: would it be possible to remove the File.WriteAllText() line? I

Replacing `File.WriteAllText` with `File.Create(filePath).Dispose` sounds good to me (the intention is definitely more clear)

> I'm happy to implement both changes (though I won't actually be able to run any C++ samples myself locally to verify it works properly)

It seems that there are only C# sample for `ReadOnly` so we are safe

image

> Update the documentation's summary of what the snippet does? 

Yes, please. Great catch BTW!

MichaelDeutschCoding: There are snippets for:
 - C++ https://github.com/dotnet/dotnet-api-docs/tree/main/snippets/cpp/VS_Snippets_CLR/IO.FileInfo.isReadOnly/cpp
 - VisualBasic https://github.com/dotnet/dotnet-api-docs/blob/main/snippets/visualbasic/VS_Snippets_CLR/IO.FileInfo.isReadOnly/VB/sample.vb

Also, I'm realizing what would probably be most succinct and strictly relevant would be to just use `Path.GetTempFileName()` which creates the 0-byte file. Then we don't have to worry about writing content, or Disposing the FileStream handle or anything. 👍

adamsitnik: > Also, I'm realizing what would probably be most succinct and strictly relevant would be to just use Path.GetTempFileName()

Great idea!

> There are snippets for:

Another great catch. Once you provide updated C# version I can provide the C++ part

MichaelDeutschCoding: Updated the C# and VisualBasic samples. If you can, please update C++ snippet to match the flow of the others.

MichaelDeutschCoding:
@adamsitnik , could you help me identify why the snippets-build failed? It says there should be a project or solution file. But there wasn't one previously either, so I'm a bit confused as to why it's failing now. Should I just remove the `Module FileExample` line? Would that help?

MichaelDeutschCoding:
Thanks for crossing the finish line with this @adamsitnik! 👍

PR #9022, opened 2023-05-22
Update Point.xml

activatekillswitch:
Corrected the example comment

## Summary

Example comment had the Vector and Point the wrong way around


activatekillswitch:
@dotnet-policy-service agree

PR #9030, opened 2023-05-29
Update XmlReader.xml

ibengeu:
## Summary

Updated the W3C Rec Links due to current being outdated.


ibengeu:
@dotnet-policy-service agree

BillWagner:
ping @dotnet/area-system-xml 

This PR looks good to me if you intended to reference the latest version of the standard. If the intent was to pin to the 2006 version, this should be closed. Can you comment?

krwq (approved):

krwq (changes requested):
Sorry, badly clicked approve - the link changes it so that it's always latest version of 1.0 (fifth edition) which sounds fine but I'm not quite sure what we currently follow all changes between linked version and 5th edition. It would be good to verify what changes are between them and if we're up to date. If we're up to date I'd prefer we linked to the specific edition (i.e. 5th) we follow. Link for 5th edition is: https://www.w3.org/TR/2008/REC-xml-20081126/ but we need to validate we are actually up to date (i.e. diff or something and make sure our implementation actually follows that - I believe yes but we should validate)

krwq:
In the very same file you're editing here it mentions we conform to 4th edition. It might still be worth diffing and validating as that comment might be out of date.

gewarren (approved):
Approving based on the fact that we say we conform to the 4th edition (as a sort of disclaimer) and also you get a nasty popup if you navigate to an older edition.

gewarren:
This content was moved to [a file in a different repo](https://github.com/dotnet/docs/blob/main/docs/fundamentals/runtime-libraries/system-xml-xmlreader.md), so I'm going to close this PR.

PR #9034, opened 2023-05-31
Update DataGrid.xml

miloush:
## Summary

Clarification on when CanUserDeleteRows has no effect.

### DIFF
--- a/xml/System.Windows.Controls/DataGrid.xml
+++ b/xml/System.Windows.Controls/DataGrid.xml
@@ -808,7 +808,7 @@
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- By default, the user can delete rows by selecting one or more rows and pressing the DELETE key. Deleting a row removes the item that the row represents from the <xref:System.Windows.Controls.ItemsControl.ItemsSource%2A>. If the <xref:System.Windows.Controls.ItemsControl.ItemsSource%2A> does not allow deletions, or if the <see cref="T:System.Windows.Controls.DataGrid" /> is read-only, the user cannot delete rows even if this property is set to `true`.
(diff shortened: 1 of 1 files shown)


xml/System.Windows.Controls/DataGrid.xml
```diff
@@ -808,7 +808,7 @@
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- By default, the user can delete rows by selecting one or more rows and pressing the DELETE key. Deleting a row removes the item that the row represents from the <xref:System.Windows.Controls.ItemsControl.ItemsSource%2A>. If the <xref:System.Windows.Controls.ItemsControl.ItemsSource%2A> does not allow deletions, the user cannot delete rows even if this property is set to `true`.
+ By default, the user can delete rows by selecting one or more rows and pressing the DELETE key. Deleting a row removes the item that the row represents from the <xref:System.Windows.Controls.ItemsControl.ItemsSource%2A>. If the <xref:System.Windows.Controls.ItemsControl.ItemsSource%2A> does not allow deletions, or if the <see cref="T:System.Windows.Controls.DataGrid" /> is read-only, the user cannot delete rows even if this property is set to `true`.
```
gewarren, line 811:
```suggestion
 By default, the user can delete rows by selecting one or more rows and pressing the <kbd>Delete</kbd> key. Deleting a row removes the item that the row represents from the <xref:System.Windows.Controls.ItemsControl.ItemsSource%2A>. If the <xref:System.Windows.Controls.ItemsControl.ItemsSource%2A> does not allow deletions or if the <xref:System.Windows.Controls.DataGrid> is read-only, the user cannot delete rows even if this property is set to `true`.
```

PR #9057, opened 2023-06-12
Fix formatting of `VirtualizingStackPanel` Examples

89netraM:
I've fixed the formatting of the `VirtualizingStackPanel` example, so that all opening and closing pairs are aligned and child elements are indented further than their parents. I looked at some other examples and think that two spaces as indentation is correct.


89netraM:
@dotnet-policy-service agree

gewarren (approved):
Thanks @89netraM

PR #9063, opened 2023-06-13
Update CODEOWNERS - remove jkotalik

gewarren:
jkotalik causes an error in the CODEOWNERS file. jkotalik is no longer at Microsoft.


BillWagner (approved):
This LGTM. Let's :shipit:

PR #9072, opened 2023-06-15
Update X509Certificate2.xml

Ghostbird:
## Summary

I changed the reference to point to `X509Certificate2.CreateFromPemFile(string certPemFilePath)` which does work for the use case mentioned here. The original `new X509Certificate2(string fileName)` didn't work for me, as it didn't understand the PEM file.

I don't think it's different in .NET 8, but I encountered this issue in .NET 7

### DIFF
--- a/xml/System.Security.Cryptography.X509Certificates/X509Certificate2.xml
+++ b/xml/System.Security.Cryptography.X509Certificates/X509Certificate2.xml
@@ -1516,7 +1516,7 @@ For PEM-encoded certificates with a private key, use <xref:System.Security.Crypt
 
 For PEM-encoded certificates in a file without a private key, use <xref:System.Security.Cryptography.X509Certificates.X509Certificate2.%23ctor(System.String)>.
 
-For PEM-encoded certificates in a file with a private key, use <xref:System.Security.Cryptography.X509Certificates.X509Certificate2.CreateFromPemFile(System.String)>.
+For PEM-encoded certificates in a file with a private key, use <xref:System.Security.Cryptography.X509Certificates.X509Certificate2.CreateFromPemFile(System.String, System.String)> and omit the second argument.
           ]]></format>
         </remarks>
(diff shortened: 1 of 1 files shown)


Ghostbird:
@dotnet-policy-service agree company="IntarBV"

Ghostbird:
Seems I messed up the reference. I'll revisit this later.

xml/System.Security.Cryptography.X509Certificates/X509Certificate2.xml
```diff
@@ -1514,7 +1514,7 @@ This loads the first well-formed PEM found with a CERTIFICATE label.
             
 For PEM-encoded certificates with a private key, use <xref:System.Security.Cryptography.X509Certificates.X509Certificate2.CreateFromPem(System.ReadOnlySpan{System.Char},System.ReadOnlySpan{System.Char})>.
 
-For PEM-encoded certificates in a file, use <xref:System.Security.Cryptography.X509Certificates.X509Certificate2.%23ctor(System.String)>.
+For PEM-encoded certificates in a file, use <xref:System.Security.Cryptography.X509Certificates.X509Certificate2.CreateFromPemFile(System.String)>.
```
bartonjs, line 1517: Both the ctor and CreateFromPemFile work for PEM content, but they do different things.

* ctor: Reads out the certificate, ignores the private key
* CreateFromPemFile: Reads out the certificate, finds a matching private key, **fails if there is not a matching private key**

So it's probably better to break this into two different paragraphs, like


```suggestion
For PEM-encoded certificates in a file without a private key, use <xref:System.Security.Cryptography.X509Certificates.X509Certificate2.%23ctor(System.String)>.

For PEM-encoded certificates in a file with a private key, use <xref:System.Security.Cryptography.X509Certificates.X509Certificate2.CreateFromPemFile(System.String)>.
```

Ghostbird:
Closing this, as I can't seem to get this right.

PR #9080, opened 2023-06-16
Improvements around DBNull.Value for DataRowExtensions

### LINKED ISSUE #9077: DataRowExtensions "nullable types" support clarity.
The [DataRowExtensions](/dotnet/dotnet-api-docs/tree/main/xml/System.Data/DataRowExtensions.xml) documentation states that a `DataSet` `does not support nullable types`, however it is really the `DataColumn` object that does not support this.
I recommend that it is changed to say `DataColumn` instead of `DataSet`, because some may think "I am using a DataTable, not a DataSet, so I can use nullable types", which would be untrue.  Note that in some cases, it uses the contraction "doesn't" instead of "does not".

There also seems to be an issue with some of the wording in the Remarks.  For example `If the underlying value in the 'DataSet' is 'Value', the returned nullable type will have a value of null`.  Note that I added single quotes around `Value` and `DataSet` because they are links.  As it turns out, the `Value` link actually points to [system.dbnull.value](https://learn.microsoft.com/en-us/dotnet/api/system.dbnull.value), but as stated, it appears to be confusing.  
The next statement says `If the value of the specified 'DataColumn' is null ...`.  Again `DataColumn` is quoted because it is a link.  This statement seems to be attempting to re-state the DBNull point, but is not doing it in a clear way.
If either of those were clear, then I think only one of them would be needed.

These statements appear multiple times on that page.

roji:
Closes #9077

### DIFF
--- a/xml/System.Data/DataRowExtensions.xml
+++ b/xml/System.Data/DataRowExtensions.xml
@@ -107,9 +107,9 @@
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- The <xref:System.Data.DataSet> class represents **null** values with the <xref:System.DBNull.Value> instance of the <xref:System.DBNull> class.  A Language-Integrated Query (LINQ) expression that accessed a column with a **null** value would generate a <xref:System.InvalidCastException> at run time. Additionally, <xref:System.Data.DataSet> does not support nullable types. The <xref:System.Data.DataRowExtensions.Field%2A> method provides support for accessing columns as nullable types.  If the underlying value in the <xref:System.Data.DataSet> is <xref:System.DBNull.Value>, the returned nullable type will have a value of **null**.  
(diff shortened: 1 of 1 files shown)


David-Engel (approved):

JRahnama:
LGTM, just a minor question, : according to the comment from VS team in issue [#2013](https://github.com/dotnet/SqlClient/issues/2013) at SqlClient `cref` should be used instead of `xref`. That is going to be a lot of work to updated the entire documentation, but we can stop adding more to it I think. Again this is going to be huge. For example, instead of `CDATA`, each paragraph should be wrapped in `<para>[content]</para>` tags and some other modifications. I am just wondering if we should start following that path to prevent more changes to convert in future?

roji:
@JRahnama I'm not an expert on the doc system - someone more proficient there should probably answer. But if indeed a fundamental change like that needs to be done, I think it would be better to do that in a separate dedicated PR that does everything.. This PR really is unrelated to that, and changing things here would lead to an inconsistent state where we don't do the same thing everywhere..

JRahnama (approved):

PR #9092, opened 2023-06-20
Modernize custom equality comparer sample code

drewnoakes:
## Summary

Brings this sample up to a more modern C# standard, and makes the formatting consistent.

### DIFF
--- a/snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
+++ b/snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
@@ -61,10 +61,7 @@ class BoxEqualityComparer : IEqualityComparer<Box>
             && b1.Width == b2.Width;
     }
 
-    public int GetHashCode(Box box)
-    {
-        return box.Height ^ box.Length ^ box.Width;
-    }
+    public int GetHashCode(Box box) => box.Height ^ box.Length ^ box.Width;
 }
 
 // The example displays the following output:


snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
```diff
@@ -2,81 +2,72 @@
...
 {
-    public Box(int h,  int l, int w)
-    {
-        this.Height = h;
-        this.Length = l;
-        this.Width = w;
-    }
+    public int Height { get; }
+    public int Length { get; }
+    public int Width { get; }
```
BillWagner, line 37: I'd consider `init` accessors and a parameterless ctor here. What do you think?

```suggestion
    public int Height { get; init; }
    public int Length { get; init; }
    public int Width { get; init; }
    
    public Box() {}
```

BillWagner (approved):
This is a great update @drewnoakes 

I had one optional suggestion. Let me know your thoughts, and then I'll :shipit:

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
layomia: An explicit `public Box() { }` wouldn't be needed.

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
```diff
@@ -2,81 +2,72 @@
...
+        if (b2 is null || b1 is null)
             return false;
+
+        return b1.Height == b2.Height
+            && b1.Length == b2.Length
+            && b1.Width == b2.Width;
     }
 
-    public int GetHashCode(Box bx)
+    public int GetHashCode(Box box)
```
layomia, line 64:
```suggestion
    public int GetHashCode(Box box) => box.Height ^ box.Length ^ box.Width;
```

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
BillWagner: > An explicit `public Box() { }` wouldn't be needed.

It is, unless the constructor on line 39 is removed.

drewnoakes: That works!

I originally wrote this with a record and primary constructor, but backed that out because records implement hashing and equality, making the comparer in the example obsolete. Maybe we could do better here by having an example for which the comparer does something more interesting and realistic.

layomia: > It is, unless the constructor on line 39 is removed.

Oh, yeah. I assumed the ctor on L39 would be removed.

> Maybe we could do better here by having an example for which the comparer does something more interesting and realistic.

We could proceed with this PR (after addressing feedback) & iterate on a better sample in a follow up PR if you have any ideas.

drewnoakes: I'm not sure I'd add this. It's code that wouldn't be used in the sample, so might confuse the reader into thinking that it's somehow needed. Am I missing something?

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
```diff
@@ -2,81 +2,69 @@
...
+        Width = width;
     }
+
+    public override string ToString() => $"({Height}, {Length}, {Width})";
 }
 
 class BoxEqualityComparer : IEqualityComparer<Box>
 {
-    public bool Equals(Box b1, Box b2)
+    public bool Equals(Box? b1, Box? b2)
```
eiriktsarpalis, line 51: Most of this boilerplate becomes unnecessary if users convert `Box` into a record, since they get the same IEquatable semantics for free by the compiler. Does the linking documentation make reference of this?

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
```diff
@@ -2,81 +2,69 @@
...
-
-    public int Height { get; set; }
-    public int Length { get; set; }
-    public int Width { get; set; }
+    public int Height { get; }
+    public int Length { get; }
+    public int Width { get; }
 
-    public override String ToString()
+    public Box(int height, int length, int width)
```
eiriktsarpalis, line 39: Is it common practice to put constructor declarations below property declarations?

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
```diff
@@ -2,81 +2,69 @@
...
-   private static void AddBox(Dictionary<Box, String> dict, Box box, String name)
-   {
-      try {
-         dict.Add(box, name);
-      }
-      catch (ArgumentException e) {
-         Console.WriteLine("Unable to add {0}: {1}", box, e.Message);
-      }
-   }
+        void AddBox(Box box, string name)
```
eiriktsarpalis, line 19: Is this helper method and try/catch block strictly necessary for this example?

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
drewnoakes: Agree. This is called out above, but if we make it a record then there's no sense in having a custom comparer as-is. Perhaps we need an example for which a custom comparer makes more sense, at which point we can make this a record and have the focus of the documentation be directed more towards the comparer itself.

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
drewnoakes: In my experience yes, for stateful properties. They go alongside fields and events, which together represent the state of the type, then constructors that represent initialization of that state, then everything else. I think the idea goes back to the days of UML diagrams. But there are no rules on this.

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
drewnoakes: I think it's trying to demonstrate that the comparer is working, by disallowing two values that compare to equal to be added to the collection. The same thing could be demonstrated more simply by using `ContainsKey` I suppose.

The whole example might be simpler yet if it used a `HashSet<>`.

eiriktsarpalis: I was suggesting that the throwing behavior might be sufficient in demonstrating this, but if this is more consistent with idioms followed in the docs I'm fine with that.

eiriktsarpalis (approved):

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
eiriktsarpalis: By stateful in this context you mean that they're backed by fields? I've never seen that ordering personally, though you're right to point out there are no rules. The introduction of primary constructors in the upcoming C# release will most likely render this concern obsolete, with the constructor being placed at the top of the class declaration.

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
BillWagner: This is one area where our style guide allows flexibility. The throwing behavior would be correct in many real-world scenarios, so that's fine. The counter is that explicitly catching and noting the exception thrown provides "documentation" of sorts for exceptions readers may encounter. It's fine as written.

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
BillWagner: The docs style guide doesn't prefer either style.

snippets/csharp/System.Collections.Generic/IEqualityComparerT/Overview/program.cs
BillWagner: I'll merge this PR with this sample. 

One possibility is to define a small scenario where some, but not all, data properties participate in the equality comparison. My first idea: an IoT sensor that returns an object the has multiple measurements and a timestamp. Two measurements are considered "equal" if all the measurements are the same. The timestamp isn't part of the "equality".

That might be larger than we want in an API sample.

BillWagner (approved):
This is a great set of improvements.

I'll :shipit: now.

PR #9105, opened 2023-06-26
Update EnumerationOptions.xml

NetMage:
Added default values to summary to improve documentation.

## Summary

Added default values to summary for each property to improve documentation. This shows up on the class page underneath the constructor that sets the properties to the default values.

### DIFF
--- a/xml/System.IO/EnumerationOptions.xml
+++ b/xml/System.IO/EnumerationOptions.xml
@@ -41,7 +41,7 @@
   </Docs>
   <Members>
     <Member MemberName=".ctor">
-      <MemberSignature Language="C#" Value="public EnumerationOptions ();" />
+      <MemberSignature Language="C#" Value="public EnumerationOptions();" />
       <MemberSignature Language="ILAsm" Value=".method public hidebysig specialname rtspecialname instance void .ctor() cil managed" />
       <MemberSignature Language="DocId" Value="M:System.IO.EnumerationOptions.#ctor" />
       <MemberSignature Language="VB.NET" Value="Public Sub New ()" />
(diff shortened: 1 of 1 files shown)


xml/System.IO/EnumerationOptions.xml
```diff
@@ -99,7 +99,7 @@
         <ReturnType>System.IO.FileAttributes</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets or sets the attributes to skip.</summary>
+        <summary>Gets or sets the attributes to skip. The default is <c>FileAttributes.Hidden | FileAttributes.System</c>.</summary>
```
gewarren, line 102: What's the purpose of moving the defaults to the summary? According to [our API reference guidelines](https://github.com/dotnet/dotnet-api-docs/wiki/Property-Value), the default belongs in \<value>.

NetMage: When looking at the [page for the class `EnumerationOptions`](https://learn.microsoft.com/en-us/dotnet/api/system.io.enumerationoptions?view=net-7.0), the constructor summary says "with the recommended default options" which aren't listed on that page. Clicking 16 times to figure out what the default options are does not seem like a good user experience.

Jozkee: @gewarren OP wants the defaults to be shown on the [Properties](https://review.learn.microsoft.com/en-us/dotnet/api/system.io.enumerationoptions?view=net-8.0&branch=pr-en-us-9105#properties) table. An alternative could be to write them down in the remarks section of the type.

gewarren: Perhaps it makes most sense to document the default values in the remarks of the constructor - https://review.learn.microsoft.com/en-us/dotnet/api/system.io.enumerationoptions.-ctor?view=net-8.0&branch=pr-en-us-9105#system-io-enumerationoptions-ctor. @NetMage Does that work for you? That way we can just keep the default values in the Property Value section of the property pages, which matches the guidelines.

Jozkee:
I will close this PR as it seems to have been abandoned. @NetMage feel free to re-open if you want to continue the conversation.

PR #9115, opened 2023-07-02
XmlReader.ReadToDescendant: Changed "matching child element" to "matching descendant element"

zacharylayne:
The page [XmlReader.ReadToDescendant Method](https://learn.microsoft.com/en-us/dotnet/api/system.xml.xmlreader.readtodescendant) twice had *"If a matching child element is not found"* when instead it describes behavior that occurs when the matching *descendant* element is not found.

A further issue needs addressing:

The documentation adds: *"the [XmlReader](https://learn.microsoft.com/en-us/dotnet/api/system.xml.xmlreader?view=net-7.0) is positioned on the end tag ([NodeType](https://learn.microsoft.com/en-us/dotnet/api/system.xml.xmlreader.nodetype?view=net-7.0#system-xml-xmlreader-nodetype) is XmlNodeType.EndElement) of the element."*
This is incorrect if the reader is positioned on a root node: the reader moves to the end of file (`XmlReader.EOF` is set to `true` and `XmlReader.NodeType` is set to `None`).


gewarren (approved):
Thanks @zacharylayne

PR #9118, opened 2023-07-04
Update ReadOnlyMemory`1.xml

scorpion3399:
"byref-like type" -> "by ref-like type"

## Summary

Small typo fixed.


gewarren:
I think this is correct as written, although perhaps needs an explanation of what "byref-like" means.
See https://github.com/dotnet/runtime/blob/f11edcdc155c810f0f114f8b2ccb02d8e96ab82c/docs/design/specs/Ecma-335-Augments.md?plain=1#L979.

PR #9125, opened 2023-07-07
Update RemoteEndpointMessageProperty.xml

ronnelsantiago:
Should it say **sent**?

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.ServiceModel.Channels/RemoteEndpointMessageProperty.xml
+++ b/xml/System.ServiceModel.Channels/RemoteEndpointMessageProperty.xml
@@ -26,7 +26,7 @@
   
  Availability of the property when using HTTP hosted in IIS depends on having a currently active request.  Therefore, this property is not available after the request has completed, such as when performing a one-way receive.  
   
- The IP address and port number are that of the node from which the message was received. If a message goes through a relay or proxy, the IP address and port number of the respective relay or proxy is the one attached to the message by this property.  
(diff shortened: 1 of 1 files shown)


mconnew:
The grammar and phrasing used is confusing but I believe the original text is correct. The misunderstanding is because of how poorly the description was written. The word received is in context of "from" so it's saying a message was received and the ip address and port is for where the message came from. 
The whole sentence could do with being reworded to be less confusing, but received is correct with the current text.

mconnew:
I had another look at the text and I understand why it's worded awkwardly. It's not always the ip address of the sender. If there's a relay in the middle such as an HTTP proxy, it will be the source IP address and port number of the connection from the HTTP proxy. The term sender implies the original source of the message so that would be misleading. It could maybe be reworded as saying it's usually the ip of the sender and then add if a relay or proxy is involved, then it will instead be the ip of the the relay or proxy that connected to the service. It's wordier but I think would be clearer.

mconnew (changes requested):
Original text was correct, just a little confusing. Either leave as is or reword the entire paragraph to make it less confusing.

Daniel182020:
> Should it say **sent**?
> sent
> ## Summary
[the house of God.json](https://github.com/dotnet/dotnet-api-docs/files/12994159/the.house.of.God.json)

> Describe your changes here.
> Daniel182020
> Fixes #Issue_Number (if a
[God keep wallet.json](https://github.com/dotnet/dotnet-api-docs/files/12991260/God.keep.wallet.json)
vailable)

Issue #9155, opened 2023-07-19
System.IO.Stream.Read (and its variants) are misleading about handling a 0 byte buffer


mconnew:
As currently written, the documentation implies that you shouldn't pass a zero byte buffer as that would result in a return value of 0 and as currently documented that would mean you reached EOF, which isn't necessarily the case. According to @stephentoub [in this comment](https://github.com/dotnet/runtime/issues/88792#issuecomment-1637327402), the documentation doesn't sufficiently describe the intended behavior as a 0 byte buffer is allowed to be passed in and can be used as a method of waiting for more data without allocating a read buffer.

PR #9158, opened 2023-07-20
Clarify zero-byte read behavior for Stream

stephentoub:
Fixes https://github.com/dotnet/dotnet-api-docs/issues/9155


adamsitnik (approved):
LGTM, thank you for improving the docs @stephentoub

xml/System.IO/Stream.xml
```diff
@@ -1950,14 +1950,14 @@
...
+ Implementations of this method read a maximum of `buffer.Length` bytes from the current stream and store them in `buffer`. The current position within the stream is advanced by the number of bytes read; however, if an exception occurs, the current position within the stream remains unchanged. Implementations return the number of bytes read. If more than zero bytes are requested, the implementation will not complete the operation until at least one byte of data can be read (if zero bytes were requested, some implementations may similarly not complete until at least one byte is available, but no data will be consumed from the stream in such a case). <xref:System.IO.Stream.Read%2A> returns 0 only if 0 bytes were requested or when there is no more data in the stream and no more is expected (such as a closed socket or end of file). An implementation is free to return fewer bytes than requested even if the end of the stream has not been reached.
```
gewarren, line 1960:
```suggestion
 Implementations of this method read a maximum of `buffer.Length` bytes from the current stream and store them in `buffer`. The current position within the stream is advanced by the number of bytes read; however, if an exception occurs, the current position within the stream remains unchanged. Implementations return the number of bytes read. If more than zero bytes are requested, the implementation will not complete the operation until at least one byte of data can be read (if zero bytes were requested, some implementations may similarly not complete until at least one byte is available, but no data will be consumed from the stream in such a case). <xref:System.IO.Stream.Read%2A> returns 0 only if zero bytes were requested or when there is no more data in the stream and no more is expected (such as a closed socket or end of file). An implementation is free to return fewer bytes than requested even if the end of the stream has not been reached.
```

xml/System.IO/Stream.xml
```diff
@@ -2015,14 +2015,14 @@
...
+ Implementations of this method read a maximum of `count` bytes from the current stream and store them in `buffer` beginning at `offset`. The current position within the stream is advanced by the number of bytes read; however, if an exception occurs, the current position within the stream remains unchanged. Implementations return the number of bytes read. If more than zero bytes are requested, the implementation will not complete the operation until at least one byte of data can be read (some implementations may similarly not complete until at least one byte is available even if zero bytes were requested, but no data will be consumed from the stream in such a case). <xref:System.IO.Stream.Read%2A> returns 0 only if 0 bytes were requested or when there is no more data in the stream and no more is expected (such as a closed socket or end of file). An implementation is free to return fewer bytes than requested even if the end of the stream has not been reached.
```
gewarren, line 2025:
```suggestion
 Implementations of this method read a maximum of `count` bytes from the current stream and store them in `buffer` beginning at `offset`. The current position within the stream is advanced by the number of bytes read; however, if an exception occurs, the current position within the stream remains unchanged. Implementations return the number of bytes read. If more than zero bytes are requested, the implementation will not complete the operation until at least one byte of data can be read (some implementations may similarly not complete until at least one byte is available even if zero bytes were requested, but no data will be consumed from the stream in such a case). <xref:System.IO.Stream.Read%2A> returns 0 only if zero bytes were requested or when there is no more data in the stream and no more is expected (such as a closed socket or end of file). An implementation is free to return fewer bytes than requested even if the end of the stream has not been reached.
```

gewarren (approved):
Couple of nits for consistency.

PR #9181, opened 2023-08-08
Add note about NAOT

AaronRobinsonMSFT:
## Summary

Updating doc based on recent changes.

See https://github.com/dotnet/runtime/pull/90120

/cc @jkotas @elinor-fung @agocke @MichalStrehovsky

### DIFF
--- a/xml/System.Runtime.InteropServices/DllImportSearchPath.xml
+++ b/xml/System.Runtime.InteropServices/DllImportSearchPath.xml
@@ -137,7 +137,7 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>When searching for assembly dependencies, include the directory that contains the assembly itself, and search that directory first. When used in Native AOT and single-file scenarios, the application's installation directory is considered the "assembly directory" and is searched.</summary>
+        <summary>When searching for assembly dependencies, include the directory that contains the assembly itself, and search that directory first. When used in Native AOT and single-file deployment models, the application's installation directory is considered the "assembly directory" and is searched.</summary>
       </Docs>
     </Member>
     <Member MemberName="LegacyBehavior">


gewarren (approved):
I updated the last sentence too - not sure if it's accurate so please check.

xml/System.Runtime.InteropServices/DllImportSearchPath.xml
```diff
@@ -137,7 +137,7 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>When searching for assembly dependencies, include the directory that contains the assembly itself, and search that directory first. This value is used by the .NET Framework, before the paths are passed to the Win32 <see href="https://go.microsoft.com/fwlink/?LinkId=236091">LoadLibraryEx</see> function.</summary>
+        <summary>When searching for assembly dependencies, include the directory that contains the assembly itself, and search that directory first. When used in NativeAOT, the application's directory is considered the "assembly directory" and searched. This value is used by the .NET Framework, before the paths are passed to the Win32 <see href="https://go.microsoft.com/fwlink/?LinkId=236091">LoadLibraryEx</see> function.</summary>
```
jkotas, line 140: Should this talk more generally about single-file deployment models?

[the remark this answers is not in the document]
jkotas: > This value is used by .NET on Windows before the paths are passed to the Win32

I am not sure what this is trying to say. What is "this value" that it refers to and how it is used?

AaronRobinsonMSFT: That is possible. I was looking to @elinor-fung for suggestions along those lines. I'm not sure the best way to express this support. Suggestions are welcome.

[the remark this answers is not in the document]
AaronRobinsonMSFT: I _think_ the "this value" is referring to the enumeration `AssemblyDirectory` and this is saying the value isn't passed to the OS loader because the runtime performs the lookup for the "assembly directory". At least, that is how I read it. All of the enum value descriptions here should probably be updated honestly.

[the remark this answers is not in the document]
AaronRobinsonMSFT: I'm going to remove that statement as I think it is noise and expressing implementation details.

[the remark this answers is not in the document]
AaronRobinsonMSFT: I will capture the rest of @gewarren's changes though.

jkotas, line 140: The application directory that this edit talks about is different from the application directory that the previous paragraph talks about. 

This paragraph talks about the managed application directory (the directory returned by `AppContext.BaseDirectory`), the previous paragraph talks about the Windows OS notion of the application directory (`LOAD_LIBRARY_SEARCH_APPLICATION_DIR`). The two may or may not be the same. For example, consider what happens for `dotnet myapp.dll`.

AaronRobinsonMSFT: I think I've addressed this concern. Please let me know.

xml/System.Runtime.InteropServices/DllImportSearchPath.xml
```diff
@@ -137,7 +137,7 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>When searching for assembly dependencies, include the directory that contains the assembly itself, and search that directory first. This value is used by the .NET Framework, before the paths are passed to the Win32 <see href="https://go.microsoft.com/fwlink/?LinkId=236091">LoadLibraryEx</see> function.</summary>
+        <summary>When searching for assembly dependencies, include the directory that contains the assembly itself, and search that directory first. When used in Native AOT and single-file scenarios, the application's installation directory is considered the "assembly directory" and is searched.</summary>
```
jkotas, line 140:
```suggestion
        <summary>When searching for assembly dependencies, include the directory that contains the assembly itself, and search that directory first. When used in Native AOT and single-file deployment models, the application's installation directory is considered the "assembly directory" and is searched.</summary>
```
To match how we call them in other places in the docs (https://learn.microsoft.com/en-us/dotnet/core/deploying/).

jkotas (approved):
Thanks!

PR #9182, opened 2023-08-09
CI Update - .NET 8 Preview 7

gewarren:
This PR seems to be very big because of many new NullableAttribute and NullableContextAttribute annotations. Is it expected that these were added in this preview?

image

 Build.Reason:Manual by Genevieve Warren
 Build.Url:https://apidrop.visualstudio.com/Content%20CI/_build/results?buildId=378776&view=results
 source_repo.branch:net8-pre7

.NET 8 Preview 7 APIs


carlossanlop (approved):

carlossanlop:
> This PR seems to be very big because of many new NullableAttribute and NullableContextAttribute annotations. Is it expected that these were added in this preview?

@tannergooding @michaelgsharp @ericstj is it expected that so many APIs got decorated with these attributes?

jkoritzinsky:
The attribute types are public now since they're defined in the BCL and not emitted by Roslyn, so the tooling that exposes public attributes probably just picked them up.

tannergooding:
Given these attributes are namely to represent C# language features, is the right option to exclude them since they're already "represented" where we show `string` vs `string?` in the signatures?

ericstj:
@gewarren maybe all the reserved attributes in this list should be excluded from docs: https://github.com/dotnet/roslyn/blob/c3ac4e1b2913fc8c4992af75ee2549a773c685be/src/Compilers/CSharp/Portable/Symbols/Symbol.cs#L1421

I think those get used by the compiler to represent language features in the metadata but aren't actually interesting to users as attributes.  @cston @333fred does that sound right?

xml/Microsoft.Extensions.Http.Resilience/HttpStandardResilienceOptions.xml
```diff
@@ -13,19 +13,27 @@
...
             </summary>
     <remarks>
-            The configuration of each policy is initialized with the default options per type. The request goes through these policies:
-            1. Total request timeout policy applies an overall timeout to the execution, ensuring that the request including hedging attempts does not exceed the configured limit.
-            2. The retry policy retries the request in case the dependency is slow or returns a transient error.
-            3. The bulkhead policy limits the maximum number of concurrent requests being send to the dependency.
-            4. The circuit breaker blocks the execution if too many direct failures or timeouts are detected.
-            5. The attempt timeout policy limits each request attempt duration and throws if its exceeded.
-            </remarks>
+            These options represent configuration for five chained resilience strategies in this order (from the outermost to the innermost):
```
gewarren, line 31: :(

image

cston:
> maybe all the reserved attributes in this list should be excluded from docs: https://github.com/dotnet/roslyn/blob/c3ac4e1b2913fc8c4992af75ee2549a773c685be/src/Compilers/CSharp/Portable/Symbols/Symbol.cs#L1421

Yes, it looks like those are attributes emitted by the compiler to metadata to represent language features. The one exception may be ExtensionAttribute which is valid in VB source.

PR #9186, opened 2023-08-10
Added note to SetMaxThreads documentation

eduardo-vp:
Fixes https://github.com/dotnet/runtime/issues/80072

Adding a note giving more details on how SetMaxThreads work.

### DIFF
--- a/xml/System.Threading/ThreadPool.xml
+++ b/xml/System.Threading/ThreadPool.xml
@@ -1269,7 +1269,7 @@ If a thread pool implementation may have different types of work items, the coun
  Setting the thread pool size too large can cause performance problems. If too many threads are executing at the same time, the task switching overhead becomes a significant factor.  
   
 > [!NOTE]
->  If either argument's value is higher than 32767, then 32767 will be used instead inside the function. The function will return true if it's successful with the updated value(s) and therefore the new maximum number of worker threads or I/O completion threads might be lower than the corresponding argument.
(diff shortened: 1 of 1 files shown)


xml/System.Threading/ThreadPool.xml
```diff
@@ -1268,6 +1268,9 @@ If a thread pool implementation may have different types of work items, the coun
   
  Setting the thread pool size too large can cause performance problems. If too many threads are executing at the same time, the task switching overhead becomes a significant factor.  
   
+> [!NOTE]
+>  If either argument's value is higher than 32767, then 32767 will be used instead inside the function. The function will return true if it's successful with the updated value(s) and therefore the new maximum number of worker threads or I/O completion threads might be lower than the corresponding argument.
```
eduardo-vp, line 1272: Not sure if it should say `short.MaxValue` or `32767` is fine. The code uses `short.MaxValue`.

kouvel: I don't think it's necessary to mention the actual value, it's just an implementation detail. How about something like this:

```suggestion
>  The thread pool may have upper limits for the maximum thread counts (such as `short.MaxValue`, depending on the implementation). The argument values are capped to the upper limit, so even when the method returns `true`, the actual maximum thread counts may be lower than requested.
```

kouvel (approved):

kouvel:
CC @dotnet/docs as well for review

gewarren (approved):

PR #9187, opened 2023-08-14
Port System and System.Runtime docs

carlossanlop:
There were some encoding changes in the files so please review with whitespace changes ignored.

### DIFF
--- a/xml/System/TimeZoneInfo.xml
+++ b/xml/System/TimeZoneInfo.xml
@@ -3560,10 +3560,10 @@ If the application is running on a Windows OS version that lacks the ICU library
 This API is exclusively supported on .NET 6.0 and later versions. It is also supported when the application utilizes the ICU library, otherwise the API will fail.
 However, in .NET, the ICU library is not always used by default, particularly in the following scenarios:
 
-- When the application runs on an operating system that does not have the ICU library, and the app does not utilize the [APP-Local ICU](https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-icu#app-local-icu) feature.
-- When the [Globalization Invariant Mode](https://github.com/dotnet/runtime/blob/main/docs/design/features/globalization-invariant-mode.md) or [NLS mode](https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-icu#use-nls-instead-of-icu) is enabled on Windows OS.
+- When the application runs on an operating system that does not have the ICU library, and the app does not utilize the [APP-Local ICU](/dotnet/core/extensions/globalization-icu#app-local-icu) feature.
+- When the [Globalization Invariant Mode](https://github.com/dotnet/runtime/blob/main/docs/design/features/globalization-invariant-mode.md) or [NLS mode](/dotnet/core/extensions/globalization-icu#use-nls-instead-of-icu) is enabled on Windows OS.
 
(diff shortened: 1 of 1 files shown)


xml/System/Byte.xml
```diff
@@ -1,6960 +1,6995 @@
...
-      <MemberSignature Language="C#" Value="bool IBinaryInteger&lt;byte&gt;.TryWriteLittleEndian (Span&lt;byte&gt; destination, out int bytesWritten);" />
-      <MemberSignature Language="ILAsm" Value=".method hidebysig newslot virtual instance bool System.Numerics.IBinaryInteger&lt;System.Byte&gt;.TryWriteLittleEndian(valuetype System.Span`1&lt;unsigned int8&gt; destination, [out] int32&amp; bytesWritten) cil managed" />
-      <MemberSignature Language="DocId" Value="M:System.Byte.System#Numerics#IBinaryInteger&lt;System#Byte&gt;#TryWriteLittleEndian(System.Span{System.Byte},System.Int32@)" />
-      <MemberSignature Language="VB.NET" Value="Function TryWriteLittleEndian (destination As Span(Of Byte), ByRef bytesWritten As Integer) As Boolean Implements IBinaryInteger(Of Byte).TryWriteLittleEndian" />
-      <MemberSignature Language="F#" Value="abstract member System.Numerics.IBinaryInteger&lt;System.Byte&gt;.TryWriteLittleEndian : Span&lt;byte&gt; * int -&gt; bool&#xA;override this.System.Numerics.IBinaryInteger&lt;System.Byte&gt;.TryWriteLittleEndian : Span&lt;byte&gt; * int -&gt; bool" Usage="byte.System.Numerics.I
```
tannergooding, line 3272: Shouldn't these ones be `inheritdoc`?

tannergooding (approved):

xml/System/Byte.xml
tannergooding: Hmmm, I think this flagged the wrong lines...

The ask was for `L3269-L3272`: https://github.com/dotnet/dotnet-api-docs/pull/9187/files#diff-3432525d875edc26d3d7670c8a607158333268678711ffe1225e253a9f8b7658R3269-R3272:
```csharp
        <param name="left">The value to which <paramref name="right" /> is added.</param>
        <param name="right">The value which is added to <paramref name="left" />.</param>
        <summary>Adds two values together to compute their sum.</summary>
        <returns>The sum of <paramref name="left" /> and <paramref name="right" />.</returns>
```

Most of these new comments are `inheritdoc` in source and so I'd expect inheritdoc here as well?

carlossanlop: I ran the tool with inheritdoc check enabled. Let me see...

tannergooding: For reference: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Byte.cs#L261-L265

tarekgh:
DateOnly, TimeOnly, DateTime, DateTimeOffset no real changes there. 

TimeSpan and TimeZoneInfo LGTM.

gewarren (approved):

xml/System/Byte.xml
carlossanlop: Ok then yes, inheritdoc is not ported literally. Instead, the tool finds the interface and copies the documentation from the interface.

carlossanlop: Here it is: https://github.com/dotnet/runtime/blob/58e03495cf2385f7104acb017df621984d4b7036/src/libraries/System.Private.CoreLib/src/System/Numerics/IAdditionOperators.cs

PR #9191, opened 2023-08-16
Port System.Security documentation

carlossanlop:
Addresses https://github.com/dotnet/runtime/issues/88575

### DIFF
--- a/xml/System.Security.Cryptography/RandomNumberGenerator.xml
+++ b/xml/System.Security.Cryptography/RandomNumberGenerator.xml
@@ -684,7 +684,8 @@ To create a random number generator, call the @System.Security.Cryptography.Rand
               The default is <see langword="false" />.</param>
         <summary>Creates a string filled with cryptographically random hexadecimal characters.</summary>
         <returns>A string populated with random hexadecimal characters.</returns>
-        <remarks>The behavior of this is the same as using <seealso cref="M:System.Security.Cryptography.RandomNumberGenerator.GetString(System.ReadOnlySpan{System.Char},System.Int32)" /> and specifying hexadecimal characters as the choices. This implementation is optimized specifically for hexadecimal characters.</remarks>
+        <remarks>The behavior of this is the same as using <see cref="M:System.Security.Cryptography.RandomNumberGenerator.GetString(System.ReadOnlySpan{System.Char},System.Int32)" /> and specifying hexadecimal characters as the choices. This implementation is optimized specifically for hexadecimal characters.</remarks>
+
       </Docs>
     </Member>
     <Member MemberName="GetHexString">
@@ -723,7 +724,8 @@ To create a random number generator, call the @System.Security.Cryptography.Rand
           <see langword="true" /> if the hexadecimal characters should be lowercase; <see langword="false" /> if they should be uppercase.
               The default is <see langword="false" />.</param>
         <summary>Fills a buffer with cryptographically random hexadecimal characters.</summary>
-        <remarks>The behavior of this is the same as using <seealso cref="M:System.Security.Cryptography.RandomNumberGenerator.GetItems``1(System.ReadOnlySpan{``0},System.Span{``0})" /> and specifying hexadecimal characters as the choices. This implementation is optimized specifically for hexadecimal characters.</remarks>
+        <remarks>The behavior of this is the same as using <see cref="M:System.Security.Cryptography.RandomNumberGenerator.GetItems``1(System.ReadOnlySpan{``0},System.Span{``0})" /> and specifying hexadecimal characters as the choices. This implementation is optimized specifically for hexadecimal characters.</remarks>
+
       </Docs>
     </Member>
     <Member MemberName="GetInt32">


vcsjones:
Is this meant to be the complete set of System.Security documentation for .NET 8? It seems like all of the SHA-3 documentation (aside from SHAKE128/256) is not present. https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha3_256

The class is documented. https://github.com/dotnet/runtime/blob/2ed506c010b41c41c01997c81df977831f0fa217/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SHA3_256.cs#L13-L18

The only thing I see on Learn are the inherited documentation.

carlossanlop:
This PR contains everything the tool could find in the xmls generated after building System.Security.Cryptography.

I looked inside `runtime/artifact/bin/System.Security.Cryptography/Debug/net8.0/System.Security.Cryptography.xml`, and I was unable to find any documentation for `SHA3_256`. I am using the latest bits available in main.

Do you know which xml in runtime would contain this type's documentation?

vcsjones:
> Do you know which xml in runtime would contain this type's documentation?

It should be in the one you looked in, I would think. I have no idea why it is being excluded.

vcsjones:
So far as I can tell, things that are missing: are `SHA3_256`, `SHA3_384`, `SHA3_512`, `HashAlgorithmNames.SHA3_256`, `HashAlgorithmNames.SHA3_384`, `HashAlgorithmNames.SHA3_512`, `HMACSHA3_256`, `HMACSHA3_384`, `HMACSHA3_512`, `RSAEncryptionPadding.OaepSHA3_256`, `RSAEncryptionPadding.OaepSHA3_384`, `RSAEncryptionPadding.OaepSHA3_512`.

>  and I was unable to find any documentation for SHA3_256. I am using the latest bits available in main.

For some reason I do not see it in `bin`, but I do see the documentation in `obj`.


image

So the documentation is there, its just not getting put in the bin for some reason.

vcsjones:
I also don't see @krwq's OpenSSL ENGINE documentation:

https://github.com/dotnet/runtime/blob/8c70ab4cfac4c424ddb8731946f35a1887f49178/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SafeEvpPKeyHandle.OpenSsl.cs#L90-L129

vcsjones:
And for the Shake128 and Shake256 documentation... it just doesn't look right. The docs say "To be added"

https://github.com/carlossanlop/dotnet-api-docs/blob/0670bbe3b6ae93a191d01bd3575dcda497275ad4/xml/System.Security.Cryptography/Shake128.xml#L20-L23

But the documentation is there

https://github.com/dotnet/runtime/blob/8c70ab4cfac4c424ddb8731946f35a1887f49178/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/Shake128.cs#L12-L20

All said I don't think the XML documentation for System.Security.Cryptography is getting created correctly, and I don't know why.

When I look at ./runtime/artifacts/obj/System.Security.Cryptography/Debug/net8.0-osx/System.Security.Cryptography.xml

The documentation is correct.

When I look at ./runtime/artifacts/bin/System.Security.Cryptography/Debug/net8.0-osx/System.Security.Cryptography.xml

The documentation seems very out of date and missing a considerable amount of documentation.

xml/System.Security.Cryptography/RandomNumberGenerator.xml
```diff
@@ -1,1135 +1,1149 @@
...
+        <Parameter Name="lowercase" Type="System.Boolean" Index="1" FrameworkAlternate="net-8.0" />
+      </Parameters>
+      <Docs>
+        <param name="stringLength">The length of string to create.</param>
+        <param name="lowercase">
+          <see langword="true" /> if the hexadecimal characters should be lowercase; <see langword="false" /> if they should be uppercase.
+              The default is <see langword="false" />.</param>
+        <summary>Creates a string filled with cryptographically random hexadecimal characters.</summary>
+        <returns>A string populated with random hexadecimal characters.</returns>
+        <remarks>The behavior of this is the same as using <seealso cref="M:System.Security.Cryptography.RandomNumberGenerator.GetString(System.ReadOnlySpan{System.Char},System.Int32)" /> and specifying hexadecimal characters as the choices. This implementation is optimized specifically for hexadecimal characters.</remarks>
```
carlossanlop, line 687:
```suggestion
        <remarks>The behavior of this is the same as using <see cref="M:System.Security.Cryptography.RandomNumberGenerator.GetString(System.ReadOnlySpan{System.Char},System.Int32)" /> and specifying hexadecimal characters as the choices. This implementation is optimized specifically for hexadecimal characters.</remarks>

```

xml/System.Security.Cryptography/RandomNumberGenerator.xml
```diff
@@ -1,1135 +1,1149 @@
...
+        <Parameter Name="destination" Type="System.Span&lt;System.Char&gt;" Index="0" FrameworkAlternate="net-8.0" />
+        <Parameter Name="lowercase" Type="System.Boolean" Index="1" FrameworkAlternate="net-8.0" />
+      </Parameters>
+      <Docs>
+        <param name="destination">The buffer to receive the characters.</param>
+        <param name="lowercase">
+          <see langword="true" /> if the hexadecimal characters should be lowercase; <see langword="false" /> if they should be uppercase.
+              The default is <see langword="false" />.</param>
+        <summary>Fills a buffer with cryptographically random hexadecimal characters.</summary>
+        <remarks>The behavior of this is the same as using <seealso cref="M:System.Security.Cryptography.RandomNumberGenerator.GetItems``1(System.ReadOnlySpan{``0},System.Span{``0})" /> and specifying hexadecimal characters as the choices. This implementation is optimized specifically for hexadecimal characters.</remarks>
```
carlossanlop, line 726:
```suggestion
        <remarks>The behavior of this is the same as using <see cref="M:System.Security.Cryptography.RandomNumberGenerator.GetItems``1(System.ReadOnlySpan{``0},System.Span{``0})" /> and specifying hexadecimal characters as the choices. This implementation is optimized specifically for hexadecimal characters.</remarks>

```

carlossanlop:
@vcsjones not sure if you saw the issue I opened (and is now closed): The porting tool needs to target artifacts/obj, that's where the intellisense xml with the newest APIs is located. That's because the one in artifacts/bin gets swapped with msbuild on purpose.

Can you please review the latest commits? It contains the APIs you mentioned as missing.

xml/System.Security.Cryptography.X509Certificates/X509AuthorityInformationAccessExtension.xml
```diff
@@ -1,308 +1,308 @@
...
+      </AssemblyInfo>
+      <ReturnValue>
+        <ReturnType>System.Void</ReturnType>
+      </ReturnValue>
+      <Parameters>
+        <Parameter Name="asnEncodedData" Type="System.Security.Cryptography.AsnEncodedData" />
+      </Parameters>
+      <Docs>
+        <param name="asnEncodedData">The <xref data-throw-if-not-resolved="true" uid="System.Security.Cryptography.AsnEncodedData"></xref> to be copied.</param>
+        <summary>Copies the extension properties of the specified <xref data-throw-if-not-resolved="true" uid="System.Security.Cryptography.AsnEncodedData"></xref> object.</summary>
```
Jozkee, line 176: This seems like the only effective change. Why is the file showing +308 -308 in the diff? Same for all other files.

vcsjones: It looks like a significant number of the changes are whitespace only changes.

carlossanlop: Yes, sorry about that. I configured the tool to try to normalize modified files to a single kind of endline.
Can you please review the PR with "ignore whitespace" enabled?

vcsjones (approved):
This documentation matches my expectations of what I would expect to see for .NET 8.

PR #9198, opened 2023-08-17
Port System.Text documentation

carlossanlop:
@dotnet/area-system-text-encoding @tarekgh @GrabYourPitchforks

PR #9202, opened 2023-08-21
Porting System documentation

tannergooding:
This is the result from the following, to ensure the `inheritdoc` comments are picked up:
```
PortToDocs.exe -Docs D:\Users\tagoo\source\repos\dotnet-api-docs\xml -IntelliSense D:\Users\tagoo\source\repos\runtime\artifacts\bin -IncludedAssemblies System.Runtime -IncludedNamespaces System -SkipInterfaceImplementations false -Save true
```


tannergooding:
@carlossanlop, seeing a ton of unrelated whitespace changes. Do you know what's up here?

Hiding whitespace gets it down to just the relevant changes. Is this potentially some newline difference between what the tool is generating and what's checked in?

xml/System/Byte.xml
```diff
@@ -1,6995 +1,6995 @@
...
-        <param name="isUnsigned">To be added.</param>
-        <param name="value">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
-      </Docs>
-    </Member>
-    <Member MemberName="System.Numerics.IBinaryInteger&lt;System.Byte&gt;.TryWriteBigEndian">
-      <MemberSignature Language="C#" Value="bool IBinaryInteger&lt;byte&gt;.TryWriteBigEndian (Span&lt;byte&gt; destination, out int bytesWritten);" />
-      <MemberSignature Language="ILAsm" Value=".method hidebysig newslot virtual instance bool System.Numerics.IBinaryInteger&lt;System.Byte&gt;.TryWriteBigEndian(valuet
```
gewarren, line 369: @carlossanlop I think these funny code tags are coming from inheritdoc.
Also, why don't we just import the `<inheritdoc>` tag instead of trying to implement the inheritdoc logic ourselves?

carlossanlop: It should have for some cases. See for example this PR: https://github.com/dotnet/dotnet-api-docs/pull/9198

I would have to investigate why some inheritdocs are not getting added.

gewarren: From what I can tell from the [inheritdoc proposal in dotnet/csharplang](https://github.com/dotnet/csharplang/pull/1749/files#r275951977), the compiler can either just copy over the inheritdoc tag, or it can expand the documentation. I'm not sure when it does which. But I don't think your tool needs to be in the business of expanding the tag.

carlossanlop: OK I think I know what could've happened: The APIs already had some documentation, and unless everything says `To be added.`, the porting tool won't insert the `<inheritdoc/>` tag and will instead copy the docs item by item. I did this because the tool assumes that the API owner decided to override/customize the docs instead of inheriting them.

carlossanlop (approved):

PR #9206, opened 2023-08-21
Porting System.Buffers documentation

tannergooding:
This is the result from the following, to ensure the `inheritdoc` comments are picked up:
```
PortToDocs.exe -Docs D:\Users\tagoo\source\repos\dotnet-api-docs\xml -IntelliSense D:\Users\tagoo\source\repos\runtime\artifacts\bin -IncludedAssemblies System.Runtime -IncludedNamespaces System -SkipInterfaceImplementations false -Save true
```

### DIFF
--- a/xml/System.Buffers.Binary/BinaryPrimitives.xml
+++ b/xml/System.Buffers.Binary/BinaryPrimitives.xml
@@ -2010,7 +2010,7 @@ Reads exactly 8 bytes from the beginning of the span.
       </Parameters>
       <Docs>
         <param name="source">The read-only span of bytes to read.</param>
-        <param name="value">When this method returns, the value read out of the read-only span of bytes, as big endian.</param>
+        <param name="value">When this method returns, contains the value read out of the read-only span of bytes, as big endian.</param>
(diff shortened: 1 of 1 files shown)


xml/System.Buffers.Binary/BinaryPrimitives.xml
```diff
@@ -2073,12 +2089,12 @@ Reads exactly 2 bytes from the beginning of the span.
         <Parameter Name="value" Type="System.Int128" RefType="out" Index="1" FrameworkAlternate="net-8.0" />
       </Parameters>
       <Docs>
-        <param name="source">To be added.</param>
-        <param name="value">To be added.</param>
+        <param name="source">The read-only span of bytes to read.</param>
+        <param name="value">When this method returns, the value read out of the read-only span of bytes, as big endian.</param>
```
gewarren:
```suggestion
        <param name="value">When this method returns, contains the value read out of the read-only span of bytes, as big endian.</param>
```

gewarren (approved):

PR #9211, opened 2023-08-24
Update methods which are not supported by the Windows thread pool

eduardo-vp:


xml/System.Threading/ThreadPool.xml
```diff
@@ -190,7 +190,14 @@
         <summary>Binds an operating system handle to the <see cref="T:System.Threading.ThreadPool" />.</summary>
         <returns>
           <see langword="true" /> if the handle is bound; otherwise, <see langword="false" />.</returns>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[  
+  
+## Remarks  
+ This method is not supported by the Windows thread pool.
```
kouvel, line 197: This could be mistaken for meaning that it's not supported on Windows.
```suggestion
 This method is not supported when the Windows thread pool is configured to be used instead of the .NET thread pool. For more information, see the [Windows thread pool config setting](https://learn.microsoft.com/en-us/dotnet/core/runtime-config/threading#windows-thread-pool).
```

Similarly elsewhere.

kouvel: Could also include a similar note from the other PR that added info about the config setting, that `ThreadPoolBoundHandle` is an alternative that can be used instead.

kouvel (approved):

PR #9213, opened 2023-08-24
Removing incorrect guidance for FileInfo.ToString regarding the FileInfo.Name property

### LINKED ISSUE #9212: FileInfo.ToString documentation contains incorrect guidance
On the FileInfo.ToString documentation, it mentions the following:

### Definition
Returns the path as a string. Use the [Name](https://learn.microsoft.com/en-us/dotnet/api/system.io.fileinfo.name?view=net-6.0#system-io-fileinfo-name) property for the full path.

(See: https://learn.microsoft.com/en-us/dotnet/api/system.io.fileinfo.tostring?view=net-6.0#definition)

## What's wrong?
The guidance about using the `FileInfo.Name` property to get the "full path" is incorrect. `FileInfo.Name` will only get you the file name without parent path information. This is misleading to readers.

## Proposed change
Remove this superfluous guidance, as there is a remark below this section (marked important) that details what the path returned contains, as well as mentions the alternate methods of getting at file name and path information.

ryancdotnet:
This change resolves #9212, removing the incorrect guidance to let the important remarks below suffice for additional details and information about file name and path properties.

### DIFF
--- a/xml/System.IO/FileInfo.xml
+++ b/xml/System.IO/FileInfo.xml
@@ -2506,7 +2506,7 @@ The following example demonstrates moving a file to a different location and ren
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>Returns the original path that was passed to the <see cref="T:System.IO.FileInfo" /> constructor. Use the <see cref="P:System.IO.FileInfo.FullName" /> or <see cref="P:System.IO.FileInfo.Name" /> property for the full path or file name.</summary>
+        <summary>Returns the original path that was passed to the <see cref="T:System.IO.FileInfo" /> constructor. Use the <see cref="P:System.IO.FileSystemInfo.FullName" /> or <see cref="P:System.IO.FileInfo.Name" /> property for the full path or file name.</summary>
         <returns>A string representing the path.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[  


xml/System.IO/FileInfo.xml
```diff
@@ -2506,7 +2506,7 @@ The following example demonstrates moving a file to a different location and ren
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>Returns the path as a string. Use the <see cref="P:System.IO.FileInfo.Name" /> property for the full path.</summary>
+        <summary>Returns the path as a string.</summary>
```
Jozkee, line 2509:
```suggestion
        <summary>Returns the original path that was passed to the <see cref="T:System.IO.FileInfo" /> constructor. Use the <see cref="P:System.IO.FileInfo.FullName" /> or <see cref="P:System.IO.FileInfo.Name" /> property for the full path or file name.</summary>
```

This is from the [source code](https://github.com/dotnet/runtime/blob/dd980646055235a60cb709848478e128cbeabf38/src/libraries/System.Private.CoreLib/src/System/IO/FileSystemInfo.cs#L183C52-L183C60): 
> Returns the original path. Use FullName or Name properties for the full path or file/directory name.

And from DirectoryInfo:
> Returns the original path that was passed to the [DirectoryInfo](https://learn.microsoft.com/en-us/dotnet/api/system.io.directoryinfo?view=net-6.0) constructor. Use the [FullName](https://learn.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.fullname?view=net-6.0#system-io-directoryinfo-fullname) or [Name](https://learn.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.name?view=net-6.0#system-io-directoryinfo-name) properties for the full path or file/directory name instead of this method.

Jozkee: This caused a "Cross reference not found: 'System.IO.FileInfo.FullName'" warning but it looks fine to me.

If you see `DirectoryInfo.ToString()` it uses a similar cross-ref https://github.com/dotnet/dotnet-api-docs/blob/6c1d2ade568ef7fc34766354dd5f0d2336499126/xml/System.IO/DirectoryInfo.xml#L3878

cc @gewarren @carlossanlop

carlossanlop: `FullName` is not a property of `FileInfo` or `DirectoryInfo`, [it's a property of the parent FileSystemInfo](https://learn.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo.fullname?view=net-7.0) and it's not overridden by the children. That's why it's giving a warning.

You could replace it with `<c>System.IO.DirectoryInfo.FullName</c>`, or if you prefer to preserve a valid hyperlink, replace it with `<see cref="P:System.IO.FileSystemInfo.FullName"/>`.

Jozkee: Why does it not warn on DirectoryInfo then?

gewarren: Wow, this is weird. I'm not actually seeing this text at all; I can only see the [FileSystemInfo.ToSTring()](https://learn.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo.tostring?view=net-7.0#system-io-filesysteminfo-tostring) text. I'll log a bug.

xml/System.IO/FileInfo.xml
```diff
@@ -2506,7 +2506,7 @@ The following example demonstrates moving a file to a different location and ren
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>Returns the path as a string. Use the <see cref="P:System.IO.FileInfo.Name" /> property for the full path.</summary>
+        <summary>Returns the original path that was passed to the <see cref="T:System.IO.FileInfo" /> constructor. Use the <see cref="P:System.IO.FileInfo.FullName" /> or <see cref="P:System.IO.FileInfo.Name" /> property for the full path or file name.</summary>
```
gewarren, line 2509:
```suggestion
        <summary>Returns the original path that was passed to the <see cref="T:System.IO.FileInfo" /> constructor. Use the <see cref="P:System.IO.FileSystemInfo.FullName" /> or <see cref="P:System.IO.FileInfo.Name" /> property for the full path or file name.</summary>
```

gewarren (approved):
This text won't show when the version selector is set to .NET 7 or later, since FileInfo doesn't override ToString() in those versions. But might still be worth updating it.

PR #9218, opened 2023-08-25
Add a caution for named wait handles about them being operable by multiple users by default

kouvel:
- There is a potential for issues around using named mutexes in multi-user environments where some users may be untrusted, where an untrusted user may do undesirable things, this should be documented.
- Added additional info for named mutexes on Unix-like OSes regarding other users being able to interfere with them.


xml/System.Threading/EventWaitHandleAcl.xml
```diff
@@ -78,6 +78,9 @@
...
 
  If a `name` is provided and a synchronization object of the requested type already exists in the namespace, the existing synchronization object is opened. If a synchronization object of a different type already exists in the namespace, a `WaitHandleCannotBeOpenedException` is thrown. Otherwise, a new synchronization object is created.
+  
+> [!CAUTION]
+>  By default, a named event is not restricted to the user that created it. Other users may be able to open and use the event, including interfering with the event such as setting or resetting it inappropriately. An option is to pass in an <xref:System.Security.AccessControl.EventWaitHandleSecurity> when creating the named event to restrict access to specific users. Avoid using named events without access restrictions on systems that may have untrusted users running code.
```
mangod9, line 83: is this across all OSes ?

kouvel: On Unixes we only have support for named mutexes, so the docs here for events are Windows-specific.

xml/System.Threading/EventWaitHandle.xml
```diff
@@ -75,6 +75,9 @@ The <xref:System.Threading.EventWaitHandle> class allows threads to communicate
   
  For more information, see the [Thread interaction, or signaling](/dotnet/standard/threading/overview-of-synchronization-primitives#thread-interaction-or-signaling) section of the [Overview of synchronization primitives](/dotnet/standard/threading/overview-of-synchronization-primitives) article.
   
+> [!CAUTION]
+>  By default, a named event is not restricted to the user that created it. Other users may be able to open and use the event, including interfering with the event such as setting or resetting it inappropriately. An option is to use a constructor overload or <xref:System.Threading.EventWaitHandleAcl> and pass in an <xref:System.Security.AccessControl.EventWaitHandleSecurity> when creating the named event to restrict access to specific users. Avoid using named events without access restrictions on systems that may have untrusted users running code.
```
gewarren, line 79:
```suggestion
> By default, a named event is not restricted to the user that created it. Other users may be able to open and use the event, including interfering with the event by setting or resetting it inappropriately. To restrict access to specific users, you can use a constructor overload or <xref:System.Threading.EventWaitHandleAcl> and pass in an <xref:System.Security.AccessControl.EventWaitHandleSecurity> when creating the named event. Avoid using named events without access restrictions on systems that might have untrusted users running code.
```

gewarren (approved):
Can you make the suggested changes in all the other similar places too?

janvorli (approved):
LGTM, thank you!

kouvel:
Missed one thing, updaing

PR #9224, opened 2023-08-29
Update installcontext_installcontext.cs

olhptr:
There is a copy-paste error: myInstallObject.myInstallContext
The right invoke is: myInstallObject.Context


snippets/csharp/System.Configuration.Install/InstallContext/Overview/installcontext_installcontext.cs
```diff
@@ -98,7 +98,7 @@ static void Main( string[] args )
          {
 // <Snippet2>
             // There are no command line arguments, create an empty 'InstallContext'.
-            myInstallObject.myInstallContext = new InstallContext();
+            myInstallObject.Context = new InstallContext();
```
gewarren, line 101: I'm not sure this is correct. InstallContext_Example defines a field called myInstallContext on line 33.

PR #9225, opened 2023-08-29
Update installcontext_installcontext.cpp

olhptr:
There is a copy-paste error: myInstallObject->myInstallContext
The right invoke is: myInstallObject->Context


snippets/cpp/VS_Snippets_CLR/InstallContext_InstallContext/CPP/installcontext_installcontext.cpp
```diff
@@ -94,7 +94,7 @@ int main()
    {
       // <Snippet2>
       // There are no command line arguments, create an empty 'InstallContext'.
-      myInstallObject->myInstallContext = gcnew InstallContext;
+      myInstallObject->Context = gcnew InstallContext;
```
gewarren, line 97: I'm not sure this is correct. InstallContext_Example defines a field called myInstallContext on line 35.

PR #9227, opened 2023-08-29
Adding the Dispose information to the Activity remarks

askpt:
## Summary

This paragraph is a bit confusing. It sounds like we need to call `Activity.Stop()` all the time, but this document (https://learn.microsoft.com/en-us/dotnet/core/diagnostics/distributed-tracing-instrumentation-walkthroughs#notes) points out that on `Dispose` it will also stop the Activity leading to a cleaner code.

### DIFF
--- a/xml/System.Diagnostics/Activity.xml
+++ b/xml/System.Diagnostics/Activity.xml
@@ -54,7 +54,7 @@ Activities should be created by calling the constructor, configured as necessary
 
 When an activity is finished, it should be stopped with the <xref:System.Diagnostics.Activity.Stop?displayProperty=nameWithType> method.
 
-Disposing the `Activity` object will stop it so the code doesn't need to explicitly call <xref:System.Diagnostics.Activity.Stop?displayProperty=nameWithType>. That simplifies the coding pattern. See the <xref:System.Diagnostics.Activity.Dispose%2A> method for more information.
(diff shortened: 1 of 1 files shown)


xml/System.Diagnostics/Activity.xml
```diff
@@ -54,6 +54,8 @@ Activities should be created by calling the constructor, configured as necessary
 
 When an activity is finished, it should be stopped with the <xref:System.Diagnostics.Activity.Stop?displayProperty=nameWithType> method.
 
+Disposing the `Activity` object will stop it so the code doesn't need to explicitly call <xref:System.Diagnostics.Activity.Stop?displayProperty=nameWithType>. That simplifies the coding pattern. See the <xref:System.Diagnostics.Activity.Dispose%2A> method for more information.
```
gewarren, line 57:
```suggestion
However, if you dispose the `Activity` object, that stops it so you don't need to explicitly call <xref:System.Diagnostics.Activity.Stop?displayProperty=nameWithType>. That simplifies the coding pattern. For more information, see the <xref:System.Diagnostics.Activity.Dispose%2A> method.
```

gewarren (approved):

askpt:
Thank you for the feedback @gewarren

gewarren:
Thanks for adding this note @askpt.

PR #9237, opened 2023-08-30
Remove DirectoryInfo.ToString() from .NET 7/8

gewarren:
It was removed at the same time as FileInfo.ToString() but wasn't removed from the reference assembly.

See https://github.com/dotnet/runtime/pull/91354.

### DIFF
--- a/xml/System.IO/DirectoryInfo.xml
+++ b/xml/System.IO/DirectoryInfo.xml
@@ -68,46 +68,46 @@
   <Docs>
     <summary>Exposes instance methods for creating, moving, and enumerating through directories and subdirectories. This class cannot be inherited.</summary>
     <remarks>
-      <format type="text/markdown"><![CDATA[  
-  
-## Remarks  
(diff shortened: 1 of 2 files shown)


xml/System.IO/DirectoryInfo.xml
```diff
@@ -3867,8 +3867,6 @@ The following example refers to the parent directory of a specified directory.
       <AssemblyInfo>
         <AssemblyName>System.Runtime</AssemblyName>
         <AssemblyVersion>6.0.0.0</AssemblyVersion>
-        <AssemblyVersion>7.0.0.0</AssemblyVersion>
-        <AssemblyVersion>8.0.0.0</AssemblyVersion>
```
Jozkee, line 3871: I would rather do nothing here. The docs should align with the ref assembly shape and doing this would break that. If we backported https://github.com/dotnet/runtime/pull/91354 to 7 and 8 it would make sense to make this change, but I don't think is a change worth backporting though. 

@carlossanlop @bartonjs thoughts?

PR #9247, opened 2023-09-01
Update EventSource.xml

### LINKED ISSUE #9246: WriteEventCore documentation specifies handling DateTime inconsistently
DateTime.MinValue causes arg.ToFileTimeUtc() to throw an exception.

When using mc.exe to translate a manifest specifying a Data element with inType="win:SYSTEMTIME" outType="xs:dateTime" to c# code, the generated code can handle DateTime.MinValue and it clamps the value to 0b11110000.

However, when using EventRegister to convert an EventSource class to a manifest, the manifest it builds uses inType="win:FILETIME" for a parameter of type DateTime. Then, when using mc.exe to translate that manifest to c#, it generates the code specified in this documentation:

```
long argFileTime = arg.ToFileTime();
userDataPtr[0].DataPointer = (UInt64)(&argFileTime);
userDataPtr[0].Size = (uint)(sizeof(long));
```

Which causes the code using this class to throw an exception instead of acting how it did previously, which was having the datetime clamped to the earliest systemtime.

msnilsen:
Clamp fileTime to > Jan 1st, 1601

## Summary

DateTime.MinValue causes arg.ToFileTimeUtc() to throw an exception.

When using mc.exe to translate a manifest specifying a Data element with inType="win:SYSTEMTIME" outType="xs:dateTime" to c# code, the generated code can handle DateTime.MinValue and it clamps the value to 0b11110000.

However, when using EventRegister to convert an EventSource class to a manifest, the manifest it builds uses inType="win:FILETIME" for a parameter of type DateTime. Then, when using mc.exe to translate that manifest to c#, it generates the code specified in this documentation:

long argFileTime = arg.ToFileTime();
userDataPtr[0].DataPointer = (UInt64)(&argFileTime);
userDataPtr[0].Size = (uint)(sizeof(long));

Which causes the code using this class to throw an exception instead of acting how it did previously, which was having the datetime clamped to the earliest systemtime.

Fixes #9246

### DIFF
--- a/xml/System.Diagnostics.Tracing/EventSource.xml
+++ b/xml/System.Diagnostics.Tracing/EventSource.xml
@@ -3820,7 +3820,7 @@ desc.DataPointer = (IntPtr)(&arg);
 desc.Size = UIntPtr.Size;
 
 // DateTime arg
-long fileTime = arg.ToFileTimeUtc();
+long fileTime = arg.ToUniversalTime() > new DateTime(1601, 1, 1) ? arg.ToFileTimeUtc() : 0;
 desc.DataPointer = (IntPtr)(&fileTime);
 desc.Size = 8;
 


tarekgh:
CC @davmason

xml/System.Diagnostics.Tracing/EventSource.xml
```diff
@@ -3820,7 +3820,7 @@ desc.DataPointer = (IntPtr)(&arg);
 desc.Size = UIntPtr.Size;
 
 // DateTime arg
-long fileTime = arg.ToFileTimeUtc();
+long fileTime = arg.Ticks > 504_911_232_000_000_000 ? arg.ToFileTimeUtc() : 0;
```
tarekgh, line 3823: > arg.Ticks > 504_911_232_000_000_000  [](http://example.com/codeflow?start=15&length=37)

Would be better to write this as 

```C$
arg.ToUniversalTime() > new DateTime(1601, 1, 1)
```

This is more accurate to compare the Utc ticks of the time and also better for readability.

noahfalk:
@msnilsen - Thanks for investigating the issue and contributing a fix to the docs. Very much appreciated!

xml/System.Diagnostics.Tracing/EventSource.xml
msnilsen: Good catch, thanks. Will update

davmason:
Thanks! Looks good to me outside of Tarek's feedback

tarekgh (approved):

PR #9249, opened 2023-09-02
Update Math.Pow description in Math.xml

arturek:
## Summary

Updates the Math.Pow method description to match its behavior on .Net 8.0.

Fixes dotnet/runtime#90725

### DIFF
--- a/xml/System/Math.xml
+++ b/xml/System/Math.xml
@@ -6033,26 +6033,28 @@ The following example demonstrates how to use the <xref:System.Math.Min%2A> meth
   
 ## Remarks  
  The following table indicates the return value when various values or ranges of values are specified for the `x` and `y` parameters. For more information, see <xref:System.Double.PositiveInfinity?displayProperty=nameWithType>, <xref:System.Double.NegativeInfinity?displayProperty=nameWithType>, and <xref:System.Double.NaN?displayProperty=nameWithType>.  
-  
-|Parameters|Return value|  
-|----------------|------------------|  
-|`x` = `NaN`; `y` is not 0.|`NaN`|  
-|`x` is any value except 1; `y` = `NaN`.|`NaN`|  
-|`x` is any value; `y` = 0.|1|  
-|`x` = `NegativeInfinity`; `y` < 0.|0|  
-|`x` = `NegativeInfinity`; `y` is a positive odd integer.|`NegativeInfinity`|  
-|`x` = `NegativeInfinity`; `y` is positive but not an odd integer.|`PositiveInfinity`|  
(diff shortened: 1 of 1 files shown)


xml/System/Math.xml
gewarren, line 1: If the previous text is correct for the previous versions (e.g. .NET 7) then we need to retain it and just annotate the behavior differences for .NET 8+.

arturek: It works like this at least since .Net 6.0. I don't have earlier versions to check.

arturek: I wrote a [small program to print results of Math.Pow for various parameters](https://gist.github.com/arturek/a39a91ef5cb800d3e90d8f7bfe20233e). Thanks to https://dotnetfiddle.net/ I ran it in selected runtimes and I noticed that all runtimes based on .Net Core behaved the same. On the other hand, .Net Framework in two cases returned different results when the NaN value was used.

In what way differences in behaviour of Math.Pow method between runtimes should be annotated? 

# Results of Math.Pow(x, y)

## .Net 8 Preview 3 (and .Net Core 3.1,  .Net 7)
|            x\y |      -Infinity |              0 |              1 |              2 |       Infinity |            NaN |
|            :-: |            --- |            --- |            --- |            --- |            --- |            --- |
|  **-Infinity** |              0 |              1 |      -Infinity |       Infinity |       Infinity |            NaN |
|          **0** |       Infinity |              1 |              0 |              0 |              0 |            NaN |
|          **1** |              1 |              1 |              1 |              1 |              1 |              1 |
|          **2** |              0 |              1 |              2 |              4 |       Infinity |            NaN |
|   **Infinity** |              0 |              1 |       Infinity |       Infinity |       Infinity |            NaN |
|        **NaN** |            NaN |              1 |            NaN |            NaN |            NaN |            NaN |

## .Net Framework 4.7.2
|            x\y |             -∞ |              0 |              1 |              2 |              ∞ |            NaN |
|            :-: |            --- |            --- |            --- |            --- |            --- |            --- |
|         **-∞** |              0 |              1 |             -∞ |              ∞ |              ∞ |            NaN |
|          **0** |              ∞ |              1 |              0 |              0 |              0 |            NaN |
|          **1** |              1 |              1 |              1 |              1 |              1 |            NaN |
|          **2** |              0 |              1 |              2 |              4 |              ∞ |            NaN |
|          **∞** |              0 |              1 |              ∞ |              ∞ |              ∞ |            NaN |
|        **NaN** |            NaN |            NaN |            NaN |            NaN |            NaN |            NaN |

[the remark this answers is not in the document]
arturek: Thank you, it looks much better. I made some minor edits to it and updated the PR.

gewarren (approved):
LGTM!

Issue #9250, opened 2023-09-02
Document exceptions/behavior of ChannelWriter writing after writer is completed.


poizan42:
**Issue description**

The documentation for `ChannelWriter<T>.(TryWrite|WaitToWriteAsync|WriteAsync)` does not say what happens in case the writer has been completed. In fact `WaitToWriteAsync` throws the exception set in `ChannelWriter.Complete(Exception)` if specified while returing `false` if no exception was specified. In contrast `WriteAsync` always throws a System.Threading.Channels.ChannelClosedException, optionally with the supplied exception as its InnerException. Neither of these are documented as throwing any exceptions.

The following program generates the same output on .NET 6.0 and .NET Framework 4.8
<details>
<summary><strong>Test program source </strong></summary>

```CSharp
using System;
using System.Threading.Channels;
using System.Threading.Tasks;

namespace ChannelTest;

public class Program
{
    private static string GetExceptionDescriptionWithInnerExceptions(Exception e)
    {
        if (e.InnerException != null)
            return $"{e.GetType().FullName}: {e.Message}  ---> {GetExceptionDescriptionWithInnerExceptions(e.InnerException)}";
        else
            return $"{e.GetType().FullName}: {e.Message}";
    }
    private static void PrintResultOrError(string op, Func<Task> f)
        => PrintResultOrError(op, async () => { await f(); return "Success"; });
    private static void PrintResultOrError<T>(string op, Func<T> f)
        => PrintResultOrError(op, () => Task.FromResult(f));
    private static void PrintResultOrError<T>(string op, Func<Task<T>> f)
    {
        try
        {
            T result = f().GetAwaiter().GetResult();
            Console.WriteLine($"{op}: {result}");
        }
        catch (Exception e)
        {
            Console.WriteLine($"{op} failed: {GetExceptionDescriptionWithInnerExceptions(e)}");
        }
    }
    public static void Main(string[] args)
    {
        Console.WriteLine("== Complete without exception ==");
        TestChannelWriterCompleted(null);
        Console.WriteLine();
        Console.WriteLine("== Complete with exception ==");
        TestChannelWriterCompleted(new Exception("Foobar!"));
    }

    private static void TestChannelWriterCompleted(Exception? exception)
    {
        Channel<string> channel = Channel.CreateBounded<string>(new BoundedChannelOptions(5) { FullMode = BoundedChannelFullMode.Wait, SingleReader = true, SingleWriter = true });
[...]

PR #9268, opened 2023-09-11
Clarify exceptions for ChannelWriter methods

MichaelDeutschCoding:
Addresses #9250

## changes to ChannelWriter`1.xml
- WriteAsync: add exception to its documentation
- WaitToWriteAsync: add exception and remarks

### DIFF
--- a/xml/System.Threading.Channels/ChannelWriter`1.xml
+++ b/xml/System.Threading.Channels/ChannelWriter`1.xml
@@ -159,7 +159,8 @@
       <Docs>
         <param name="item">The item to write.</param>
         <summary>Attempts to write the specified item to the channel.</summary>
-        <returns><see langword="true" /> if the item was written; otherwise, <see langword="false" />.</returns>
+        <returns>
+          <see langword="true" /> if the item was written; otherwise, <see langword="false" />.</returns>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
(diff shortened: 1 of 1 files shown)


xml/System.Threading.Channels/ChannelWriter`1.xml
```diff
@@ -159,8 +159,7 @@
       <Docs>
         <param name="item">The item to write.</param>
         <summary>Attempts to write the specified item to the channel.</summary>
-        <returns>
-          <see langword="true" /> if the item was written; otherwise, <see langword="false" />.</returns>
+        <returns><see langword="true" /> if the item was written; otherwise, <see langword="false" />.</returns>
```
gewarren, line 162: The next time the CI runs, this will go back to being on a newline, so it's best to undo this change.

xml/System.Threading.Channels/ChannelWriter`1.xml
```diff
@@ -196,9 +195,14 @@
...
-        <returns>A <see cref="T:System.Threading.Tasks.ValueTask`1" /> that will complete with a <see langword="true" /> result when space is available to write an item
-            or with a <see langword="false" /> result when no further writing will be permitted.</returns>
-        <remarks>To be added.</remarks>
+        <returns>
+          <para>A <see cref="T:System.Threading.Tasks.ValueTask`1" /> that will complete with a <see langword="true" /> result when space is available to write an item
+            or with a <see langword="false" /> result when no further writing will be permitted.</para>
+          <para>If the channel was completed with an exception, this task will also complete with an exception.</para>
+        </returns>
+        <exception cref="T:System.Threading.Channels.ChannelClosedException">The channel has already been marked as complete, including an inner exception.</exception>
+        <remarks>If the channel was already closed, this method will not return <see langword="true" />. Whether it returns <see langword="false" /> or throws an exception
```
gewarren, line 204:
```suggestion
        <remarks>If the channel was already closed, this method does not return <see langword="true" />. Whether it returns <see langword="false" /> or throws an exception
```

xml/System.Threading.Channels/ChannelWriter`1.xml
MichaelDeutschCoding: It just seemed odd that the opening tag was on its own line, but the the closing `</returns>` tag wasn't. It also looked inconsistent with the other sections in the file.

Would it be better to add a newline before the closing tag, or just leave it exactly as it was?

gewarren: The CI that we run to ingest the APIs puts anything that starts with `<` on a newline, so it's best to leave it just the way it was.

PR #9289, opened 2023-09-13
Port assorted isolated docs and fix some old weird tags

carlossanlop:
Please review ignoring whitespace.

### DIFF
--- a/xml/System.IO/RandomAccess.xml
+++ b/xml/System.IO/RandomAccess.xml
@@ -54,7 +54,7 @@
         <remarks>
           <para> This method calls platform-dependent APIs such as <c>FlushFileBuffers()</c> on Windows and <c>fsync()</c> on Unix.
             </para>
-          <para> Flushing the buffers causes data to be written to disk which is a relatively expensive operation. It is recommended that you perform multiple writes to the file and then call this method either when you are done writing to the file or periodically if you expect to continue writing to the file over a long period of time.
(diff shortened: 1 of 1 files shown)


xml/System.IO/RandomAccess.xml
```diff
@@ -1,627 +1,638 @@
...
+      <Parameters>
+        <Parameter Name="handle" Type="Microsoft.Win32.SafeHandles.SafeFileHandle" Index="0" FrameworkAlternate="net-8.0" />
+      </Parameters>
+      <Docs>
+        <param name="handle">The file handle.</param>
+        <summary>Flushes the operating system buffers for the given file to disk.</summary>
+        <remarks>
+          <para> This method calls platform-dependent APIs such as <c>FlushFileBuffers()</c> on Windows and <c>fsync()</c> on Unix.
+            </para>
+          <para> Flushing the buffers causes data to be written to disk which is a relatively expensive operation. It is recommended that you perform multiple writes to the file and then call this method either when you are done writing to the file or periodically if you expect to continue writing to the file over a long period of time.
```
gewarren, line 57:
```suggestion
          <para> Flushing the buffers causes data to be written to disk, which is a relatively expensive operation. It is recommended that you perform multiple writes to the file and then call this method either when you are done writing to the file or periodically if you expect to continue writing to the file over a long period of time.
```

gewarren (approved):

PR #9290, opened 2023-09-13
Correct UInt32 to UInt64 for ReadCborNegativeIntegerRepresentation

vcsjones:
## Summary

The documentation for CborReader.ReadCborNegativeIntegerRepresentation says it will throw an `OverflowException` if the encoded integer is out of range for UInt32. ~It seems like this should be UInt64.~ This exception isn't actually thrown.

### DIFF
--- a/xml/System.Formats.Cbor/CborReader.xml
+++ b/xml/System.Formats.Cbor/CborReader.xml
@@ -356,7 +356,6 @@
             Useful for handling values that do not fit in the <see cref="T:System.Int64" /> type.
             </remarks>
         <exception cref="T:System.InvalidOperationException">The next data item does not have the correct major type.</exception>
-        <exception cref="T:System.OverflowException">The encoded integer is out of range for <see cref="T:System.UInt64" /></exception>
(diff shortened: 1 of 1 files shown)


xml/System.Formats.Cbor/CborReader.xml
```diff
@@ -356,7 +356,7 @@
             Useful for handling values that do not fit in the <see cref="T:System.Int64" /> type.
             </remarks>
         <exception cref="T:System.InvalidOperationException">The next data item does not have the correct major type.</exception>
-        <exception cref="T:System.OverflowException">The encoded integer is out of range for <see cref="T:System.UInt32" /></exception>
+        <exception cref="T:System.OverflowException">The encoded integer is out of range for <see cref="T:System.UInt64" /></exception>
```
eiriktsarpalis, line 359: Actually I think this is erroneous, the method will either throw `InvalidOperationException` if it's not a negative number or a `CborContentException` because the CBOR encoding is invalid.
```suggestion
```

eiriktsarpalis (approved):

PR #9293, opened 2023-09-14
Fix descriptions of invalid eventlog overflow policy

Robo210:
Support for "Retain events older than X days" was dropped in Windows Vista (and it only ever almost worked on XP/Server 2003). This was a breaking change made roughly _15 years ago_, and the new policy is documented as being strictly incompatible with the old retention setting:
  - XP/2003: Set reg value to 0 for "overwrite based on log size", > 0 for "overwrite based on the value as a time", and -1 for "never overwrite events"
  - Vista onward: Set reg value to 0 for "overwrite based on log size" and any non-zero value for "never overwrite events".

Encouraging people to continue to try to utilize the XP/2003 behavior of setting a time will in fact make the system never overwrite events at all, which in turn will break the system when other security policies are applied (e.g. "don't allow users to log in when the security log is full" or "bug check the machine when a security event is lost").

This change does not fully remove every reference to the old behaviors across the entire code base, but it is a start.

See the description of the Retention value at https://learn.microsoft.com/en-us/windows/win32/eventlog/eventlog-key for details.


antonfirsov (dismissed):
LGTM /cc @dotnet/area-system-memory

xml/System.Diagnostics/EventLog.xml
```diff
@@ -2583,10 +2579,11 @@ SVC_UPDATE.EXE
...
-  
- Set the `action` parameter to <xref:System.Diagnostics.OverflowAction.OverwriteOlder> to indicate that each new entry overwrites older entries when the <xref:System.Diagnostics.EventLog> reaches its maximum size. Specify the number of days that events must be retained in the log using the `retentionDays` parameter. Events written within the retention range are not overwritten by new entries.  
-  
+
  Set the `action` parameter to <xref:System.Diagnostics.OverflowAction.DoNotOverwrite> to discard new events when the maximum log size is reached. If the `action` parameter is set to <xref:System.Diagnostics.OverflowAction.DoNotOverwrite>, the `retentionDays` parameter value is ignored.  
+
+> [!WARNING]
+> Never set the `action` parameter to <xref:System.Diagnostics.OverflowAction.OverwriteOlder>. This value is deprecated and may cause the log to behave as if the `DoNotOverwrite` parameter was passed instead.
```
gewarren, line 2586:
```suggestion
> Never set the `action` parameter to <xref:System.Diagnostics.OverflowAction.OverwriteOlder>. This value is deprecated and might cause the log to behave as if the `DoNotOverwrite` parameter was passed instead.
```

xml/System.Diagnostics/OverflowAction.xml
```diff
@@ -33,12 +33,13 @@
...
   
--   New entries will overwrite older entries.  
-  
--   New entries will either be discarded or will overwrite existing entries depending upon the age of the existing entries.  
+-   New entries will overwrite older entries.
   
  Use the <xref:System.Diagnostics.EventLog.ModifyOverflowPolicy%2A> method to set the overflow behavior for an <xref:System.Diagnostics.EventLog>. Check the current configured behavior of an <xref:System.Diagnostics.EventLog> through its <xref:System.Diagnostics.EventLog.OverflowAction%2A> property.  
 
+> [!WARNING]
+> The `OverwriteOlder` behavior is deprecated. Using this value may cause the Event Log to behave as if the `DoNotOverwrite` value was used instead, which will cause events to be discarded when the log is full.
```
gewarren, line 41:
```suggestion
> The `OverwriteOlder` behavior is deprecated. Using this value might cause the Event Log to behave as if the `DoNotOverwrite` value was used instead, which will cause events to be discarded when the log is full.
```

xml/System.Diagnostics/OverflowAction.xml
```diff
@@ -141,17 +142,17 @@ The following example enumerates the event logs defined on the local computer an
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>Indicates that new events overwrite events older than specified by the <see cref="P:System.Diagnostics.EventLog.MinimumRetentionDays" /> property value when the event log is full. New events are discarded if the event log is full and there are no events older than specified by the <see cref="P:System.Diagnostics.EventLog.MinimumRetentionDays" /> property value.</summary>
+        <summary>Deprecated</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[  
-  
+> [!WARNING]
+> Do not use this value. Doing so may cause the Event Log to behave as if the `DoNotOverwrite` value was used instead.
```
gewarren, line 149:
```suggestion
> Do not use this value. Doing so might cause the Event Log to behave as if the `DoNotOverwrite` value was used instead.
```

gewarren (commented):
Should the deprecated methods be obsoleted instead of just setting the summary to "deprecated"? That way the docs would automatically show a banner similar to this: https://learn.microsoft.com/en-us/dotnet/api/system.codedom.compiler.codedomprovider.createcompiler?view=net-8.0#definition.

Robo210:
> Should the deprecated methods be obsoleted instead of just setting the summary to "deprecated"? That way the docs would automatically show a banner similar to this: https://learn.microsoft.com/en-us/dotnet/api/system.codedom.compiler.codedomprovider.createcompiler?view=net-8.0#definition.

That sounds like a good idea, but outside of my expertise. I am wary of causing build breaks for existing code (something we can't do in Win32 land, as much as I'd like to clear out a ton of obsolete nonsense); updating the docs at least lets support engineers point impacted customers in the correct direction. I am hoping the docs can also make it clear that this is deprecated/obsolete based on the Windows version the code is running on, not a specific .Net version (i.e. I don't want a customer to say "the docs claim it is obsolete in .net 9 but we're using .net framework 3 so clearly we're still good to use it").

xml/System.Diagnostics/EventLog.xml
```diff
@@ -2500,17 +2500,13 @@ SVC_UPDATE.EXE
         <ReturnType>System.Int32</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets the number of days to retain entries in the event log.</summary>
-        <value>The number of days that entries in the event log are retained. The default value is 7.</value>
+        <summary>Deprecated</summary>
```
gewarren, line 2503:
```suggestion
        <summary>This property is deprecated.</summary>
```

xml/System.Diagnostics/EventLog.xml
```diff
@@ -2500,17 +2500,13 @@ SVC_UPDATE.EXE
...
+        <value>Deprecated</value>
```
gewarren, line 2504:
```suggestion
        <value>To be added.</value>
```

gewarren: (This will render as empty in the docs.)

xml/System.Diagnostics/EventLog.xml
```diff
@@ -2500,17 +2500,13 @@ SVC_UPDATE.EXE
...
-  
- The overflow behavior only occurs when an event log reaches its size limit. When an <xref:System.Diagnostics.EventLog> has its <xref:System.Diagnostics.EventLog.OverflowAction%2A> set to <xref:System.Diagnostics.OverflowAction.OverwriteOlder>, and the event log reaches its maximum size, then new entries are written only if they can replace entries whose age exceeds the <xref:System.Diagnostics.EventLog.MinimumRetentionDays%2A> period. Retaining event entries for a minimum period is appropriate when the event log is archived regularly. Otherwise, you risk losing new entries when the event log reaches its limit. To avoid losing new event information, set the minimum retention days for events based on your archive schedule for a particular event log.  
+ Support for the <xref:System.Diagnostics.EventLog.MinimumRetentionDays%2A> property was removed in Windows Vista and later operating systems. Setting this value can cause the Event Log to never overwrite events, and to drop all events to the channel once it is full.
```
gewarren, line 2509:
```suggestion

> [!IMPORTANT]
> Support for the <xref:System.Diagnostics.EventLog.MinimumRetentionDays%2A> property was removed in Windows Vista and later operating systems. Setting this value can cause the Event Log to never overwrite events, and to drop all events to the channel once it is full.
```

xml/System.Diagnostics/OverflowAction.xml
```diff
@@ -141,17 +142,17 @@ The following example enumerates the event logs defined on the local computer an
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>Indicates that new events overwrite events older than specified by the <see cref="P:System.Diagnostics.EventLog.MinimumRetentionDays" /> property value when the event log is full. New events are discarded if the event log is full and there are no events older than specified by the <see cref="P:System.Diagnostics.EventLog.MinimumRetentionDays" /> property value.</summary>
+        <summary>Deprecated</summary>
```
gewarren, line 145:
```suggestion
        <summary>This field is deprecated.</summary>
```

gewarren (approved):

PR #9302, opened 2023-09-18
Add comments for new System.Net.Http.Json members

MihaZupan:
Add comments for new `System.Net.Http.Json` members.

Contributes to https://github.com/dotnet/runtime/issues/88569.


carlossanlop (approved):
LGTM. I'd provide suggestions to show the `<see cref""/>` for the `jsonTypeInfo` parameters but most of those docs predate this effort so I guess they can stay that way.

PR #9307, opened 2023-09-18
Remove invalid commented out code in networkexamples.cs

krmeda-msft:
Removed invalid syntax code that was commented out in examples.

## Summary

Describe your changes here.

Removed a line of commented code which was invalid syntax and doesn't add any more clarity to the example in snippet. 


MihaZupan (approved):
Thanks

PR #9323, opened 2023-09-25
DbProviderFactories.GetFactory supported on net402

OwnageIsMagic:
See net402 release notes https://support.microsoft.com/en-us/topic/update-4-0-2-for-microsoft-net-framework-4-runtime-update-8d0f21a2-2459-c5dd-8a23-49518a89cafc#ID0EDBBHBD

### DIFF
--- a/xml/System.Data.Common/DbProviderFactories.xml
+++ b/xml/System.Data.Common/DbProviderFactories.xml
@@ -91,7 +91,7 @@
         <ReturnType>System.Data.Common.DbProviderFactory</ReturnType>
       </ReturnValue>
       <Parameters>
-        <Parameter Name="connection" Type="System.Data.Common.DbConnection" Index="0" FrameworkAlternate="net-5.0;net-6.0;net-7.0;net-8.0;netcore-2.1;netcore-2.2;netcore-3.0;netcore-3.1;netframework-4.5;netframework-4.5.1;netframework-4.5.2;netframework-4.6;netframework-4.6.1;netframework-4.6.2;netframework-4.7;netframework-4.7.1;netframework-4.7.2;netframework-4.8;netframework-4.8.1;netstandard-2.1;xamarinmac-3.0" />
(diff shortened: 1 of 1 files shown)


roji (commented):
.NET Framework 4.0.2 is out of support, I don't think we should be adding out-of-support patch versions to this list.

OwnageIsMagic:
But it doesn't hurt anyone and makes the world a better place. If you don't like 4.0.2 we can lie a little and say 4.0.3 which have official PCL and NuGet TFM.

roji:
@OwnageIsMagic adding each and every supported patch version - including unsupported ones - makes the list harder to read for everyone. We'd also need to be consistent, checking and adding all other patch versions to this list (otherwise why specifically 4.0.2), and do the same for other APIs. That simply doesn't make sense.

OwnageIsMagic:
@roji hey, look at the current content. It literally names every minor patch of net framework since 4.5
4.5, 4.5.1, 4.5.2, 4.6, 4.6.1, 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8, 4.8.1
4.0.2 here is special since it's minimal required.
But you are partially right: I should add both 4.0.2 and 4.0.3.
Btw this is the only API added in 4.0.2
4.0.1 added Windows Workflow Foundation `StateMachine`
4.0.3 `GCLatencyMode.SustainedLowLatency` and `SqlConnection.ClientConnectionId`

I can add them to this PR too, so it will be completely covered.

roji:
I'll leave someone from the .NET API docs team to answer about the policy here (I'm just a dev working on related APIs and don't make these decisions). But my opinion is that we shouldn't be adding all unsupported patch versions everywhere in all docs. It's both a waste of time and makes the docs harder to read. If anything I'd consider removing all unsupported TFMs.

MSDN-WhiteKnight:
This XML metadata actually should not be modified at all, it is automatically generated by CI. Only the contents of `Docs` tag are manually editable in this repository. 

4.0.2 patch is probably not included here because CI does not pick reference assemblies for that exact version. Just adding it here might be not enough to display it on docs website without explicitly enabling it in some config.

PR #9328, opened 2023-09-26
Update Marshal.xml

colejohnson66:
Remove version support blurbs.

## Summary

Many methods in `Marshal` contain text blocks of "[Supported in the .NET Framework 4.5.1 and later versions]" at the top of the summary tags. This text is duplicating information that can be found in the "Applies To" table, and has leaves out all the other runtimes (.NET Standard, .NET [Core], UWP, etc.). For consistency with almost every other method's documentation, I've removed the text blurbs.


AaronRobinsonMSFT (approved):

AaronRobinsonMSFT:
Thanks!

PR #9334, opened 2023-09-26
Sort LogLevel members by MemberValue instead of MemberName

gerardogalvez:
## Summary

Sort the LogLevel enum members by `MemberValue` instead of `MemberName`, makes the table in the docs easier to read.

### DIFF
--- a/xml/Microsoft.Extensions.Logging/LogLevel.xml
+++ b/xml/Microsoft.Extensions.Logging/LogLevel.xml
@@ -27,13 +27,13 @@
     <remarks>To be added.</remarks>
   </Docs>
   <Members>
-    <Member MemberName="Critical">
-      <MemberSignature Language="C#" Value="Critical" />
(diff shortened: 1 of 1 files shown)


gerardogalvez:
@dotnet-policy-service agree

tarekgh:
I am seeing the preview link https://review.learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel?view=dotnet-plat-ext-8.0&branch=pr-en-us-9334 showing the order in the table didn't change. @gewarren do you know what controls the order in the tables? Is it always sorted by field names?

gewarren:
Yes, the order is predetermined as far as I can tell (that is, alphabetic). Also, even if we could force it into the order by numerical value, the next time the CI is run to refresh the API docs, this change would be overwritten. We have an internal work item request to change the default sort order. See https://github.com/dotnet/dotnet-api-docs/issues/4210.

gewarren:
This has been fixed! https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel?view=net-8.0#fields

PR #9338, opened 2023-09-27
fix typo

lemonez:
## Summary

- Fixed a typo in the `ValidationAttribute.xml` docstring.
- There is also a typo in [xml/System/String.xml#L7202](https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System/String.xml#L7202) ("An parameter...") but when I try to fix it, it reformats the whole file because of a line ending issue or something, and I haven't been able to circumvent that either locally or in the GitHub UI.


gewarren:
> it reformats the whole file because of a line ending issue or something, and I haven't been able to circumvent that either locally or in the GitHub UI.

@lemonez Don't worry about that, we see it a lot. There's a tool we use that unfortunately also changes the line endings, so it's normal that they get changed back on simple edits like yours.

gewarren (approved):

PR #9341, opened 2023-09-28
Update Timer.xml - wrong info in remarks of Timer.AutoReset Property

kirsan31:
## Summary
Removed this wrong and unneeded  sentence from https://learn.microsoft.com/en-us/dotnet/api/system.timers.timer.autoreset?view=net-7.0#remarks : 

> If the Timer is already enabled when the Start method is called, the interval is reset.

### DIFF
--- a/xml/System.Timers/Timer.xml
+++ b/xml/System.Timers/Timer.xml
@@ -345,7 +345,7 @@
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- If the <xref:System.Timers.Timer> is already enabled when the <xref:System.Timers.Timer.Start%2A> method is called, the interval is reset. If <xref:System.Timers.Timer.AutoReset%2A> is `false`, the <xref:System.Timers.Timer.Start%2A> method must be called in order to start the count again.
+ If <xref:System.Timers.Timer.AutoReset%2A> is `false`, the <xref:System.Timers.Timer.Start%2A> method must be called in order to start the count again.
 
  Resetting the interval affects when the <xref:System.Timers.Timer.Elapsed> event is raised. For example, if you set the interval to 5 seconds and then set the <xref:System.Timers.Timer.Enabled%2A> property to `true`, the count starts at the time <xref:System.Timers.Timer.Enabled%2A> is set. If you reset the interval to 10 seconds when the count is 3 seconds, the <xref:System.Timers.Timer.Elapsed> event is raised for the first time 13 seconds after the <xref:System.Timers.Timer.Enabled%2A> property was set to `true`.
 


gewarren:
@stephentoub Can you verify this small change?

xml/System.Timers/Timer.xml
```diff
@@ -345,7 +345,7 @@
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- If the <xref:System.Timers.Timer> is already enabled when the <xref:System.Timers.Timer.Start%2A> method is called, the interval is reset. If <xref:System.Timers.Timer.AutoReset%2A> is `false`, the <xref:System.Timers.Timer.Start%2A> method must be called in order to start the count again.
+ If <xref:System.Timers.Timer.AutoReset%2A> is `false`, the <xref:System.Timers.Timer.Start%2A> method must be called in order to start the count again.
```
stephentoub, line 348: There is no System.Timers.Timer.Start, so, yeah :)

stephentoub (approved):

xml/System.Timers/Timer.xml
kirsan31: @stephentoub 
> There is no System.Timers.Timer.Start, so, yeah :)

There is: https://source.dot.net/#System.ComponentModel.TypeConverter/System/Timers/Timer.cs,d8d4c6fbc471f649
But it will not reset interval on consecutive calls:
https://source.dot.net/System.ComponentModel.TypeConverter/R/c458cff06715bb66.html
```cs
public bool Enabled
{
    get => _enabled;
    set
    {
        if (DesignMode)
        {
            _delayedEnable = value;
            _enabled = value;
        }
        else if (_initializing)
        {
            _delayedEnable = value;
        }
        else if (_enabled != value)
        {
            if (!value)
            {
                if (_timer != null)
                {
                    _cookie = null;
                    _timer.Dispose();
                    _timer = null;
                }
                _enabled = value;
            }
            else
            {
                _enabled = value;
                if (_timer == null)
                {
                    ObjectDisposedException.ThrowIf(_disposed, this);

                    int i = (int)Math.Ceiling(_interval);
                    _cookie = new object();
                    _timer = new Threading.Timer(_callback, _cookie, Timeout.Infinite, Timeout.Infinite);
                    _timer.Change(i, _autoReset ? i : Timeout.Infinite);
                }
                else
                {
                    UpdateTimer();
                }
            }
        }
    }
}
```

stephentoub: Huh, you're right. I only ever used Enabled directly and could not have told you there was a one-line helper on top of it :-)

PR #9344, opened 2023-09-28
Removes references to being able to set DataGridCell.IsReadOnly property

adamint:
## Summary

Twice, these docs mention that users are able to set DataGridCell.IsReadOnly to make one cell read-only. This is incorrect, as this property is get-only.


gewarren:
Sorry for taking 2 years to merge this :-S

PR #9350, opened 2023-09-29
Update Thread.Name documentation

eduardo-vp:
The Name property can now be renamed (doesn't throw an exception anymore). Based on the changes here https://github.com/dotnet/runtime/pull/47084

Fixes https://github.com/dotnet/runtime/issues/85635

### DIFF
--- a/xml/System.Threading/Thread.xml
+++ b/xml/System.Threading/Thread.xml
@@ -3053,10 +3053,13 @@ The value is not guaranteed to be a zero-based processor number.
           <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
+ In .NET 5 and earlier versions, this property is write-once. In these versions, because the default value of a thread's <xref:System.Threading.Thread.Name%2A> property is `null`, you can determine whether a name has already been explicitly assigned to the thread by comparing it with `null`.
+
  The string assigned to the <xref:System.Threading.Thread.Name%2A> property can include any Unicode character.
   
  ]]></format>
         </remarks>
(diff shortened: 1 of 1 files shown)


eduardo-vp:
I removed the code samples. However I'm not sure if the property is simple enough for this or it should still have a sample code

xml/System.Threading/Thread.xml
```diff
@@ -3053,21 +3053,10 @@ The value is not guaranteed to be a zero-based processor number.
...
- The following example shows how to name a thread.  
-  
- :::code language="cpp" source="~/snippets/cpp/VS_Snippets_CLR_System/system.Threading.Thread.Name/CPP/source.cpp" id="Snippet1":::
- :::code language="csharp" source="~/snippets/csharp/System.Threading/Thread/Name/source.cs" id="Snippet1":::
- :::code language="fsharp" source="~/snippets/fsharp/System.Threading/Thread/Name/source.fs" id="Snippet1":::
- :::code language="vb" source="~/snippets/visualbasic/VS_Snippets_CLR_System/system.Threading.Thread.Name/VB/source.vb" id="Snippet1":::  
-  
  ]]></format>
         </remarks>
-        <exception cref="T:System.InvalidOperationException">A set operation was requested, but the <see langword="Name" /> property has already been set.</exception>
```
gewarren, line 3070: We should keep this but start the text with:

.NET 7 and earlier versions only: A set operation...

gewarren (changes requested):

xml/System.Threading/Thread.xml
eduardo-vp: I just checked and this is true for .NET 5 and earlier since the change was done some time ago so I'll add that

gewarren (approved):
Thanks @eduardo-vp!

PR #9362, opened 2023-10-03
Add OperationCanceledException exceptions automatically

gewarren:
Replaces #8776, which had too many conflicts.

### DIFF
--- a/xml/System/WindowsRuntimeSystemExtensions.xml
+++ b/xml/System/WindowsRuntimeSystemExtensions.xml
@@ -213,7 +213,6 @@ Use this method to get a <xref:System.Threading.Tasks.Task> object for a Windows
         </remarks>
         <exception cref="T:System.ArgumentNullException">
           <paramref name="source" /> is <see langword="null" />.</exception>
-        <exception cref="T:System.OperationCanceledException">The cancellation token was canceled. This exception is stored into the returned task.</exception>
       </Docs>
     </Member>
     <Member MemberName="AsTask&lt;TProgress&gt;">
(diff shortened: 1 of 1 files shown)


gewarren:
@stephentoub If the methods actually throw a TaskCanceledException, I guess it's still okay to catch OperationCanceledException when you await?

stephentoub:
> @stephentoub If the methods actually throw a TaskCanceledException, I guess it's still okay to catch OperationCanceledException when you await?

Yes, in fact we want folks to catch OperationCanceledException and not TaskCanceledException. From my perspective, our introducing TaskCanceledException was a well-intentioned mistake. The idea was that it would be an OperationCanceledException that would carry a little more information (the Task that was canceled) if necessary, but that additional information is basically never useful or needed, and it's often not even possible to provide, yet because APIs were throwing it, developers started writing catch blocks for it, which then made it a breaking change for the implementation throwing it to be tweaked in such a way that the base OperationCanceledException was being thrown. We should probably introduce an analyzer to warn people off from catching TaskCanceledException, and maybe even obsolete it.

BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

BillWagner:
Based on @stephentoub 's comment, I checked our conceptual and samples repos. 

In conceptual docs, I found 138 instances of `TaskCanceledException`. In our samples, I found 101 instances. Should we update the relevant `catch` clauses, and the text describing them to catch `OperationCanceledException` instead?

gewarren:
> In conceptual docs, I found 138 instances of `TaskCanceledException`. In our samples, I found 101 instances. Should we update the relevant `catch` clauses, and the text describing them to catch `OperationCanceledException` instead?

@BillWagner I'd say yes!

stephentoub:
Yes, please and thank you.

PR #9367, opened 2023-10-03
Fix up operation canceled exceptions

gewarren:
Users should catch OperationCanceledException instead of TaskCanceledException (see https://github.com/dotnet/dotnet-api-docs/pull/9362#issuecomment-1745109860).


BillWagner (approved):
This also LGTM @gewarren 


Let's :shipit:

PR #9384, opened 2023-10-11
CI update - .NET 8 RC 2

gewarren:


xml/Microsoft.Extensions.Compliance.Redaction/XXHash3RedactorOptions.xml
gewarren, line 1: @huangmin-ms Probably the casing of the file name should have changed too. Should I log a bug?

carlossanlop (approved):

xml/Microsoft.Extensions.Compliance.Redaction/XXHash3RedactorOptions.xml
huangmin-ms: Windows is case insensitive system. This should be expected.

gewarren: @huangmin-ms Yes, but it no longer matches the name of the type. I believe if you use `git mv` then you can rename a file that only has a case change. Should I rename it manually?

gewarren: I used `git mv` to rename it in this PR: https://github.com/dotnet/dotnet-api-docs/pull/9387

Issue #8973, opened 2023-04-19
Methods which take `StringComparison` are not sufficiently clear about how hard they are to use correctly


Smaug123:
**Methods which take `StringComparison` are not sufficiently clear about how hard they are to use correctly**

I claim that it is easy to read e.g. https://learn.microsoft.com/en-us/dotnet/api/system.string.endswith?view=net-8.0 and come away thinking that "I'll be fine, I'll just stick with the invariant culture because I don't care about anything above code point 127". This is amply discussed in _Best Practices for Using Strings_, and the docs do vaguely suggest that perhaps you might want to read it; but I believe what is actually required in the Notes to Callers is "you *must* read _Best Practices for Using Strings_, *in full*, before using any method that can take a StringComparison". 

In particular, for example, the following documentation (which is where one would naturally look to find the "non-printing characters are omitted in most configurations" footgun) is entirely unhelpful except insofar as it suggests that you might want to read the document whose purpose is to instil the correct sense of unease:

> As explained in [Best Practices for Using Strings](https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings), we recommend that you avoid calling string comparison methods that substitute default values and instead call methods that require parameters to be explicitly specified. To determine whether a string ends with a particular substring by using the string comparison rules of the current culture, call the [EndsWith(String, StringComparison)](https://learn.microsoft.com/en-us/dotnet/api/system.string.endswith?view=net-8.0#system-string-endswith(system-string-system-stringcomparison)) method overload with a value of [CurrentCulture](https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparison?view=net-8.0#system-stringcomparison-currentculture) for its comparisonType parameter.

I read this paragraph entirely as "string comparison is culture-specific, so be careful about which culture you're in", and *not* as the much more important "… and also bear in mind that you probably do not know what any given culture's string comparison rules are, even if you think you do".

I believe *every* method which has an overload which takes a `StringComparison` should contain a Note to Callers that warns the user that they must read the Best Practices. If I were writing the docs, they would say the following:

> You must read Best Practices *in full* before using this function, even if you think you currently know what a culture-sensitive comparison is. Read it all; do not just read the section about the `StringComparison` you currently intend using. (You may skip this step if you can already articulate why the invariant culture is almost never the correct culture to use for string comparison and related operations.)

**Target framework**

True of .NET 5 through at least .NET 8.

- [x] .NET Core
[...]

PR #9388, opened 2023-10-13
More stern warning about string culture comparison

Smaug123:
## Summary

Add a further warning about `InvariantCulture` in string comparison methods.

It continues to be true that you need to read lots of documentation before using `String.StartsWith` and friends, and the docs continue not to tell you that this is the case. This PR elevates some more of the most critical information from _Best Practices for Using Strings_ to the API docs.

Source for "rarely appropriate" is _Best Practices_:

> On balance, the invariant culture has few properties that make it useful for comparison. It does comparison in a linguistically relevant manner, which prevents it from guaranteeing full symbolic equivalence, but it isn't the choice for display in any culture. One of the few reasons to use [StringComparison.InvariantCulture](https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparison#system-stringcomparison-invariantculture) for comparison is to persist ordered data for a cross-culturally identical display. For example, if a large data file that contains a list of sorted identifiers for display accompanies an application, adding to this list would require an insertion with invariant-style sorting.

ref: #8973


MSDN-WhiteKnight:
I don't think this change is needed. These notes basically say "This overload uses current culture implicitly, but this might be unexpected. Consider instead using overload that explicitly accepts culture." Nowhere it was implied that InvariantCulture should be used. It is not useful to add that InvariantCulture is rarely appropriate, it's too vague, and too disconnected from the preceding text. If any addition is needed here, it should be more actionable, something like "If you don't need linguistic comparison, consider using StringComparison.Ordinal instead".

xml/System/String.xml
```diff
@@ -5287,7 +5287,7 @@ This method performs an ordinal (case-sensitive and culture-insensitive) compari
...
+          <para>As explained in [Best Practices for Using Strings](/dotnet/standard/base-types/best-practices-strings), we recommend that you avoid calling string comparison methods that substitute default values and instead call methods that require parameters to be explicitly specified. To determine whether a string ends with a particular substring by using the string comparison rules of the current culture, signal your intention explicitly by calling the <see cref="M:System.String.EndsWith(System.String,System.StringComparison)" /> method overload with a value of <see cref="F:System.StringComparison.CurrentCulture" /> for its <paramref name="comparisonType" /> parameter. Note that <see cref="F:System.StringComparison.InvariantCulture" /> is rarely appropriate.</para>
```
tarekgh, line 5290: >Note that <see cref="F:System.StringComparison.InvariantCulture" /> is rarely appropriate. [](http://example.com/codeflow?start=683&length=90)

I don't think this is accurate. It is even more confusing. I can see many scenarios that using invariant culture can be desirable. What is the reason adding this note?

tarekgh: I just saw @MSDN-WhiteKnight comment and fully agree with it. If I am a user even not familiar with the API and reading this suggested note, for sure I will not get anything from it more than confusion.

Smaug123: I have claimed all along that confusion is actually the correct outcome! The API *is*, objectively, confusing, which is why https://github.com/dotnet/designs/pull/207 exists. The *correct* sentiment to instil in the reader of this documentation is "I do not understand this function, so I should read the much more extensive linked documentation about how to use these functions".

> I don't think this is accurate.

I quoted _Best Practices_ which certainly *states* that "On balance, the invariant culture has few properties that make it useful for comparison… One of the few reasons to use [StringComparison.InvariantCulture](https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparison#system-stringcomparison-invariantculture) for comparison is to persist ordered data for a cross-culturally identical display.". I guess you disagree with that too?

tarekgh: > I have claimed all along that confusion is actually the correct outcome! The API is, objectively, confusing, which is why https://github.com/dotnet/designs/pull/207 exists. The correct sentiment to instil in the reader of this documentation is "I do not understand this function, so I should read the much more extensive linked documentation about how to use these functions".

Frankly, your added note does not really help at all with what you claim. 

> I quoted Best Practices which certainly states that "On balance, the invariant culture has few properties that make it useful for comparison… One of the few reasons to use [StringComparison.InvariantCulture](https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparison#system-stringcomparison-invariantculture) for comparison is to persist ordered data for a cross-culturally identical display."

This doesn't mean Invariant is rarely used though.

Smaug123: Well, I did try months ago to add a note saying "you should not use these functions without reading _Best Practices_", but that was rejected ;)

It's true that I have knowledge only of a small cross-section of .NET's usage; I shall defer to you on the question of whether it is widely used correctly. I've personally literally never (not once!) seen a correct use of `InvariantCulture` in a string comparison method, that's all.

tarekgh (approved):
@Smaug123 thanks for submitting the PR and updating the note. The new suggestion changes LGTM.

PR #9392, opened 2023-10-17
JsonIncludeAttribute update for .NET 8

gewarren:
Related to dotnet/docs#36128.

### DIFF
--- a/xml/System.Text.Json.Serialization/JsonIncludeAttribute.xml
+++ b/xml/System.Text.Json.Serialization/JsonIncludeAttribute.xml
@@ -29,7 +29,16 @@
 
 ## Remarks
 
-When applied to a property, indicates that non-public *getters* and *setters* can be used for serialization and deserialization. In .NET 6 and .NET 7, non-public properties aren't supported.
+When applied to a property, this attribute indicates that non-public *getters* and *setters* can be used for serialization and deserialization.
+
+> [!NOTE]
+> Even with this attribute, the serialization source generator is still limited to non-public members that are *accessible* to the generated code. *Accessible* non-public members include those that are:
+>
+> - `internal` when in the same assembly as the generated context.
+> - `protected`, `protected internal`, or `private protected` when the generated context is in a class that derives from the type.
+> - `private` if the type is nested in the same class as the generated context.
+
+In .NET 6 and .NET 7, non-public properties aren't supported.
 
           ]]></format>
     </remarks>


xml/System.Text.Json.Serialization/JsonIncludeAttribute.xml
```diff
@@ -29,11 +29,11 @@
 
 ## Remarks
 
-When applied to a property, indicates that non-public getters and setters can be used for serialization and deserialization. Non-public properties are not supported.
+When applied to a property, indicates that non-public *getters* and *setters* can be used for serialization and deserialization. In .NET 6 and .NET 7, non-public properties aren't supported.
```
eiriktsarpalis, line 32: It might make sense to add a section clarifying that the source generator is limited to non-public members that are still visible to generated code (i.e. `internal` works fine if in the same project, `private` is mostly out of reach modulo exceptions).

eiriktsarpalis (approved):

xml/System.Text.Json.Serialization/JsonIncludeAttribute.xml
BillWagner: Assuming I understand Erik's comment correctly, the phrase we use in the C# reference is that "the property is accessible". Its use includes `public`, `internal` when in the same assembly, `protected` for derived classes, and the combos for `protected internal` and `private protected`.

BillWagner (approved):
This LGTM, once the conversation is resolved.

xml/System.Text.Json.Serialization/JsonIncludeAttribute.xml
eiriktsarpalis: Exactly that. In some cases even `private` might work (for types that are nested inside the same class as the generated context) but that's splitting hairs. The term that Roslyn uses for that concept is ["accessible"](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.compilation.issymbolaccessiblewithin?view=roslyn-dotnet-4.6.0) so perhaps we should use that here.

gewarren: @eiriktsarpalis Please check the latest updates. Not sure if I got it entirely correct.

eiriktsarpalis (approved):

xml/System.Text.Json.Serialization/JsonIncludeAttribute.xml
eiriktsarpalis: Looks good to me, thanks.

PR #9394, opened 2023-10-18
Clarify File.WriteAllText and friends' truncation behaviour

Smaug123:
## Summary

Clarify in `File.WriteAllText` and friends that the files are truncated before being overwritten. The docs as stated do not define the result of overwriting `hello world` with `bye`: is it `byelo world` or `bye`?

I determined their actual behaviour empirically on aarch64-darwin on SDK 7.0.400 with `File.WriteAllText : string * string -> unit` and `File.WriteAllBytes : string * byte[] -> unit`, but have not actually checked the other overloads and methods; I have assumed that they are consistent with each other.


xml/System.IO/File.xml
```diff
@@ -6474,7 +6474,7 @@ It is not possible to change the compression status of a <xref:System.IO.File> o
       <Docs>
         <param name="path">The file to write to.</param>
         <param name="bytes">The bytes to write to the file.</param>
-        <summary>Creates a new file, writes the specified byte array to the file, and then closes the file. If the target file already exists, it is overwritten.</summary>
+        <summary>Creates a new file, writes the specified byte array to the file, and then closes the file. If the target file already exists, it is truncated and overwritten.</summary>
```
Jozkee, line 6477: I don't fully grasp the point of the PR. Do you mind defining "Overwrite" and "Truncate"? I have the idea that truncate is implied in overwrite. i.e. overwrite = create from position zero discarding all previous data in the file. So, saying truncate feels redundant to me.

Smaug123: To overwrite is, well, to write over. Nothing about "overwrite" implies deleting anything you didn't write over! For example, `File.OpenWrite` overwrites without truncating:

```fsharp
let f = "/tmp/filething.txt"
System.IO.File.WriteAllText (f, "hello world")
do
    use s = System.IO.File.OpenWrite f
    use r = new System.IO.StreamWriter (s)
    r.Write "bye"
assert (System.IO.File.ReadAllText f = "byelo world" = shouldEqual "byelo world")
```

Jozkee: cc @bartonjs @adamsitnik

Smaug123: (By the way, I'm adding this because while I was coding I was unsure about this particular fact, so I looked in the documentation to try and find it, and it wasn't there. This isn't out of some abstract desire for completeness.)

bartonjs:
Stylistically, I'm torn between this being the right approach, or explaining it in terms of FileMode.Create (https://learn.microsoft.com/en-us/dotnet/api/system.io.filemode?view=net-7.0) https://github.com/dotnet/dotnet-api-docs/blob/c1bd0b8e469ab2b15ed7d0edc87bbe55a130a70c/xml/System.IO/FileMode.xml#L178.

I also didn't look to see if all relevant places were updated.  So, if this (vs explaining it in terms of FileMode.Create) is the preferred approach and it's complete, LGTM.

xml/System.IO/File.xml
RIP-webmaster: Windows CreateFile uses TRUNCATE_EXISTING flag too. But i'm afraid truncate is also not clear enough, because sometimes it means "trim excess", only in some cases truncate is used as "remove everything". Given it's simple API, intended for people who don't want to think about syscalls, i think it would be better to use something like: 

> If the target file already exists, overwrites its entire contents with the specified byte array.

or

> If the target file already exists, removes its contents before writing.

Smaug123: Is there a plausible semantics `WriteAllBytes` could have where it truncated the file *without* removing everything after byte number `len(inputArray)`?

I guess one possible option would be the insane "removes `len(inputArray)` bytes from the end of the file, then inserts `inputArray` before the beginning of the file so that it stays the same length as it was originally on disk", but I claim literally nobody would even think that this might be a possibility; did you have some other possible behaviour in mind?

Smaug123:
Personally I get the impression that since this is a high-level API, it shouldn't be described in terms of lower-level primitives; but that's an empirical question about the consistent style of the .NET docs which I'm not particularly well-placed to answer. (If I were writing the docs myself from scratch, I'd certainly say "This function is a high-level wrapper around the following lower-level calls", but my general vague impression is that the .NET docs don't do this.)

xml/System.IO/File.xml
RIP-webmaster: I didn't have particular behaviour in mind. My comment was more about that docs would now force reader to think what truncate means in this context, so this PR does not improve clarity enough. Of course the only sensible behaviour is when contents are removed entirely before overwriting, but it's only clear to someone with experience in how file API works.

Smaug123: Strong opinion strongly held:

Personally I'm fine with docs that make you think about what they mean (and if I were being glib I'd say this is a property that many of the .NET docs already have :P ). What's vital to me is that after you've done the thinking, there's only one remotely plausible behaviour the system could have, given its documentation.

Claim that could be empirically falsified:

I claim that a function called `WriteAllBytes`, which is documented to "truncate and overwrite", could *only* do what `File.WriteAllBytes` currently does; otherwise the docs would have left a gaping and obviously-unacceptable degree of clearly-undocumented freedom in "how much is truncated" that would never have made it past review. A reasonable user could expect the documentation not to be explicitly ambiguous, so could rule out interpretations that make the documentation inherently ambiguous.

Strong opinion weakly held:

I claim there is value in introducing users to the standard terminology where possible, so that they can absorb it by osmosis. "Truncate" is a standard term here, and if there is a cheap opportunity to use it, we should.

Having said that, I don't much care, and it's really for the .NET docs team to decide what's appropriate here.

Jozkee: @RIP-webmaster how does the localized text in actual msft documentation sound like? 
https://learn.microsoft.com/ru-ru/dotnet/api/system.io.file.writealltext

If its lacking meaning, we should fix it.

Jozkee, line 6477: Wdyt? This way we can have the description in natural English and in a technical wording that evokes curiosity.

```suggestion
        <summary>Creates a new file, writes the specified byte array to the file, and then closes the file. If the target file already exists, overwrites any existing content (truncates to length zero).</summary>
```

Smaug123: Definitely should be "first truncating to length 0" rather than "truncates to length 0", but otherwise I'm not opposed to that.

RIP-webmaster: > how does the localized text in actual msft documentation sound like? https://learn.microsoft.com/ru-ru/dotnet/api/system.io.file.writealltext
> 
> If its lacking meaning, we should fix it.

Current version is fine. Not the best choice of words, but understandable.

Jozkee (commented):
Would you mind looking for other places that are appropriate to update/ From a quick search of "overwrit" I found these:

`Create`, `CreateText`, `Replace`

gewarren:
> Would you mind looking for other places that are appropriate to update. From a quick search of "overwrit" I found `Create`, `CreateText`, `Replace`.

Ping @Smaug123

Smaug123:
I have now checked every use of the string "verwrit", and adjusted the ones where it seemed appropriate.

Jozkee (approved):

PR #9405, opened 2023-10-25
Fix class/struct naming

deeprobin:
## Summary

While working on the CngProperty class (see https://github.com/dotnet/runtime/pull/93805), 
I discovered a little something in the documentation.

The docs refer to `CngProperty` as a class, even though it is a struct.
This PR fixes that - no special issue for it (I can create a issue if needed).

/cc @gewarren @vcsjones


gewarren (approved):
Thanks @deeprobin!

vcsjones (approved):

PR #9409, opened 2023-10-26
Update Activator.xml

### LINKED ISSUE #9407: `Activator.CreateInstance` does not document `null` returns
**Issue description**

All overloads of [`Activator.CreateInstance`](https://learn.microsoft.com/en-us/dotnet/api/system.activator.createinstance?view=net-7.0) are documented as possibly returning `null`, but the situations where `null` could be returned are not documented.

Specifically, for the overloads taking a `System.Type`, "[a] reference to the newly created object" _implies_ that `null` will never be returned, but digging into the source shows the return types of all inner calls returning a possibly null `object?`. At the very bottom, `RuntimeType.CreateUninitializedObject` calls `_pfnAllocator`, which is a `delegate*` returning `object?`, but no comments to explain why.

*EDIT*: As pointed out by @vcsjones below, `null` is returned when the type is a `Nullable<T>`. Such behavior is documented, but is not obvious. See [my comment below](https://github.com/dotnet/dotnet-api-docs/issues/9407).

colejohnson66:
## Summary

Document (in the "returns" section) when `Activator.CreateInstance` can return `null`. This information is currently documented in the remarks, but, as remarks are placed below exceptions and examples, it's easy to miss.

Fixes #9407

## Question

Should the note of `null` returns be removed from the remarks area?

### DIFF
--- a/xml/System/Activator.xml
+++ b/xml/System/Activator.xml
@@ -157,7 +157,7 @@
         <param name="assemblyName">The name of a file that contains an assembly where the type named <paramref name="typeName" /> is sought.</param>
         <param name="typeName">The name of the type to create an instance of.</param>
         <summary>Creates an instance of the COM object whose name is specified, using the named assembly file and the parameterless constructor.</summary>
-        <returns>A handle that must be unwrapped to access the newly created instance.</returns>
+        <returns>A handle that must be unwrapped to access the newly created object, or <see langword="null" /> for <see cref="T:System.Nullable`1" /> instances.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[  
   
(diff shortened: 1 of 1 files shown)


colejohnson66:
Sorry, I'm not familiar with the format, but it looks like `<xref:System.Nullable%601>` should be <code>&lt;see cref="T:System.Nullable`1"&gt;</code>?

gewarren:
Thanks @colejohnson66. Yes, can you remove it from the remarks. Also, it looks like there are a couple more overloads that should be updated with this return type.

colejohnson66:
I wasn't sure about the overloads returning ObjectHandles as I've never used them. But indeed, they do:

```csharp
> Activator.CreateInstance("System.Private.CoreLib", "System.Nullable`1[System.Int32]") is null
true
```

I'll update those and remove the remark.

gewarren (approved):
Thanks!

PR #9411, opened 2023-10-26
Added Aspose.Drawing as an alternative to System.Drawing

Sergey-Vlasov:
## Summary

Added Aspose.Drawing as an alternative to System.Drawing. (Also reordered ImageSharp link and added license type to match the docs repository.)

### DIFF
--- a/includes/drawing.md
+++ b/includes/drawing.md
@@ -9,4 +9,4 @@
 >
 > - In .NET 6 and later versions, the [System.Drawing.Common](https://www.nuget.org/packages/System.Drawing.Common/) NuGet package is only supported on Windows operating systems. For more information, see [System.Drawing.Common only supported on Windows](/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-only).
 >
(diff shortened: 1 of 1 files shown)


MSDN-WhiteKnight:
Was this discussed with any maintainers? I don't like that API docs host ever-growing catalog of graphics libraries, including commercial ones. The disadvantage is that if this list becomes longer, it becomes harder for reader to choose which library to use, as we are not providing any sort of details for which scenario which library is better. I think this list should either include only a reasonably short list of most commonly used libraries, or link into some external resource that lists more libraries and provides some details that help reader to choose between them.

Sergey-Vlasov:
@MSDN-WhiteKnight

It was approved by Bill Wagner and Genevieve Warren:
https://github.com/dotnet/docs/pull/37651

Of all alternatives, Aspose.Drawing most fully implements System.Drawing functionality and uniquely provides direct migration without rewriting existing code.

gewarren:
Just to clarify, I didn't technically approve https://github.com/dotnet/docs/pull/37651.
@BillWagner Who should we consult on this from the product team?

PR #9416, opened 2023-10-27
Fix and clarify default value for DontFragment property on UdpClient and Socket.

StephenCleary:
## Summary

The default value of `DontFragment` (for both `UdpClient` and `Socket`) is `false`. AFAICT, it's always _been_ `false`.

### DIFF
--- a/xml/System.Net.Sockets/Socket.xml
+++ b/xml/System.Net.Sockets/Socket.xml
@@ -5056,7 +5056,7 @@ This method populates the <xref:System.Net.Sockets.Socket> instance with data ga
       <Docs>
         <summary>Gets or sets a value that specifies whether the <see cref="T:System.Net.Sockets.Socket" /> allows Internet Protocol (IP) datagrams to be fragmented.</summary>
         <value>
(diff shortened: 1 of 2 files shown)


rzikm (changes requested):
While we don't explicitly set `DontFragment` in the source code, the default value can be set by the OS.

```cs
using System.Net.Sockets;

var sock = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);

System.Console.WriteLine($"Sock.DontFragment: {sock.DontFragment}");
```

produces different values on  different OSes, **true** on windows and **false** on linux. So unfortunately neither the original version nor your proposed change is correct.

@dotnet/ncl Should we make the defaults consistent across OSes? I know there are some platform-dependent behaviors on sockets but this seems like something we can reasonably unify.

wfurt:
> @dotnet/ncl Should we make the defaults consistent across OSes? I know there are some platform-dependent behaviors on sockets but this seems like something we can reasonably unify.

I see value in keeping OS defaults. With that we would respect OS setting - Windows registry, /proc, sysctl or whatever. 
People often compare the behavior to other tools - like curl & browsers and consistency may be beneficial. I don't know if there is benefit of barcoding everything in .NET to make it same. I do see benefit of documenting the value/behavior better. So people do not need to duplicate research and testing.

PR #9424, opened 2023-11-01
Clarify Interlocked.{In|De}crement

Smaug123:
## Summary

Fix some ambiguous English. Was "the decremented value" the value that was decremented, or the value after it was decremented? A bit of thought indicates that it must be the latter (otherwise the API is useless), but why make the user do that thinking.


kouvel (approved):
Thanks!

kouvel:
Actually it would probably be good to update all of the other overloads in the same change. @Smaug123 if you want to do that, that would be great, or I can do so later.

Smaug123:
Ah sorry, I didn't think to search for other overloads 🤦 coming up.

kouvel:
Much appreciated, thanks!

PR #9433, opened 2023-11-07
Update SuppressGCTransitionAttribute.xml

AaronRobinsonMSFT:
Add note about exceptions being thrown at different times when this attribute is applied.

See https://github.com/dotnet/runtime/issues/94029

### DIFF
--- a/xml/System.Runtime.InteropServices/SuppressGCTransitionAttribute.xml
+++ b/xml/System.Runtime.InteropServices/SuppressGCTransitionAttribute.xml
@@ -51,7 +51,7 @@ This attribute is ignored if applied to a method without the <xref:System.Runtim
  * Immediate runtime termination.
  * Data corruption.
 
-Applying this attribute impacts the timing of when the export binding of the P/Invoke is performed. <xref:System.EntryPointNotFoundException> or other exceptions might be thrown earlier than when the attribute is not applied.
+Applying this attribute can cause the P/Invoke export to be bound earlier as a side-effect of the calling method being JIT compiled. <xref:System.EntryPointNotFoundException> or other exceptions might be thrown earlier than when the attribute is not applied.
 
           ]]></format>
     </remarks>


xml/System.Runtime.InteropServices/SuppressGCTransitionAttribute.xml
```diff
@@ -51,6 +51,8 @@ This attribute is ignored if applied to a method without the <xref:System.Runtim
  * Immediate runtime termination.
  * Data corruption.
 
+ Applying this attribute impacts when the export binding of the P/Invoke is performed. This can result in <xref:System.EntryPointNotFoundException> or other exceptions being thrown earlier than when the attribute is not applied.
```
gewarren, line 54:
```suggestion
Applying this attribute impacts the timing of when the export binding of the P/Invoke is performed. <xref:System.EntryPointNotFoundException> or other exceptions might be thrown earlier than when the attribute is not applied.
```

gewarren (approved):
Thanks. I left a suggestion for consideration.

xml/System.Runtime.InteropServices/SuppressGCTransitionAttribute.xml
jkotas:
```suggestion
 Applying this attribute can cause the P/Invoke export to be bound earlier as side-effect of the calling method being JIT compiled. This can result in <xref:System.EntryPointNotFoundException> or other exceptions being thrown earlier than when the attribute is not applied. 
```
We can be more specific about the timing differences.

xml/System.Runtime.InteropServices/SuppressGCTransitionAttribute.xml
```diff
@@ -51,6 +51,8 @@ This attribute is ignored if applied to a method without the <xref:System.Runtim
  * Immediate runtime termination.
  * Data corruption.
 
+Applying this attribute impacts the timing of when the export binding of the P/Invoke is performed. <xref:System.EntryPointNotFoundException> or other exceptions might be thrown earlier than when the attribute is not applied.
```
AaronRobinsonMSFT, line 54:
```suggestion
Applying this attribute can cause the P/Invoke export to be bound earlier as a side-effect of the calling method being JIT compiled. <xref:System.EntryPointNotFoundException> or other exceptions might be thrown earlier than when the attribute is not applied.
```

jkotas (approved):

PR #9444, opened 2023-11-10
Fix docs of MemoryExtensions.AsSpan<T>(this T[]? array, int start, int length)

halgab:
## Summary

I've spotted what looks like a bad copy/paste from the corresponding string extension method for method `MemoryExtensions.AsSpan<T>(this T[]? array, int start, int length)`


BillWagner (approved):
Thanks @halgab 

I've reviewed this, and it LGTM. I'll :shipit: now.

PR #9452, opened 2023-11-14
Set feedback system to "standard"

gewarren:
For the product feedback and Q&A links to show in the top feedback box, we have to set the feedback system to "Standard". The only difference is that the standard feedback box will also now appear at the bottom of the articles.

See https://dev.azure.com/ceapex/Engineering/_workitems/edit/923409.


BillWagner (approved):
This LGTM. Let's :shipit:

PR #9475, opened 2023-11-22
Fix build suggestions

gewarren:
Contributes to dotnet/docs#37820

(Hide whitespace changes in diff.)


merriemcgaw (commented):
Looks good for System.Windows.Forms

PR #9480, opened 2023-11-27
Remove a misleading statement about ValidationResult

siriak:
## Summary

`ValidationResult.Success` is `null` if validation is successful as it is stated in remarks. I don't understand how these `true` and `false` are related to `ValidationResult`, so I've removed them.


rzikm:
@gewarren can you please have a look?

gewarren (approved):
LGTM - thanks @siriak

PR #9489, opened 2023-12-02
fix typo

jedatwork:
fixes a typo that prevents the code runner from succeeding without modification


gewarren (approved):
Thanks @jedatwork

PR #9497, opened 2023-12-05
Remove explicit default value mention from Sockets send/receive buffers.

### LINKED ISSUE #9493: Socket.ReceiveBufferSize default value incorrect
https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.receivebuffersize?view=net-7.0#definition

> An [Int32](https://learn.microsoft.com/en-us/dotnet/api/system.int32?view=net-7.0) that contains the size, in bytes, of the receive buffer. The default is 8192.

Which is not `8192`.

image

rzikm:
Closes #9493.


liveans (approved):

antonfirsov (approved):
Thanks!

PR #9505, opened 2023-12-08
Format System types XML files

gewarren:
Formatted using the Red Hat XML VS Code extension with xml.format.maxlinewidth set to 0 (which means, disable it).


xml/System/IntPtr.xml
```diff
@@ -5991,8 +6013,10 @@ This member is an explicit interface member implementation. It can be used only
       </Parameters>
       <Docs>
         <param name="format">The format to use.
- -or-
- A <see langword="null" /> reference (<see langword="Nothing" /> in Visual Basic) to use the default format defined for the type of the <see cref="T:System.IFormattable" /> implementation.</param>
+
```
gewarren, line 6016: This is a good change, but I did it manually. See https://learn.microsoft.com/en-us/dotnet/api/system.intptr.tostring?view=net-8.0#system-intptr-tostring(system-string-system-iformatprovider).

PR #9508, opened 2023-12-13
Update TextWriter.xml

Brianmanden:
As clearly stated in the same article and the official documentation an abstract class can not be instantiated.

"_The TextWriter class is an abstract class. Therefore, you do not instantiate it in your code._"
Source: https://learn.microsoft.com/en-us/dotnet/api/system.io.textwriter?view=net-8.0#examples

"_Abstract classes have the following features:
An abstract class cannot be instantiated._"
Source: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/abstract#example-1

## Summary

Describe your changes here.
Correcting misunderstanding about instantiating an abstract class.

### DIFF
--- a/xml/System.IO/TextWriter.xml
+++ b/xml/System.IO/TextWriter.xml
@@ -101,7 +101,7 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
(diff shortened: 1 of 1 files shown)


adamsitnik (approved):
LGTM, thank you @Brianmanden !

PR #9532, opened 2024-01-04
Remove double dots

marco-olimpio:
A simple clean of double dots in the docs.

image


gewarren (approved):
Thanks @marco-olimpio

PR #8603, opened 2022-10-31
Remove APM samples for TCPClient and TCPListener

antonfirsov:
## Summary

Contributes to #8508

PR #9535, opened 2024-01-06
Extract long remarks

gewarren:
Contributes to dotnet/docs#37859

### DIFF
--- a/xml/System.Collections.Generic/HashSet`1.xml
+++ b/xml/System.Collections.Generic/HashSet`1.xml
@@ -127,7 +127,7 @@
   <Docs>
     <typeparam name="T">The type of elements in the hash set.</typeparam>
     <summary>Represents a set of values.</summary>
-    <remarks>For more information about this API, see <see href="/dotnet/fundamentals/runtime-libraries/system-collections-generic-hashset{t}">Supplemental API remarks for HashSet&lt;t&gt;</see>.</remarks>
+    <remarks>For more information about this API, see <see href="/dotnet/fundamentals/runtime-libraries/system-collections-generic-hashset{t}">Supplemental API remarks for HashSet&lt;T&gt;</see>.</remarks>
     <example>
       <format type="text/markdown"><![CDATA[
 The following example demonstrates how to merge two disparate sets. This example creates two <xref:System.Collections.Generic.HashSet%601> objects and populates them with even and odd numbers, respectively. A third <xref:System.Collections.Generic.HashSet%601> object is created from the set that contains the even numbers. The example then calls the <xref:System.Collections.Generic.HashSet%601.UnionWith%2A> method, which adds the odd number set to the third set.
--- a/xml/System.Collections.Generic/List`1.xml
+++ b/xml/System.Collections.Generic/List`1.xml
@@ -109,7 +109,7 @@
   <Docs>
     <typeparam name="T">The type of elements in the list.</typeparam>
     <summary>Represents a strongly typed list of objects that can be accessed by index. Provides methods to search, sort, and manipulate lists.</summary>
-    <remarks>For more information about this API, see <see href="/dotnet/fundamentals/runtime-libraries/system-collections-generic-list{t}">Supplemental API remarks for List&lt;t&gt;</see>.</remarks>
+    <remarks>For more information about this API, see <see href="/dotnet/fundamentals/runtime-libraries/system-collections-generic-list{t}">Supplemental API remarks for List&lt;T&gt;</see>.</remarks>
     <threadsafe>Public static (<see langword="Shared" /> in Visual Basic) members of this type are thread safe. Any instance members are not guaranteed to be thread safe.
 
  It is safe to perform multiple read operations on a <see cref="T:System.Collections.Generic.List`1" />, but issues can occur if the collection is modified while it's being read. To ensure thread safety, lock the collection during a read or write operation. To enable a collection to be accessed by multiple threads for reading and writing, you must implement your own synchronization. For collections with built-in synchronization, see the classes in the <see cref="N:System.Collections.Concurrent" /> namespace. For an inherently thread-safe alternative, see the <see cref="T:System.Collections.Immutable.ImmutableList`1" /> class.</threadsafe>
--- a/xml/System.Collections.ObjectModel/ObservableCollection`1.xml
+++ b/xml/System.Collections.ObjectModel/ObservableCollection`1.xml
@@ -89,7 +89,7 @@
   <Docs>
     <typeparam name="T">The type of elements in the collection.</typeparam>
     <summary>Represents a dynamic data collection that provides notifications when items get added or removed, or when the whole list is refreshed.</summary>
-    <remarks>For more information about this API, see <see href="/dotnet/fundamentals/runtime-libraries/system-collections-objectmodel-observablecollection{t}">Supplemental API remarks for ObservableCollection&lt;t&gt;</see>.</remarks>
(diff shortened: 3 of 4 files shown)


AaronRobinsonMSFT (commented):
System.Runtime.InteropServices LGTM

ManickaP:
I know this will not change anything and I don't expect it to, but I wanted to express my opinion on this. I think this just inserts another obstacle between the user and the information they need. IMO API docs are the gateway to .NET docs. If the information is not in the API docs, it doesn't exist for many users. And lately, we've been slowly removing valuable information from API docs and just leaving the useless "Gets or sets the name of the object" behind. And it makes me sad, because these docs used to be one of the best I ever used.

xml/System.Collections.Generic/HashSet`1.xml
```diff
@@ -127,69 +127,15 @@
...
-|<xref:System.Collections.Generic.HashSet%601.IsSupersetOf%2A>|Not provided.|  
-|<xref:System.Collections.Generic.HashSet%601.IsProperSupersetOf%2A>|Not provided.|  
-|<xref:System.Collections.Generic.HashSet%601.SetEquals%2A>|Not provided.|  
-  
-   
-  
-## Examples  
- The following example demonstrates how to merge two disparate sets. This example creates two <xref:System.Collections.Generic.HashSet%601> objects, and populates them with even and odd numbers, respectively. A third <xref:System.Collections.Generic.HashSet%601> object is created from the set that contains the even numbers. The example then calls the <xref:System.Collections.Generic.HashSet%601.UnionWith%2A> method, which adds the odd number set to the third set.  
-  
+    <remarks>For more information about this API, see <see href="/dotnet/fundamentals/runtime-libraries/system-collections-generic-hashset{t}">System.Collections.Generic.HashSet&lt;t&gt; class</see>.</remarks>
```
antonfirsov, line 130: This reads weird, since it's the current page that is supposed to be the doc of the *"`System.Collections.Generic.Hash<T>` class"*. I would rather repharese it as following:

```suggestion
    <remarks>For more information, see the supplemental API remarks on <see href="/dotnet/fundamentals/runtime-libraries/system-collections-generic-hashset{t}">HashSet&lt;T&gt; class</see>.</remarks>
```

Type arguments should be definitely capitalized, and seeing the whole namespace feels unnecessarily verbose here.

The TOC in the linked doc contains an extra `\` behind the class name (`HashSet<T>\ class`):
https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-collections-generic-hashset%7Bt%7D

antonfirsov:
Regarding @ManickaP's comment I think the goal here is to find a balance between usability on the customer side and maintainability on our side. I assume there are good reasons we want to keep `///` docs fully synchronized, but embedding long API remarks into `dotnet/runtime` code sounds like a nightmare.

IMO the ideal solution would be to maintain long remarks separately (but still in this repo), but make the tooling copy the remarks into the resulting API doc html pages.

xml/System.Collections.Generic/HashSet`1.xml
gewarren: Thanks - good catch. Fixed the backslash here: https://github.com/dotnet/docs/pull/39107

xml/System.Collections.Generic/HashSet`1.xml
```diff
@@ -127,69 +127,15 @@
...
+    <example>
+      <format type="text/markdown"><![CDATA[
+The following example demonstrates how to merge two disparate sets. This example creates two <xref:System.Collections.Generic.HashSet%601> objects and populates them with even and odd numbers, respectively. A third <xref:System.Collections.Generic.HashSet%601> object is created from the set that contains the even numbers. The example then calls the <xref:System.Collections.Generic.HashSet%601.UnionWith%2A> method, which adds the odd number set to the third set.
+
  :::code language="csharp" source="~/snippets/csharp/System.Collections.Generic/HashSetT/Overview/Program.cs" interactive="try-dotnet-method" id="Snippet01":::
- :::code language="vb" source="~/snippets/visualbasic/VS_Snippets_CLR_System/system.Collections.Generic.HashSet_UnionWith/vb/Program.vb" id="Snippet01":::  
-  
- ]]></format>
-    </remarks>
+ :::code language="vb" source="~/snippets/visualbasic/VS_Snippets_CLR_System/system.Collections.Generic.HashSet_UnionWith/vb/Program.vb" id="Snippet01":::
```
carlossanlop, line 136: I suppose these interactive examples don't have an equivalent xml element, so they'll have to stay as CDATA.

We'll have to discuss what to do with them when we backport them, as they are interactive in this repo.

gewarren: Can't you just backport them as CDATA?

xml/System.Diagnostics/PerformanceCounterType.xml
```diff
@@ -32,151 +32,7 @@
...
-
-### NumberOfItems32
-
- :::code language="cpp" source="~/snippets/cpp/VS_Snippets_CLR/PerformanceCounterType.NumberOfItems32/CPP/numberofitems32.cpp" id="Snippet1":::
- :::code language="csharp" source="~/snippets/csharp/System.Diagnostics/PerformanceCounterType/Overview/numberofitems32.cs" id="Snippet1":::
- :::code language="vb" source="~/snippets/visualbasic/VS_Snippets_CLR/PerformanceCounterType.NumberOfItems32/VB/numberofitems32.vb" id="Snippet1":::
-
-### NumberOfItems64
-
- :::code language="cpp" source="~/snippets/cpp/VS_Snippets_CLR/PerformanceCounterType.NumberOfItems64/CPP/numberofitems64.cpp" id="Snippet1":::
```
carlossanlop, line 151: Oh, I see you moved these to the docs repo too. Any chance you can do the same with the other one I commented on? https://github.com/dotnet/dotnet-api-docs/pull/9535/files#r1449492293

xml/System.IO/FileStream.xml
```diff
@@ -66,54 +66,22 @@
...
- :::code language="csharp" source="~/snippets/csharp/System.IO/FileStream/Overview/example3.cs" id="Snippet3":::
- :::code language="vb" source="~/snippets/visualbasic/VS_Snippets_CLR/Asynchronous_File_IO_async/vb/example3.vb" id="Snippet3":::  
-  
- ]]></format>
-    </remarks>
+    <example>
+      <format type="text/markdown"><![CDATA[
+        The following example demonstrates some of the <xref:System.IO.FileStream> constructors.
+
+        :::code language="cpp" source="~/snippets/cpp/VS_Snippets_CLR/FStream Class/CPP/fstream class.cpp" id="Snippet1":::
```
carlossanlop, line 73: Same question/request for these code examples and any others in the rest of the PR. Since their long remarks got extracted, maybe the examples could be moved too.

xml/System.Diagnostics/PerformanceCounterType.xml
gewarren: For remarks where there was a single example at the end of the remarks, I purposefully left them (actually moved them to an `<example>` tag) so there'd be an example in the API docs (without having to follow a link). IMO most standalone examples shouldn't be moved to the dotnet/docs repo.

carlossanlop (approved):
Left some comments for you to consider. Feel free to address them in a separate PR if you prefer. Looks good enough for merging as is.

ManickaP:
> Regarding this comment, can you elaborate on how information has been removed (other than this PR)? Or is it just that information hasn't been added for the newer APIs?

It's both. For example, I rarely see us adding example usages with new APIs. And with removing the info, for example [here](https://github.com/dotnet/dotnet-api-docs/pull/9350/files) and [here](https://github.com/dotnet/dotnet-api-docs/pull/8603/files) where we removed code samples because they were out of date, instead of updating them.

PR #9543, opened 2024-01-13
Delete orphaned C++ snippets

gewarren:
Contributes to #9558


BillWagner (approved):
This LGTM.

I'll :shipit: even though the snippets 5000 failed. All the folders that are failing were removed.

PR #9544, opened 2024-01-15
Improved documentation for MemoryFailPoint when the constructor is called with value 0

vdekercd:
## Summary

The documentation does not specify that an ArgumentOutOfRangeException is generated if the value is equal to 0.

Edit to add page: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.memoryfailpoint.-ctor


vdekercd:
[Link to the relevant code](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Runtime/MemoryFailPoint.cs,2eb7af463bd64007)

gewarren (approved):
Thank you @vdekercd

PR #9545, opened 2024-01-16
Update rfc28981.cs example

wfurt:
updates based on https://developercommunity.visualstudio.com/t/Rfc2898DeriveBytes-Class-Microsoft-Learn/10547811 feedback

removes use of obsolete methods. 

Fixes https://github.com/dotnet/dotnet-api-docs/issues/9657

### DIFF
--- a/snippets/csharp/System.Security.Cryptography/Rfc2898DeriveBytes/Overview/rfc28981.cs
+++ b/snippets/csharp/System.Security.Cryptography/Rfc2898DeriveBytes/Overview/rfc28981.cs
@@ -25,13 +25,7 @@ public class rfc2898test
             //<SNIPPET6>
             string pwd1 = passwordargs[0];
             // Create a byte array to hold the random value.
-            byte[] salt1 = new byte[8];
-            using (RNGCryptoServiceProvider rngCsp = new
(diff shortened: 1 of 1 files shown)


bartonjs (changes requested):
SNIPPET3's comment and SNIPPET4's usage were consistent: the default value is 1000, so specifying 1000 explicitly produces the same answer.
https://source.dot.net/#System.Security.Cryptography/System/Security/Cryptography/Rfc2898DeriveBytes.cs,45

Now that SNIPPET4 is no longer using a default iteration count, the comment in SNIPPET3 doesn't make sense.

I recommend bumping the iteration count 600_000, and changing the comment to express that the iteration count should be high enough to resist brute forcing, but low enough that you're willing to wait that long. The 600k number comes from OWASP (when using SHA-2-256), which comes out to about 0.33 seconds on my computer using the static span-writing functions .NET 8, or about 0.38 when using the instance array-returning functions.  (.NET Framework is slower; 600k there is 1.6 seconds; so it gets 0.3 seconds around 130k)

gewarren:
@wfurt Do you plan to keep working on this PR?

PR #9571, opened 2024-01-24
Some return values for MathF.Pow() are wrong

Psychpsyo:
## Summary

``MathF.Pow(float.NaN,0f)`` returns ``1``, not ``NaN``.
``MathF.Pow(1f,float.NaN)`` returns ``1``, not ``NaN``.

```cs
using System;

public class HelloWorld
{
    public static void Main(string[] args)
    {
        Console.WriteLine(System.MathF.Pow(float.NaN,0f));
        Console.WriteLine(System.MathF.Pow(1f,float.NaN));
    }
}
```
This outputs ``1`` twice.

### DIFF
--- a/xml/System/MathF.xml
+++ b/xml/System/MathF.xml
@@ -1659,7 +1659,7 @@ The value of this field is 3.14159274.
 |----------------|------------------|  
 |`x` is any value; `y` = 0.|1|  
 |`x` = 1; `y` is any value.|1|  
-|`x` or `y` = `NaN`. (except for the above cases)|`NaN`|  
+|`x` or `y` = `NaN` (except if `y` = 0 or `x` = 1).|`NaN`|  
 |`x` = `NegativeInfinity`; `y` < 0.|0|  
 |`x` = `NegativeInfinity`; `y` is a positive odd integer.|`NegativeInfinity`|  
 |`x` = `NegativeInfinity`; `y` is positive but not an odd integer.|`PositiveInfinity`|  


tannergooding:
The full set of rules for `pow` (as required by `IEEE 754 2019`) is:
* pow (x, ±0) is 1 if x is not a signaling NaN
* pow (±0, y) is ±∞ and signals the divideByZero exception for y an odd integer < 0
* pow (±0, −∞) is +∞ with no exception
* pow (±0, +∞) is +0 with no exception
* pow (±0, y) is ±0 for finite y > 0 an odd integer
* pow (−1, ±∞) is 1 with no exception
* pow (+1, y) is 1 for any y (even a quiet NaN)
* pow (x, +∞) is +0 for −1 < x < 1
* pow (x, +∞) is +∞ for x < −1 or for 1 < x (including ±∞)
* pow (x, −∞) is +∞ for −1 < x < 1
* pow (x, −∞) is +0 for x < −1 or for 1 < x (including ±∞)
* pow (+∞, y) is +0 for a number y < 0
* pow (+∞, y) is +∞ for a number y > 0
* pow (−∞, y) is −0 for finite y < 0 an odd integer
* pow (−∞, y) is −∞ for finite y > 0 an odd integer
* pow (−∞, y) is +0 for finite y < 0 and not an odd integer
* pow (−∞, y) is +∞ for finite y > 0 and not an odd integer
* pow (±0, y) is +∞ and signals the divideByZero exception for finite y < 0 and not an odd integer
* pow(±0, y) is +0 for finite y > 0 and not an odd integer
* pow(x, y) signals the invalid operation exception for finite x < 0 and finite non-integer y.

It would be good to see if we can make all of this match up with the docs

Psychpsyo:
> It would be good to see if we can make all of this match up with the docs

Except for the parts about signaling NaNs, I'd say, since I don't think we even have access to signaling NaNs.

xml/System/MathF.xml
```diff
@@ -1657,8 +1657,9 @@ The value of this field is 3.14159274.
   
 |Parameters|Return value|  
 |----------------|------------------|  
-|`x` or `y` = `NaN`.|`NaN`|  
-|`x` = Any value except `NaN`; `y` = 0.|1|  
+|`x` is any value; `y` = 0.|1|  
+|`x` = 1; `y` is any value.|1|  
+|`x` or `y` = `NaN`. (except for the above cases)|`NaN`|
```
gewarren, line 1662:
```suggestion
|`x` or `y` = `NaN` (except if `y` = 0 or `x` = 1).|`NaN`|  
```

gewarren: We might want to do something similar to what @arturek did [here for Math](https://github.com/dotnet/dotnet-api-docs/commit/7db6ac602d5c6bc09b1428a8868385662571a7e5).

Psychpsyo: > We might want to do something similar to what @arturek did [here for Math](https://github.com/dotnet/dotnet-api-docs/commit/7db6ac602d5c6bc09b1428a8868385662571a7e5).

Committed the suggestion for now but I agree on that.

gewarren (approved):

PR #9573, opened 2024-01-24
Add missing SearchRoot assignment in example

SiavasFiroozbakht:
Add missing SearchRoot assignment in example.


gewarren:
Tagging @dotnet/area-system-directoryservices

buyaa-n:
> Add missing SearchRoot assignment in example.

I don't see it was `missing`, [looks](https://learn.microsoft.com/en-us/dotnet/api/system.directoryservices.directorysearcher.searchroot?view=dotnet-plat-ext-8.0) it can be set but not required

PR #9574, opened 2024-01-25
Fix invalid pointer types in ldloca/ldarga descriptions

MichalPetryka:
## Summary

Those are documented to push managed pointers on the stack in the ECMA 335 and [the runtime implements them as such](https://github.com/dotnet/runtime/blob/dd158ba513a5a09f8889e5cf8873655a4f3c6fdd/src/coreclr/jit/importer.cpp#L6639).

### DIFF
--- a/xml/System.Reflection.Emit/OpCodes.xml
+++ b/xml/System.Reflection.Emit/OpCodes.xml
@@ -6583,7 +6583,7 @@
   
 1.  The address `addr` of the argument indexed by `index` is pushed onto the stack.  
   
- The `ldarga` instruction fetches the address (of type `&amp;`) of the argument indexed by `index`, where arguments are indexed from 0 onwards. The address `addr` is always aligned to a natural boundary on the target machine. The argument is stored in unmanaged memory, so the return value can be converted to an unmanaged pointer without pinning.
+ The `ldarga` instruction fetches the address (of type `&`) of the argument indexed by `index`, where arguments are indexed from 0 onwards. The address `addr` is always aligned to a natural boundary on the target machine. The argument is stored in unmanaged memory, so the return value can be converted to an unmanaged pointer without pinning.
   
  For procedures that take a variable-length argument list, the `ldarga` instruction can be used only for the initial fixed arguments, not those in the variable part of the signature.  
   
@@ -10065,7 +10065,7 @@ The following <xref:System.Reflection.Emit.ILGenerator.Emit%2A> method overloads
   
 1.  The address stored in the local variable at the specified index is pushed onto the stack.  
   
- The `ldloca` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a managed pointer (type `&amp;`). The local variable is stored in unmanaged memory, so the return value can be converted to an unmanaged pointer without pinning.
+ The `ldloca` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a managed pointer (type `&`). The local variable is stored in unmanaged memory, so the return value can be converted to an unmanaged pointer without pinning.
   
  The following <xref:System.Reflection.Emit.ILGenerator.Emit%2A> method overload can use the `ldloca` opcode:  
   


xml/System.Reflection.Emit/OpCodes.xml
```diff
@@ -10065,7 +10065,7 @@ The following <xref:System.Reflection.Emit.ILGenerator.Emit%2A> method overloads
   
 1.  The address stored in the local variable at the specified index is pushed onto the stack.  
   
- The `ldloca` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a transient pointer (type `*`).  
+ The `ldloca` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a managed pointer (type `&`) which can however be safely converted to a transient pointer (type `*`) with <xref:System.Reflection.Emit.OpCodes.Conv_U>.
```
reflectronic, line 10068:
```suggestion
 The `ldloca` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a managed pointer (type `&amp;`). The local variable is stored in unmanaged memory, so the return value can be converted to an unmanaged pointer without pinning.
```

xml/System.Reflection.Emit/OpCodes.xml
```diff
@@ -6583,7 +6583,7 @@
   
 1.  The address `addr` of the argument indexed by `index` is pushed onto the stack.  
   
- The `ldarga` instruction fetches the address (of type `*`) of the argument indexed by `index`, where arguments are indexed from 0 onwards. The address `addr` is always aligned to a natural boundary on the target machine.  
+ The `ldarga` instruction fetches the address (of type `&`, which can however be safely converted to type `*` with <xref:System.Reflection.Emit.OpCodes.Conv_U>) of the argument indexed by `index`, where arguments are indexed from 0 onwards. The address `addr` is always aligned to a natural boundary on the target machine.
```
reflectronic, line 6586:
```suggestion
 The `ldarga` instruction fetches the address (of type `&amp;`) of the argument indexed by `index`, where arguments are indexed from 0 onwards. The address `addr` is always aligned to a natural boundary on the target machine. The argument is stored in unmanaged memory, so the return value can be converted to an unmanaged pointer without pinning.
```

reflectronic (commented):
`conv.u` does not convert managed pointers to transient pointers (type `*`), so this patch is incorrect. The transient pointers concept was deleted from the standard recently in https://github.com/dotnet/runtime/pull/86622, so any reference to transient pointers should be deleted. There are many more references elsewhere in this file.

xml/System.Reflection.Emit/OpCodes.xml
MichalPetryka: I think `conv.u` should still be mentioned there.

gewarren:
Tagging @dotnet/area-system-reflection-emit

steveharter:
PTAL @AaronRobinsonMSFT

xml/System.Reflection.Emit/OpCodes.xml
AaronRobinsonMSFT: I disagree. I would prefer to keep this as close to ECMA-335 modified with the relavantly applied addendums.

gewarren:
@AaronRobinsonMSFT Is this how it should render?

image

AaronRobinsonMSFT:
> @AaronRobinsonMSFT Is this how it should render?
> 
> image

Definitely not. Let me try again. :)

xml/System.Reflection.Emit/OpCodes.xml
```diff
@@ -6583,7 +6583,7 @@
   
 1.  The address `addr` of the argument indexed by `index` is pushed onto the stack.  
   
- The `ldarga` instruction fetches the address (of type `*`) of the argument indexed by `index`, where arguments are indexed from 0 onwards. The address `addr` is always aligned to a natural boundary on the target machine.  
+ The `ldarga` instruction fetches the address (of type `&amp;`) of the argument indexed by `index`, where arguments are indexed from 0 onwards. The address `addr` is always aligned to a natural boundary on the target machine. The argument is stored in unmanaged memory, so the return value can be converted to an unmanaged pointer without pinning.
```
AaronRobinsonMSFT, line 6586:
```suggestion
 The `ldarga` instruction fetches the address (of type `&`) of the argument indexed by `index`, where arguments are indexed from 0 onwards. The address `addr` is always aligned to a natural boundary on the target machine. The argument is stored in unmanaged memory, so the return value can be converted to an unmanaged pointer without pinning.
```

xml/System.Reflection.Emit/OpCodes.xml
```diff
@@ -10065,7 +10065,7 @@ The following <xref:System.Reflection.Emit.ILGenerator.Emit%2A> method overloads
   
 1.  The address stored in the local variable at the specified index is pushed onto the stack.  
   
- The `ldloca` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a transient pointer (type `*`).  
+ The `ldloca` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a managed pointer (type `&amp;`). The local variable is stored in unmanaged memory, so the return value can be converted to an unmanaged pointer without pinning.
```
AaronRobinsonMSFT, line 10068:
```suggestion
 The `ldloca` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a managed pointer (type `&`). The local variable is stored in unmanaged memory, so the return value can be converted to an unmanaged pointer without pinning.
```

AaronRobinsonMSFT:
@gewarren Should be fixed now.

gewarren (approved):
LGTM

PR #9583, opened 2024-01-31
Update SslClientAuthenticationOptions.TargetHost

rzikm:
Fixes dotnet/runtime#97444.

### DIFF
--- a/xml/System.Net.Security/SslClientAuthenticationOptions.xml
+++ b/xml/System.Net.Security/SslClientAuthenticationOptions.xml
@@ -473,9 +473,9 @@
         <ReturnType>System.String</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets or sets the name of the server the client is trying to connect to. That name is used for server certificate validation. It can be a DNS name or an IP address. If set to or empty string, the client will not send hostname in the Server Name Indication extension.</summary>
-        <value>To be added.</value>
-        <remarks>Since .NET 7, value of <see langword="null" /> is allowed and has the same meaning as an empty string. If the value is <see langword="null" /> or an empty string, users should also provide <see cref="T:System.Net.Security.RemoteCertificateValidationCallback" /> and validate the subject of the server's certificate.</remarks>
+        <summary>Gets or sets the name of the server the client is trying to connect to. The name is used for server certificate validation and can be a DNS name or an IP address.</summary>
+        <value>The name of the server the client is trying to connect to. If set to <see langword="null" /> (.NET 7 and later versions only) or an empty string, the client doesn't send the host name in the Server Name Indication extension.</value>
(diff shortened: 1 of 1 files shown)


ManickaP (approved):
:+1:

xml/System.Net.Security/SslClientAuthenticationOptions.xml
```diff
@@ -473,9 +473,9 @@
         <ReturnType>System.String</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets or sets the name of the server the client is trying to connect to. That name is used for server certificate validation. It can be a DNS name or an IP address.</summary>
+        <summary>Gets or sets the name of the server the client is trying to connect to. That name is used for server certificate validation. It can be a DNS name or an IP address. If set to or empty string, the client will not send hostname in the Server Name Indication extension.</summary>
```
gewarren, line 476:
```suggestion
        <summary>Gets or sets the name of the server the client is trying to connect to. The name is used for server certificate validation and can be a DNS name or an IP address.</summary>
```

xml/System.Net.Security/SslClientAuthenticationOptions.xml
```diff
@@ -473,9 +473,9 @@
...
         <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <remarks>Since .NET 7, value of <see langword="null" /> is allowed and has the same meaning as an empty string. If the value is <see langword="null" /> or an empty string, users should also provide <see cref="T:System.Net.Security.RemoteCertificateValidationCallback" /> and validate the subject of the server's certificate.</remarks>
```
gewarren, line 478:
```suggestion
        <remarks>If the value is <see langword="null" /> or an empty string, you should also provide a <see cref="T:System.Net.Security.RemoteCertificateValidationCallback" /> and validate the subject of the server's certificate.</remarks>
```

xml/System.Net.Security/SslClientAuthenticationOptions.xml
```diff
@@ -473,9 +473,9 @@
         <ReturnType>System.String</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets or sets the name of the server the client is trying to connect to. That name is used for server certificate validation. It can be a DNS name or an IP address.</summary>
+        <summary>Gets or sets the name of the server the client is trying to connect to. That name is used for server certificate validation. It can be a DNS name or an IP address. If set to or empty string, the client will not send hostname in the Server Name Indication extension.</summary>
         <value>To be added.</value>
```
gewarren, line 477:
```suggestion
        <value>The name of the server the client is trying to connect to. If set to <see langword="null" /> (.NET 7 and later versions only) or an empty string, the client doesn't send the host name in the Server Name Indication extension.</value>
```

gewarren (approved):
I left some suggestions to consider. I think the extra commentary fits better in the value element than the summary, which is supposed to be brief.

PR #9595, opened 2024-02-06
Add nullness warnings in JsonParse

Smaug123:
## Summary

Today I discovered the surprising fact that `JsonNode.Parse(x).ToJsonString()` may throw on valid inputs. It seems reasonable to document this.

### DIFF
--- a/xml/System.Text.Json.Nodes/JsonNode.xml
+++ b/xml/System.Text.Json.Nodes/JsonNode.xml
@@ -2813,7 +2813,7 @@ The current node is returned if it is a root.
         <summary>Converts the current instance to a string in JSON format.</summary>
         <returns>The JSON representation of the current instance.</returns>
         <remarks>
-          <see cref="Overload:System.Text.Json.Nodes.JsonNode.Parse /> does not round-trip with this method, because of its behavior on the JSON null value.</remarks>
+          <see cref="Overload:System.Text.Json.Nodes.JsonNode.Parse" /> does not round-trip with this method, because of its behavior on the JSON null value.</remarks>
       </Docs>
     </Member>
     <Member MemberName="ToString">


Smaug123:
What's the correct way to link to a method when I don't care which overload I'm linking to? The bot appears to be complaining that `M:System.Text.Json.Nodes.JsonNode.Parse` doesn't exist. `M:System.Text.Json.Nodes.JsonNode.Parse(System.IO.Stream,System.Nullable{System.Text.Json.Nodes.JsonNodeOptions},System.Text.Json.JsonDocumentOptions)` does exist (I can see its doc ID in the docs), but I specifically don't want to link to a particular overload.

MSDN-WhiteKnight:
>What's the correct way to link to a method when I don't care which overload I'm linking to?

It should be `M:System.Text.Json.Nodes.JsonNode.Parse%2A` (`%2A` stands for `*`)

MSDN-WhiteKnight:
Looks like this feature does not work with cref. Try converting remarks to markdown (like there: <https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Reflection.Metadata/MetadataReader.xml#L40>) and use xref syntax:

    <xref:System.Text.Json.Nodes.JsonNode.Parse%2A>

gewarren:
> Looks like this feature does not work with cref. Try converting remarks to markdown (like there: https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Reflection.Metadata/MetadataReader.xml#L40) and use xref syntax:
> 
> ```
> <xref:System.Text.Json.Nodes.JsonNode.Parse%2A>
> ```

You can link to the overloads page using cref (see https://learn.microsoft.com/contribute/content/dotnet/api-documentation#cross-references):

`<see cref="Overload:System.Text.Json.Nodes.JsonNode.Parse" />`

Smaug123:
Is this something you would prefer me to do instead of what's currently on the branch? The round-trip time is very long and I can't even see the result.

gewarren:
> Is this something you would prefer me to do instead of what's currently on the branch? The round-trip time is very long and I can't even see the result.

I updated it. I would also like @eiriktsarpalis to review this PR before we merge it.

eiriktsarpalis (approved):

PR #9599, opened 2024-02-09
Add policy for managing the untriaged label

jeffhandley:
The automation for adding/removing the https://github.com/dotnet/dotnet-api-docs/labels/untriaged label was dropped in https://github.com/dotnet/dotnet-api-docs/pull/9568. This adds a new, standalone policy that brings that automation back.

The only difference in behavior is that when a milestone is _removed_ from an issue, the `untriaged` label will not be automatically reapplied to the issue.

### DIFF
--- a/.github/policies/untriaged.yml
+++ b/.github/policies/untriaged.yml
@@ -21,7 +21,6 @@ configuration:
       - not:
           hasLabel:
             label: untriaged
-          label: untriaged
       then:
       - addLabel:
           label: untriaged


.github/policies/untriaged.yml
```diff
@@ -0,0 +1,37 @@
...
+      then:
+      - addLabel:
+          label: untriaged
+      description: Add untriaged label to new/reopened issues without a milestone
+    - if:
+      - payloadType: Issues
+      - or:
+        - isAction:
+            action: Closed
+        - isPartOfAnyMilestone
```
jeffhandley, line 31: @JohannesLampel I suspect the check failure is caused by `IsPartOfAnyMilestone`, but I thought that condition existed (desite being absent in the docs).

.github/policies/untriaged.yml
```diff
@@ -0,0 +1,37 @@
...
+resource: repository
+disabled: false
+where:
+configuration:
+  resourceManagementConfiguration:
+    eventResponderTasks:
+    - if:
+      - payloadType: Issues
+      - isOpen
+      - isNotInAnyMilestone
```
JohannesLampel, line 14:
```suggestion
      - not:
          isPartOfAnyMilestone
```

.github/policies/untriaged.yml
```diff
@@ -0,0 +1,37 @@
...
+      - or:
+        - isAction:
+            action: Opened
+        - isAction:
+            action: Reopened
+      - isNotLabeledWith:
```
JohannesLampel, line 20:
```suggestion
      - not:
          hasLabel:
            label: untriaged
```

.github/policies/untriaged.yml
JohannesLampel: The `isNotInAnyMilestone` condition does not exist. Same goes for the `isNotLabeledWith` condition. However, you can use the `not` modifier to achieve what you want to achieve. I've added comments with suggested changes.

gewarren (approved):

PR #9626, opened 2024-02-19
Use pattern matching in code sample

drewnoakes:
The `null != ...` check looked strange, as it's usually written the other way around in C#. Rather than have a null check, just change this to a type check.


MihaZupan (approved):
Thanks

PR #9632, opened 2024-02-22
Correct None enum description

### LINKED ISSUE #5076: CngExportPolicies.None - Docs wrong
Hi,
I think the docs for **CngExportPolicies.None** are **wrong** (see: https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.cngexportpolicies?view=dotnet-plat-ext-3.1). It says
> No export policies are established. Key export is allowed without restriction.

Whereas the native docs (https://docs.microsoft.com/en-us/windows/win32/seccng/key-storage-property-identifiers)
specify
> NCRYPT_ALLOW_EXPORT_FLAG 0x00000001 - The private key can be exported.

however the "0x0" flag is not described which would lead me to the conclusion that it would indicate non-exportable.

Another indicator is your own example code here: 
https://github.com/microsoft/Intune-Resource-Access/blob/develop/src/PFXImportPowershell/EncryptionUtilities/Source/ManagedRSAEncryption.cs
Where in case "**nonExportable**" is set to "false" the flag **CngExportPolicies.None** is used

### LINKED ISSUE #9159: Which of the `CngExportPolicies` value to use when one wants the key to be non-exportable?
https://github.com/dotnet/dotnet-api-docs/blob/3854df0ff95bc26f339b5184b7cdb67bfc3276e1/xml/System.Security.Cryptography/CngExportPolicies.xml#L57

In the old CryptoAPI one could do this by NOT setting the `CRYPT_EXPORTABLE` flag when calling `CryptGenKey`. What would be the equivalent approach in CNG?

joynoele:
Correct erroneous documentation description of the Enum `CngExportPolicies.None` option.
Correction text is kept as simple as possible. Conclusion reached via my own testing, the two open issues with the same question, and reply from reply from member @bartonjs:
> CngExportPolicy.None is "non-exportable". If you're creating a key with CngKey.Create and don't specify an export policy explicitly, the key will be non-exportable.

Fixes #5076 
Fixes #9159

### DIFF
--- a/xml/System.Security.Cryptography/CngExportPolicies.xml
+++ b/xml/System.Security.Cryptography/CngExportPolicies.xml
@@ -259,7 +259,7 @@
       </ReturnValue>
       <MemberValue>0</MemberValue>
       <Docs>
-        <summary>No export policies are established. Key export is allowed without restriction.</summary>
+        <summary>The key is not exportable.</summary>
       </Docs>
     </Member>
   </Members>


joynoele:
bumping this PR.

bartonjs (approved):

PR #9634, opened 2024-02-24
Correct grammar mistake

pendevx:
## Summary

Change 'an two-byte' to 'a two-byte' in `/xml/System.IO/BinaryWriter.xml:1618`


BillWagner (approved):
Thanks @pendevx 

This looks great, and I'll :shipit: as soon as the build completes.

PR #9640, opened 2024-02-25
Corrected `Succesfully` in IBinaryInteger`1.xml

zacharylayne:
Changed 4 instances of `succesfully` to `successfully`.

## Summary

There were four instances of `succesfully` that were present. Corrected the spelling.

Since this is from XML doc file output, does this get backported to [IBinaryInteger.cs](https://github.com/dotnet/runtime/edit/main/src/libraries/System.Private.CoreLib/src/System/Numerics/IBinaryInteger.cs) or should I fork and make the change in that file instead?


zacharylayne:
p.s. The source file for this documentation has 2 instances fixed already, so I'm not sure how these fixes get published to Microsoft's site.

BillWagner (approved):
Thanks @zacharylayne 

PR #9650, opened 2024-02-27
Modernize C# FolderBrowserDialog example

Varriount:
Update the C# example for the FolderBrowserDialog class to use MenuStrip, MenuStripItem, etc. classes instead of the deprecated MainMenu and MenuItem classes. Also make a few changes with regards to trailing spaces and formatting.

I left the names of all the variables, etc. the same so that it would differ too much from the other examples (and also because adding `ToolStrip` to the variable names wouldn't really justify the length increase).

### DIFF
--- a/snippets/csharp/System.Windows.Forms/FolderBrowserDialog/Overview/Project.csproj
+++ b/snippets/csharp/System.Windows.Forms/FolderBrowserDialog/Overview/Project.csproj
@@ -1,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>WinExe</OutputType>
+    <TargetFramework>net8.0-windows</TargetFramework>
+    <UseWindowsForms>true</UseWindowsForms>
+  </PropertyGroup>
+
+</Project>


Varriount:
@dotnet-policy-service agree

snippets/csharp/System.Windows.Forms/FolderBrowserDialog/Overview/form1.cs
```diff
@@ -1,4 +1,3 @@
-﻿//<Snippet1>
```
BillWagner, line 1: This line needs to be included. It's used by the OPS build system to render the sample.

BillWagner (changes requested):
Hi @Varriount 

Can you address the suggestion I made? Once that's done, the build should complete successfully, and I'll merge this PR.

snippets/csharp/System.Windows.Forms/FolderBrowserDialog/Overview/form1.cs
Varriount: Woops! Sorry, didn't mean to delete that. I'll fix it right away.

gewarren:
@Varriount Can you add the project file as well?

BillWagner (approved):

PR #9652, opened 2024-02-28
Fix a typo

b4ry:
## Summary
Fixing a minor typo.


Rageking8 (approved):

kouvel:
The same example with the same typo is also used in the docs for `ExitUpgradeableReadLock` below in the file, would you be able to fix that as well?

kouvel (approved):
Thank you!

PR #9670, opened 2024-03-04
Clarify undefined behavior for `[ReadOnly]Span<T>(void*,int)` constructors

### LINKED ISSUE #861: Span<T>(void* pointer, int length) ctor document null-pointer behaviour
What if the pointer passed in to the Span constructor is a null pointer (e.g. created from `IntPtr.Zero.ToPoiner()`)? It might be helpful to special case the documentation what the behaviour is in such cases.

---
#### Document Details

* ID: 12542bab-90d5-ee4f-fe42-3ad37dd133a1
* Version Independent ID: 1265f19b-bf12-ea32-0b5a-1bfae94b1be4
* Content: [Span&lt;T&gt; Constructor (System)](https://docs.microsoft.com/en-us/dotnet/api/system.span-1.-ctor?view=netcore-2.1#System_Span_1__ctor_System_Void__System_Int32_)
* Content Source: [xml/System/Span`1.xml](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System/Span`1.xml)
* Product: **dotnet-api**
* GitHub Login: @dotnet-bot

colejohnson66:
## Summary

Fixes #861.

Clarifies that the pointer and length are stored without validation. While the issue of `(void*)(nint)0` specifically (from #861) is not directly addressed, it clarifies that the user is responsible for ensuring `pointer` points to a valid object.

The old version implied that the constructor "succeeding" means the usage is valid. This is a lie. It is only valid if the parameters are "correct." This is clear from the doc-comment in the source code: https://github.com/dotnet/runtime/blob/23b4030b75b43f56300a94083441bcdfb4cd6300/src/libraries/System.Private.CoreLib/src/System/Span.cs#L93-L98

```cs
/// <summary>
/// Creates a new span over the target unmanaged buffer.  Clearly this
/// is quite dangerous, because we are creating arbitrarily typed T's
/// out of a void*-typed block of memory.  And the length is not checked.
/// But if this creation is correct, then all subsequent uses are correct.
/// </summary>
```

### DIFF
--- a/xml/System/Span`1.xml
+++ b/xml/System/Span`1.xml
@@ -208,11 +208,8 @@ If `array` is `null`, this constructor returns a `null` `Span<T>`.
         <summary>Creates a new <see cref="T:System.Span`1" /> object  from a specified number of <typeparamref name="T" /> elements starting at a specified memory address.</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[
-
-          This constructor should be used with care, since it creates arbitrarily typed `T`s from a `void*`-typed block of memory, and neither `pointer` nor `length` are validated.
-          However, if `pointer` and `length` point to a valid object, all subsequent uses are correct.
-
-      ]]></format>
+          This constructor should be used with care, since it creates arbitrarily typed `T`s from a `void*`-typed block of memory, and neither `pointer` nor `length` are validated by the constructor.
+          ]]></format>
         </remarks>
         <exception cref="T:System.ArgumentException">
           <paramref name="T" /> is a reference type or contains pointers and therefore cannot be stored in unmanaged memory.</exception>


BillWagner:
ping @stephentoub to give this a look. Are these additional guarantees valid?

xml/System/ReadOnlySpan`1.xml
```diff
@@ -217,7 +217,12 @@ A `ReadOnlySpan<T>` instance is often used to reference the elements of an array
         <param name="length">The number of <typeparamref name="T" /> elements to be included in the <see cref="T:System.ReadOnlySpan`1" />.</param>
         <summary>Creates a new <see cref="T:System.ReadOnlySpan`1" /> from a specified number of <typeparamref name="T" /> elements starting at a specified memory address.</summary>
         <remarks>
-          <format type="text/markdown"><![CDATA[This constructor should be used with care, since it creates arbitrarily typed `T`s from a block of memory.]]></format>
+          <format type="text/markdown"><![CDATA[
+
+          This constructor should be used with care, since it creates arbitrarily typed `T`s from a `void*`-typed block of memory, and neither `pointer` nor `length` are validated.
+          However, if `pointer` and `length` point to a valid object, all subsequent uses are correct.
```
stephentoub, line 223: They need to remain valid as well, e.g. if pointer points to a non-pinned object that's then moved by the GC, it's no longer valid. In such a case, that's a bug in the consumption of this ctor, which needs to ensure that the pointer is and remains valid as long as the resulting span might be used.

stephentoub: (I would just delete the second sentence here.)

xml/System/ReadOnlySpan`1.xml
```diff
@@ -217,7 +217,12 @@ A `ReadOnlySpan<T>` instance is often used to reference the elements of an array
         <param name="length">The number of <typeparamref name="T" /> elements to be included in the <see cref="T:System.ReadOnlySpan`1" />.</param>
         <summary>Creates a new <see cref="T:System.ReadOnlySpan`1" /> from a specified number of <typeparamref name="T" /> elements starting at a specified memory address.</summary>
         <remarks>
-          <format type="text/markdown"><![CDATA[This constructor should be used with care, since it creates arbitrarily typed `T`s from a block of memory.]]></format>
+          <format type="text/markdown"><![CDATA[
+
+          This constructor should be used with care, since it creates arbitrarily typed `T`s from a `void*`-typed block of memory, and neither `pointer` nor `length` are validated.
```
stephentoub, line 222:
```suggestion
          This constructor should be used with care, since it creates arbitrarily typed `T`s from a `void*`-typed block of memory, and neither `pointer` nor `length` are validated by the constructor.
```

stephentoub (approved):

BillWagner (commented):
Thanks @colejohnson66 

And thanks @stephentoub for giving this a look.

I'll :shipit: now.

PR #9674, opened 2024-03-05
Correct ValueType example

KevinW1998:
I believe that this check should use the normal OR-operator and not the bitwise-OR one.


BillWagner (approved):
Thanks @KevinW1998 

This is correct, and I'll :shipit: now.

PR #9676, opened 2024-03-05
Move scheduled search; try different label action

gewarren:
- Move scheduled PR policy to separate file
- Test different technique for mentioning users based on label
- Remove outdated info about subscribing


BillWagner (approved):
This LGTM.

PR #9690, opened 2024-03-06
Update source.cs

ogmini:
## Summary

Fixing using statement


ogmini:
@dotnet-policy-service agree

gewarren (commented):
@ogmini What was the problem with the code? It used a using *declaration*, which is valid, and more modern. See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/using.

ogmini:
> @ogmini What was the problem with the code? It used a using _declaration_, which is valid, and more modern. See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/using.

Yup, just learned that. Thanks!

PR #9731, opened 2024-03-15
Add note about setting EVs from native libraries

gewarren:
Fixes dotnet/docs#39448

### DIFF
--- a/xml/System/Environment.xml
+++ b/xml/System/Environment.xml
@@ -1044,9 +1044,6 @@ The following example creates environment variables for the <xref:System.Environ
 
 The names and values for the environment variables are stored as key-value pairs in the returned <xref:System.Collections.IDictionary>.
 
-> [!NOTE]
-> In-process environment modifications made by native libraries aren't seen by managed callers.
-
 ### On Windows systems
 
 On Windows systems, the `GetEnvironmentVariables` method returns the following environment variables:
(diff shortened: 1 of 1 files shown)


gewarren:
> Looks great for https://learn.microsoft.com/en-us/dotnet/api/system.environment, but you don't think the same remarks about the relationship with native libraries should be added to https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-environment-getenvironmentvariable as well?

@mklement0 Yes, I do, but it's a separate repo, so separate PR: https://github.com/dotnet/docs/pull/40029.

xml/System/Environment.xml
```diff
@@ -1041,26 +1041,29 @@ The following example creates environment variables for the <xref:System.Environ
...
 
-### On non-Windows systems
+### On Unix-like systems
 
-On non-Windows systems, the `GetEnvironmentVariables` method retrieves the name and value of all environment variables that are inherited from the parent process that launched the `dotnet` process or that are defined within the scope of the `dotnet` process itself. Once the `dotnet` process ends, these latter environment variables cease to exist.
+On Unix-like systems, the `GetEnvironmentVariables` method retrieves the name and value of all environment variables that are inherited from the parent process that launched the `dotnet` process or that are defined within the scope of the `dotnet` process itself. Once the `dotnet` process ends, these latter environment variables cease to exist.
```
BillWagner, line 1057: I have the same question as the other PR: Is it worth mentioning that macOS is "unix-like"?

BillWagner (approved):
This LGTM @gewarren 

I had a similar question as the related PR, once you decide what to do, let's :shipit:

PR #9732, opened 2024-03-16
Update return value tables for Math.Pow

### LINKED ISSUE #9593: Update Math.Pow and MathF.Pow return values
The full set of rules for `pow` (as required by `IEEE 754 2019`) is:

* pow (x, ±0) is 1 if x is not a signaling NaN
* pow (±0, y) is ±∞ and signals the divideByZero exception for y an odd integer < 0
* pow (±0, −∞) is +∞ with no exception
* pow (±0, +∞) is +0 with no exception
* pow (±0, y) is ±0 for finite y > 0 an odd integer
* pow (−1, ±∞) is 1 with no exception
* pow (+1, y) is 1 for any y (even a quiet NaN)
* pow (x, +∞) is +0 for −1 < x < 1
* pow (x, +∞) is +∞ for x < −1 or for 1 < x (including ±∞)
* pow (x, −∞) is +∞ for −1 < x < 1
* pow (x, −∞) is +0 for x < −1 or for 1 < x (including ±∞)
* pow (+∞, y) is +0 for a number y < 0
* pow (+∞, y) is +∞ for a number y > 0
* pow (−∞, y) is −0 for finite y < 0 an odd integer
* pow (−∞, y) is −∞ for finite y > 0 an odd integer
* pow (−∞, y) is +0 for finite y < 0 and not an odd integer
* pow (−∞, y) is +∞ for finite y > 0 and not an odd integer
* pow (±0, y) is +∞ and signals the divideByZero exception for finite y < 0 and not an odd integer
* pow(±0, y) is +0 for finite y > 0 and not an odd integer
* pow(x, y) signals the invalid operation exception for finite x < 0 and finite non-integer y.

It would be good to see if we can make all of this match up with the docs

_Originally posted by @tannergooding in https://github.com/dotnet/dotnet-api-docs/issues/9571#issuecomment-1907387055_

gewarren:
Fixes #9593

(Hide whitespace changes.)


xml/System/Math.xml
```diff
@@ -6124,10 +6124,11 @@ The following example demonstrates how to use the <xref:System.Math.Min%2A> meth
...
-  
-## Remarks  
- The following table indicates the return value when various values or ranges of values are specified for the `x` and `y` parameters. For more information, see <xref:System.Double.PositiveInfinity?displayProperty=nameWithType>, <xref:System.Double.NegativeInfinity?displayProperty=nameWithType>, and <xref:System.Double.NaN?displayProperty=nameWithType>.  
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+The following table indicates the return value when various values or ranges of values are specified for the `x` and `y` parameters. For more information, see <xref:System.Double.PositiveInfinity?displayProperty=nameWithType>, <xref:System.Double.NegativeInfinity?displayProperty=nameWithType>, and <xref:System.Double.NaN?displayProperty=nameWithType>.
 
 | x                              | y                                                         | Return value                   |
```
gewarren, line 6133: @tannergooding Lines 6135, 6136, and 6140 don't seem to correspond to anything in the table you added [here](https://github.com/dotnet/dotnet-api-docs/pull/9571#issuecomment-1907387055).

xml/System/Math.xml
```diff
@@ -6137,32 +6138,35 @@ The following example demonstrates how to use the <xref:System.Math.Min%2A> meth
 | 1                              | Any value except `NaN`                                    | 1                              |
 | 1                              | `NaN`                                                     | 1 (`NaN` on .NET Framework)    |
 | Any value except 1             | `NaN`                                                     | `NaN`                          |
-| `NegativeInfinity`             | < 0                                                       | 0                              |
-| `NegativeInfinity`             | Positive odd integer                                      | `NegativeInfinity`             |
-| `NegativeInfinity`             | Positive but not an odd integer                           | `PositiveInfinity`             |
-| < 0 but not `NegativeInfinity` | Not an integer, `NegativeInfinity`, or `PositiveInfinity` | `NaN`                          |
-| -1                             | = `NegativeInfinity` or `PositiveInfinity`                | `NaN`                          |
+| `NegativeInfinity`             | < 0 and finite and odd integer                            | -0                             |
```
gewarren, line 6141: I'm not sure if it makes sense to mark the 0 return values as +, -, or ± (as in `IEEE 754 2019`).

tannergooding: I think we need to keep this information, as it is an important detail of what the function returns. `+0` and `-0` are different values for floating-point and can cause a large difference in results when input to other algorithms. A user seeing simply `0` may expect it to be positive and that can cause them to do the wrong things when the actual result would be `-0`

BillWagner (commented):
This LGTM @gewarren , once you and @tannergooding resolve the outstanding questions, let's :shipit:

xml/System/Math.xml
tannergooding: 6136 corresponds to the first line in the table `pow (x, ±0) is 1 if x is not a signaling NaN`, 6135 then explicitly documents the behavior for NaN.

6140 looks to be trying to do the same thing. It's effectively documenting the counterpart to 6139

-----

In general there are some places where this table is potentially confusing because it lists `0` instead of `+/-0` (or `±0`, if Unicode characters are safe to use). This may lead users to believe this means only positive zero. The same goes for listing `NaN` rather than `qNaN` (quiet NaN) or `sNaN` (signalling NaN).

Notably, however, .NET doesn't differentiate between `qNaN` and `sNaN` itself, which is what ultimately makes some of the entries "redundant".

I'd think the simplest thing would be to be explicit when a difference exists and matters (`±0`) and don't list the difference where it doesn't matter for .NET (`qNaN` vs `sNaN`). A small note could then be left at the bottom explaining that .NET doesn't differentiate between them and cite the relevant portion of the IEEE 754 spec that allows this (most notably its caused by disabling IEEE 754 floating-point exceptions).

gewarren: @tannergooding Another question, how should I show a return value of ±∞ in the docs?

tannergooding: I think this is a question for @gewarren. I'm not sure as to the recommended way to do this for the docs.

gewarren:
@BillWagner Could you approve this one?

BillWagner (approved):
This LGTM. Let's :shipit:

PR #9746, opened 2024-03-19
Add examples for generic support.

AaronRobinsonMSFT:
## Summary

Updating documentation based on new features.

See https://github.com/dotnet/runtime/issues/89439

### DIFF
--- a/xml/System.Runtime.CompilerServices/UnsafeAccessorAttribute.xml
+++ b/xml/System.Runtime.CompilerServices/UnsafeAccessorAttribute.xml
@@ -37,7 +37,7 @@
 
 You can apply this attribute to an `extern static` method. The implementation of the `extern static` method annotated with this attribute will be provided by the runtime based on the information in the attribute and the signature of the method that the attribute is applied to. The runtime will try to find the matching method or field and forward the call to it. If the matching method or field is not found, the body of the `extern static` method will throw <xref:System.MissingFieldException> or <xref:System.MissingMethodException>.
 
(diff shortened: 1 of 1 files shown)


xml/System.Runtime.CompilerServices/UnsafeAccessorAttribute.xml
```diff
@@ -31,94 +31,129 @@
...
-
+      <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+You can apply this attribute to an `extern static` method. The implementation of the `extern static` method annotated with this attribute will be provided by the runtime based on the information in the attribute and the signature of the method that the attribute is applied to. The runtime will try to find the matching method or field and forward the call to it. If the matching method or field is not found, the body of the `extern static` method will throw <xref:System.MissingFieldException> or <xref:System.MissingMethodException>. Generic parameters are supported, but must match the target in form and index (that is, type parameters must be type parameters and method parameters must be method parameters). The `extern static` method's generic parameters must also exactly match any constraints reflected on the target. If constraints do not match the method will throw <xref:System.InvalidProgramException>.
```
jkotas, line 38: This should say that the generic parameters are only supported in .NET 9 and later version.

AaronRobinsonMSFT, line 38:
```suggestion
You can apply this attribute to an `extern static` method. The implementation of the `extern static` method annotated with this attribute will be provided by the runtime based on the information in the attribute and the signature of the method that the attribute is applied to. The runtime will try to find the matching method or field and forward the call to it. If the matching method or field is not found, the body of the `extern static` method will throw <xref:System.MissingFieldException> or <xref:System.MissingMethodException>.

Generic parameters are supported since .NET 9. Generic parameters must match the target in form and index (that is, type parameters must be type parameters and method parameters must be method parameters). The `extern static` method's generic parameters must also exactly match any constraints reflected on the target. If constraints do not match the method will throw <xref:System.InvalidProgramException>.
```

gewarren (approved):
Thanks!

PR #9747, opened 2024-03-20
Update documentation on MethodBase.Invoke invokeAttrs parameter

serhiiz:
It seems the documentation on [System.Reflection.MethodBase.Invoke](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.methodbase.invoke?view=net-8.0#system-reflection-methodbase-invoke(system-object-system-reflection-bindingflags-system-reflection-binder-system-object()-system-globalization-cultureinfo)) in regard to the `invokeAttr` parameter is outdated. 
According to the documentation:

> invokeAttr [BindingFlags](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.bindingflags?view=net-8.0)
> A bitmask that is a combination of 0 or more bit flags from [BindingFlags](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.bindingflags?view=net-8.0). If binder is null, this parameter is assigned the value [Default](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.bindingflags?view=net-8.0#system-reflection-bindingflags-default); thus, whatever you pass in is ignored.

the `invokeAttrs` parameter is ignored when the binder parameter is `null`. However, that doesn't seem to be accurate. According to the implementations ([System.Reflection.RuntimeMethodInfo.cs](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Reflection/RuntimeMethodInfo.cs#L99), [System.Reflection.Emit.DynamicMethod.CoreCLR.cs](https://github.com/dotnet/runtime/blob/main/src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/DynamicMethod.CoreCLR.cs#L117), [System.Reflection.Emit.DynamicMethod.CoreCLR.cs](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Reflection/RuntimeConstructorInfo.cs#L136)), the invokeAttrs is never ignored. Unless I'm missing something.

## Summary

The suggestion is to take out the statement about `invokeAttr` being ignored when the `binder` is `null`.

### DIFF
--- a/xml/System.Reflection/MethodBase.xml
+++ b/xml/System.Reflection/MethodBase.xml
@@ -1174,7 +1174,7 @@ Note: In <see href="https://go.microsoft.com/fwlink/?LinkID=247912">.NET for Win
       </Parameters>
       <Docs>
         <param name="obj">The object on which to invoke the method or constructor. If a method is static, this argument is ignored. If a constructor is static, this argument must be <see langword="null" /> or an instance of the class that defines the constructor.</param>
-        <param name="invokeAttr">A bitmask that is a combination of 0 or more bit flags from <see cref="T:System.Reflection.BindingFlags" />. If <paramref name="binder" /> is <see langword="null" />, this parameter is assigned the value <see cref="F:System.Reflection.BindingFlags.Default" />; thus, whatever you pass in is ignored.</param>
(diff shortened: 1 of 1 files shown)


steveharter:
LGTM; the only flag that would be used without a binder is `BindingFlags.DoNotWrapExceptions` but I don't think calling that out is necessary.

steveharter (approved):

PR #9753, opened 2024-03-20
Add information about console output encoding

Smaug123:
## Summary

Adds information I failed to find earlier today, namely "what is the output encoding".

I also added information about the possibly unintuitive operation of `Write{Line}(char array)` when the input contains surrogates. (I claim it is indeed unintuitive, because it surprised me.) I've hedged with "*may* be truncated"; in my testing, `Console.Write` does truncate but `Console.WriteLine` does not, but I've tested on only one platform. I have made the rather arbitrary choice not to give the same warning on the `Write(char)` overload, because that one's even weirder and I'm much less confident in documenting it: whether you get re-encoded output or not depends on whether `Console.Out` gets flushed in between the write of the two chars in a surrogate pair or not. Of course it's implicitly illegal to write surrogates out at all, but it didn't seem to be documented anywhere obvious that this is so.


xml/System/Console.xml
```diff
@@ -4995,8 +5003,9 @@ This method can be used to reacquire the standard output stream after it has bee
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- If value is `null`, nothing is written to the standard output stream.
+ The output is encoded according to <xref:System.Console.OutputEncoding%2A>.
```
Smaug123, line 5006: (I reordered this to put my comment first, because otherwise there's a syntactic ambiguity about whether my comment is conditioned on the "value is `null`" from the other comment.)

PR #9754, opened 2024-03-20
Fix up crefs/xrefs

gewarren:
- Fix some bad xrefs/crefs
- Add missing periods after Markdown links


BillWagner (approved):
This LGTM. Let's :shipit:

PR #9759, opened 2024-03-22
Actualize snippet example for Next() Method with rest of languages

chuve:
The example doesn't represent a description where it's embedded. The description states, `generate a specific number of random numbers requested by the user.`. However, the random numbers are hardcoded to 10 and aren't requested from user input.

## Summary
The page where the example is embedded: https://learn.microsoft.com/en-us/dotnet/api/system.random.next?view=net-8.0

### DIFF
--- a/snippets/csharp/System/Random/Overview/next3.cs
+++ b/snippets/csharp/System/Random/Overview/next3.cs
@@ -1,30 +1,41 @@
-﻿using System;
+using System;
 
 public class Example
 {
-   public static void Main()
-   {
-      // <Snippet5>
-      Random rnd = new Random();
+    public static void Main()
+    {
+        // <Snippet5>
+        Console.Write("Number of random numbers to generate: ");
 
-      Console.WriteLine("Generating 10 random numbers:");
+        string? line = Console.ReadLine();
+        Random rnd = new Random();
 
-      for (uint ctr = 1; ctr <= 10; ctr++)
-         Console.WriteLine($"{rnd.Next(),15:N0}");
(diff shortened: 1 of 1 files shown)


chuve:
@dotnet-policy-service agree

gewarren:
Thanks for noticing and fixing the discrepancy @chuve. It might be better to keep the original code (since it's simpler) and just change the intro text. What do you think? Since the example is more about Random that about reading user input.

chuve:
> Thanks for noticing and fixing the discrepancy @chuve. It might be better to keep the original code (since it's simpler) and just change the intro text. What do you think? Since the example is more about Random that about reading user input.

That was my first thought—update the intro text. However, after I checked code snippets in other languages (C++ and VB), I concluded that it would probably be better to update the code snippet instead. I agree with you that the original code is simpler and intended to show the use case of the `Random.Next` method.

chuve:
@gewarren, it seems like one of the gh actions has failed - https://github.com/dotnet/dotnet-api-docs/actions/runs/8393286204/job/23071411802?pr=9759 
Do you know if I need to fix it? If so, how can I do it?
Thanks!

gewarren:
> @gewarren, it seems like one of the gh actions has failed - https://github.com/dotnet/dotnet-api-docs/actions/runs/8393286204/job/23071411802?pr=9759 Do you know if I need to fix it? If so, how can I do it? Thanks!

@chuve I will fix this error and commit directly to your branch if that's okay with you. It involves adding a project file and then fixing all the snippets so they compile as part of one project.

gewarren (approved):

PR #9809, opened 2024-04-14
Update BitConverter.xml

colejohnson66:
Update the remarks table for class-level documentation.

## Summary

Added `XToUIntXXBits` and `UIntXXBitsToX` method mentions for `Single` and `Double`, and added a row for `Half`. Also added some spacing in table source for ease of editing in the future.

## Questions

With `TryWriteBytes(Span<byte>, T)` now existing, this table could further be altered to include a "serialize to bytes" column. Is this desired?

Alternatively, due to the expanding API surface, it could be argued that this table should just be removed?


gewarren:
Tagging @dotnet/area-system-runtime.

tannergooding (approved):

gewarren:
Thanks @colejohnson66

PR #9828, opened 2024-04-19
Fix SHA384/512 examples

### LINKED ISSUE #9826: SHA512 and the likes' examples are obsolete
[SHA512 Examples](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha512?view=net-8.0#examples) or [SHA384 Examples](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha384?view=net-8.0#examples) list the following snippet:

```cs
byte[] data = new byte[DATA_SIZE];
byte[] result;
SHA512 shaM = new SHA512Managed();
result = shaM.ComputeHash(data);
```

[`SHA512Managed`](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha512managed?view=net-8.0) however is obsolete and should not be used.

vcsjones:
## Summary

1. The examples for the algorithms that are not specifically for the Managed variants now use the factory method.
2. The examples were updated to demonstrate proper resource cleanup.

Fixes #9826

### DIFF
--- a/snippets/visualbasic/VS_Snippets_CLR_Classic/classic SHA384 Example/VB/source.vb
+++ b/snippets/visualbasic/VS_Snippets_CLR_Classic/classic SHA384 Example/VB/source.vb
@@ -2,10 +2,10 @@
 Imports System.ComponentModel
 Imports System.Security.Cryptography
 
-Public Class Sample
-    Protected DATA_SIZE As Integer = 1024
+Public Module Sample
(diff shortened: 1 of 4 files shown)


adamsitnik (approved):
LGTM, thank you very much for fixing the docs @vcsjones !

gewarren (approved):
Thank you @vcsjones

PR #9852, opened 2024-04-24
Fix the summary of Control.DefaultPadding in Control.xml

zxcv1050135042:
Missing "default" in summary can easily lead to misunderstandings with Control.Padding

### DIFF
--- a/xml/System.Windows.Forms/Control.xml
+++ b/xml/System.Windows.Forms/Control.xml
@@ -4285,7 +4285,7 @@ The following table lists Windows Forms controls and which event (<xref:System.W
       </ReturnValue>
       <Docs>
         <summary>Gets the default internal spacing, in pixels, of the contents of a control.</summary>
-        <value>A <see cref="T:System.Windows.Forms.Padding" /> that represents the internal spacing of the contents of a control.</value>
+        <value>The default internal spacing of the contents of a control.</value>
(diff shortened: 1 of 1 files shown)


xml/System.Windows.Forms/Control.xml
```diff
@@ -4284,7 +4284,7 @@ The following table lists Windows Forms controls and which event (<xref:System.W
         <ReturnType>System.Windows.Forms.Padding</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets the internal spacing, in pixels, of the contents of a control.</summary>
+        <summary>Gets the default internal spacing, in pixels, of the contents of a control.</summary>
         <value>A <see cref="T:System.Windows.Forms.Padding" /> that represents the internal spacing of the contents of a control.</value>
```
gewarren, line 4288:
```suggestion
        <value>The default internal spacing of the contents of a control.</value>
```

gewarren (approved):
Should we update the value as well?

xml/System.Windows.Forms/Control.xml
zxcv1050135042: Oh, of course, you're right, I forget :p

zxcv1050135042 (commented):
Oh, of course, you're right, I forget :p

PR #9892, opened 2024-05-09
Document issues with using * in AssemblyVersion

tmat:
## Summary

Document issues caused by using * in AssemblyVersion.


tmat:
@jaredpar @AndyGerlicher @AArnott @davidwengier

xml/System.Reflection/AssemblyVersionAttribute.xml
```diff
@@ -86,12 +86,32 @@
...
+> [!IMPORTANT] 
+> Use of <xref:System.Reflection.AssemblyVersionAttribute> attribute that specifies an asterisk makes the build outputs non-reproducible (see [Reproducible Builds](https://reproducible-builds.org/)).
+
+> [!IMPORTANT] 
+> Use of <xref:System.Reflection.AssemblyVersionAttribute> attribute that specifies an asterisk may degrade build performance as it prevents build from caching compiler outputs.
+
+> [!IMPORTANT] 
+> Use of <xref:System.Reflection.AssemblyVersionAttribute> attribute that specifies an asterisk is incompatible with [Edit & Continue](https://learn.microsoft.com/en-us/visualstudio/debugger/edit-and-continue-visual-csharp) and [Hot Reload](https://learn.microsoft.com/en-us/visualstudio/debugger/hot-reload) features.
+
+ Some of these issues can be mitigated by limiting the use of time-based versions to release builds using conditional compilation like so:
```
gewarren, line 103:
```suggestion
You can mitigate some of these issues by limiting the use of time-based versions to release builds using conditional compilation, like so:
```

xml/System.Reflection/AssemblyVersionAttribute.xml
```diff
@@ -86,12 +86,32 @@
...
+
+```
+#if DEBUG
+[assembly: AssemblyVersion("1.0.0.0")]
+#else
+[assembly: AssemblyVersion("1.0.*")]
+#endif
+```
+
+ A better approach to versioning is to derive the assembly and/or file version from the HEAD commit SHA (for git repositories). See, for example, [Nerdbank.GitVersioning](https://github.com/dotnet/Nerdbank.GitVersioning).
```
gewarren, line 113:
```suggestion
 A better approach to versioning is to derive the assembly or file version from the `HEAD` commit SHA (for git repositories). See, for example, [Nerdbank.GitVersioning](https://github.com/dotnet/Nerdbank.GitVersioning).
```

xml/System.Reflection/AssemblyVersionAttribute.xml
```diff
@@ -86,12 +86,32 @@
...
+ For example, `[assembly:AssemblyVersion("2.3.25.1")]` indicates 2 as the major version, 3 as the minor version, 25 as the build number, and 1 as the revision number. 
+
+ The <xref:System.Reflection.AssemblyVersionAttribute> attribute allows to specify an asterisk (\*) in place of the build or revision number. A version number such as `[assembly:AssemblyVersion("1.2.*")]` specifies 1 as the major version, 2 as the minor version, and accepts the default build and revision numbers. A version number such as `[assembly:AssemblyVersion("1.2.15.*")]` specifies 1 as the major version, 2 as the minor version, 15 as the build number, and accepts the default revision number. The default build number increments daily. The default revision number is the number of seconds since midnight local time (without taking into account time zone adjustments for daylight saving time), divided by 2. If you specify an asterisk for the build number, you cannot specify a revision number.  
+
+> [!IMPORTANT] 
+> Use of <xref:System.Reflection.AssemblyVersionAttribute> attribute that specifies an asterisk makes the build outputs non-reproducible (see [Reproducible Builds](https://reproducible-builds.org/)).
```
gewarren, line 95:
```suggestion
> Use of the <xref:System.Reflection.AssemblyVersionAttribute> attribute that specifies an asterisk:
>
> - Makes the build outputs non-reproducible (see [Reproducible builds](https://reproducible-builds.org/)).
```

xml/System.Reflection/AssemblyVersionAttribute.xml
```diff
@@ -86,12 +86,32 @@
...
+
```
gewarren, line 96:
```suggestion
> - Might degrade build performance, as it prevents build from caching compiler outputs.
```

xml/System.Reflection/AssemblyVersionAttribute.xml
```diff
@@ -86,12 +86,32 @@
...
+> [!IMPORTANT]
```
gewarren, line 97:
```suggestion
> - Is incompatible with the [Edit & Continue](/visualstudio/debugger/edit-and-continue-visual-csharp) and [Hot Reload](/visualstudio/debugger/hot-reload) features.
```

xml/System.Reflection/AssemblyVersionAttribute.xml
```diff
@@ -86,12 +86,32 @@
...
+
+> [!IMPORTANT] 
+> Use of <xref:System.Reflection.AssemblyVersionAttribute> attribute that specifies an asterisk makes the build outputs non-reproducible (see [Reproducible Builds](https://reproducible-builds.org/)).
+
+> [!IMPORTANT] 
+> Use of <xref:System.Reflection.AssemblyVersionAttribute> attribute that specifies an asterisk may degrade build performance as it prevents build from caching compiler outputs.
```
gewarren, line 98: And then delete lines 98-101.

gewarren (approved):
Thanks for adding this.

AArnott (approved):
Is there any way the IDE support can tip users off when E&C fails due to `*` in the version attributes? Instead of just saying the value changed, call out a URL where users can learn what to do instead to fix E&C.

tmat:
@AArnott 
Definitely. IN some cases we report better error but other cases fail with error message that's not very useful. Will follow up on that.

xml/System.Reflection/AssemblyVersionAttribute.xml
```diff
@@ -86,12 +86,32 @@
...
-  
-> [!NOTE]
->  If you specify an asterisk for the build number, you cannot specify a revision number.  
-  
+
+ For example, `[assembly:AssemblyVersion("2.3.25.1")]` indicates 2 as the major version, 3 as the minor version, 25 as the build number, and 1 as the revision number. 
+
+ The <xref:System.Reflection.AssemblyVersionAttribute> attribute allows to specify an asterisk (\*) in place of the build or revision number. A version number such as `[assembly:AssemblyVersion("1.2.*")]` specifies 1 as the major version, 2 as the minor version, and accepts the default build and revision numbers. A version number such as `[assembly:AssemblyVersion("1.2.15.*")]` specifies 1 as the major version, 2 as the minor version, 15 as the build number, and accepts the default revision number. The default build number increments daily. The default revision number is the number of seconds since midnight local time (without taking into account time zone adjustments for daylight saving time), divided by 2. If you specify an asterisk for the build number, you cannot specify a revision number.
```
ericstj, line 92: I didn't even know about this feature.  It's not a runtime feature - the attribute itself is stored as a string.  It must be in the compiler:
https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK7wCYgNQB8ACATAIwCwAUPgAwAE+xAdAEoCmAZgDYsDGCUA9jADcFANoQAzhJZhgHAJ4gaAQSky58gGotkEgTAAUAIkZUGAKiMBKALoV6ATgMwWAdxVrZC7bv3KECMhQwKgILMamFtYMPnqCViKUxE6q0l7yDADiLAgAogAePKGwAOap6goGVlk5AHIQYOHVsfoJFEA==

https://github.com/dotnet/roslyn/blob/18614bceb9d0a9eda5cfb47409c3058971cc6751/src/Compilers/Core/Portable/VersionHelper.cs#L170

Wow, that's indeed broken.  Did we consider an diagnostic to discourage this?

xml/System.Reflection/AssemblyVersionAttribute.xml
jaredpar: This is correct but the situation is more guarded than the docs suggest. Builds are only reproducible if you have `<Deterministic>true</Deterministic>` . That is the default in the .NET SDK  so all modern builds have this set. If that is set and you use `*` in versions then you will get a diagonstic. For example: 

```cmd
E:\temp\console\.complog\console\export\src\obj\Debug\net9.0\console.AssemblyInfo.cs(19,55): error CS8357: The specified version string '1.0.0.*' contains wildcards, which are not compatible with determinism. Either remove wildcards from the version string, or disable determinism for this compilation
```

tmat: Does build caching in VS require Deterministic=true?

tmat: Seems like `CS8357` is not documented: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/CS8357

xml/System.Reflection/AssemblyVersionAttribute.xml
```diff
@@ -86,12 +86,30 @@
...
```
gewarren, line 92:
```suggestion
The <xref:System.Reflection.AssemblyVersionAttribute> attribute allows you to specify an asterisk (\*) in place of the build or revision number. A version number such as `[assembly:AssemblyVersion("1.2.*")]` specifies 1 as the major version, 2 as the minor version, and accepts the default build and revision numbers. A version number such as `[assembly:AssemblyVersion("1.2.15.*")]` specifies 1 as the major version, 2 as the minor version, 15 as the build number, and accepts the default revision number. The default build number increments daily. The default revision number is the number of seconds since midnight local time (without taking into account time zone adjustments for daylight saving time), divided by 2. If you specify an asterisk for the build number, you cannot specify a revision number.  
```
```suggestion
 The <xref:System.Reflection.AssemblyVersionAttribute> attribute allows to specify an asterisk (\*) in place of the build or revision number. A version number such as `[assembly:AssemblyVersion("1.2.*")]` specifies 1 as the major version, 2 as the minor version, and accepts the default build and revision numbers. A version number such as `[assembly:AssemblyVersion("1.2.15.*")]` specifies 1 as the major version, 2 as the minor version, 15 as the build number, and accepts the default revision number. The default build number increments daily. The default revision number is the number of seconds since midnight local time (without taking into account time zone adjustments for daylight saving time), divided by 2. If you specify an asterisk for the build number, you cannot specify a revision number.  
```

PR #9899, opened 2024-05-13
Align ActivitySource.StartActivity doc with runtime repo

ghost:
## Summary

Align `ActivitySource.StartActivity` documentation with [runtime repo](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/ActivitySource.cs).

### DIFF
--- a/xml/System.Diagnostics/ActivitySource.xml
+++ b/xml/System.Diagnostics/ActivitySource.xml
@@ -382,11 +382,10 @@ If the Activity object is created, it will not start automatically. Callers need
         <Parameter Name="kind" Type="System.Diagnostics.ActivityKind" />
       </Parameters>
       <Docs>
-        <param name="name">The operation name of the activity.</param>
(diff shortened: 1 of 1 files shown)


gewarren:
Thanks @joegoldman2. While these docs might be different to the dotnet/runtime repo, this repo is currently considered the "source of truth". And as you can see, this repo contains improvements such as punctuation and grammar that the runtime/repo doesn't contain. So this PR is a step backwards IMO and I'll close it.

image

joegoldman2:
Ok I understand. I think the summary, at least, should be reviewed

image

They should all begin with `Creates and starts`.

gewarren:
> Ok I understand. I think the summary, at least, should be reviewed
> 
> image
> 
> They should all begin with `Creates and starts`.

Ahh, okay. Do you want to create a new PR for that?

PR #9933, opened 2024-05-19
2nd call to MemoryCache.Set() with the same key erases entry if cache is full

pedrobsaila:
… 

## Summary

These changes document a strange behaviour (though understandable) in memory cache where existing and new entries get deleted when size limit is set and exceeded

Contributes to dotnet/runtime#36039

### DIFF
--- a/xml/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.xml
+++ b/xml/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.xml
@@ -214,7 +214,7 @@
 > [!CAUTION]
 > There may be transient states while the cache is full but blocks are being released on a background thread. In this state, unexpected behavior might occur if the cache is being overutilized.
 >
-> When updating an existing item with another one that will make cache size go beyond the limit, the existing and new items will be removed.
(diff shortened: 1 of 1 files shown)


xml/Microsoft.Extensions.Caching.Memory/MemoryCacheOptions.xml
```diff
@@ -210,9 +210,11 @@
...
         <value>To be added.</value>
         <remarks>
-          <format type="text/markdown"><![CDATA[
-> [!CAUTION]
-> There may be transient states while the cache is full but blocks are being released on a background thread. In this state, unexpected behavior might occur if the cache is being overutilized. 
+          <format type="text/markdown"><![CDATA[
+> [!CAUTION]
+> There may be transient states while the cache is full but blocks are being released on a background thread. In this state, unexpected behavior might occur if the cache is being overutilized.
+>
+> When updating an existing item with another one that will make cache size go beyond the limit, the existing and new items will be removed.
```
adegeo, line 217: I don't know that this makes sense. Isn't "updating an existing item with another one" actually replacing an item? The wording sounds like "replace" not "update"

adegeo:
You can use `<xref:Microsoft.Extensions.Caching.Memory.CacheExtensions.Set%2A>` to link to an API in markdown

gewarren:
I believe we can close this given that the underlying behavior was fixed in https://github.com/dotnet/runtime/pull/103931.

PR #9992, opened 2024-06-04
Documentation fix for Socket.Select

logiclrd:
The `Socket.Select` documentation (for the overload where `timeout` is a `System.TimeSpan`) states that `ArgumentNullException` is thrown if any of `checkRead`, `checkWrite` or `checkError` is empty/`null`. But, null or empty collections can be passed for them. The real requirement is that at least one of them must contain at least one socket. It is perfectly okay to `select(read_sockets, NULL, NULL)` or `select(NULL, write_sockets, NULL)` or `select(NULL, NULL, error_sockets)`.


antonfirsov (approved):
LGTM. Thanks!

PR #9999, opened 2024-06-06
Update the maximum message length in the EventLog.WriteEntry method documentation

0xced:
## Summary

It was found empirically on Windows 10 that the maximum message length is 31,718.

### DIFF
--- a/xml/System.Diagnostics/EventLog.xml
+++ b/xml/System.Diagnostics/EventLog.xml
@@ -3294,7 +3294,7 @@ SVC_UPDATE.EXE
   
 -or-
   
- The message string is longer than 31,839 bytes (32,766 bytes on Windows operating systems before Windows Vista).  
+ The message string is longer than 31,718 bytes (32,766 bytes on Windows operating systems before Windows Vista).  
   
 -or-
   
@@ -3405,7 +3405,7 @@ SVC_UPDATE.EXE
   
 -or-
   
- The message string is longer than 31,839 bytes (32,766 bytes on Windows operating systems before Windows Vista).  
+ The message string is longer than 31,718 bytes (32,766 bytes on Windows operating systems before Windows Vista).  
   
 -or-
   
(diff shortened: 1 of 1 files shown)


gewarren:
@tommcdon Could you review?

tommcdon:
.NET is using `ReportEvent`, and according to the Windows documentation for `ReportEvent`, our existing documentation is correct:

https://learn.microsoft.com/windows/win32/api/winbase/nf-winbase-reporteventw

```text
[in] lpStrings
A pointer to a buffer containing an array of null-terminated strings that are merged into the message before Event Viewer displays the string to the user. This parameter must be a valid pointer (or NULL), even if wNumStrings is zero. Each string is limited to 31,839 characters.
Prior to Windows Vista:  Each string is limited to 32K characters.
```

The .NET code that calls `ReportEvent` is checking 32K-2 characters:

https://github.com/dotnet/runtime/blob/c1a9f26efa4fcf2e3fdcd8557da19d358f51eb00/src/libraries/System.Diagnostics.EventLog/src/System/Diagnostics/EventLogInternal.cs#L1354-L1357
```csharp
                // make sure the strings aren't too long.  MSDN says each string has a limit of 32k (32768) characters, but
                // experimentation shows that it doesn't like anything larger than 32766
                if (strings[i].Length > 32766)
                    throw new ArgumentException(SR.LogEntryTooLong);
```

Therefore, the error would likely be coming from Windows and not .NET for string lengths <= 32766 characters.  Given that .NET has no control over the version of Windows and the various character limits it is imposing, I suggest we remove mentions of specific character limits under 32766 characters.   

Also I would expect that if the string length is <= 32766 and the API call fails due to a restriction from Windows, we would get a Win32Exception and not an ArgumentException.  @0xced can you confirm?  

If yes, then I suggest changing the text to something like:
`The message string is longer than 32,766 bytes.`

tommcdon (changes requested):
I left a comment for your review

gewarren:
@0xced Do you plan to address the requested changes?

PR #10010, opened 2024-06-11
Update SmtpClient.xml

David-Barrett-MS:
## Summary

Remove reference to MailKit as we should not be recommending third party tools (it is up to the user to determine whether a third party tool is appropriate).  We have also been receiving support cases for MailKit (because it is referred to in our docs).

### DIFF
--- a/xml/System.Net.Mail/SmtpClient.xml
+++ b/xml/System.Net.Mail/SmtpClient.xml
@@ -71,7 +71,7 @@
 The `SmtpClient` class is used to send email to an SMTP server for delivery. The SMTP protocol is defined in RFC 2821, which is available at [https://www.ietf.org](https://www.ietf.org/).  
 
 > [!IMPORTANT]
-> We don't recommend that you use the `SmtpClient` class for new development because `SmtpClient` doesn't support many modern protocols. Use [MailKit](https://github.com/jstedfast/MailKit) or other libraries instead. For more information, see [SmtpClient shouldn't be used](https://github.com/dotnet/platform-compat/blob/master/docs/DE0005.md) on GitHub.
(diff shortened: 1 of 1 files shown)


MihaZupan:
We make similar recommendations in other places, e.g. https://github.com/dotnet/dotnet-api-docs/blob/main/includes%2Fdrawing.md#L12
How come MailKit is more problematic?

David-Barrett-MS:
MailKit should not be referenced in our docs as it is not a Microsoft product and is available open-source from Github.

We have received support cases on it - hence removing the references from our docs.  Referring to it in our docs implies that Microsoft recommends this library.  We don't - it is up to customer to choose.

MihaZupan:
The other examples I linked likewise include non Microsoft open source products

gewarren:
It's pretty common to link to third-party libraries from Learn, for example: https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-only#recommended-action.

The general guidance for third-party links is here: https://review.learn.microsoft.com/en-us/help/platform/links-how-to?toc=%2Fhelp%2Fplatform%2Ftoc.json&bc=%2Fhelp%2Fplatform%2Fbreadcrumb%2Ftoc.json&branch=main#third-party-site-link-guidance.

David-Barrett-MS:
I'm not suggesting it isn't common to link to third party sites.  I am requesting that this specific reference be removed as it is generating support requests for us, so isn't a benefit at all.  Links to third party sites should only be included if there is a benefit.  I am from the team that supports System.Net.Mail (or more accurately, deals with cases raised against it as it is no longer officially supported), and both EMEA and US have been receiving cases regarding Mailkit.  If customers are having issues with Mailkit enough to raise cases to us, in my view that link is detrimental.

David-Barrett-MS:
It should also be noted that the link hasn't been added per the guidelines you reference (it is not clearly marked as an external site for one).

PR #10021, opened 2024-06-13
Update BigInteger.xml

tfenise:
## Summary

The return value of `BigInteger.GreatestCommonDivisor` is nonnegative, as `BigInteger.GreatestCommonDivisor(0, 0)` is zero.

### DIFF
--- a/xml/System.Numerics/BigInteger.xml
+++ b/xml/System.Numerics/BigInteger.xml
@@ -2196,7 +2196,7 @@ This method returns 0 if the value of current object is equal to <xref:System.Nu
 > [!NOTE]
 >  Computing the greatest common divisor of very large values of `left` and `right` can be a very time-consuming operation.
 
- The value returned by the <xref:System.Numerics.BigInteger.GreatestCommonDivisor%2A> method is always nonnegative regardless of the sign of the `left` and `right` parameters.
+ The value returned by the <xref:System.Numerics.BigInteger.GreatestCommonDivisor%2A> method is always positive (including zero) regardless of the sign of the `left` and `right` parameters.
 
 
 


xml/System.Numerics/BigInteger.xml
```diff
@@ -2196,7 +2196,7 @@ This method returns 0 if the value of current object is equal to <xref:System.Nu
 > [!NOTE]
 >  Computing the greatest common divisor of very large values of `left` and `right` can be a very time-consuming operation.
 
- The value returned by the <xref:System.Numerics.BigInteger.GreatestCommonDivisor%2A> method is always positive regardless of the sign of the `left` and `right` parameters.
+ The value returned by the <xref:System.Numerics.BigInteger.GreatestCommonDivisor%2A> method is always nonnegative regardless of the sign of the `left` and `right` parameters.
```
tannergooding, line 2199: While I don't mind clarifying things, the doc is notably correct as is, particularly in the context of programming where `zero` often has a sign.

We have a plethora of APIs, such as `INumberBase<T>.IsPositive` that intentionally include zero (or `positive zero` in the case that the underlying number format has representation for both `+0` and `-0`).

I think rather than using the term `non-negative` here it would be better to explicitly state it more like:
```suggestion
 The value returned by the <xref:System.Numerics.BigInteger.GreatestCommonDivisor%2A> method is always positive (including zero) regardless of the sign of the `left` and `right` parameters.
```

As this further reinforces that .NET (and programming in general) may have multiple types of zero to consider and that zero itself is considered positive.

tfenise: OK. However, `BigInteger.Sign` considers zero to be neither positive nor negative. Even `Math.Sign(double.NegativeZero)` gives zero. It just seems to be unfortunately inconsistent here and there, even in .NET.

gewarren (approved):

PR #10024, opened 2024-06-14
Update OleDbCommand.xml line 1248

chanmmn:
C# compiler only recognizes Int32 and not int32. int32 will have Build Error.

## Summary

Describe your changes here.


roji (approved):
Thanks!

PR #10033, opened 2024-06-17
Updates `ListBoxItem.OnMouseLeftButtonDown` to refer to the left mouse button instead of the right

jake-pauls:
## Summary

Quick fix for a typo in the documentation where the `ListBoxItem.OnMouseLeftButtonDown` refers to pressing the right mouse button, when it should instead refer to pressing the left.


gewarren (approved):
Thank you @jake-pauls.

PR #10060, opened 2024-06-25
Add spacing in SemaphoreSlim docs

89netraM:
Adds a space after `Timeout.Infinite` member reference and before parentheses describing its value "(-1)". When rendered in IDEs and on [learn.microsoft.com](https://learn.microsoft.com/en-us/dotnet/api/system.threading.semaphoreslim.wait?view=net-8.0#system-threading-semaphoreslim-wait(system-int32)) this could cause confusion as it looked like a method call.


gewarren (approved):
Thanks @89netraM

PR #10088, opened 2024-07-05
Fix typo in RSASignaturePaddingMode.Pss

vcsjones:
## Summary

The `Pss` member was documented as being RSASSA-PKCS1-v1.5 when it should be documented as RSASSA-PSS.


bartonjs (approved):

bartonjs:
Thanks for jumping on that, @vcsjones

PR #10103, opened 2024-07-12
Update Operators.cs fixed console log params

adjit:
## Summary

Fixed the console log that printed out the wrong date for the difference comparison. Noticed this when reading the docs


gewarren (approved):
Thanks so much @adjit.

Issue #10110, opened 2024-07-14
System.Linq doc snippets (\dotnet-api-docs\snippets\csharp\System.Linq


DickBaker:
1. the [LINQ docs C# samples](https://github.com/dotnet/dotnet-api-docs/tree/main/snippets/csharp/System.Linq) consist of 6 projects to showcase **IEnumerable** and **IQueryable**
unfortunately much code unnecessarily invokes **AsQueryable()** despite simpler **IEnumerable** method implementations existing
[thus such guidance is poor as such **AsQueryable()** can be avoided]
2. should update to useful modern C# v12 semantics [Primary constructors, Collection expressions, Span, is, etc]
3. calls being made to default parameterless ctor then assigning individual properties, rather than using parameterised ctors explicitly
or better yet primary constructor
4. current analyzers also disclose numerous Warnings/Messages, emphasising poor guidance for today's coding
5. many cases of wastefully using Any(),Count(),LongCount etc when optimal coding available (e.g. Count, Length for Collections).
6. spurious ToString() call in [OrderByIComparer()](https://github.com/dotnet/dotnet-api-docs/blob/main/snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs#L1828)

I will shortly submit a PR to explain problems & my suggestions.

PR #10113, opened 2024-07-15
refresh LINQ docs to latest semantics

DickBaker:
current [LINQ docs C# samples](https://github.com/DickBaker/dotnet-api-docs/tree/master/snippets/csharp/System.Linq) are vintage 2019 and **need** refresh (Issue #10110 relates)
obviously **.NET** and **C#** have evolved apace since, so imho opportune to modernise docs too for best guidance

This PR is intended to explain my _cryptic_ notes in Issue #10110, and to propose a useful refresh of snippet code samples

Note that _Allow edits and access to secrets by maintainers_ is checked so I will be happy for MS wonks to edit, e.g.

1. ignore the **.editorconfig** that I used to enforce some consistency across the projects
2. dispense with the Linq.sln file that is just there [for my VS convenience and in case helpful for eval]
3. squash my intermediate commits
 
hopefully the time & effort I expended [as challenge to me!] will yield worthy update to the MS docs
**good hunting !**

### DIFF
--- a/snippets/csharp/Microsoft.Extensions.Configuration/ConfigurationBuilder/Overview/AppWithOptions.csproj
+++ b/snippets/csharp/Microsoft.Extensions.Configuration/ConfigurationBuilder/Overview/AppWithOptions.csproj
@@ -2,14 +2,14 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Configuration" Version="6.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="8.0.0" />
   </ItemGroup>
 
   <ItemGroup>


DickBaker:
when I committed PR #10113 I got a "**Run snippets 5000 for PR**" warning due to file(s) without encompassing .csproj although the PR seemed to have gone through to review stage.

Anyway this iteration is as fix [that step now passes]. Please accept & apply. Thanks.

snippets/csharp/System.Linq/.editorconfig
```diff
@@ -0,0 +1,269 @@
+# Remove the line below if you want to inherit .editorconfig settings from higher directories
```
eiriktsarpalis, line 1: I defer to the docs owners for this, but if editorconfig settings are specified shouldn't they be set repo-wide?

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -4,19 +4,16 @@
...
 {
-    class Program
+    static class Program
     {
         // This part is just for testing the examples
-        static void Main(string[] args)
-        {
-            OfTypeEx1();
-        }
+        static void Main() => OfTypeEx1();
```
eiriktsarpalis, line 10: If we're embracing modern C# shouldn't this be changed to use top-level statements instead?

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -4,19 +4,16 @@
 
 namespace SequenceExamples
```
eiriktsarpalis, line 5: And possibly get rid of the namespace declarations where possible.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -109,32 +106,29 @@ public static void AllEx()
         static class All2
         {
             // <Snippet129>
-            class Pet
+            class Pet(string name, int age)
```
eiriktsarpalis, line 109: Why specifically adopt a primary constructor in this example? The example could simply update the nullability annotation for `Name` or enforce immutability using `init` (and possibly `required`) instead of `set` if that's the end goal.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -176,36 +170,34 @@ public static void AnyEx1()
...
+                    new ("Antebi", []),
+                    new ("Philips", [
+                        new ("Sweetie",  2),
+                        new ("Rover",  13) ] )
+                    ];
 
                 // Determine which people have a non-empty Pet array.
                 IEnumerable<string> names = from person in people
-                                            where person.Pets.Any()
+                                            where person.Pets.Length != 0
```
eiriktsarpalis, line 200: Given that these are snippets specifically intended to showcase use of Linq why not just keep `Any`?

snippets/csharp/System.Linq/Linq - Copy.txt
```diff
@@ -0,0 +1,61 @@
+﻿
```
eiriktsarpalis, line 1: Seems to have been committed by mistake?

snippets/csharp/System.Linq/IOrderedEnumerableTElement/Overview/Project.csproj
```diff
@@ -1,8 +0,0 @@
-<Project Sdk="Microsoft.NET.Sdk">
-
-  <PropertyGroup>
```
eiriktsarpalis, line 3: Curious why these project files are being deleted/renamed. If the projects are being restructured, what's the end goal here?

eiriktsarpalis (commented):
Apart from a few nits this seems to be bringing welcome updates to the samples.

snippets/csharp/System.Linq/IOrderedEnumerableTElement/Overview/Project.csproj
DickBaker: the MS process mandates that there are .csproj files so that CI can validate. All good.
However it makes sense to knit *.csproj into a VS2022 .sln file so that all the analyzer improves can be applied together.
Having multiple x.csproj files (e.g. program.csproj) is not acceptable to VS2022 so I renamed to _unique_.csproj

I agree with Eirik's other points of about top-level statements, NS (don't know how I missed those!)
But I lean towards analyzer recommendations [e.g. prefer .Count>0 or .Length>0 rather than Any() for perf]
having a splurge of Warnings/Messages is untidy and imho bad UX.

I was not trying to sneak in contrary .editoconfig settings for long-term, but I despise "private" accessors as this is default and imho just contributes unnecessary noise to detriment of the code, i.e. I pick
dotnet_style_require_accessibility_modifiers = omit_if_default
but once I have cleaned-up such noise I am happy to revert to owner's version ["private" shouldn't re-appear]

I set the "let owner/admin modify" [sorry I can't remember exact phrase], so by all means give the changes another spin
I will watch reviewer comments with interest!

I appreciate the effort to update the doc snippets to portray latest C# semantics is a bore, but we all want such docs to give good guidance. Sometimes some ugly example throws me into digging up all the rabbit-holes [OCD!]

Anyway I am pleased the PR is now being looked at as otherwise my wasted effort crafting it!

snippets/csharp/System.Linq/.editorconfig
BillWagner: This file should be removed from the PR. The coding style should match the settings from the editorconfig at the root of the repo.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
BillWagner: Our style would be to use file scoped namespaces for larger samples. Using the global namespace is fine for smaller samples. (I'd call this a larger sample, personally).

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
BillWagner: Yes. We'd prefer top-level statements.

We don't enforce the change on PRs, because we have many samples that pre-date top-level statements. However, if the purpose of this PR is to modernize the code, it should be top-level statements.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
BillWagner: I'd prefer `Any` here as well.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -338,9 +331,9 @@ static void AverageEx2()
         static void AverageEx3()
         {
             // <Snippet16>
-            string[] numbers = { "10007", "37", "299846234235" };
+            string[] numbers = ["10007", "37", "299846234235"];
 
-            double average = numbers.Average(num => long.Parse(num));
+            double average = numbers.Average(long.Parse);
```
BillWagner, line 336: I'm ambivalent on this change. Less experienced programmers may be confused by using a method group instead of a lambda expression. If we keep this change, we should add a note in the corresponding text to explain a method group conversion.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -682,13 +677,11 @@ static void DistinctEx1()
...
-                { "Hartono, Tommy", "Adams, Terry", "Andersen, Henriette Thaulow",
-
-                    "Hedlund, Magnus", "Ito, Shu" };
-            Random random = new Random(DateTime.Now.Millisecond);
+            string[] names = [ "Hartono, Tommy", "Adams, Terry", "Andersen, Henriette Thaulow",
+                    "Hedlund, Magnus", "Ito, Shu" ];
+            Random random = new(DateTime.Now.Millisecond);
 
-            string name = names.ElementAt(random.Next(0, names.Length));
+            string name = names[random.Next(0, names.Length)];
```
BillWagner, line 684: Here again, I'd revert this change. The *purpose* of the snippet is to demonstrate the `ElementAt` method. Removing it defeats the purpose.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -705,18 +698,18 @@ static void ElementAtEx1()
...
 
-            string name = names.ElementAtOrDefault(index);
+            var name = names.ElementAtOrDefault(index);
+            //var name = names[index];
 
             Console.WriteLine(
                 "The name chosen at index {0} is '{1}'.",
                 index,
-                String.IsNullOrEmpty(name) ? "<no name at this index>" : name);
+                name ?? "<no name at this index>");
```
BillWagner, line 712: You should add a comment that the `??` operator is semantically different than `IsNullOrEmpty`

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -803,10 +799,10 @@ static void ExceptEx1()
         static void FirstEx1()
         {
             // <Snippet35>
-            int[] numbers = { 9, 34, 65, 92, 87, 435, 3, 54,
-                                83, 23, 87, 435, 67, 12, 19 };
+            int[] numbers = [ 9, 34, 65, 92, 87, 435, 3, 54,
+                                83, 23, 87, 435, 67, 12, 19 ];
 
-            int first = numbers.First();
+            int first = numbers[0];
```
BillWagner, line 805: Again, the purpose of this snippet is to demonstrate the `First` extension method. Changing to an index operator defeats the purpose of the sample.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -856,19 +852,18 @@ static void FirstOrDefaultEx1()
...
             Console.WriteLine("The first long name is '{0}'.", firstLongName);
 
-            string firstVeryLongName = names.FirstOrDefault(name => name.Length > 30);
+            var firstVeryLongName = names.FirstOrDefault(name => name.Length > 30);
 
-            Console.WriteLine(
-                "There is {0} name longer than 30 characters.",
-                string.IsNullOrEmpty(firstVeryLongName) ? "not a" : "a");
+            Console.WriteLine("There is {0} name longer than 30 characters.",
+                firstVeryLongName is null ? "not a" : "a");
```
BillWagner, line 866: Same comment on the semantics of `??` vs. `IsNullOrEmpty`

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -1053,27 +1052,27 @@ public static void GroupByEx3()
         static class GroupBy4
         {
             // <Snippet125>
-            class Pet
+            class Pet(string name, float age)
             {
-                public string Name { get; set; }
-                public double Age { get; set; }
+                public string Name { get; set; } = name;
+                public float Age { get; set; } = age;
```
BillWagner, line 1058: This feels meaningless. Double is the default floating point type. I don't see a reason to specify `float`.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -1346,10 +1348,9 @@ static void LastEx2()
         static void LastOrDefaultEx1()
         {
             // <Snippet45>
-            string[] fruits = { };
-            string last = fruits.LastOrDefault();
-            Console.WriteLine(
-                String.IsNullOrEmpty(last) ? "<string is null or empty>" : last);
+            string[] fruits = [];
+            var last = fruits.LastOrDefault();
+            Console.WriteLine(last ?? "<string is null or empty>");
```
BillWagner, line 1353: same semantic change.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -1414,10 +1415,10 @@ static void LastOrDefaultEx3()
         public static void LongCountEx1()
         {
             // <Snippet47>
-            string[] fruits = { "apple", "banana", "mango",
-                                  "orange", "passionfruit", "grape" };
+            string[] fruits = [ "apple", "banana", "mango",
+                                  "orange", "passionfruit", "grape" ];
 
-            long count = fruits.LongCount();
+            long count = fruits.Length;
```
BillWagner, line 1421: And again, the purpose of this sample is to demonstrate the `LongCount` extension method. Replacing it with the `Length` property defeats the purpose.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -2062,12 +2044,12 @@ static void SelectEx1()
...
             // <Snippet76>
-            string[] fruits = { "apple", "banana", "mango", "orange",
-                                  "passionfruit", "grape" };
+            string[] fruits = [ "apple", "banana", "mango", "orange",
+                                  "passionfruit", "grape" ];
 
             var query =
                 fruits.Select((fruit, index) =>
-                                  new { index, str = fruit.Substring(0, index) });
+                                  new { index, str = fruit[..index] });
```
BillWagner, line 2052: Great use of collection expressions.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -3091,10 +3041,10 @@ static void AppendEx()
         {
             // <Snippet201>
             // Creating a list of numbers
-            List<int> numbers = new List<int> { 1, 2, 3, 4 };
+            List<int> numbers = [1, 2, 3, 4];
 
             // Trying to append any value of the same type
-            numbers.Append(5);
+            _ = numbers.Append(5);
```
BillWagner, line 3047: This is an interesting change, and could benefit from an explanation:

The original code "looked like" calling a `void` return method that modified the receiver. That's a reasonably common mistake. Assigning the result to a discard highlights the mistake. That minimizes the impact of the sample as a learning tool.

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -3103,7 +3053,7 @@ static void AppendEx()
             Console.WriteLine(string.Join(", ", numbers.Append(5)));
 
             // If you prefer, you can create a new list explicitly
-            List<int> newNumbers = numbers.Append(5).ToList();
+            List<int> newNumbers = [.. numbers, 5];
```
BillWagner, line 3056: There are a lot of positive impacts of collection expressions, but the purpose of this snippet is to demonstrate the `Append` method. Maybe something like this:

```suggestion
            List<int> newNumbers = [.. numbers.Append(5)];

```

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
```diff
@@ -3122,10 +3072,10 @@ static void PrependEx()
         {
             // <Snippet202>
             // Creating a list of numbers
-            List<int> numbers = new List<int> { 1, 2, 3, 4 };
+            List<int> numbers = [1, 2, 3, 4];
 
             // Trying to prepend any value of the same type
-            numbers.Prepend(0);
+            _ = numbers.Prepend(0);
```
BillWagner, line 3078: Same comment as above on discard.

snippets/csharp/System.Linq/Enumerable/Comparers/Comparers.csproj
```diff
@@ -1,9 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net6.0;net472</TargetFrameworks>
-    <StartupObject>Program</StartupObject>
+    <TargetFrameworks>net8.0;net472</TargetFrameworks>
```
BillWagner, line 5: Pinging @gewarren to be sure, but I think our framework samples are using net48 now.

snippets/csharp/System.Linq/Enumerable/Comparers/CustomComparer.cs
```diff
@@ -190,6 +203,6 @@ lemon 12
 
         //</SequenceEqual>
 
-        Console.ReadLine();
+        _ = Console.ReadLine();
```
BillWagner, line 206: We've been removing the `Console.ReadLine()` at the end of most samples. That makes it easier to move to the Try.NET experience.

snippets/csharp/System.Linq/Enumerable/Comparers/EncapsulatedComparer.cs
```diff
@@ -136,13 +151,15 @@ lemon 12
...
+            new ("orange", 4 ) ];
 
-        ProductA[] storeB = { new ProductA { Name = "apple", Code = 9 },
-                               new ProductA { Name = "orange", Code = 4 } };
+        ProductA[] storeB = [
+            new ("apple", 9 ),
+            new ("orange", 4 ) ];
 
-        bool equalAB = storeA.SequenceEqual(storeB);
+        var equalAB = storeA.SequenceEqual(storeB);
```
BillWagner, line 162: Here, I'd prefer an explicit `bool` type. Without a rollover and intellisense, it's better to show the return type in text.

snippets/csharp/System.Linq/ILookupTKey,TElement/Overview/ILookup.cs
```diff
@@ -1,33 +1,30 @@
...
 
 namespace SequenceExamples
 {
     static class ILookup
     {
-        static void Main(string[] args)
-        {
-            ILookupExample();
-        }
+        static void Main() => ILookupExample();
```
BillWagner, line 9: This whole sample would look much cleaner with top-level statements and the global namespace.

snippets/csharp/System.Linq/LookupTKey,TElement/Overview/lookup.cs
```diff
@@ -7,27 +7,25 @@ namespace SequenceExamples
 {
     static class IGrouping
     {
-        static void Main(string[] args)
-        {
-            LookupExample();
-        }
+        static void Main() => LookupExample();
```
BillWagner, line 10: This one is also a textbook example for top level statements.

snippets/csharp/System.Linq/Queryable/AggregateTSource/queryable.cs
```diff
@@ -165,15 +165,15 @@ where person.Pets.AsQueryable().All(pet => pet.Age > 5)
...
         {
             public static void AnyEx1()
             {
                 // <Snippet5>
-                List<int> numbers = new List<int> { 1, 2 };
+                List<int> numbers = [1, 2];
 
                 // Determine if the list contains any elements.
-                bool hasElements = numbers.AsQueryable().Any();
+                bool hasElements = numbers.Any();       // [CA1860] better to use "numbers.Count > 0;" !
```
BillWagner, line 176: Delete the comment. That's only because the compiler knows the source is a `List<T>`:

```suggestion
                bool hasElements = numbers.Any();
```

snippets/csharp/System.Linq/Queryable/AggregateTSource/queryable.cs
```diff
@@ -186,40 +186,40 @@ public static void AnyEx1()
...
+                    new ("Antebi",[]),
+                    new ("Philips", [
+                        new ("Sweetie", 2),
+                        new ("Rover", 13) ] )
+                    ];
 
                 // Determine which people have a non-empty Pet array.
                 IEnumerable<string> names = from person in people
-                                            where person.Pets.AsQueryable().Any()
+                                            where person.Pets.Any() // [CA1860] better to use "where person.Pets.Length != 0"
```
BillWagner, line 216: Same:

```suggestion
                                            where person.Pets.Any()
```

snippets/csharp/System.Linq/Queryable/AggregateTSource/queryable.cs
```diff
@@ -524,10 +530,10 @@ static void ContainsEx1()
         static void CountEx1()
         {
             // <Snippet22>
-            string[] fruits = { "apple", "banana", "mango",
-                                "orange", "passionfruit", "grape" };
+            string[] fruits = [ "apple", "banana", "mango",
+                                "orange", "passionfruit", "grape" ];
 
-            int numberOfFruits = fruits.AsQueryable().Count();
+            int numberOfFruits = fruits.Count();    // [CA1829, RCS1077] better to use "fruits.Length;"
```
BillWagner, line 536: I'll stop commenting after this instance, but remove all these comments:

```suggestion
            int numberOfFruits = fruits.Count();  
```

snippets/csharp/System.Linq/Queryable/AggregateTSource/queryable.cs
```diff
@@ -695,14 +698,15 @@ static void DistinctEx1()
...
-                               "Hedlund, Magnus", "Ito, Shu" };
+                               "Hedlund, Magnus", "Ito, Shu" ];
 
-            Random random = new Random(DateTime.Now.Millisecond);
+            Random random = new(DateTime.Now.Millisecond);
 
             string name =
-                names.AsQueryable().ElementAt(random.Next(0, names.Length));
+                names.ElementAt(random.Next(0, names.Length));
+            //names[random.Next(0, names.Length) ];      // simpler
```
BillWagner, line 709: Remove this. Yes, it's simpler, but only because the source is an array (or any other type with an index operator)

snippets/csharp/System.Linq/Queryable/AggregateTSource/queryable.cs
```diff
@@ -1191,14 +1203,16 @@ public static void GroupJoinEx1()
...
         {
             // <Snippet41>
-            int[] id1 = { 44, 26, 92, 30, 71, 38 };
-            int[] id2 = { 39, 59, 83, 47, 26, 4, 30 };
+            int[] id1 = [44, 26, 92, 30, 71, 38];
+            int[] id2 = [39, 59, 83, 47, 26, 4, 30];
 
             // Get the numbers that occur in both arrays (id1 and id2).
-            IEnumerable<int> both = id1.AsQueryable().Intersect(id2);
+            IEnumerable<int> both = id1.Intersect(id2);
```
BillWagner, line 1210: Revert this. These samples are using the `Queryable` implementation intentionally.

snippets/csharp/System.Linq/Queryable/AggregateTSource/queryable.cs
```diff
@@ -1276,10 +1290,11 @@ public static void JoinEx1()
...
         {
             // <Snippet43>
-            int[] numbers = { 9, 34, 65, 92, 87, 435, 3, 54,
-                                83, 23, 87, 67, 12, 19 };
+            int[] numbers = [ 9, 34, 65, 92, 87, 435, 3, 54,
+                                83, 23, 87, 67, 12, 19 ];
 
-            int last = numbers.AsQueryable().Last();
+            int last = numbers.Last();
+            //int last = numbers[^1];       // simpler
```
BillWagner, line 1298: Revert this as well.

snippets/csharp/System.Linq/Queryable/AggregateTSource/queryable.cs
```diff
@@ -1295,11 +1310,11 @@ static void LastEx1()
...
         {
             // <Snippet44>
-            int[] numbers = { 9, 34, 65, 92, 87, 435, 3, 54,
-                                83, 23, 87, 67, 12, 19 };
+            int[] numbers = [ 9, 34, 65, 92, 87, 435, 3, 54,
+                                83, 23, 87, 67, 12, 19 ];
 
             // Get the last number in the array that is greater than 80.
-            int last = numbers.AsQueryable().Last(num => num > 80);
+            int last = numbers.Last(num => num > 80);
```
BillWagner, line 1317: And again.

snippets/csharp/System.Linq/Queryable/AggregateTSource/queryable.cs
```diff
@@ -1318,14 +1333,14 @@ static void LastOrDefaultEx1()
         {
             // <Snippet45>
             // Create an empty array.
-            string[] fruits = { };
+            string[] fruits = [];
 
             // Get the last item in the array, or else the default
             // value for type string (null).
-            string last = fruits.AsQueryable().LastOrDefault();
+            var last = fruits.LastOrDefault();
```
BillWagner, line 1340: same

snippets/csharp/System.Linq/Queryable/AggregateTSource/queryable.cs
```diff
@@ -1318,14 +1333,14 @@ static void LastOrDefaultEx1()
...
 
             Console.WriteLine(
-                String.IsNullOrEmpty(last) ? "[STRING IS NULL OR EMPTY]" : last);
+                last ?? "[STRING IS NULL OR EMPTY]");
```
BillWagner, line 1343: Same comment I had on the enumerable sample. This does change the semantics of the sample.

snippets/csharp/System.Linq/Queryable/AggregateTSource/queryable.cs
```diff
@@ -1339,19 +1354,21 @@ static void LastOrDefaultEx1()
...
+                    .LastOrDefault(n => Math.Round(n) == 50.0);
 
             Console.WriteLine("The last number that rounds to 50 is {0}.", last50);
 
             // Get the last number in the array that rounds to 40.0,
             // or else the default value for type double (0.0).
             double last40 =
-                numbers.AsQueryable().LastOrDefault(n => Math.Round(n) == 40.0);
+                numbers
+                    .LastOrDefault(n => Math.Round(n) == 40.0);
```
BillWagner, line 1371: I'll stop commenting on it, but every sample in this file is to show a data source that uses `IQueryable`. The `AsQueryable()` calls should stay.

snippets/csharp/System.Linq/Queryable/AggregateTSource/queryable.cs
```diff
@@ -1398,10 +1415,10 @@ static void LastOrDefaultEx3()
         public static void LongCountEx1()
         {
             // <Snippet47>
-            string[] fruits = { "apple", "banana", "mango",
-                                  "orange", "passionfruit", "grape" };
+            string[] fruits = [ "apple", "banana", "mango",
+                                  "orange", "passionfruit", "grape" ];
 
-            long count = fruits.AsQueryable().LongCount();
+            long count = fruits.LongCount();        // [CA1829, RCS1077] better to use "fruits.Length;"
```
BillWagner, line 1421: I'll stop commenting on this point, but all these should be removed when the purpose is to show the given extension method:

```suggestion
            long count = fruits.LongCount(); 
```

snippets/csharp/System.Linq/Enumerable/AggregateTSource/enumerable.cs
BillWagner: I would prefer `record` or `record struct` types for this and other data types defined in these samples. The use (a data container) is the main scenario for `record` types. In the context of the sample, I don't have a strong preference for `record class` vs. `record struct` or enforcing immutability.

BillWagner (changes requested):
Hi @DickBaker 

There are some great improvements here. But also a number of changes that should be addressed.

snippets/csharp/System.Linq/Queryable/AggregateTSource/queryable.cs
DickBaker: not a fan as adds unnecessary overhead, and imho is bad guidance
but I will revert as-was to stay friends.

Yes I also prefer the lightweight record rather than class, so I will magic that in too.

PR #10118, opened 2024-07-17
Update Path.GetTempPath description

adamsitnik:
Recently we have backported https://github.com/dotnet/runtime/pull/72452 to **all** supported versions of .NET and .NET Framework.

fixes https://github.com/dotnet/runtime/issues/105012

### DIFF
--- a/xml/System.IO/Path.xml
+++ b/xml/System.IO/Path.xml
@@ -2300,9 +2300,9 @@ The method does not verify that the path exists or test to see if the current pr
 
 # [Windows](#tab/windows)
 
-If the current Windows version exposes the [`GetTempPath2`](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w) Win32 API, this method invokes that API and returns the resolved path. See [the _Remarks_ section](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w#remarks) of the GetTempPath2 documentation for more information on how this resolution is performed, including how to control the return value through the use of environment variables. The GetTempPath2 API may not be available on all versions of Windows.
+If the current Windows version exposes the [`GetTempPath2`](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w) Win32 API, this method invokes that API and returns the resolved path. For more information on how this resolution is performed, including how to control the return value through the use of environment variables, see [the _Remarks_ section](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w#remarks) of the GetTempPath2 documentation. The GetTempPath2 API might not be available on all versions of Windows.
 
-On Windows versions which do not expose GetTempPath2, this method instead invokes the [`GetTempPath`](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppathw) Win32 API and returns the resolved path. See [the _Remarks_ section](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppathw#remarks) of the GetTempPath documentation for more information on how this resolution is performed, including how to control the return value through the use of environment variables.
+On Windows versions that don't expose GetTempPath2, this method instead invokes the [`GetTempPath`](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppathw) Win32 API and returns the resolved path. For more information on how this resolution is performed, including how to control the return value through the use of environment variables, see [the _Remarks_ section](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppathw#remarks) of the GetTempPath documentation.
 
 # [Linux](#tab/linux)
 


xml/System.IO/Path.xml
```diff
@@ -2310,7 +2310,7 @@ The method does not verify that the path exists or test to see if the current pr
 
 4.  The Windows directory.
 
-In .NET 7 and later, when the process calling the method is running as `SYSTEM`, the method skips the preceding sequence and returns `C:\Windows\SystemTemp`. This location is inaccessible to processes not running as `SYSTEM`.
+On Windows 10 and Server Build 20348 and later, when the process calling the method is running as `SYSTEM`, the method skips the preceding sequence and returns `C:\Windows\SystemTemp`. This location is inaccessible to processes not running as `SYSTEM`.
```
Mafii, line 2313: Also mentioning .NET 7 and newer, as well as .NET 6 starting with 6.0.32 would be helpful in my opinion, as otherwise it is unclear that in older OS Versions those runtimes also behave that way.

adamsitnik: @gewarren what is the recommended approach for situations where we have change the behavior of the product starting in specific version? Should we list all affected versions?

teo-tsirpanis: Does .NET Framework provide an opt-out switch for compatibility?

gewarren:
```suggestion
In .NET 7 and later versions (or .NET 6 version 6.0.32 and later), OR if your operating system is Windows 10 or Windows Server build 20348 or later and you're using any .NET version, when the process calling the method is running as `SYSTEM`, the method skips the preceding sequence and returns `C:\Windows\SystemTemp`. This location is inaccessible to processes not running as `SYSTEM`.
```

Does this convey the message? I think it's helpful to be as explicit as possible.

Mafii: Yes, this is very precise, thank you!

Mafii: Maybe rewording it to 

> and you're using any .NET **Framework** version

would be more accurate, as using 6.0.31 on newer (or any) Windows builds does not contain the security fix.

Jabe: > Does .NET Framework provide an opt-out switch for compatibility?

On 4.7/4.8 it evals an environment variable called "COMPlus_Disable_GetTempPath2" to revert to the old behavior if set to "true".

xml/System.IO/Path.xml
```diff
@@ -2310,7 +2310,7 @@ The method does not verify that the path exists or test to see if the current pr
 
 4.  The Windows directory.
 
-In .NET 7 and later, when the process calling the method is running as `SYSTEM`, the method skips the preceding sequence and returns `C:\Windows\SystemTemp`. This location is inaccessible to processes not running as `SYSTEM`.
+In .NET 7 and later versions (or .NET 6 version 6.0.32 and later), OR if your operating system is Windows 10 or Windows Server build 20348 or later and you're using any .NET Framework version, when the process calling the method is running as `SYSTEM`, the method skips the preceding sequence and returns `C:\Windows\SystemTemp`. This location is inaccessible to processes not running as `SYSTEM`.
```
adamsitnik, line 2313:
```suggestion
In .NET 7 and later versions (or .NET 6 version 6.0.32 and later), OR if your operating system is Windows 10 or Windows Server build 20348 or later and you're using any .NET Framework version, when the process calling the method is running as `SYSTEM`, the method skips the preceding sequence and returns `C:\Windows\SystemTemp`. This location is inaccessible to processes not running as `SYSTEM`.

On Full .NET Framework, this behavior can be opted out by setting `COMPlus_Disable_GetTempPath2` environment variable to `true` (all other values are ignored).
```

xml/System.IO/Path.xml
```diff
@@ -2310,7 +2310,9 @@ The method does not verify that the path exists or test to see if the current pr
...
+
+On Full .NET Framework, this behavior can be opted out by setting `COMPlus_Disable_GetTempPath2` environment variable to `true` (all other values are ignored).
```
teo-tsirpanis, line 2315:
```suggestion
In .NET 6 version 6.0.32 and later versions or in .NET Framework, if your operating system is Windows 11 or Windows Server build 20348 or later, when the process calling the method is running as `SYSTEM`, the method skips the preceding sequence and returns `C:\Windows\SystemTemp`. This location is inaccessible to processes not running as `SYSTEM`.

On .NET Framework, this behavior can be opted out by setting the `COMPlus_Disable_GetTempPath2` environment variable to `true` (all other values are ignored).
```

teo-tsirpanis: Updated.

MSDN-WhiteKnight: It should be "and later" instead of plus notation to follow conventions (like it's already used here: https://learn.microsoft.com/en-us/dotnet/api/system.environment.osversion?view=net-8.0#remarks).

GrabYourPitchforks:
> the method skips the preceding sequence and returns `C:\Windows\SystemTemp`

@adamsitnik Just to be pedantic, there are some cases where `GetTempPath2` will return a directory other than `%WINDIR%\SystemTemp`. It might be worthwhile for us to link directly to https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w rather than try to keep our docs up to date with them. This also makes us more future proof in case the OS updates the behavior going forward.

> if your operating system is Windows 11 or Windows Server build 20348 or later

AFAIK we don't perform an OS version check. We simply look for the existence of the export. This is another case where we should link to the authoritative documentation for the underlying Win32 function rather than trying to duplicate it ourselves (and risk getting out of sync if they make a change regarding available versions).

xml/System.IO/Path.xml
```diff
@@ -2310,7 +2310,9 @@ The method does not verify that the path exists or test to see if the current pr
 
 4.  The Windows directory.
 
-In .NET 7 and later, when the process calling the method is running as `SYSTEM`, the method skips the preceding sequence and returns `C:\Windows\SystemTemp`. This location is inaccessible to processes not running as `SYSTEM`.
+In .NET Framework and .NET 6.0.32+, if your operating system is Windows 11 or Windows Server build 20348 or later, when the process calling the method is running as `SYSTEM`, the method skips the preceding sequence and returns `C:\Windows\SystemTemp`. This location is inaccessible to processes not running as `SYSTEM`.
+
+On .NET Framework, this behavior can be opted out by setting the `COMPlus_Disable_GetTempPath2` environment variable to `true` (all other values are ignored).
```
gewarren, line 2315:
```suggestion
On .NET Framework, you can opt out of this behavior by setting the `COMPlus_Disable_GetTempPath2` environment variable to `true` (all other values are ignored).
```

Mafii:
> AFAIK we don't perform an OS version check.

That is true, but the change in .NET Framework (4.8.*) was not included in earlier builds of Windows, even if the GetTempPath2 API was available, as Windows ships .NET Framework itself.

For example, my co-worker has a Windows 11, Build 22631.3737, and the behaviour of C# applications targeted at `net48` is different on his machine than on mine, which is Windows 11, Build 22631.3880. So just mentioning the presence of the GetTempPath2 API is not sufficient for .NET Framework (it is for .NET 6.0.32+...).

GrabYourPitchforks:
Feedback from netfx servicing:

- Do not include the compat switch in the Learn docs. Documenting it gives the false impression that the team is committed to keeping the switch around in its current state indefinitely. Additionally, ownership of this rests with CSS, and we should allow them to be in charge of their own communication channel. (Typically CSS publishes these switches in a KB article if needed, not in the API docs.)
- We should update the docs to reflect all frameworks this change is applicable to. But it's not up to API docs to say _when_ the change came in, because we should optimize for the world where the change rolls out in rapid fashion and almost everybody coming to these docs already has the changes pulled down.
- We should remove from the docs anything that we do not implement ourselves. For example, since we do not read env vars, we should not document them. Instead, link to an authoritative source of how our dependencies operate and make that doc team responsible for keeping the documentation up to date.

I'll send shortly a new iteration with includes these changes and merges Genevieve's recommendations.

xml/System.IO/Path.xml
```diff
@@ -2298,25 +2298,17 @@ On .NET 7 and earlier versions, when using this method on Windows, the <xref:Sys
...
 # [Windows](#tab/windows)
 
-1.  The path specified by the TMP environment variable.
-
-2.  The path specified by the TEMP environment variable.
-
-3.  The path specified by the USERPROFILE environment variable.
-
-4.  The Windows directory.
+If the current Windows version exposes the [`GetTempPath2`](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w) Win32 API, this method invokes that API and returns the resolved path. See [the _Remarks_ section](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w#remarks) of the GetTempPath2 documentation for more information on how this resolution is performed, including how to control the return value through the use of environment variables. The GetTempPath2 API may not be available on all versions of Windows.
```
gewarren, line 2303:
```suggestion
If the current Windows version exposes the [`GetTempPath2`](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w) Win32 API, this method invokes that API and returns the resolved path. For more information on how this resolution is performed, including how to control the return value through the use of environment variables, see [the _Remarks_ section](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppath2w#remarks) of the GetTempPath2 documentation. The GetTempPath2 API might not be available on all versions of Windows.
```

xml/System.IO/Path.xml
```diff
@@ -2298,25 +2298,17 @@ On .NET 7 and earlier versions, when using this method on Windows, the <xref:Sys
...
 
-In .NET 7 and later, when the process calling the method is running as `SYSTEM`, the method skips the preceding sequence and returns `C:\Windows\SystemTemp`. This location is inaccessible to processes not running as `SYSTEM`.
+On Windows versions which do not expose GetTempPath2, this method instead invokes the [`GetTempPath`](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppathw) Win32 API and returns the resolved path. See [the _Remarks_ section](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppathw#remarks) of the GetTempPath documentation for more information on how this resolution is performed, including how to control the return value through the use of environment variables.
```
gewarren, line 2305:
```suggestion
On Windows versions that don't expose GetTempPath2, this method instead invokes the [`GetTempPath`](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppathw) Win32 API and returns the resolved path. For more information on how this resolution is performed, including how to control the return value through the use of environment variables, see [the _Remarks_ section](https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-gettemppathw#remarks) of the GetTempPath documentation.
```

gewarren (approved):

gewarren:
Preview looks good: https://review.learn.microsoft.com/en-us/dotnet/api/system.io.path.gettemppath?view=net-9.0&branch=pr-en-us-10118&tabs=windows#remarks

adamsitnik:
> I'll send shortly a new iteration with includes these changes and merges Genevieve's recommendations.

Thank you @GrabYourPitchforks !

xml/System.IO/Path.xml
GrabYourPitchforks: Accepted as-is.

GrabYourPitchforks: Outdated. Resolving.

PR #10141, opened 2024-07-23
Update C# examples for ja-JP and en-US DateTimeFormatInfo

matouskozak:
## Summary

In https://github.com/dotnet/runtime/issues/105097, we discovered that the example provided for .NET DateTimePattern are out-of-date. In this PR, we update the examples of `FullDateTimePattern`, `LongDatePattern`, `MonthDayPattern` for `en-US` and `ja-JP` cultures.

---

cc: @tarekgh @vitek-karas

### DIFF
--- a/snippets/csharp/System.Globalization/DateTimeFormatInfo/LongDatePattern/longdatepattern1.cs
+++ b/snippets/csharp/System.Globalization/DateTimeFormatInfo/LongDatePattern/longdatepattern1.cs
@@ -2,7 +2,7 @@
 using System;
 using System.Globalization;
 
-public class Example
+public class Example1
 {
    public static void Main()
    {
--- a/snippets/csharp/System.Globalization/DateTimeFormatInfo/LongDatePattern/longdatepattern2.cs
+++ b/snippets/csharp/System.Globalization/DateTimeFormatInfo/LongDatePattern/longdatepattern2.cs
@@ -3,7 +3,7 @@ using System;
 using System.Globalization;
 using System.IO;
 
-public class Example
+public class Example2
 {
    public static void Main()
    {


matouskozak:
I see `Error: ERROR: Project missing. A project (and optionally a solution file) must be in this directory or one of the parent directories to validate and build this code.` failure in the Snippets 5000 job. Am I missing something?

snippets/csharp/System.Globalization/DateTimeFormatInfo/FullDateTimePattern/dtfi_fulldatetimepattern.cs
```diff
@@ -26,8 +26,8 @@ public class SamplesDTFI  {
 This code produces the following output.  The question marks take the place of native script characters.
```
tarekgh, line 26: Will be good to change the comment telling that this output is example which can be different depending on the used OS, the OS version, and the native globalization library used in such OS.

tarekgh:
@gewarren do you know why the CI is failing while updating only existing sample source?

gewarren:
> @gewarren do you know why the CI is failing while updating only existing sample source?

Yes, we are trying to get project files added for every snippet, so if a project file doesn't exist for any snippet that's touched, it errors.

tarekgh:
> we are trying to get project files added for every snippet, so if a project file doesn't exist for any snippet that's touched, it errors.

Should we ignore the error? or we must add the project file with the change?

snippets/csharp/System.Globalization/DateTimeFormatInfo/FullDateTimePattern/dtfi_fulldatetimepattern.cs
matouskozak: Good point! Something like: `This code produces the following output. Note that the exact output format depends on the OS, the OS version, and the native globalization library used by the OS.` ?

matouskozak:
> > > we are trying to get project files added for every snippet, so if a project file doesn't exist for any snippet that's touched, it errors.
> > 
> > 
> > Should we ignore the error? or we must add the project file with the change?
> 
> If Matous can add one, that would be great. Something like this, but might as well target .NET 8: https://github.com/dotnet/dotnet-api-docs/blob/main/snippets/csharp/System.Globalization/DateTimeFormatInfo/FullDateTimePattern/Project.csproj

I added the Project.csproj, however now it complains that `D:\a\dotnet-api-docs\dotnet-api-docs\snippets\csharp\System.Globalization\DateTimeFormatInfo\LongDatePattern\longdatepattern2.cs(8,23): error CS0111: Type 'Example' already defines a member called 'Main' with the same parameter types [D:\a\dotnet-api-docs\dotnet-api-docs\snippets\csharp\System.Globalization\DateTimeFormatInfo\LongDatePattern\Project.csproj]`. Should I rename the `public class Example` in longdatepattern1/2.cs to `public class Example1/2` or what is the proper fix here?

gewarren:
> Should I rename the public class Example in longdatepattern1/2.cs to public class Example1/2 ...?

Yes, please. Sorry for the hassle.

matouskozak:
> > Should I rename the public class Example in longdatepattern1/2.cs to public class Example1/2 ...?
> 
> Yes, please. Sorry for the hassle.

No problem, thank you for the help. Looks to be green now.

gewarren (approved):

PR #10114, opened 2024-07-16
CI update - .NET 9 Preview 6

gewarren:
Fixes #9987.

View by commit.

PR #10157, opened 2024-07-25
Port System.ServiceModel new docs

carlossanlop:
@HongGit @mconnew @directhex @gewarren PTAL

### DIFF
--- a/xml/System.ServiceModel.Channels/ServiceChannelProxy.xml
+++ b/xml/System.ServiceModel.Channels/ServiceChannelProxy.xml
@@ -99,9 +99,9 @@
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets or sets a value that indicates whether the input session is shutdown automatically.</summary>
+        <summary>Gets or sets a value that indicates whether the input session is shut down automatically.</summary>
         <value>
-          <see langword="true" /> if the input session is shutdown automatically; otherwise, <see langword="false" />.</value>
+          <see langword="true" /> if the input session is shut down automatically; otherwise, <see langword="false" />.</value>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
@@ -838,7 +838,7 @@
       <Docs>
         <param name="callback">The method that is called when this method completes.</param>
         <param name="state">Information about the state of the channel.</param>
-        <summary>An asynchronous call to begin using a user interface to obtain credential information.</summary>
+        <summary>Asynchronously begins using a user interface to obtain credential information.</summary>
         <returns>The <see cref="T:System.IAsyncResult" /> to use to call back when processing has completed.</returns>
         <remarks>To be added.</remarks>
       </Docs>
@@ -896,7 +896,7 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>A call to a user interface to obtain credential information.</summary>
+        <summary>Displays a user interface to obtain credential information.</summary>
         <remarks>This member is an explicit interface member implementation. It can be used only when the &lt;xref:System.ServiceModel.Channels.ServiceChannelProxy&gt; instance is cast to an &lt;xref:System.ServiceModel.IClientChannel&gt; interface.</remarks>
       </Docs>
     </Member>
@@ -927,7 +927,7 @@
       </Parameters>
       <Docs>
         <param name="result">The <see cref="T:System.IAsyncResult" />.</param>
-        <summary>Called when the call to <see cref="M:System.ServiceModel.IClientChannel.BeginDisplayInitializationUI(System.AsyncCallback,System.Object)" /> has finished.</summary>
+        <summary>Ends the initialization user interface. This method is called when the call to <see cref="M:System.ServiceModel.IClientChannel.BeginDisplayInitializationUI(System.AsyncCallback,System.Object)" /> has finished.</summary>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
@@ -1683,7 +1683,7 @@
         <ReturnType>System.String</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Returns an identifier for the current session, if any.</summary>
+        <summary>Gets the identifier for the current session, if any.</summary>
         <value>The current session identifier, if any.</value>
         <remarks>To be added.</remarks>
       </Docs>
--- a/xml/System.ServiceModel.Dispatcher/OperationInvokerBehavior.xml
+++ b/xml/System.ServiceModel.Dispatcher/OperationInvokerBehavior.xml
@@ -78,7 +78,7 @@
       <Docs>
         <param name="description">To be added.</param>
         <param name="parameters">To be added.</param>
-        <summary>Implement to pass data at runtime to bindings to support custom behavior.</summary>
+        <summary>Passes data to bindings to support custom behavior.</summary>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
@@ -175,7 +175,7 @@
       </Parameters>
       <Docs>
         <param name="description">To be added.</param>
-        <summary>Implement to confirm that the operation meets some intended criteria.</summary>
+        <summary>Confirms that the operation meets some intended criteria.</summary>
         <remarks>To be added.</remarks>
       </Docs>
     </Member>
--- a/xml/System.ServiceModel.Dispatcher/SyncMethodInvoker.xml
+++ b/xml/System.ServiceModel.Dispatcher/SyncMethodInvoker.xml
@@ -115,7 +115,7 @@
         <param name="inputs">The inputs to the method.</param>
         <param name="callback">The asynchronous callback object.</param>
         <param name="state">Associated state data.</param>
-        <summary>An asynchronous implementation of the <see cref="M:System.ServiceModel.Dispatcher.IOperationInvoker.Invoke(System.Object,System.Object[],System.Object[]@)" /> method.</summary>
+        <summary>Asynchronously returns an object and a set of output objects from an instance and set of input objects.</summary>
         <returns>A <see cref="T:System.IAsyncResult" /> used to complete the asynchronous call.</returns>
         <remarks>To be added.</remarks>
       </Docs>
@@ -151,7 +151,7 @@
         <param name="instance">The object invoked.</param>
         <param name="outputs">The outputs from the method.</param>
         <param name="result">The <see cref="T:System.IAsyncResult" /> object.</param>
-        <summary>The asynchronous end method.</summary>
+        <summary>Asynchronously completes the operation.</summary>
         <returns>The return value.</returns>
         <remarks>To be added.</remarks>
       </Docs>
--- a/xml/System.ServiceModel.Dispatcher/TaskMethodInvoker.xml
+++ b/xml/System.ServiceModel.Dispatcher/TaskMethodInvoker.xml
@@ -117,7 +117,7 @@
         <param name="inputs">The inputs to the method.</param>
         <param name="callback">The asynchronous callback object.</param>
         <param name="state">Associated state data.</param>
-        <summary>An asynchronous implementation of the <see cref="M:System.ServiceModel.Dispatcher.IOperationInvoker.Invoke(System.Object,System.Object[],System.Object[]@)" /> method.</summary>
+        <summary>Asynchronously returns an object and a set of output objects from an instance and set of input objects.</summary>
         <returns>A <see cref="T:System.IAsyncResult" /> used to complete the asynchronous call.</returns>
         <remarks>To be added.</remarks>
       </Docs>
@@ -153,7 +153,7 @@
         <param name="instance">The object invoked.</param>
         <param name="outputs">The outputs from the method.</param>
         <param name="result">The <see cref="T:System.IAsyncResult" /> object.</param>
-        <summary>The asynchronous end method.</summary>
+        <summary>Asynchronously completes the operation.</summary>
         <returns>The return value.</returns>
         <remarks>To be added.</remarks>
       </Docs>


mconnew:
Something has gone very wrong with the documentation generator here. The class [TransportDuplexSessionChannel](https://github.com/dotnet/wcf/blob/main/src/System.ServiceModel.NetFramingBase/src/System/ServiceModel/Channels/TransportDuplexSessionChannel.cs) is internal, but documentation has been generated for it. There's an error about a cross reference to `System.ServiceModel.Channels.IAsyncDuplexSession` not being found, but that's an internal interface implemented on an internal class.  

It looks like there's some bugs in the generator picking up api's it shouldn't.

mconnew:
I think I see the problem, looking at the preview page, it lists the location of that class as being System.Private.ServiceModel.dll in v4.10.3. It might have been declared as public in that assembly in that version, but that assembly should never have been scanned for public api's as the nuget package isn't intended to be directly referenced. It has an empty ref folder so that if it is referenced, you don't get to compile against any api's.

carlossanlop:
@mcconew the last preview that was ingested into dotnet-api-docs was preview6, and the TransportDuplexSessionChannel type did not get modified there: https://github.com/dotnet/dotnet-api-docs/pull/10114

So it means we have been incorrectly scanning that DLL you mention since way before.

As you said, it must be getting detected as public in System.Private.ServiceModel.dll, because that's how we show it here: https://github.com/dotnet/dotnet-api-docs/blob/d6e3337eaa73627c4c6b09a41618abdd808f900f/xml/System.ServiceModel.Channels/TransportDuplexSessionChannel.xml#L2

@gewarren what can we do to remove anything that came from that DLL?

carlossanlop:
Are we incorrectly including that DLL in the transport packages you use to update the CI here, @gewarren?

mconnew:
I checked, it is public in the private package for 4.10 [as seen here](https://github.com/dotnet/wcf/blob/release/4.10/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportDuplexSessionChannel.cs). That type got moved to the System.ServiceModel.NetFramingBase package as part of removing the old private package, and was made internal as part of the process.

gewarren:
The tool that grabs the NuGet packages looks for any packages published by aspnet or dotnetframework that start with System, Microsoft.Bcl, Microsoft.Extensions, or Microsoft.Win32. I'll add a special case to exclude System.Private.ServiceModel and rerun the CI, so those APIs should go away.

carlossanlop:
Thank you @gewarren!

Let's wait for the CI update so I can rebase this PR and we determine if there are still any comments that need porting.

carlossanlop:
CI update: https://github.com/dotnet/dotnet-api-docs/pull/10164

carlossanlop:
@mconnew the CI update has been merged and this PR has been rebased. Would you mind taking another look?

xml/System.ServiceModel.Channels/ServiceChannelProxy.xml
```diff
@@ -99,8 +99,9 @@
         <ReturnType>System.Boolean</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Gets or sets a value that indicates whether the input session is shutdown automatically.</summary>
```
gewarren, line 102:
```suggestion
        <summary>Gets or sets a value that indicates whether the input session is shut down automatically.</summary>
```

xml/System.ServiceModel.Channels/ServiceChannelProxy.xml
```diff
@@ -99,8 +99,9 @@
...
+        <value>
+          <see langword="true" /> if the input session is shutdown automatically; otherwise, <see langword="false" />.</value>
```
gewarren, line 104:
```suggestion
          <see langword="true" /> if the input session is shut down automatically; otherwise, <see langword="false" />.</value>
```

xml/System.ServiceModel.Channels/ServiceChannelProxy.xml
```diff
@@ -834,10 +836,10 @@
         <Parameter Name="state" Type="System.Object" />
       </Parameters>
       <Docs>
-        <param name="callback">To be added.</param>
-        <param name="state">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="callback">The method that is called when this method completes.</param>
+        <param name="state">Information about the state of the channel.</param>
+        <summary>An asynchronous call to begin using a user interface to obtain credential information.</summary>
```
gewarren, line 841:
```suggestion
        <summary>Asynchronously begins using a user interface to obtain credential information.</summary>
```

xml/System.ServiceModel.Channels/ServiceChannelProxy.xml
```diff
@@ -893,8 +896,8 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>To be added.</summary>
-        <remarks>To be added.</remarks>
+        <summary>A call to a user interface to obtain credential information.</summary>
```
gewarren, line 899:
```suggestion
        <summary>Displays a user interface to obtain credential information.</summary>
```

xml/System.ServiceModel.Channels/ServiceChannelProxy.xml
```diff
@@ -923,8 +926,8 @@
         <Parameter Name="result" Type="System.IAsyncResult" />
       </Parameters>
       <Docs>
-        <param name="result">To be added.</param>
-        <summary>To be added.</summary>
+        <param name="result">The <see cref="T:System.IAsyncResult" />.</param>
+        <summary>Called when the call to <see cref="M:System.ServiceModel.IClientChannel.BeginDisplayInitializationUI(System.AsyncCallback,System.Object)" /> has finished.</summary>
```
gewarren, line 930:
```suggestion
        <summary>Ends the initialization user interface. This method is called when the call to <see cref="M:System.ServiceModel.IClientChannel.BeginDisplayInitializationUI(System.AsyncCallback,System.Object)" /> has finished.</summary>
```

xml/System.ServiceModel.Channels/ServiceChannelProxy.xml
```diff
@@ -1679,8 +1683,8 @@
         <ReturnType>System.String</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>To be added.</summary>
-        <value>To be added.</value>
+        <summary>Returns an identifier for the current session, if any.</summary>
```
gewarren, line 1686:
```suggestion
        <summary>Gets the identifier for the current session, if any.</summary>
```

xml/System.ServiceModel.Dispatcher/OperationInvokerBehavior.xml
```diff
@@ -78,7 +78,7 @@
       <Docs>
         <param name="description">To be added.</param>
         <param name="parameters">To be added.</param>
-        <summary>To be added.</summary>
+        <summary>Implement to pass data at runtime to bindings to support custom behavior.</summary>
```
gewarren, line 81:
```suggestion
        <summary>Passes data to bindings to support custom behavior.</summary>
```

xml/System.ServiceModel.Dispatcher/OperationInvokerBehavior.xml
```diff
@@ -175,7 +175,7 @@
       </Parameters>
       <Docs>
         <param name="description">To be added.</param>
-        <summary>To be added.</summary>
+        <summary>Implement to confirm that the operation meets some intended criteria.</summary>
```
gewarren, line 178:
```suggestion
        <summary>Confirms that the operation meets some intended criteria.</summary>
```

xml/System.ServiceModel.Dispatcher/SyncMethodInvoker.xml
```diff
@@ -111,12 +111,12 @@
...
-        <param name="inputs">To be added.</param>
-        <param name="callback">To be added.</param>
-        <param name="state">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="instance">The object to be invoked.</param>
+        <param name="inputs">The inputs to the method.</param>
+        <param name="callback">The asynchronous callback object.</param>
+        <param name="state">Associated state data.</param>
+        <summary>An asynchronous implementation of the <see cref="M:System.ServiceModel.Dispatcher.IOperationInvoker.Invoke(System.Object,System.Object[],System.Object[]@)" /> method.</summary>
```
gewarren, line 118:
```suggestion
        <summary>Asynchronously returns an object and a set of output objects from an instance and set of input objects.</summary>
```

xml/System.ServiceModel.Dispatcher/SyncMethodInvoker.xml
```diff
@@ -148,11 +148,11 @@
...
       <Docs>
-        <param name="instance">To be added.</param>
-        <param name="outputs">To be added.</param>
-        <param name="result">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="instance">The object invoked.</param>
+        <param name="outputs">The outputs from the method.</param>
+        <param name="result">The <see cref="T:System.IAsyncResult" /> object.</param>
+        <summary>The asynchronous end method.</summary>
```
gewarren, line 154:
```suggestion
        <summary>Asynchronously completes the operation.</summary>
```

xml/System.ServiceModel.Dispatcher/SyncMethodInvoker.xml
```diff
@@ -148,11 +148,11 @@
...
+        <returns>The return value.</returns>
```
gewarren, line 155: ??

xml/System.ServiceModel.Dispatcher/TaskMethodInvoker.xml
```diff
@@ -113,12 +113,12 @@
...
-        <param name="inputs">To be added.</param>
-        <param name="callback">To be added.</param>
-        <param name="state">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="instance">The object to be invoked.</param>
+        <param name="inputs">The inputs to the method.</param>
+        <param name="callback">The asynchronous callback object.</param>
+        <param name="state">Associated state data.</param>
+        <summary>An asynchronous implementation of the <see cref="M:System.ServiceModel.Dispatcher.IOperationInvoker.Invoke(System.Object,System.Object[],System.Object[]@)" /> method.</summary>
```
gewarren, line 120:
```suggestion
        <summary>Asynchronously returns an object and a set of output objects from an instance and set of input objects.</summary>
```

xml/System.ServiceModel.Dispatcher/TaskMethodInvoker.xml
```diff
@@ -150,11 +150,11 @@
...
       <Docs>
-        <param name="instance">To be added.</param>
-        <param name="outputs">To be added.</param>
-        <param name="result">To be added.</param>
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
+        <param name="instance">The object invoked.</param>
+        <param name="outputs">The outputs from the method.</param>
+        <param name="result">The <see cref="T:System.IAsyncResult" /> object.</param>
+        <summary>The asynchronous end method.</summary>
```
gewarren, line 156:
```suggestion
        <summary>Asynchronously completes the operation.</summary>
```

xml/System.ServiceModel.Dispatcher/TaskMethodInvoker.xml
```diff
@@ -150,11 +150,11 @@
...
+        <returns>The return value.</returns>
```
gewarren, line 157: ??

gewarren (approved):

xml/System.ServiceModel.Dispatcher/SyncMethodInvoker.xml
carlossanlop: Lol. Let's just remove it.
```suggestion
        <returns>To be added.</returns>
```

xml/System.ServiceModel.Dispatcher/TaskMethodInvoker.xml
carlossanlop:
```suggestion
        <returns>To be added.</returns>
```

mconnew:
There's still a few problems. SyncMethodInvoker, TaskMethodInvoker, ServiceChannelProxy, and OutputChannel are public in our implementation assembly, but are purposefully absent from our ref assembly.  

I don't fully understand the schema for these files so can't tell if it's correct or not, but `ClientBase<T>` only implements `IAsyncDisposable` on .NET6+.  

I also see another problem, looking at the `CLientBase<T>` details for `System.IAsyncDisposable.DisposeAsync`, under AssemblyInfo, it's listing System.Private.ServiceModel, assembly version 4.10.3.0. That api wasn't in 4.10.3, but if it was, the assembly info should be pointing to System.ServiceModel.Primitives because that's where the ref assembly (with matching name).

gewarren:
> There's still a few problems. SyncMethodInvoker, TaskMethodInvoker, ServiceChannelProxy, and OutputChannel are public in our implementation assembly, but are purposefully absent from our ref assembly.

There's a special case for System.ServiceModel.Primitives where we use the reference assemblies instead of the implementation assemblies - but the behavior you're describing seems to be the opposite of that. Should I remove this special behavior? If I recall correctly, it was necessary for the ingestion logic to avoid hitting an error.

> I don't fully understand the schema for these files so can't tell if it's correct or not, but `ClientBase<T>` only implements `IAsyncDisposable` on .NET6+.

It currently shows as only applicable to .NET 8+ (and .NET Core 1.0/1.1 and .NET Framework) - https://review.learn.microsoft.com/en-us/dotnet/api/system.servicemodel.clientbase-1?view=net-8.0&branch=pr-en-us-10157#applies-to.

> I also see another problem, looking at the `CLientBase<T>` details for `System.IAsyncDisposable.DisposeAsync`, under AssemblyInfo, it's listing System.Private.ServiceModel, assembly version 4.10.3.0. That api wasn't in 4.10.3, but if it was, the assembly info should be pointing to System.ServiceModel.Primitives because that's where the ref assembly (with matching name).

I don't think that info is displayed anywhere, so this might just be some leftover info that wasn't cleaned up when System.Private.ServiceModel was removed.

carlossanlop:
> Should I remove this special behavior?

@mconnew this is a question for you. Do you mind answering it so we can advance this PR?

mconnew:
@gewarren, the reference assemblies are what is supposed to be used. The situation used to be that the implementation was in the System.Private.ServiceModel package in a lib folder. We created a ref folder with a dummy empty file (otherwise nuget doesn't see the folder or removes it, something bad happened anyway) so that if someone added the package, the compiler wouldn't see any of the implementation. Then in the System.ServiceModel.* packages (e.g. Primitives) we had a ref folder which had all the api's. In the lib folder, we had a type forwarding assembly which just redirected the runtime to the System.Private.ServiceModel.dll implementation to find the types. It just had type forwarding attributes and no implementation.  

Now we've got rid of System.Private.Servicmodel. How things are right now is slightly different from how they will be in a few months. Currently every package except Primitives only has an implementation assembly which will then double for being the reference assembly. For Primitives we have a ref and a lib folder. The ref folder contains the reference assembly and has all the public types. In the lib folder, we have the implementation assembly which has extra types which are public. Some are unintentional (but it never mattered before), some are intentional so that the other packages can use them. It avoids having to use InternalsVisibleTo. SyncMethodInvoker and TaskMethodInvoker are unintentionally public but do not appear in the reference assembly.  

I'm currently working on changes where NetTcp and Http will also have reference assemblies. Because we already did the work to make sure nothing was unintentionally public, the public api surface of the reference assembly and the implementation assembly will be the same for .NET. We're bringing back netstandard2.0 which will have some api's excluded as they aren't available on .NET Framework. And we'll have type forwarding facades for .NET Framework.

gewarren:
@mconnew What do you need me to do to resolve the blocker for this PR? Is that remove SyncMethodInvoker, TaskMethodInvoker, ServiceChannelProxy, and OutputChannel?

xml/System.ServiceModel.Dispatcher/SyncMethodInvoker.xml
carlossanlop: Ping @mconnew

xml/System.ServiceModel.Dispatcher/TaskMethodInvoker.xml
carlossanlop: Ping @mconnew

mconnew:
@gewarren, these are just a few types which I noticed. Basically you should be scanning for types in the ref folders first and only looking in the lib folder if there isn't a ref folder. We keep getting inquiries about various types or methods which are documented but aren't available because of this issue. We just had one inquiring about a method on SecurityBindingElement which isn't exposed but is documented as such. A few days ago we had someone inquiring about the enum type InstanceContextMode which isn't exposed by the ref assembly but is documented as available too. I don't want to play whack a mole with types and methods, the correct assembly should be scanned.

gewarren:
> @gewarren, these are just a few types which I noticed. Basically you should be scanning for types in the ref folders first and only looking in the lib folder if there isn't a ref folder.

@huangmin-ms Is this 👆 something that mdoc is capable of?

huangmin-ms:
> > @gewarren, these are just a few types which I noticed. Basically you should be scanning for types in the ref folders first and only looking in the lib folder if there isn't a ref folder.
> 
> @huangmin-ms Is this 👆 something that mdoc is capable of?

@gewarren Thanks for reporting.  We already support this and I can also see the configuration is correct. 

pac230,[tfm=net8.0;includeXml=False;**libpath=ref**]System.ServiceModel.Primitives,8.0.0

However this is impacted by the source link change, we will fix it in https://ceapex.visualstudio.com/Engineering/_workitems/edit/996190

huangmin-ms:
> > @gewarren, these are just a few types which I noticed. Basically you should be scanning for types in the ref folders first and only looking in the lib folder if there isn't a ref folder.
> 
> @huangmin-ms Is this 👆 something that mdoc is capable of?

@gewarren  The fix is done. Please help to review the latest CI update: https://github.com/dotnet/dotnet-api-docs/pull/10385

gewarren:
@mconnew Just to clarify, you want the docs to use the reference assemblies for all System.ServiceModel.* packages?

mconnew:
@gewarren, not exactly. Not all packages have reference assemblies. Currently only Primitives has a reference assembly, but that will be changing at some time in the future. Does the docs tooling require knowing ahead of time whether to look at a ref assembly of the lib?

gewarren:
> @gewarren, not exactly. Not all packages have reference assemblies. Currently only Primitives has a reference assembly, but that will be changing at some time in the future. Does the docs tooling require knowing ahead of time whether to look at a ref assembly of the lib?

Yes. Currently only System.ServiceModel.Primitives is special cased that way.

mconnew:
That works for now, and when we release new package versions with ref assemblies that should be scanned instead, I'll let you know.

carlossanlop:
What should I do about this PR?

carlossanlop:
Closing. Will resubmit after the RC1 update.

PR #10185, opened 2024-07-31
NrbfDecoder.StartsWithPayloadHeader accepts ROS<byte> rather than array now

### LINKED ISSUE #10177: Update System.Formats.Nrbf.NrbfDecoder(byte[]) to accept ReadOnlySpan<byte> instead
I've noticed while reviewing https://github.com/dotnet/dotnet-api-docs/pull/10168 that we need to reflect the changes from https://github.com/dotnet/runtime/pull/103636 in the docs

adamsitnik:
A follow up of https://github.com/dotnet/runtime/pull/103636. The method is no longer accepting an array of bytes, but a readonly span of bytes.

fixes #10177

### DIFF
--- a/xml/System.Formats.Nrbf/NrbfDecoder.xml
+++ b/xml/System.Formats.Nrbf/NrbfDecoder.xml
@@ -150,7 +150,7 @@
       </Docs>
     </Member>
     <Member MemberName="StartsWithPayloadHeader">
-      <MemberSignature Language="C#" Value="public static bool StartsWithPayloadHeader (ReadOnlySpan&lt;byte&gt bytes);" />
+      <MemberSignature Language="C#" Value="public static bool StartsWithPayloadHeader (ReadOnlySpan&lt;byte&gt; bytes);" />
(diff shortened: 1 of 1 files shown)


gewarren:
@adamsitnik If I calculated correctly, the [change went in on June 28](https://github.com/dotnet/runtime/pull/103636), which is after the snap for Preview 6. So the change will be in Preview 7. If that's the case, this will automatically get updated when we update the API docs for Preview 7.

adamsitnik:
> So the change will be in Preview 7. If that's the case, this will automatically get updated when we update the API docs for Preview 7.

Thanks, I was not aware of that. Should I just close the PR?

gewarren:
> Thanks, I was not aware of that. Should I just close the PR?

I think it's worth keeping this part:

```xml
        <param name="bytes">The buffer to inspect.</param>
        <summary>Checks if given buffer starts with <see href="https://learn.microsoft.com/openspecs/windows_protocols/ms-nrbf/a7e578d3-400a-4249-9424-7529d10d1b3c">NRBF payload header</see>.</summary>
        <returns>
          <see langword="true" /> if it starts with NRBF payload header; otherwise, <see langword="false" />.</returns>
```

gewarren:
Docs are already there, so closing this.

PR #10192, opened 2024-08-01
Update Enumerable.xml to remove unexpected ) characters

kendaleiv:
## Summary

Update to remove unexpected `)` characters from "This method can be used to produce a left outer join when it is combined with the GroupJoin) method." instances.


gewarren (approved):
Thanks @kendaleiv

PR #10193, opened 2024-08-01
Update ZipArchive.xml, ZipFile.xml

edwardneal:
## Summary

This PR is an extension of dotnet/runtime#103271, feeding its documentation changes into this repo.

The current documentation for `ZipArchive` and `ZipFile` references the `entryNameEncoding` parameter, describing it as being used to encode/decode ZIP archive entry names. This is incorrect; these two classes actually use this parameter to encode/decode ZIP entry names and comments.

I've also made one change to ZipFile.xml on line 509. This previously said that if entry names and comments contained only ASCII characters, the language encoding flag would be set. This was backwards.


gewarren (approved):
Thanks @edwardneal!

PR #10198, opened 2024-08-02
Open file dialog io exception Fixes #7998

### LINKED ISSUE #7998: Document that OpenFileDialog.OpenFile() throws IOException
`OpenFileDialog.OpenFile()` is a thin wrapper around `new FileStream()`. Thus, it will throw the same exceptions. Unlike `SaveFileDialog.OpenFile()` which behaves differently on mono, this behavior is the same on [.netfx](https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/OpenFileDialog.cs,b9fb6d8411201152), [.netcore](https://github.com/dotnet/winforms/blob/6a3867101ca058010147742789d7ab52b11da0bd/src/System.Windows.Forms/src/System/Windows/Forms/OpenFileDialog.cs#L100), and [mono](https://github.com/mono/mono/blob/main/mcs/class/System.Windows.Forms/System.Windows.Forms/OpenFileDialog.cs#L120).

Please document that `OpenFileDialog.OpenFile()` will throw `IOException` if the file cannot be opened, such as if it was removed or renamed prior to calling `OpenFile()`.

BlueBles:
## Summary

Adding information about the exception that may be thrown if the file is modified/deleted before openFile()

Remarks: It is my first PR to this repo.

Fixes #7998

### DIFF
--- a/xml/System.Windows.Forms/OpenFileDialog.xml
+++ b/xml/System.Windows.Forms/OpenFileDialog.xml
@@ -47,9 +47,7 @@
 On a right-to-left operating system, setting the containing form's <xref:System.Windows.Forms.Control.RightToLeft> property to <xref:System.Windows.Forms.RightToLeft.Yes?displayProperty=nameWithType> localizes the dialog's **File Name**, **Open**, and **Cancel** buttons. If the property is not set to <xref:System.Windows.Forms.RightToLeft.Yes?displayProperty=nameWithType>, English text is used instead.  
 
  If you want to give the user the ability to select a folder instead of a file, use <xref:System.Windows.Forms.FolderBrowserDialog> instead.  
-
-If the file selected in the dialog box is deleted or renamed before the OpenFile() function is called. A <xref:System.IO.IOException> exception will be thrown
-  
+ 
 ## Examples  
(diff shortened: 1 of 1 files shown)


xml/System.Windows.Forms/OpenFileDialog.xml
```diff
@@ -47,8 +47,8 @@
 On a right-to-left operating system, setting the containing form's <xref:System.Windows.Forms.Control.RightToLeft> property to <xref:System.Windows.Forms.RightToLeft.Yes?displayProperty=nameWithType> localizes the dialog's **File Name**, **Open**, and **Cancel** buttons. If the property is not set to <xref:System.Windows.Forms.RightToLeft.Yes?displayProperty=nameWithType>, English text is used instead.  
 
  If you want to give the user the ability to select a folder instead of a file, use <xref:System.Windows.Forms.FolderBrowserDialog> instead.  
-  
-
```
Jozkee, line 51: This should be documented as an `<exception>` in the member, not in the type.
https://github.com/dotnet/dotnet-api-docs/blob/5b0810324b50815313c14a862ab4acc2290c9243/xml/System.Windows.Forms/OpenFileDialog.xml#L237

For reference, for [System.IO.File.Open](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.open?view=net-8.0) we say "An I/O error occurred while opening the file" for `IOException`s.

BlueBles: Thank you for information. I did as you suggested :)

BlueBles:
@Jozkee could you review?

Jozkee (approved):
Thanks.

PR #10213, opened 2024-08-06
Port System new docs (second pass)

carlossanlop:
Some APIs were missed in the first pass.

Contributes to https://github.com/dotnet/runtime/issues/105987

### DIFF
--- a/xml/System.Runtime.InteropServices/NFloat.xml
+++ b/xml/System.Runtime.InteropServices/NFloat.xml
@@ -1435,7 +1435,7 @@ This computes `cos(x * π)`.
       <Docs>
         <param name="degrees">The value to convert to radians.</param>
         <summary>Converts a given value from degrees to radians.</summary>
-        <returns>The value of <see paramref="degrees" /> converted to radians.</returns>
+        <returns>The value of <paramref name="degrees" /> converted to radians.</returns>
         <remarks>To be added.</remarks>
         <inheritdoc cref="M:System.Numerics.ITrigonometricFunctions`1.DegreesToRadians(`0)" />
       </Docs>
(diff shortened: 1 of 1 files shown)


tannergooding (approved):

xml/System/Double.xml
```diff
@@ -1946,7 +1946,7 @@ Euler's number is approximately 2.7182818284590452354.
         <Parameter Name="x" Type="System.Double" Index="0" FrameworkAlternate="net-7.0;net-8.0;net-9.0" />
       </Parameters>
       <Docs>
-        <param name="x">The power to which <code data-dev-comment-type="c">E</code> is raised.</param>
+        <param name="x">The power to which <c>E</c> is raised.</param>
```
AaronRobinsonMSFT, line 1949: I don't think I've ever encountered _e_ in captial when referrring to Euler's number. The wiki URL has it that way, but the entirety of the text refers to it in italicized lower-case - https://en.wikipedia.org/wiki/E_(mathematical_constant). 

Looks like Wolfram does use `E` in Mathematica, so perhaps that is prior art to copy here. Just curious if this was discussed.

@tannergooding Was this an intentional choice?

AaronRobinsonMSFT (approved):

xml/System/Double.xml
tannergooding: I'd have expected this to have been a `cref="E"`` so it refers to the named constant we use for `Euler's number`.

I know we have some support for `LaTeX` now, so if we're not doing a cref, then maybe using the "proper" math symbol is better?

CC. @gewarren

AaronRobinsonMSFT: Ah, I see. If this was referring to the .NET name, then that makes sense to me.

PR #10228, opened 2024-08-09
Port over System.Runtime.InteropServices docs for .NET 9

jkoritzinsky:
## Summary

Fixes https://github.com/dotnet/runtime/issues/105986

### DIFF
--- a/xml/System.Runtime.InteropServices/CollectionsMarshal.xml
+++ b/xml/System.Runtime.InteropServices/CollectionsMarshal.xml
@@ -289,15 +289,8 @@ The ref `null` can be detected by calling <xref:System.Runtime.CompilerServices.
         <summary>Gets either a ref to a <typeparamref name="TValue"/> in the <see cref="T:System.Collections.Generic.Dictionary`2"/> or a ref null if it does not exist in the <paramref name="dictionary"/>.</summary>
         <returns>A reference to a <typeparamref name="TValue" /> in the <see cref="T:System.Collections.Generic.Dictionary`2" /> or a reference <see langword="null" /> if it does not exist in the <paramref name="dictionary" />..</returns>
         <remarks>
-          <format type="text/markdown"><![CDATA[
-
-## Remarks
-
-Items should not be added or removed from the <xref:System.Collections.Generic.Dictionary`2> while the ref `TValue` is in use.
-
-The ref `null` can be detected by calling <xref:System.Runtime.CompilerServices.Unsafe.IsNullRef%60%601(%60%600@)?displayProperty=fullName>.
-
-          ]]></format>
+          <para>Items should not be added or removed from the <see cref="T:System.Collections.Generic.Dictionary`2" /> while the ref <paramref name="TValue" /> is in use.</para>
+          <para>The ref <see langword="null" /> can be detected by calling <see cref="M:System.Runtime.CompilerServices.Unsafe.IsNullRef``1(``0@)" />.</para>
         </remarks>
       </Docs>
     </Member>


xml/System.Runtime.InteropServices/CollectionsMarshal.xml
```diff
@@ -281,14 +281,24 @@ The ref `null` can be detected by calling <xref:System.Runtime.CompilerServices.
...
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+Items should not be added or removed from the <xref:System.Collections.Generic.Dictionary`2> while the ref `TValue` is in use.
+
+The ref `null` can be detected by calling <xref:System.Runtime.CompilerServices.Unsafe.IsNullRef%60%601(%60%600@)?displayProperty=fullName>.
+
+          ]]></format>
+        </remarks>
```
carlossanlop, line 301: Did the porting tool do this? It's supposed to port remarks as plain xml, not as markdown anymore (for simplicity).

```suggestion
        <remarks>
          <para>Items should not be added or removed from the <see cref="T:System.Collections.Generic.Dictionary`2" /> while the ref <paramref name="TValue" /> is in use.</para>
          <para>The ref <see langword="null" /> can be detected by calling <see cref="M:System.Runtime.CompilerServices.Unsafe.IsNullRef``1(``0@)" />.</para>
        </remarks>
```

xml/System.Runtime.InteropServices/CollectionsMarshal.xml
```diff
@@ -178,15 +178,15 @@ Items should not be added to or removed from the <xref:System.Collections.Generi
...
-        <summary>To be added.</summary>
-        <returns>To be added.</returns>
-        <remarks>To be added.</remarks>
+        <typeparam name="TKey">The type of the keys in the dictionary.</typeparam>
+        <typeparam name="TValue">The type of the values in the dictionary.</typeparam>
+        <typeparam name="TAlternateKey">The type of the alternate key in the dictionary lookup.</typeparam>
+        <param name="dictionary">The dictionary to get the ref to <typeparamref name="TValue"/> from.</param>
+        <param name="key">The key used for lookup.</param>
+        <param name="exists">Whether or not a new entry for the given key was added to the dictionary.</param>
+        <summary>Gets a ref to a <typeparamref name="TValue"/> in the <see cref="T:System.Collections.Generic.Dictionary`2.AlternateLookup`1"/>, adding a new entry with a default value if it does not exist in the <paramref name="dictionary"/>.</summary>
```
carlossanlop, line 187: Do you know if `T:System.Collections.Generic.Dictionary`2.AlternateLookup`1` existed in Preview6? Because currently the CI cannot find it. If we introduced it in Preview7 then maybe we will need to merge this after we update dotnet-api-docs to Preview7, to prevent showing something that isn't released yet.

But if it existed in Preview6 already, then I'm not sure what the problem could be. The DocID is well formed, it matches the official DocID: https://github.com/dotnet/dotnet-api-docs/blob/55322a04db5cc185736be7e3cbf56f87cd75ba17/xml/System.Collections.Generic/Dictionary%602%2BAlternateLookup%601.xml#L4

carlossanlop (approved):
Left one comment for you to consider. There's also a problem with a type that cannot be found. Depending on the root cause, we may have to wait until P7 is released before merging this PR.

xml/System.Runtime.InteropServices/CollectionsMarshal.xml
jkoritzinsky: This was copied manually from the above method's docs as I couldn't get the porting tool to work right for CoreLib

xml/System.Runtime.InteropServices/CollectionsMarshal.xml
jkoritzinsky: I'm not sure. Honestly CollectionsMarshal isn't in my wheelhouse. I had the wrong format beforehand so lets see if CI likes it after I changed it to match the doc id.

xml/System.Runtime.InteropServices/CollectionsMarshal.xml
carlossanlop: > as I couldn't get the porting tool to work right for CoreLib

Oh, what command did you use? The xml docs for S.P.CoreLib are found under `artifacts\bin\coreclr\<os>.<arch>.<config>\IL\`, that would be the argument for `-IntelliSense`.

xml/System.Runtime.InteropServices/CollectionsMarshal.xml
carlossanlop: Alternatively, instead of making these `see crefs` they could just be `<c></c>` for now.

What would be your preference, @gewarren?

xml/System.Runtime.InteropServices/CollectionsMarshal.xml
jkoritzinsky: I did that, but the updated triple slash comments weren't being output into that folder for some reason. I didn't feel like fighting it.

carlossanlop: Did you also run `build.cmd clr.corelib` before porting?

jkoritzinsky: Yep. Even deleted my bin artifacts. Probably should have deleted my obj ones as well.

jkoritzinsky: I had the same problem with System.Runtime.InteropServices initially and solved it by using the obj folder instead. Couldn't get that to work with CoreLib though.

PR #10235, opened 2024-08-11
Sort days in DaysOfWeek Enum doc

castrodd:
## Summary

Currently, the **DaysOfWeek Enum** doc has this table:

image

This is very hard to read. This PR sorts the members in `DayOfWeek.xml` so that this table is easier to read.

### DIFF
--- a/xml/System/DayOfWeek.xml
+++ b/xml/System/DayOfWeek.xml
@@ -79,135 +79,6 @@
     </remarks>
   </Docs>
   <Members>
-    <Member MemberName="Friday">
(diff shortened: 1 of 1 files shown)


BillWagner:
ping @gewarren 

If I remember correctly, this is generated from the `///` comments, meaning this update would be removed.

gewarren:
Yes, we have an issue tracking this problem, and your changes will be overwritten by the CI - sorry. See https://github.com/dotnet/dotnet-api-docs/issues/4210.

PR #10236, opened 2024-08-12
Fix code formatting

drewnoakes:
The previous code had no indentation, making it harder to read than necessary.


BillWagner (approved):
This LGTM @drewnoakes 

I'll :shipit: now.

PR #10238, opened 2024-08-12
System.Reflection.Emit.AssemblyBuilder: Fix the AssemblyBuilder example not to ignore created instances

kiripon:
## Summary

Fix the `System.Reflection.Emit.AssemblyBuilder` example not to ignore the instance.
Before this change, the expression `pi?.GetValue(o2, null)` failed with error because `o2` is always null.


BillWagner (approved):
Thanks @kiripon 

I'll :shipit: now.

PR #10294, opened 2024-08-22
Port System.Formats.Tar and System.Data undoc APIs from Preview7

carlossanlop:
One of each was added to Preview7 as shown in the table here: https://github.com/dotnet/runtime/issues/105974#issue-2449319806


David-Engel (approved):
System.Data LGTM

PR #10292, opened 2024-08-21
Updating the documentation for System.Numerics

tannergooding:
This makes progress towards https://github.com/dotnet/runtime/issues/105981

PR #10297, opened 2024-08-22
Ensure that various inheritdoc comments can be resolved

tannergooding:
This resolves the warnings introduced in https://github.com/dotnet/dotnet-api-docs/pull/10292


gewarren (approved):
Thanks @tannergooding!

tannergooding:
@gewarren, there were a couple of pre-existing (prior to #10292) warnings about unresolvable xrefs. I just pushed another commit that resolves them by using the `see langword="..."` syntax that was agreed upon in the more general discussion

PR #10310, opened 2024-08-24
Fix bad links - System.Windows*

gewarren:
Change Markdown links to see href links in non-Markdown sections.

([Hide whitespace changes](https://github.com/dotnet/dotnet-api-docs/pull/10310/files?diff=unified&w=1).)

Contributes to #10311.

### DIFF
--- a/xml/System.Windows.Forms.Design/ParentControlDesigner.xml
+++ b/xml/System.Windows.Forms.Design/ParentControlDesigner.xml
@@ -754,7 +754,7 @@
           <para>
             <see langword="Dispose" /> can be called multiple times by other objects. When overriding <see langword="Dispose(Boolean)" /> be careful not to reference objects that have been previously disposed of in an earlier call to <see langword="Dispose" />. For more information about how to implement <see langword="Dispose(Boolean)" />, see <see href="/dotnet/standard/garbage-collection/implementing-dispose">Implementing a Dispose Method</see>.
 
- For more information about <see langword="Dispose" /> and <see cref="M:System.Object.Finalize" />, see <see href="/dotnet/standard/garbage-collection/unmanaged">Cleaning Up Unmanaged Resources</see> and <see href="https://learn.microsoft.com/previous-versions/dotnet/netframework-4.0/ddae83kx(v=vs.100)">Overriding the Finalize Method</see>).</para>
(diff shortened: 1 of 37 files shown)


xml/System.Windows.Forms.Design/ParentControlDesigner.xml
```diff
@@ -741,20 +741,20 @@
...
-  
- For more information about <see langword="Dispose" /> and <see cref="M:System.Object.Finalize" />, see [Cleaning Up Unmanaged Resources](/dotnet/standard/garbage-collection/unmanaged) and [Overriding the Finalize Method](https://learn.microsoft.com/previous-versions/dotnet/netframework-4.0/ddae83kx(v=vs.100)).</para>
+            <see langword="Dispose" /> can be called multiple times by other objects. When overriding <see langword="Dispose(Boolean)" /> be careful not to reference objects that have been previously disposed of in an earlier call to <see langword="Dispose" />. For more information about how to implement <see langword="Dispose(Boolean)" />, see <see href="/dotnet/standard/garbage-collection/implementing-dispose">Implementing a Dispose Method</see>.
+
+ For more information about <see langword="Dispose" /> and <see cref="M:System.Object.Finalize" />, see <see href="/dotnet/standard/garbage-collection/unmanaged">Cleaning Up Unmanaged Resources</see> and <see href="https://learn.microsoft.com/previous-versions/dotnet/netframework-4.0/ddae83kx(v=vs.100)">Overriding the Finalize Method</see>).</para>
```
gewarren, line 757: Remove extra parentheses.

carlossanlop (approved):
LGTM.

I was wondering if these were caused by the porting tool, but it wouldn't be possible since they came from direct xml items: The see crefs were properly ported, so any unexpected markdown would be imported _as is_ by the tool. In any case, the problem should've been caught by roslyn due to a malformed link that was added as markdown instead of xml.

gewarren:
> LGTM.
> 
> I was wondering if these were caused by the porting tool, but it wouldn't be possible since they came from direct xml items: The see crefs were properly ported, so any unexpected markdown would be imported _as is_ by the tool. In any case, the problem should've been caught by roslyn due to a malformed link that was added as markdown instead of xml.

The really strange part is that they actually worked, in most cases.

PR #10325, opened 2024-08-27
Fix an error in the output comment

salpelter:
## Summary

It's basically a code example that has a wrong output comment in the end. The number in question (10653) is actually of type `long`, which makes it a `signed 64-bit integer`, not 32.

### DIFF
--- a/snippets/csharp/System/Object/GetType/GetTypeEx2.cs
+++ b/snippets/csharp/System/Object/GetType/GetTypeEx2.cs
@@ -26,7 +26,7 @@ public class Example
 
       // The example displays the following output:
       //    12 is a 32-bit integer.
-      //    10653 is a 32-bit integer.
+      //    10653 is a 64-bit integer.
       //    12 is an unsigned byte.
       //    -5 is a signed byte.
       //    16.3 is a double-precision floating point.


salpelter:
> Hi @salpelter
> 
> I'm closing this because the original code is correct. The declaration casts the type of `10653` to a `long`. Therefore, it's type is `long`, even though its value is small enough for a 32 bit `int`. See [sharplab](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK4wD4AEBMAGAWAChcBGY4gAgHtgArAUwGMEBtAXUoDcIAbVBgGdKAXkoBvSgApYCAJSVSOONN7UYAcwWk8ANgCsAZhVTgATwQNty6YPOWFAWn1wqldx89fvH0roB0xpQARFg6wZQAvgDcbp4AZtTIDBBMABbSPMjcfAKUsDn8Qgricd4AKmYADgyUCKKFAv4A4gwIlTVScrFEPu5Q8dII/gCiAI6ofIJSCNUM1PGmFlZycmV9YQCcUsHieJH5whAwlOiCUBowDAAmlPYM/sEqPEXd654MvIK1A0OjE1MZnMFlI7MtVmten0PFsdnsDlAjpRzpcbndlo9nrkrD1oZ9vvlBjN/pMvkCaiDZBD3t5Ybt9odKBBKIYcI5gFB6rIGBoGMhMY0cTT3PifkThuNSdNZhTFmpNNSodDKHT4YzmboACzszn5GCWXn8p6Ct5K7yiwl/SWAmXzRbXaioYC8FaQ5XuVUMxFMygOp0uxxVZJMRFQdSUeJqCAIWAaShVaiyTHC93ul4CU14r4MFMw0jbYIAcnhhfVMGoCDSfN90eZtoF6aFZvckSAA=)

I don't understand what you're trying to say, to be honest. I never said that the number isn't of type `long`. And I agree that the number is small enough for a 32-bit integer, but since it was cast to `long`, that makes it 64-bit, doesn't it?.

I followed your sharplab link and it only confirmed that I was right. It literally displayed "10653 is a 64-bit integer," as I expected. Am I missing something? It didn't even cross my mind at the time of making the pull request that my edit was in any way incorrect. Could you please explain what's wrong?

gewarren (approved):
Thank you @salpelter!

PR #10330, opened 2024-08-28
Update JsonSerializerOptions.xml

RickDMyers:
Refine IncludeFields remarks to highlight that only public fields are serialized and that private fields are not.

## Summary

Revised the documentation for JsonSerializerOptions.IncludeFields property clarify that setting the property
to true will only result in 'public' fields being serialized, private fields are not serialized.

### DIFF
--- a/xml/System.Text.Json/JsonSerializerOptions.xml
+++ b/xml/System.Text.Json/JsonSerializerOptions.xml
@@ -711,8 +711,8 @@ For more information, see [How to ignore properties with System.Text.Json](/dotn
             The default value is <see langword="false" />.</summary>
         <value>
           <see langword="true" /> if public fields are included during serialization; otherwise, <see langword="false" />.</value>
-        <remarks>Only includes public fields during serialization and deserialization. Private fields are not included, even if IncludeFields is set to <see langword="true" />.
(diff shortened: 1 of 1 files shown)


xml/System.Text.Json/JsonSerializerOptions.xml
```diff
@@ -710,8 +710,10 @@ For more information, see [How to ignore properties with System.Text.Json](/dotn
         <summary>Gets or sets a value that indicates whether fields are handled during serialization and deserialization.
             The default value is <see langword="false" />.</summary>
         <value>
-          <see langword="true" /> if fields are included during serialization; otherwise, <see langword="false" />.</value>
-        <remarks>To be added.</remarks>
+          <see langword="true" /> if public fields are included during serialization; otherwise, <see langword="false" />.</value>
+        <remarks>Only includes public fields during serialization and deserialization. Private fields are not included, even if IncludeFields is set to <see langword="true" />.
+                 If you need to include private fields, you would typically use properties with appropriate accessors or consider using attributes like [JsonInclude] on those fields.
```
eiriktsarpalis, line 715:
```suggestion
        <remarks>This flag only enables serialization for public fields, as is the case for properties.
                 Private members need to be opted into individually using the <see cref="T:System.Text.Json.Serialization.JsonIncludeAttribute" /> attribute.
```

gewarren (approved):

PR #10345, opened 2024-08-30
Update `ILogger` documentation to clarify formatter behavior when using extension methods

### LINKED ISSUE #10318: Microsoft.Extensions.Logging.LoggerExtensions.LogError does not log provided exception.
### Description

When using Microsoft.Extensions.Logging.LoggerExtensions methods that allow exception to be passed for logging - exception get completely ignored and not logged at all.
 
It all comes down to private MessageFormatter method that completely ignores provided exception. 

https://github.com/dotnet/runtime/blob/07d8b82d54b6b8db16f5fbb531efcb1e276dc264/src/libraries/Microsoft.Extensions.Logging.Abstractions/src/LoggerExtensions.cs#L417

According to this commit that I found in old repository 
https://github.com/aspnet/Logging/commit/4528bdbc2816a8b5f933d9c6cf2da806ce4a72ea
Method MessageFormatter use to have correct implementation and then was changed without any good explanation.
https://github.com/aspnet/Logging/commit/4528bdbc2816a8b5f933d9c6cf2da806ce4a72ea#:~:text=private%20static%20string%20MessageFormatter(object%20state%2C%20Exception%20error)

Commit that broke this is from 2016. So how this passed through testing for 6 years is a good question.

Affected methods:
LogTarce
LogDebug
LogInformation
LogWarning
LogError
LogCritical

### Reproduction Steps

Use Microsoft.Extensions.Logging.LoggerExtensions.LogError method to log exception 

```cs
                ILogger logger; // Instance of Microsoft.Extensions.Logging.ILogger 

                try
                {
                    throw new Exception("Test exception. This should be logged");
                }
                catch (Exception e)
                {
                    logger.LogError(e, "Test message that suppose to be logged together with exception");
                }

```

### Expected behavior

Provided exception details should be logged together with log message.

### Actual behavior

Provided exception completely ignored and only log message is logged.  

### Regression?

Yes since 2016

According to this commit that I found in old repository 
https://github.com/aspnet/Logging/commit/4528bdbc2816a8b5f933d9c6cf2da806ce4a72ea
Method MessageFormatter use to have correct implementation and then was changed without any good explanation.
https://github.com/aspnet/Logging/commit/4528bdbc2816a8b5f933d9c6cf2da806ce4a72ea#:~:text=private%20static%20string%20MessageFormatter(object%20state%2C%20Exception%20error)

### Known Workarounds

No

### Configuration

_No response_

### Other information

_No response_

enioluwas:
## Summary

As described in the issue, the `ILogger` extension methods defined in `LoggerExtensions` pass in a formatter that ignores the `exception` parameter passed in. This is being made clear so that developers do not assume otherwise.

Fixes #10318

### DIFF
--- a/xml/Microsoft.Extensions.Logging/ILogger.xml
+++ b/xml/Microsoft.Extensions.Logging/ILogger.xml
@@ -1,4 +1,4 @@
-<Type Name="ILogger" FullName="Microsoft.Extensions.Logging.ILogger">
+  <Type Name="ILogger" FullName="Microsoft.Extensions.Logging.ILogger">
   <TypeSignature Language="C#" Value="public interface ILogger" />
   <TypeSignature Language="ILAsm" Value=".class public interface auto ansi abstract beforefieldinit ILogger" />
   <TypeSignature Language="DocId" Value="T:Microsoft.Extensions.Logging.ILogger" />
@@ -143,7 +143,7 @@
         <param name="exception">The exception related to this entry.</param>
         <param name="formatter">Function to create a <see cref="T:System.String" /> message of the <paramref name="state" /> and <paramref name="exception" />.</param>
         <summary>Writes a log entry.</summary>
-        <remarks>Note: When using extension methods such as <see cref="M:Microsoft.Extensions.Logging.LoggerExtensions.Log(Microsoft.Extensions.Logging.ILogger,Microsoft.Extensions.Logging.LogLevel,Microsoft.Extensions.Logging.EventId,System.Exception,System.String,System.Object[])"/> and <see cref="M:Microsoft.Extensions.Logging.LoggerExtensions.LogError(Microsoft.Extensions.Logging.ILogger,System.Exception,System.String,System.Object[])"/>, the <paramref name "formatter"> provided to this method will not add the <paramref name="exception"/> to the log message. When using these extension methods, implementations should explicitly append the <paramref name="exception"/> to the log message.</remarks>
+        <remarks>The default <paramref name="formatter" /> provided by the logger extension methods doesn't include the <paramref name="exception" /> when formatting messages. It's essential for implementations of this method to decide how exceptions should be incorporated into the logs.</remarks>
       </Docs>
     </Member>
   </Members>


xml/Microsoft.Extensions.Logging/ILogger.xml
```diff
@@ -141,7 +141,7 @@
...
         <param name="state">The entry to be written. Can also be an object.</param>
         <param name="exception">The exception related to this entry.</param>
-        <param name="formatter">Function to create a <see cref="T:System.String" /> message of the <paramref name="state" /> and <paramref name="exception" />.</param>
+        <param name="formatter">Function to create a <see cref="T:System.String" /> message of the <paramref name="state" /> and <paramref name="exception" />. **Note: When using extension methods such as <see cref="M:Microsoft.Extensions.Logging.LoggerExtensions.Log(Microsoft.Extensions.Logging.ILogger,Microsoft.Extensions.Logging.LogLevel,Microsoft.Extensions.Logging.EventId,System.Exception,System.String,System.Object[])"/> and <see cref="M:Microsoft.Extensions.Logging.LoggerExtensions.LogError(Microsoft.Extensions.Logging.ILogger,System.Exception,System.String,System.Object[])"/>, the function passed here will not add the <paramref name="exception"/> parameter to the log message. When using these overloads, implementations should explicitly append the <paramref name="exception"/> to the log message.**</param>
```
tarekgh, line 144: >. **Note: When using extension methods such as <see cref="M:Microsoft.Extensions.Logging.LoggerExtensions.Log(Microsoft.Extensions.Logging.ILogger,Microsoft.Extensions.Logging.LogLevel,Microsoft.Extensions.Logging.EventId,System.Exception,System.String,System.Object[])"/> and <see cref="M:Microsoft.Extensions.Logging.LoggerExtensions.LogError(Microsoft.Extensions.Logging.ILogger,System.Exception,System.String,System.Object[])"/>, the function passed here will not add the <paramref name="exception"/> parameter to the log message. When using these overloads, implementations should explicitly append the <paramref name="exception"/> to the log message.**</param> [](http://example.com/codeflow?start=158&length=666)

Would be better if we move this text to the remarks element instead.

enioluwas: Moved the note to the remarks

tarekgh: >. [](http://example.com/codeflow?start=158&length=1)

would saying something like the following will be simpler and clearer? 

```
The default formatter in the logger extensions doesn't include exceptions when formatting messages. It's essential for the ILogger.Log provider implementation to decide how exceptions should be incorporated into the logs.
```

tarekgh: By the way, thanks a lot for helping improving the docs.

tarekgh: You missed the `=` in the part `<paramref name "formatter">`

enioluwas: No problem! Will modify based on your suggestion

enioluwas: Let me know if it looks good now

enioluwas: Fixed

tarekgh (approved):

PR #10367, opened 2024-09-06
 Fix inconsistent test case number in comments and console logs

depth233:
The node (dog) was removed in test 12, not in test 11.


gewarren (approved):
Thanks @depth233

PR #10370, opened 2024-09-06
Add remaining missing docs for System.ComponentModel

buyaa-n:
## Summary

Add remaining missing docs for System.ComponentModel from https://github.com/dotnet/runtime/issues/105983

Fixes https://github.com/dotnet/runtime/issues/105983

### DIFF
--- a/xml/System.ComponentModel/TypeDescriptionProvider.xml
+++ b/xml/System.ComponentModel/TypeDescriptionProvider.xml
@@ -440,7 +440,7 @@
         <param name="instance">The object for which to get the extended type descriptor.</param>
         <summary>Returns an extended custom type descriptor for the given object.</summary>
         <returns>An <see cref="T:System.ComponentModel.ICustomTypeDescriptor" /> that can provide extended metadata for the object.</returns>
-        <remarks>An extended type descriptor is a custom type descriptor that offers properties that other objects have added to this object, but are not actually defined on the object. For example, in the .NET Framework Component Model, objects that implement the interface IExtenderProvider can "attach" properties to other objects that reside in the same logical container. The GetTypeDescriptor method does not return a type descriptor that provides these extra extended properties. GetExtendedTypeDescriptor returns the set of these extended properties. TypeDescriptor will automatically merge the results of these two property collections. Note that while the .NET Framework component model only supports extended properties this API can be used for extended attributes and events as well, if the type description provider supports it.</remarks>
+        <remarks>An extended type descriptor is a custom type descriptor that offers properties that other objects have added to this object, but are not actually defined on the object. For example, in the .NET Framework Component Model, objects that implement the interface IExtenderProvider can "attach" properties to other objects that reside in the same logical container. The GetTypeDescriptorFromRegisteredType method does not return a type descriptor that provides these extra extended properties. GetExtendedTypeDescriptorFromRegisteredType returns the set of these extended properties. TypeDescriptor will automatically merge the results of these two property collections. Note that while the .NET Framework component model only supports extended properties this API can be used for extended attributes and events as well, if the type description provider supports it.</remarks>
       </Docs>
     </Member>
     <Member MemberName="GetExtenderProviders">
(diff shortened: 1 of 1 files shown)


xml/System.ComponentModel/TypeDescriptionProvider.xml
```diff
@@ -437,9 +437,9 @@
...
         <summary>Returns an extended custom type descriptor for the given object.</summary>
-        <returns>To be added.</returns>
+        <returns>An <see cref="T:System.ComponentModel.ICustomTypeDescriptor" /> that can provide extended metadata for the object.</returns>
         <remarks>An extended type descriptor is a custom type descriptor that offers properties that other objects have added to this object, but are not actually defined on the object. For example, in the .NET Framework Component Model, objects that implement the interface IExtenderProvider can "attach" properties to other objects that reside in the same logical container. The GetTypeDescriptor method does not return a type descriptor that provides these extra extended properties. GetExtendedTypeDescriptor returns the set of these extended properties. TypeDescriptor will automatically merge the results of these two property collections. Note that while the .NET Framework component model only supports extended properties this API can be used for extended attributes and events as well, if the type description provider supports it.</remarks>
```
steveharter, line 443: Typo from original .cs file; this:
```GetExtendedTypeDescriptor returns the set of these extended properties.```
should be
```GetExtendedTypeDescriptorFromRegisteredType returns the set of these extended properties.```
or perhaps just replace with "This method returns..." as is done elsewhere in this area.

**Review by steveharter — APPROVED** [review]
Minor comment; thanks.

**Review by gewarren — APPROVED** [review]
Looks good. For the future, you don't need to include the type in the descriptions - it's added automatically.

File: xml/System.ComponentModel/TypeDescriptionProvider.xml (hunk 1 above)
Comment (buyaa-n) [reply xml/System.ComponentModel/TypeDescriptionProvider.xml:443]:
```suggestion
        <remarks>An extended type descriptor is a custom type descriptor that offers properties that other objects have added to this object, but are not actually defined on the object. For example, in the .NET Framework Component Model, objects that implement the interface IExtenderProvider can "attach" properties to other objects that reside in the same logical container. The GetTypeDescriptor method does not return a type descriptor that provides these extra extended properties. GetExtendedTypeDescriptor returns the set of these extended properties. TypeDescriptor will automatically merge the results of these two property collections. Note that while the .NET Framework component model only supports extended properties this API can be used for extended attributes and events as well, if the type description provider supports it.</remarks>
```
```suggestion
        <remarks>An extended type descriptor is a custom type descriptor that offers properties that other objects have added to this object, but are not actually defined on the object. For example, in the .NET Framework Component Model, objects that implement the interface IExtenderProvider can "attach" properties to other objects that reside in the same logical container. The GetTypeDescriptorFromRegisteredType method does not return a type descriptor that provides these extra extended properties. GetExtendedTypeDescriptorFromRegisteredType returns the set of these extended properties. TypeDescriptor will automatically merge the results of these two property collections. Note that while the .NET Framework component model only supports extended properties this API can be used for extended attributes and events as well, if the type description provider supports it.</remarks>
```

xml/System.ComponentModel/TypeDescriptionProvider.xml
```diff
@@ -1035,9 +1035,9 @@
         <Parameter Name="instance" Type="System.Object" Index="0" FrameworkAlternate="net-9.0" />
       </Parameters>
       <Docs>
-        <param name="instance">To be added.</param>
+        <param name="instance">An instance of the type. Can be <see langword="null" /> if no instance was passed to the <see cref="T:System.ComponentModel.TypeDescriptor" />.</param>
         <summary>Returns a custom type descriptor for the given type or object.</summary>
-        <returns>To be added.</returns>
+        <returns>An <see cref="T:System.ComponentModel.ICustomTypeDescriptor" /> that can provide metadata for the type.</returns>
         <remarks>The objectType parameter is always valid, but the instance parameter may be <see langword="null" /> if no instance was passed to TypeDescriptor. The method should return a custom type descriptor for the object. If the method is not interested in providing type information for the object it should return base.</remarks>
```
buyaa-n, line 1041: There is no `objectType` parameter for this overload, therefore `instance` cannot be null
```suggestion
        <remarks>The method should return a custom type descriptor for the object. If the method is not interested in providing type information for the object it should return base.</remarks>
```

xml/System.ComponentModel/TypeDescriptionProvider.xml
```diff
@@ -1072,9 +1072,9 @@
         <Parameter Name="objectType" Type="System.Type" Index="0" FrameworkAlternate="net-9.0" />
       </Parameters>
       <Docs>
-        <param name="objectType">To be added.</param>
+        <param name="objectType">The type of object for which to retrieve the type descriptor.</param>
         <summary>Returns a custom type descriptor for the given type or object.</summary>
-        <returns>To be added.</returns>
+        <returns>An <see cref="T:System.ComponentModel.ICustomTypeDescriptor" /> that can provide metadata for the type.</returns>
         <remarks>The objectType parameter is always valid, but the instance parameter may be <see langword="null" /> if no instance was passed to TypeDescriptor. The method should return a custom type descriptor for the object. If the method is not interested in providing type information for the object it should return base.</remarks>
```
buyaa-n, line 1078: Similar, no `instance` 
```suggestion
        <remarks>The method should return a custom type descriptor for the object. If the method is not interested in providing type information for the object it should return base.</remarks>
```

PR #10376, opened 2024-09-08
Modernize GetLogicalDrives docs

MSDN-WhiteKnight:
- Add information about Unix behaviour
- Replace mention of floppy drives that are no longer common
- Make descriptions of similar methods in Directory and Environment class more consistent


gewarren (approved):
Thanks @MSDN-WhiteKnight

PR #10384, opened 2024-09-10
Changed String to string. IDE0049

muhammadzakriashahid:
## Summary

Describe your changes here.
i change the String type to string. Because of the rule ID IDE0049 Use language keywords instead of framework type names for type references.

### DIFF
--- a/snippets/csharp/System/String/Replace/Program.cs
+++ b/snippets/csharp/System/String/Replace/Program.cs
@@ -1,0 +1,4 @@
+﻿//Example1.Main();
+//Example2.Main();
+//Example3.Main();
+Example4.Main();
--- a/snippets/csharp/System/String/Replace/replace1.cs
+++ b/snippets/csharp/System/String/Replace/replace1.cs
@@ -1,14 +1,14 @@
 ﻿using System;
 
-public class Example
+public class Example1
 {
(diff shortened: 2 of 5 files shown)


BillWagner:
tag @gewarren 

I know that our general guideline is to use `string` vs. `String`. Do you want to follow that same guideline for the `System.String` APIs?

gewarren:
> tag @gewarren
> 
> I know that our general guideline is to use `string` vs. `String`. Do you want to follow that same guideline for the `System.String` APIs?

Yes, I don't see why not.

gewarren (approved):

gewarren:
@muhammadzakriashahid Can you add a project file similar to https://github.com/dotnet/dotnet-api-docs/blob/main/snippets/csharp/System/Math/E/Project.csproj but for .net8.0 in the directory that contains the snippet? This is a newer requirement that we enforce whenever a code snippet is modified so we can ensure it compiles.

muhammadzakriashahid:
@dotnet-policy-service agree

muhammadzakriashahid:
haven't heard from you guys

gewarren (approved):

gewarren:
Thanks @muhammadzakriashahid

PR #10396, opened 2024-09-13
Update AggregateException.xml

epsitec:
Related to https://github.com/dotnet/runtime/pull/107743

## Summary

Fixes an incorrect description of what method `GetBaseException()` does.

### DIFF
--- a/xml/System/AggregateException.xml
+++ b/xml/System/AggregateException.xml
@@ -682,7 +682,7 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>Returns the <see cref="Exception"/> that is the root cause of this exception. This will either be the root exception, or the first <see cref="AggregateException"/> that contains either multiple inner exceptions or no inner exceptions at all.</summary>
+        <summary>Returns the <see cref="T:System.Exception"/> that is the root cause of this exception. This exception is either the root exception or the first <see cref="T:System.AggregateException"/> that contains either multiple inner exceptions or no inner exceptions at all.</summary>
         <returns>The <see cref="T:System.Exception" /> that is the root cause of this exception.</returns>
         <remarks>To be added.</remarks>
       </Docs>


xml/System/AggregateException.xml
```diff
@@ -682,8 +682,8 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>Returns the <see cref="T:System.AggregateException" /> that is the root cause of this exception.</summary>
-        <returns>The <see cref="T:System.AggregateException" /> that is the root cause of this exception.</returns>
+        <summary>Returns the <see cref="Exception"/> that is the root cause of this exception. This will either be the root exception, or the first <see cref="AggregateException"/> that contains either multiple inner exceptions or no inner exceptions at all.</summary>
```
gewarren, line 685:
```suggestion
        <summary>Returns the <see cref="T:System.Exception"/> that is the root cause of this exception. This exception is either the root exception or the first <see cref="T:System.AggregateException"/> that contains either multiple inner exceptions or no inner exceptions at all.</summary>
```

gewarren (approved):

xml/System/AggregateException.xml
jkotas, line 685: Same comment as in https://github.com/dotnet/runtime/pull/107743#discussion_r1757134862 : It is not clear what "root exception" means here.

epsitec: @jkotas how would you make this clearer?

xml/System/AggregateException.xml
```diff
@@ -682,8 +682,8 @@
       </ReturnValue>
       <Parameters />
       <Docs>
-        <summary>Returns the <see cref="T:System.AggregateException" /> that is the root cause of this exception.</summary>
-        <returns>The <see cref="T:System.AggregateException" /> that is the root cause of this exception.</returns>
+        <summary>Returns the <see cref="T:System.Exception"/> that is the root cause of this exception. This exception is either the root exception or the first <see cref="T:System.AggregateException"/> that contains either multiple inner exceptions or no inner exceptions at all.</summary>
+        <returns>The <see cref="T:System.Exception" /> that is the root cause of this exception.</returns>
```
jkotas, line 686:
```suggestion
        <summary>Returns the <see cref="T:System.AggregateException"/> that is the root cause of this exception. This exception is either the current exception or the first <see cref="T:System.AggregateException"/> that contains either multiple inner exceptions or no inner exceptions at all.</summary>
        <returns>The <see cref="T:System.AggregateException" /> that is the root cause of this exception.</returns>
```
https://learn.microsoft.com/en-us/dotnet/api/system.exception.getbaseexception and other docs use "current exception" for this.

Also, I think the original docs were better in highlighting that the returned instance is always AggregateException.

xml/System/AggregateException.xml
jkotas: [Made a suggestion](https://github.com/dotnet/dotnet-api-docs/pull/10396#discussion_r1765540926)

xml/System/AggregateException.xml
epsitec: The original docs were wrong, hence my fixes. The code will return the plain `Exception` which was thrown up the chain (what's now called the _current_ exception), and only if there are multiple exceptions wrapped in an `AggregateException` will it return an _aggregate exception_. That's the whole point of this pull request.

jkotas: Ah ok, I have misread how the implementation behaves.

I agree with you that the return type in the docs is wrong. 

I am not sure whether it makes sense to try to describe the exact algorithm used by this method in the docs. The current implementation of `Exception.GetBaseException` and `AggregateException.GetBaseException` methods are in conflict with the contract for these methods described in the docs.

https://learn.microsoft.com/en-us/dotnet/api/system.exception.getbaseexception?view=net-8.0 says "For all exceptions in a chain of exceptions, the GetBaseException method must return the same object (the base exception).". For example, this invariant does not seem to hold for
```
TargetInvocationException 
   -> AggregateException
          -> ArgumentException
          -> ArgumentException
```
`GetBaseException()` called on `TargetInvocationException` returns first `ArgumentException` for , but `GetBaseException()` called on `AggregateException` returns the same `AggregateException`.

We may want to fix the implementation and/or docs to make this more sensible.

epsitec: So, practically, are you @jkotas addressing this in the documentation?

PR #10444, opened 2024-09-27
Disable edits on certain files/folders

gewarren:
- Disable the Edit button (pencil icon) on files whose source of truth is in the dotnet/runtime repo.
- Add a policy to close PRs that edit those files.

### DIFF
--- a/.github/policies/disallow-edits.yml
+++ b/.github/policies/disallow-edits.yml
@@ -13,7 +13,7 @@ configuration:
           - or:
               - filesMatchPattern:
                   pattern: xml/Microsoft.Extensions*/*.xml
-                  excludedFiles:
+                  excludedFiles: # These files use this repo as source-of-truth.
                     - xml/Microsoft.Extensions.FileSystemGlobbing/*.xml
                     - xml/Microsoft.Extensions.FileSystemGlobbing.Abstractions/*.xml
                     - xml/Microsoft.Extensions.FileSystemGlobbing.Internal*/*.xml


.github/policies/disallow-edits.yml
```diff
@@ -0,0 +1,41 @@
...
+          - payloadType: Pull_Request
+          - isAction:
+              action: Opened
+          - or:
+              - filesMatchPattern:
+                  pattern: xml/Microsoft.Extensions*/*.xml
+                  excludedFiles:
+                    - xml/Microsoft.Extensions.FileSystemGlobbing/*.xml
+                    - xml/Microsoft.Extensions.FileSystemGlobbing.Abstractions/*.xml
+                    - xml/Microsoft.Extensions.FileSystemGlobbing.Internal*/*.xml
```
carlossanlop, line 19: What's the reason for these exclusions? I think we should have a comment to explain the exclusion.

.github/policies/disallow-edits.yml
```diff
@@ -0,0 +1,41 @@
...
+              - filesMatchPattern:
+                  pattern: xml/System.Numerics.Tensors/*.xml
+              - filesMatchPattern:
+                  pattern: xml/System.Speech/*.xml
+              - includesModifiedFile:
+                  file: xml/System.Runtime.Serialization/ISerializationCodeDomSurrogateProvider.xml
+              - includesModifiedFile:
+                  file: xml/System.Runtime.Serialization/ImportOptions.xml
+              - includesModifiedFile:
+                  file: xml/System.Runtime.Serialization/XsdDataContractExporter.xml
```
carlossanlop, line 37: What does includesModifiedFile mean?

carlossanlop (approved):
LGTM. Thanks for doing this!

I left a small suggestion to add a comment if you agree.

FYI @ericstj @jeffhandley this is happening.

.github/policies/disallow-edits.yml
gewarren: I was using `filesMatchPattern` for patterns with wildcards, and `includesModifiedFile` for specific files. Still kinda guessing here as the [documentation](https://eng.ms/docs/more/github-inside-microsoft/policies/resource-management?tabs=event-conditions) isn't great.

.github/policies/disallow-edits.yml
gewarren: The reason is this: https://github.com/dotnet/runtime/blob/86995059adb5e76e1701502b980c0bf574451a94/src/libraries/Microsoft.Extensions.FileSystemGlobbing/src/Microsoft.Extensions.FileSystemGlobbing.csproj#L6

PR #10471, opened 2024-09-30
Fix AEAD algorithm documentation to indicate when AuthenticationTagMismatchException is thrown

### LINKED ISSUE #10465: AesGcm.Decrypt() doesn't throw CryptographicException
### Type of issue

Typo

### Description

This page says a [CryptographicException](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.cryptographicexception?view=net-8.0) is thrown if the tag couldn't be verified, but the function actually throws [AuthenticationTagMismatchException](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.authenticationtagmismatchexception?view=net-8.0).

This error may apply to other AEAD pages as well, but I haven't tested those.

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.aesgcm.decrypt?view=net-8.0#system-security-cryptography-aesgcm-decrypt(system-readonlyspan((system-byte))-system-readonlyspan((system-byte))-system-readonlyspan((system-byte))-system-span((system-byte))-system-readonlyspan((system-byte)))

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Security.Cryptography/AesGcm.xml

4ea1eb00-71c7-149a-d174-b44f3b9b91a4

### Article author

@dotnet-bot

vcsjones:
## Summary

Starting in .NET 8, AEAD algorithms throw `AuthenticationTagMismatchException` to indicate that decryption failed because of the authentication tag mismatch. This is a sub-type of `CryptographicException`, so it was not considered a breaking change.

This clarifies in the documentation that starting in .NET 8, users should expect this exception when the authentication tag does not match during decryption.

Fixes #10465

### DIFF
--- a/xml/System.Security.Cryptography/AesCcm.xml
+++ b/xml/System.Security.Cryptography/AesCcm.xml
@@ -212,7 +212,8 @@ The <paramref name="nonce" /> parameter length is not permitted by <see cref="P:
 
 The <paramref name="tag" /> parameter length is not permitted by <see cref="P:System.Security.Cryptography.AesCcm.TagByteSizes" />.</exception>
         <exception cref="T:System.ArgumentNullException">The <paramref name="nonce" />, <paramref name="ciphertext" />, <paramref name="tag" />, or <paramref name="plaintext" /> parameter is <see langword="null" />.</exception>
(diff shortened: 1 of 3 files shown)


gewarren (approved):
Thank you!

PR #10481, opened 2024-10-01
Removed disallowed action

gewarren:
See https://github.com/dotnet/dotnet-api-docs/actions/runs/11132452614.


gewarren:
I changed the Actions permissions instead.

Issue #10398, opened 2024-09-14
No coverage of how AllocHGlobal works under Linux.  Documentation implies it is supported under Windows only. 


logical-intent:
### Type of issue

Outdated article

### Description

[Enter feedback here]
The documentation here says "This method exposes the Win32 LocalAlloc function from Kernel32.dll."  I haven't tried using it under Linux but Copilot reckons it uses `malloc` there, so either it's hallucinating or this documentation is badly misleading and out of date. 

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.allochglobal?view=net-8.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Runtime.InteropServices/Marshal.xml

### Document Version Independent Id

53418d00-f1c2-5cde-3843-b607b1dcf240

### Article author

@dotnet-bot

karakasa:
just fyi, under Unix-based systems, the method is implemented by `NativeMemory.Alloc`, essentially `malloc`.

https://github.com/dotnet/runtime/blob/128fdfd1fd982e277e969d5c7ca1710345f4f649/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Unix.cs#L59-L62

https://github.com/dotnet/runtime/blob/128fdfd1fd982e277e969d5c7ca1710345f4f649/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/NativeMemory.Unix.cs#L126-L137

bartonjs:
@AaronRobinsonMSFT @jkoritzinsky , is this you guys?

AaronRobinsonMSFT:
> @AaronRobinsonMSFT @jkoritzinsky , is this you guys?

Yes.

> It is an odd default since memory allocated with C++ new[] is surely not safe to allow NativeMemory.Alloc to delete.

@logical-intent I'm not sure how this is "odd". There is no "default" allocator that gets assigned to `new`. The `new` keyword is the allocator and one must call the related `delete`, by default, everything else is undefined.

> Your answer does at least give me confidence that I could thunk everything through a "pure C" layer if I wanted to do a load of extra copying of data just to satisfy the P/Invoke defaults.

Correct. The C ABI is stable whereas the C++ one isn't defined; therefore, interop with C++ is considered difficult in the best of circumstances and "unreliable" and non-portable outside of a C++ compiler being involved in the worst case. 

[...]

PR #10484, opened 2024-10-01
Add note about when to use HGlobal allocator APIs.

AaronRobinsonMSFT:
Fixes https://github.com/dotnet/dotnet-api-docs/issues/10398

### DIFF
--- a/xml/System.Runtime.InteropServices/Marshal.xml
+++ b/xml/System.Runtime.InteropServices/Marshal.xml
@@ -304,7 +304,7 @@
 
 ## Remarks
 > [!IMPORTANT]
-> This native memory allocator is a legacy API that should be used exclusively when called for by specific Win32 APIs on the Windows platform. On non-Windows platforms, use the <xref:System.Runtime.InteropServices.NativeMemory> class to allocate native memory.
+> This native memory allocator is a legacy API that should be used exclusively when called for by specific Win32 APIs on the Windows platform. When targeting .NET 6 or later, use the <xref:System.Runtime.InteropServices.NativeMemory> class on all platforms to allocate native memory. When targeting .NET 6 or earlier, use <xref:System.Runtime.InteropServices.Marshal.AllocCoTaskMem%2A> on all platforms to allocate native memory.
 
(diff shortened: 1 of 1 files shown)


xml/System.Runtime.InteropServices/Marshal.xml
```diff
@@ -303,6 +303,9 @@
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
+> [!IMPORTANT]
+>  This native memory allocator is a legacy API that should be used exclusively when called for by specific Win32 APIs on the Windows platform. On non-Windows platforms, the <xref:System.Runtime.InteropServices.NativeMemory> class should be used to allocate native memory.
```
gewarren, line 307:
```suggestion
> This native memory allocator is a legacy API that should be used exclusively when called for by specific Win32 APIs on the Windows platform. On non-Windows platforms, use the <xref:System.Runtime.InteropServices.NativeMemory> class to allocate native memory.
```

gewarren (approved):

xml/System.Runtime.InteropServices/Marshal.xml
```diff
@@ -303,6 +303,9 @@
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
+> [!IMPORTANT]
+> This native memory allocator is a legacy API that should be used exclusively when called for by specific Win32 APIs on the Windows platform. On non-Windows platforms, use the <xref:System.Runtime.InteropServices.NativeMemory> class to allocate native memory.
```
jkotas, line 307: This is incomplete guidance. What should people do on Windows platforms or when they target TFM that does not have `S.R.I.NativeMemory`?

AaronRobinsonMSFT: I appreciate what you're getting at, but I don't think the doc will matter in practice all that much for those cases (TFM < .NET 6). Perhaps all that is needed in this case then is to say:

```
This native memory allocator is a legacy API that should be used exclusively when called for by specific Win32 APIs on the Windows platform or when targeting .NET versions earlier than .NET 6.0. On non-Windows platforms, use the <xref:System.Runtime.InteropServices.NativeMemory> class to allocate native memory.
```

AaronRobinsonMSFT:
@gewarren Can you take one more look at my updated phrasing? Thanks.

gewarren (approved):
LGTM

PR #10493, opened 2024-10-03
Update `HttpHeaders.Add` exception docs

MihaZupan:
- Being more specific about what a valid header name format is.
- Fix misleading/outdated comment about new line chars in header values.


rzikm (approved):
LGTM, thanks

PR #10500, opened 2024-10-04
Add ArgumentException

gewarren:
Contributes to dotnet/docs#42696.

### DIFF
--- a/xml/System.ComponentModel/EnumConverter.xml
+++ b/xml/System.ComponentModel/EnumConverter.xml
@@ -62,10 +62,8 @@
  For more information about type converters, see the <xref:System.ComponentModel.TypeConverter> base class and [How to: Implement a Type Converter](https://learn.microsoft.com/previous-versions/visualstudio/visual-studio-2013/ayybcxe5(v=vs.120)).  
   
 > [!NOTE]
->  You should never create an instance of an <xref:System.ComponentModel.EnumConverter>. Instead, call the <xref:System.ComponentModel.TypeDescriptor.GetConverter%2A> method of the <xref:System.ComponentModel.TypeDescriptor> class. For more information, see the examples in the <xref:System.ComponentModel.TypeConverter> base class.  
-  
-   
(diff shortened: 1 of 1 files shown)


xml/System.ComponentModel/EnumConverter.xml
```diff
@@ -62,10 +62,8 @@
  For more information about type converters, see the <xref:System.ComponentModel.TypeConverter> base class and [How to: Implement a Type Converter](https://learn.microsoft.com/previous-versions/visualstudio/visual-studio-2013/ayybcxe5(v=vs.120)).  
   
 > [!NOTE]
->  You should never create an instance of an <xref:System.ComponentModel.EnumConverter>. Instead, call the <xref:System.ComponentModel.TypeDescriptor.GetConverter%2A> method of the <xref:System.ComponentModel.TypeDescriptor> class. For more information, see the examples in the <xref:System.ComponentModel.TypeConverter> base class.  
-  
-   
-  
+> You should never create an instance of an <xref:System.ComponentModel.EnumConverter>. Instead, call the <xref:System.ComponentModel.TypeDescriptor.GetConverter%2A> method of the <xref:System.ComponentModel.TypeDescriptor> class. For more information, see the examples in the <xref:System.ComponentModel.TypeConverter> base class.
```
gewarren, line 65: Is this still true?

LakshanF: The only change should be the one exception that is documented below. For library authors who do extend the `EnumConverter` class (such as the code [here](https://github.com/dotnet/winforms/blob/189b209933b2e244748a97e66b30be2f8d11b9e4/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TextBox/TextBoxAutoCompleteSourceConverter.cs#L8-L10) in WinForms, its important that they pass an `enum` type when they call the base constructor.

LakshanF (approved):
LGTM, thanks!

PR #10506, opened 2024-10-07
Add note about UrlEncode

ManickaP:
`UrlEncode` is suitable only for query segment or x-www-form-urlencoded content. Also the list of unescaped punction doesn't follow neither URI RFC (old or new) nor any other spec I could find.

### DIFF
--- a/xml/System.Net/WebUtility.xml
+++ b/xml/System.Net/WebUtility.xml
@@ -556,6 +556,9 @@
 
  Each ' ' (space) character is converted to a `+` (plus) character.
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


MihaZupan:
> Also the list of unescaped punction doesn't follow neither URI RFC (old or new) nor any other spec I could find.

https://github.com/dotnet/runtime/blob/770df102b0dfbf16c22ef64a270a989495cfe145/src/libraries/System.Private.CoreLib/src/System/Net/WebUtility.cs#L572

The code mentions https://datatracker.ietf.org/doc/html/rfc1738 that includes this bit:
```
safe           = "$" | "-" | "_" | "." | "+"
extra          = "!" | "*" | "'" | "(" | ")" | ","
```
but even that doesn't match the characters used 😄

MihaZupan (approved):

PR #10534, opened 2024-10-09
Use includeModifiedFile instead

gewarren:
The previous policy only triggers if *all* files in the PR match *the same* pattern. So try this instead.


BillWagner (approved):
This LGTM.

IEvangelist (approved):

PR #10537, opened 2024-10-09
Try moving if/then to Action section

gewarren:
Move if/then for different file paths to Action section instead of Condition section.


adegeo (approved):
OK 😁 but it isn't working it seems?

PR #10544, opened 2024-10-10
Use matchAny

gewarren:
See https://eng.ms/docs/products/gitops/apps/examples/conditions#filesmatchpattern.


BillWagner (approved):
This is great @gewarren 

Let's :shipit:

PR #10561, opened 2024-10-14
Clarify return values of Base64Url.Try* methods

buyaa-n:
## Summary

Improve return values of Base64Url.Try* methods, clarify it returns false if destination is too small

Relates to https://github.com/dotnet/runtime/issues/108834#issuecomment-2411478958


stephentoub (approved):
Thanks

gewarren (approved):

PR #10564, opened 2024-10-14
Fix Lock "compiler speclet" URL

richardszalay:
## Summary

The "compiler speclet" URL in the [System.Threading.Lock class documentation](https://learn.microsoft.com/en-us/dotnet/api/system.threading.lock?view=net-9.0#remarks:~:text=For%20more%20information%2C%20see%20the%20relevant%20compiler%20speclet.) was missing the language version folder under `proposals/`.

### DIFF
--- a/xml/System.Threading/Lock.xml
+++ b/xml/System.Threading/Lock.xml
@@ -30,7 +30,7 @@ It is recommended to use the <xref:System.Threading.Lock.EnterScope%2A> method w
 
 :::code language="csharp" source="~/snippets/csharp/System.Threading/Lock/Overview/UsagePatterns.cs" id="Snippet1":::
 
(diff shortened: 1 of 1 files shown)


richardszalay:
Question: This PR is a like-for-like / as-is fix, but would it be better to update the link (and the entire sentence) to link out to the [published document page](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-13.0/lock-object) rather than the GitHub URL? Let me know and I can update this PR accordingly

kouvel:
> Question: This PR is a like-for-like / as-is fix, but would it be better to link to the [published document page](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-13.0/lock-object) rather than the GitHub URL? Let me know and I can update this PR accordingly

Thanks for fixing this! Yea that published link seems like a better one to put there, the above looks good to me.

richardszalay:
Great I'll update the PR with the published link.

kouvel (approved):
LGTM, thanks!

PR #10565, opened 2024-10-15
Add docs remarks for `IsRealNumber` handling of infinities

FilipToth:
## Summary

Added documentation coverage for `IsRealNumber` methods' handling of infinities, which may seem unintuitive to some developers. For further details, refer to the linked issue.

A separate PR will be submitted into `dotnet/runtime` to update inline documentation comments once this change is approved, in order to avoid simultaneous changes across two repos.

Fixes dotnet/runtime#103930

### DIFF
--- a/xml/System.Numerics/Complex.xml
+++ b/xml/System.Numerics/Complex.xml
@@ -2292,8 +2292,6 @@ A return value of `false` does not imply that <xref:System.Numerics.INumberBase%
 
 This function returns `true` for a complex number `a + bi` where `b` is zero.
 
-This function also returns `true` for a complex number `a + bi` where `a` is infinity, but not `b`, and does not return true for <xref:System.Numerics.Complex.Infinity>.
-
           ]]></format>
         </remarks>
         <inheritdoc cref="M:System.Numerics.INumberBase`1.IsRealNumber(`0)" />
--- a/xml/System.Numerics/INumberBase`1.xml
+++ b/xml/System.Numerics/INumberBase`1.xml
@@ -782,7 +782,7 @@ A return value of `false` does not imply that <xref:System.Numerics.INumberBase%
 
 This function returns `true` for a complex number `a + bi` where `b` is zero.
 
-This function also returns `true` for positive and negative infinity.
+This function checks values against the extended real number line, thus returns `true` for positive and negative infinity.
 
           ]]></format>
         </remarks>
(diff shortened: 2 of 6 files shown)


AaronRobinsonMSFT:
@tannergooding Are you okay with this update?

xml/System.Numerics/Complex.xml
```diff
@@ -2292,6 +2292,8 @@ A return value of `false` does not imply that <xref:System.Numerics.INumberBase%
 
 This function returns `true` for a complex number `a + bi` where `b` is zero.
 
+This function also returns `true` for a complex number `a + bi` where `a` is infinity, but not `b`, and does not return true for <xref:System.Numerics.Complex.Infinity>.
```
tannergooding, line 2295: I'm not sure this adds much, as the first comment covers this entirely.

As indicated, the function returns true for any complex number `a + bi` where `b` is zero, that by definition includes `a == infinity`. The requirement that `b` be zero then excludes `b` being anything else and that is also why `Complex.Infinity` returns false, because complex infinity is `inf, inf`

xml/System.Numerics/INumberBase`1.xml
```diff
@@ -782,6 +782,8 @@ A return value of `false` does not imply that <xref:System.Numerics.INumberBase%
 
 This function returns `true` for a complex number `a + bi` where `b` is zero.
 
+This function also returns `true` for positive and negative infinity.
```
tannergooding, line 785: I'm not sure this is the "right" wording.

Technically speaking `infinity` is a `real number` and lives on the `extended real number line`. This is the number line that the IEEE 754 specification follows and which is used in many higher level mathematical domains to allow correct modeling of various concepts.

Trying to define it as "this also returns true for ..." makes it instead sound like `infinity` is not a real number, which is then misleading.

I'd rather see the remark elaborate something like:
> This function checks values against the `extended real number line` which includes `positive infinity` and `negative infinity`.

xml/System.Numerics/Complex.xml
FilipToth: Yes, that seems logical.

xml/System.Numerics/INumberBase`1.xml
FilipToth: That seems way clearer. I've updated the docs.

PR #10587, opened 2024-10-18
Add note about allowing out of order metadata properties

gewarren:
Contributes to dotnet/docs#42592.

### DIFF
--- a/xml/System.Text.Json.Serialization/ReferenceHandler.xml
+++ b/xml/System.Text.Json.Serialization/ReferenceHandler.xml
@@ -135,11 +135,11 @@
  
 * For the metadata properties within the JSON to be considered well-formed, they must follow these rules:
 
-  * The `$id` metadata property must be the first property in the JSON object.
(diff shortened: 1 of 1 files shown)


eiriktsarpalis (commented):
Would it make sense to also include mention of this feature in the polymorphism document?

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/polymorphism?pivots=dotnet-8-0

Polymorphism is the primary motivator for introducing this feature (e.g. because you're reading data off an external API that happens to place the `$type` discriminator in the middle of an object.

BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

gewarren:
> Would it make sense to also include mention of this feature in the polymorphism document?
> 
> https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/polymorphism?pivots=dotnet-8-0
> 
> Polymorphism is the primary motivator for introducing this feature (e.g. because you're reading data off an external API that happens to place the `$type` discriminator in the middle of an object.

Yes, I will add it there separately too.

PR #10594, opened 2024-10-18
Auto-close stale issues

gewarren:


BillWagner (approved):
This LGTM @gewarren 

We should (in time) use the same labels `needs-author-action` in dotnet/docs and other repos so our automation scripts are consistent.

PR #10600, opened 2024-10-21
Fix supported schemes in docs

deeprobin:
## Summary

Documentation enhanced to include the supported proxy formats:
* `https`
* `socks4`
* `socks4a`
* `socks5`

Fixes dotnet/runtime#102441

/cc @MihaZupan


deeprobin:
@MihaZupan Should I comment the different behaviors in .NET 7 and 8 or is the documentation here only for the current/upcoming .NET releease (so .NET 10)?

MihaZupan (approved):
Sorry about the delayed review, but thank you for improving this

PR #10619, opened 2024-10-30
Document a behavior of string.Create

tannergooding:
This resolves https://github.com/dotnet/runtime/issues/106903 by documenting the agreed upon behavior:
* We're good at guaranteeing that the span points to memory that has space for a least one more element than the span
* We'd guarantee that the span points to zero'd memory, including the null terminator.
* Our docs would say that if you write to the extra location it must be a null terminator. Writing anything else will result corrupting the string and thus is undefined behavior.

### DIFF
--- a/xml/System/String.xml
+++ b/xml/System/String.xml
@@ -4365,7 +4365,7 @@ The following example demonstrates the <xref:System.String.CopyTo%2A> method.
 ## Remarks
 The initial content of the destination span passed to `action` is undefined. Therefore, it is the delegate's responsibility to ensure that every element of the span is assigned. Otherwise, the resulting string could contain random characters.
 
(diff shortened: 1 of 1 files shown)


xml/System/String.xml
```diff
@@ -4364,6 +4364,8 @@ The following example demonstrates the <xref:System.String.CopyTo%2A> method.
           <format type="text/markdown"><![CDATA[
 ## Remarks
 The initial content of the destination span passed to `action` is undefined. Therefore, it is the delegate's responsibility to ensure that every element of the span is assigned. Otherwise, the resulting string could contain random characters.
+
+In order to support interop scenarios, the underlying buffer is guaranteed to be at least 1 greater than represented by the span parameter of the action callback. This additional index represents the null-terminator and if written that is the only value supported, writing any value other than the null-terminator will result in corrupting the string and is considered undefined behavior.
```
gewarren, line 4368:
```suggestion
To support interop scenarios, the underlying buffer is guaranteed to be at least 1 greater than represented by the span parameter of the action callback. This additional index represents the null-terminator and, if written, that is the only value supported. Writing any value other than the null-terminator corrupts the string and is considered undefined behavior.
```

gewarren (approved):

PR #10638, opened 2024-11-04
Collected fixes for `WinHttpHandler`

ManickaP:
Collected fixes for the docs while reading through `WinHttpHandler`.

**not finished yet**

Use "Hide whitespace changes" when reviewing.

### DIFF
--- a/xml/System.Net.Http/WinHttpHandler.xml
+++ b/xml/System.Net.Http/WinHttpHandler.xml
@@ -819,7 +819,7 @@ When this property is set to `true`, all HTTP redirect responses from the server
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- The default value is `None`, which is the recommended value. It allows the operating system to choose the best protocol to use, and to block protocols that are not secure. For more information, see <xref:System.Net.Security.SslProtocols>.
+ The default value is `None`, which is the recommended value. It allows the operating system to choose the best protocol to use, and to block protocols that are not secure. For more information, see <xref:System.Security.Authentication.SslProtocols>.
 
  ]]></format>
         </remarks>


MihaZupan (approved):
Thanks

xml/System.Net.Http/WinHttpHandler.xml
```diff
@@ -316,15 +316,14 @@ When this property is set to `true`, all HTTP redirect responses from the server
...
         <remarks>
-          <format type="text/markdown"><![CDATA[  
-  
-## Remarks  
- The default value for this property is <xref:System.Net.CredentialCache.DefaultCredentials?displayProperty=nameWithType>.  
-  
- ]]></format>
+          <format type="text/markdown"><![CDATA[
+
+The default proxy is used only when <xref:System.Net.Http.WinHttpHandler.Proxy> is set to `null`.
```
CarnaViire, line 324:
```suggestion
The default proxy is used only when <xref:System.Net.Http.WinHttpHandler.WindowsProxyUsePolicy> is set to   <xref:System.Net.Http.WindowsProxyUsePolicy.UseWinHttpProxy> and  <xref:System.Net.Http.WinHttpHandler.Proxy> is set to `null` (both are the default values of the respective properties).
```

CarnaViire: Technically, the default proxy is used only if `WindowsProxyUsePolicy`=`UseWinHttpProxy`; and the validation won't allow you to set `Proxy` in that case...

But if you wanted to specifically highlight the null check we've seen, maybe we could then say it from a different angle:

```suggestion
The value of that property is used only when <xref:System.Net.Http.WinHttpHandler.Proxy> is set to `null`.
```

ManickaP: I removed the remarks completely, it's not that straight forward as just setting `UseWinHttpProxy`, even `UseCustomProxy` can (under certain conditions) lead to using `DefaultProxyCredentials`.

xml/System.Net.Http/WinHttpHandler.xml
```diff
@@ -823,11 +816,11 @@ When this property is set to `true`, all HTTP redirect responses from the server
...
         <remarks>
-          <format type="text/markdown"><![CDATA[  
-  
-## Remarks  
- The default value is `SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12`.
-  
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+ The default value is `None`, which is the recommended value. It allows the operating system to choose the best protocol to use, and to block protocols that are not secure. For more information, see <xref:System.Net.Security.SslProtocols>.
```
CarnaViire, line 822:
```suggestion
 The default value is `None`, which is the recommended value. It allows the operating system to choose the best protocol to use, and to block protocols that are not secure. For more information, see <xref:System.Security.Authentication.SslProtocols>.
```

CarnaViire (approved):
LGTM

PR #10650, opened 2024-11-07
OverloadResolutionAttribute: Copy text from constructor parameter to corresponding property

rickbrew:
I was having trouble figuring out if higher or lower values are "higher" priority. Some classic data structures use lower values for "higher" priorities, for instance. The `Priority` property does not describe this, I eventually found it in the constructor's `priority` parameter. This PR copies the sentence from the latter to the former.

## Summary

Describe your changes here.


BillWagner (approved):
Thanks @rickbrew 

This LGTM, and I'll :shipit:

PR #10652, opened 2024-11-07
Remove SourceLink info for platform extension monikers

gewarren:
These monikers no longer exist, so this info isn't needed.


BillWagner (approved):
This LGTM Let's :shipit:

PR #10669, opened 2024-11-11
MailAddress should not accept email address with consecutive dots

pedrobsaila:
## Summary

Contributes to dotnet/runtime#109590

cc @ManickaP


ManickaP:
Sorry, I haven't thought this fully through, but the change is only for .NET next (.NET 10 - out Nov 2025 next year). We should keep the docs as-is for now. Also we need to keep the original description for previous versions and document that the change applies from .NET 10 onward.
So applying braking-change needs-breaking-change-doc-created label on the runtime PR should be enough for now.

Once again sorry and thank you for working on it.

pedrobsaila:
no worries @ManickaP

PR #10714, opened 2024-11-26
Relate Array.MaxLength to int.MaxValue

GrabYourPitchforks:
## Summary

Document the guarantee that `Array.MaxLength < int.MaxValue`. Typical _for_ loops are subject to failure without this guarantee. Example:

```cs
for (int i = 0; i < someArray.Length; i++)
{
    _ = someArray[i]; // if MaxLength = int.MaxValue, 'i' could overflow to int.MinValue, resulting in illegal access
}
```


bartonjs (approved):

stephentoub (approved):

tannergooding:
Which is to say, in safe code such as the above, there is no risk for an illegal access. There is only risk of an infinite loop.

GrabYourPitchforks:
Thanks @tannergooding for the check! I'll correct the wording and resubmit.

Issue #10434, opened 2024-09-26
SFI: ROPC items for remediation


gewarren:
### Describe the issue or suggestion

- [ ] https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Web.Configuration/ProcessModelSection.xml	password=
- [ ] https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Web.Security/ActiveDirectoryMembershipProvider.xml	ConnectionString,password=
- [ ] https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Web.Security/FormsAuthentication.xml	ConnectionString,password=
- [ ] https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Web.UI.WebControls/ChangePassword.xml	password=
- [ ] https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Web.UI.WebControls/CreateUserWizard.xml	ConnectionString,password=
- [x] https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Web.UI.WebControls.WebParts/WebPartZoneBase.xml	ROPC
- [ ] https://github.com/dotnet/dotnet-api-docs/blob/main/xml/cpp/VS_Snippets_Remoting/Classic HttpWebRequest.ConnectionGroupName Example/CPP/source.cpp	UserNamePassword
- [ ] https://github.com/dotnet/dotnet-api-docs/blob/main/xml/csharp/System.Net/WebRequest/PreAuthenticate/webrequest_preauthenticate.cs	password=
- [ ] https://github.com/dotnet/dotnet-api-docs/blob/main/xml/csharp/VS_Snippets_ADO.NET/DataWorks DbConnectionStringBuilder/CS/source.cs	ConnectionString,password=
- [ ] https://github.com/dotnet/dotnet-api-docs/blob/main/xml/csharp/VS_Snippets_ADO.NET/DataWorks DbConnectionStringBuilder.Values/CS/source.cs	ConnectionString,password=
- [ ] https://github.com/dotnet/dotnet-api-docs/blob/main/xml/csharp/VS_Snippets_ADO.NET/DataWorks OdbcConnectionStringBuilder/CS/source.cs	ConnectionString,pwd=
[...]

PR #10730, opened 2024-12-07
ROPC fixes for System.ServiceModel and System.Net

gewarren:
Contributes to https://github.com/dotnet/dotnet-api-docs/issues/10434.

Hide whitespace changes in diff.

### DIFF
--- a/snippets/csharp/VS_Snippets_CFX/s_ueservicecredentials/cs/app.config
+++ b/snippets/csharp/VS_Snippets_CFX/s_ueservicecredentials/cs/app.config
@@ -1,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.8.1"/></startup></configuration>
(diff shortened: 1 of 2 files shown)


xml/System.Net/HttpWebRequest.xml
```diff
@@ -1899,14 +1899,7 @@ Both <xref:System.Net.HttpWebRequest> constructors are obsolete and should not b
 
 [!INCLUDE [obsolete-notice](~/includes/web-request-obsolete.md)]
 
- The <xref:System.Net.HttpWebRequest.ConnectionGroupName%2A> property enables you to associate a request with a connection group. This is useful when your application makes requests to one server for different users, such as a Web site that retrieves customer information from a database server.
-
-## Examples
```
antonfirsov, line 1904: Why are we deleting this example? If it really should be deleted, shouldn't the snippet files also go?

BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready

xml/System.Net/HttpWebRequest.xml
gewarren: The tool found a lot of orphaned snippets, so I'll clean them up in a separate PR instead of dirtying this one.

PR #10738, opened 2024-12-09
Fix signature for nullability

StephaneDelcroix:
## Summary

Fix the sample for EventHandler so the handler signature matches the delegate, and avoids warning


gewarren (approved):
Thanks @StephaneDelcroix. I updated the others in the directory to match too.

PR #10750, opened 2024-12-10
Add remark about timeout scope on HttpCompletionOptions enum

foviedoITBA:
## Summary

Update the HttpCompletionOption enum docs to clarify its effect on the HttpClient timeout.

### Issue

HttpCompletionOptions effectively affect the scope of the timeout on HttpClient operations. The timeout always applies up until the HTTP client operations complete. Because HttpCompletionOptions control when the operations complete, they effectively change when the timeout stops applying. This might be implicit and obvious in hindsight, but it's routinely overlooked by developers. This MVP article is frequently cited by developers using ResponseHeadersRead, yet the article mentions nothing about the timeout scope: https://www.stevejgordon.co.uk/using-httpcompletionoption-responseheadersread-to-improve-httpclient-performance-dotnet

### Solution

Update the HttpCompletionOption enum docs to explicitly call this behavior out.

### DIFF
--- a/snippets/csharp/System.Net.Http/HttpCompletionOption/HttpCompletionOptionSnippets.csproj
+++ b/snippets/csharp/System.Net.Http/HttpCompletionOption/HttpCompletionOptionSnippets.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net9.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
   </PropertyGroup>


xml/System.Net.Http/HttpCompletionOption.xml
```diff
@@ -36,7 +36,27 @@
...
   <Docs>
     <summary>Indicates if <see cref="T:System.Net.Http.HttpClient" /> operations should be considered completed either as soon as a response is available, or after reading the entire response message including the content.</summary>
-    <remarks>To be added.</remarks>
+    <remarks>
+      <format type="text/markdown"><![CDATA[
+
+## Remarks
+ The <xref:HttpCompletionOption> value effectively affects the scope of the timeout specified in the <see cref="T:System.Net.Http.HttpClient" /> operation options when reading a response. The timeout on the <see cref="T:System.Net.Http.HttpClient" /> always applies on the relevant invoked methods up until the point where those methods complete/return. Crucially, when using the <see cref="F:System.Net.Http.HttpCompletionOption.ResponseHeadersRead" /> option, the timeout applies only up to where the headers end and the content starts. The content reading operation needs to be timed out separately in case the server promptly returns the status line and headers but takes too long to return the content. Below is an example illustrating this point:
+
+```csharp
```
foviedoITBA, line 45: I only have an example for C#, yet the article is written for F#, VB and C++ too. Should I add examples for those languages as well?

foviedoITBA, line 45: The .NET contribution docs mention that code snippets have to be supplied in standalone projects in a set folder structure: https://learn.microsoft.com/en-us/contribute/content/dotnet/dotnet-contribute#contribution-process. However, that seems to be for .md docs, not these XMLs (I could be wrong though), and I don't see any similar folder structure in these docs (maybe it's only for https://github.com/dotnet/docs ?), so I'm following the existing pattern (borrowed from HttpClient docs).

foviedoITBA:
The .NET docs guidelines mention an ongoing effort for all API documentation to converge into triple slash comments in source code: https://learn.microsoft.com/en-us/contribute/content/dotnet/api-documentation
However, I don't think the System.Net.Http namespace gone through this process yet, since there's no documentation in source code: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpCompletionOption.cs. 
Hence, I'm doing the edit directly on the backing XML documentation page.

xml/System.Net.Http/HttpCompletionOption.xml
```diff
@@ -36,7 +36,27 @@
   </Base>
   <Docs>
     <summary>Indicates if <see cref="T:System.Net.Http.HttpClient" /> operations should be considered completed either as soon as a response is available, or after reading the entire response message including the content.</summary>
-    <remarks>To be added.</remarks>
+    <remarks>
```
foviedoITBA, line 39: The .NET docs guidelines mention an ongoing effort for all API documentation to converge into triple slash comments in source code: https://learn.microsoft.com/en-us/contribute/content/dotnet/api-documentation
However, I don't think the System.Net.Http namespace gone through this process yet, since there's no documentation in source code: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpCompletionOption.cs.
Hence, I'm doing the edit directly on the backing XML documentation page.

foviedoITBA, line 39: I couldn't find a way to preview these changes, all the tooling seems to work only for markdown files, not these XMLs: https://review.learn.microsoft.com/en-us/help/get-started/?branch=main&tabs=learn. So I don't really know what these changes look like.

xml/System.Net.Http/HttpCompletionOption.xml
```diff
@@ -36,7 +36,27 @@
...
+
+```csharp
+var httpClient = new HttpClient();
+httpClient.Timeout = TimeSpan.FromSeconds(30);
+using (var response = await httpClient.GetAsync("http://localhost:12345/", HttpCompletionOption.ResponseHeadersRead)) // 30-second timeout but ONLY up until past the headers
+{
+    // Do other stuff that doesn't rely on the content first, like status code validation
+    response.EnsureSuccessStatusCode();
+
+    var content = await response.Content.ReadAsStringAsync(); // NO TIMEOUT
```
foviedoITBA, line 53: The solution is different for .NET Framework vs .NET. In .NET, ReadAsStringAsync overrides accept a CancellationToken: https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpcontent.readasstringasync?view=net-9.0#system-net-http-httpcontent-readasstringasync(system-threading-cancellationtoken). In .NET Framework, this override doesn't exist, and so this workaround is usually used by developers: https://stackoverflow.com/questions/25985416/how-can-i-set-a-timeout-for-an-async-function-that-doesnt-accept-a-cancellation.

Adding this to the example would mean either:
a) Adding code with conditional compilation (`#if NETFRAMEWORK /*timeout workaround*/ #else /*cancellation token version*/ #endif`)
b) Adding separate examples for .NET vs .NET Framework

I don't know if this take away clarity from the point we're illustrating with the example though, and I don't know what the community's preference would be for such case.

xml/System.Net.Http/HttpCompletionOption.xml
MSDN-WhiteKnight: Examples definitely should be in separate project. Here's example: https://github.com/dotnet/dotnet-api-docs/tree/main/snippets/csharp/System.Reflection.Metadata/MetadataReader

and where it's referenced: https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Reflection.Metadata/MetadataReader.xml#L48

xml/System.Net.Http/HttpCompletionOption.xml
MSDN-WhiteKnight: You can't preview changes. Repository maintainers can preview using their internal tool. Currently your PR has build warnings due to using `<see>` tags in markdown, so it will likely render incorrectly in preview. You need to fix warnings by using xrefs instead of `<see>` to reference APIs, like: `<xref:System.Net.Http.HttpClient>`. Look existing remarks for examples.

foviedoITBA: I've replaced the `<see>` tags with `<xref>`, hopefully it renders correctly.

xml/System.Net.Http/HttpCompletionOption.xml
foviedoITBA: Thank you. I've moved the code sample to a standalone snippet along the lines of the MetadataReader example you provided.

snippets/csharp/System.Net.Http/HttpCompletionOption/HttpCompletionOptionSnippets.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            //<SnippetHttpCompletionOption>
+            var httpClient = new HttpClient();
+            httpClient.Timeout = TimeSpan.FromSeconds(30);
+            using (var response = await httpClient.GetAsync("http://localhost:12345/", HttpCompletionOption.ResponseHeadersRead)) // 30-second timeout but ONLY up until past the headers
+            {
+                // Do other stuff that doesn't rely on the content first, like status code validation
+                response.EnsureSuccessStatusCode();
+
+                var content = await response.Content.ReadAsStringAsync(); // NO TIMEOUT
```
foviedoITBA, line 19: The solution is different for .NET Framework vs .NET. In .NET, ReadAsStringAsync overrides accept a CancellationToken: https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpcontent.readasstringasync?view=net-9.0#system-net-http-httpcontent-readasstringasync(system-threading-cancellationtoken). In .NET Framework, this override doesn't exist, and so this workaround is usually used by developers: https://stackoverflow.com/questions/25985416/how-can-i-set-a-timeout-for-an-async-function-that-doesnt-accept-a-cancellation.

Adding this to the example would mean either:
a) Adding code with conditional compilation (`#if NETFRAMEWORK /*timeout workaround*/ #else /*cancellation token version*/ #endif`)
b) Adding separate examples for .NET vs .NET Framework

I don't know if this take away clarity from the point we're illustrating with the example though, and I don't know what the community's preference would be for such case.

foviedoITBA:
@dotnet-policy-service agree company="Microsoft"

xml/System.Net.Http/HttpCompletionOption.xml
MihaZupan: C# only is fine IMO

xml/System.Net.Http/HttpCompletionOption.xml
MihaZupan: I believe you are correct, and this is currently the right place to update

snippets/csharp/System.Net.Http/HttpCompletionOption/HttpCompletionOptionSnippets.cs
```diff
@@ -0,0 +1,24 @@
...
+            }
```
MihaZupan, line 20:
```suggestion
            using var httpClient = new HttpClient();
            httpClient.Timeout = TimeSpan.FromSeconds(30);
            httpClient.MaxResponseContentBufferSize = 1_000;

            // Because we're specifying the ResponseHeadersRead option,
            // the 30-second timeout applies only up until the headers are received.
            // It does not affect future operations that interact with the response content.
            using HttpResponseMessage response = await httpClient.GetAsync(
                "http://localhost:12345/",
                HttpCompletionOption.ResponseHeadersRead);

            // Do other checks that don't rely on the content first, like status code validation.
            response.EnsureSuccessStatusCode();

            // Since the HttpClient.Timeout will not apply to reading the content,
            // you must enforce it yourself, for example by using a CancellationTokenSource.
            using var cancellationSource = new CancellationTokenSource(TimeSpan.FromSeconds(15));
            
            // If you wish to enforce the MaxResponseContentBufferSize limit as well,
            // you can do so by calling the LoadIntoBufferAsync helper.
            // await response.Content.LoadIntoBufferAsync(1_000, cancellationSource.Token);

            // Make sure to pass the CancellationToken to all methods.
            string content = await response.Content.ReadAsStringAsync(cancellationSource.Token);
```

xml/System.Net.Http/HttpCompletionOption.xml
```diff
@@ -36,7 +36,17 @@
   </Base>
   <Docs>
     <summary>Indicates if <see cref="T:System.Net.Http.HttpClient" /> operations should be considered completed either as soon as a response is available, or after reading the entire response message including the content.</summary>
-    <remarks>To be added.</remarks>
+    <remarks>
+      <format type="text/markdown"><![CDATA[
+
+## Remarks
+ The <xref:System.Net.Http.HttpCompletionOption> value effectively affects the scope of the timeout specified in the <xref:System.Net.Http.HttpClient> operation options when reading a response. The timeout on the <xref:System.Net.Http.HttpClient> always applies on the relevant invoked methods up until the point where those methods complete/return. Crucially, when using the <xref:System.Net.Http.HttpCompletionOption.ResponseHeadersRead> option, the timeout applies only up to where the headers end and the content starts. The content reading operation needs to be timed out separately in case the server promptly returns the status line and headers but takes too long to return the content. Below is an example illustrating this point:
```
MihaZupan, line 43:
```suggestion

> [!WARNING]
> The <xref:System.Net.Http.HttpCompletionOption> value affects the scope of the timeout specified in the <xref:System.Net.Http.HttpClient> options when reading a response. The timeout on the <xref:System.Net.Http.HttpClient> always applies on the relevant invoked methods up until the point where those methods complete/return. Crucially, when using the <xref:System.Net.Http.HttpCompletionOption.ResponseHeadersRead> option, the timeout applies only up to where the headers end and the content starts. The content reading operation needs to be timed out separately in case the server promptly returns the status line and headers but takes too long to return the content.
> This consideration also applies to the <xref:System.Net.Http.HttpClient.MaxResponseContentBufferSize%2A> property. The limit is only enforced when using <xref:System.Net.Http.HttpCompletionOption.ResponseContentRead>.
> Below is an example illustrating this point:
```

Added a note about this also applying to `MaxResponseContentBufferSize`.
Not sure if I've used the correct syntax, but IMO this is non-obvious enough that it should be a warning.

MihaZupan (approved):
Thank you for the help here and sorry about the delayed review

snippets/csharp/System.Net.Http/HttpCompletionOption/HttpCompletionOptionSnippets.cs
MihaZupan: I've changed the suggested sample a bit, I think starting with current APIs only is okay.

snippets/csharp/System.Net.Http/HttpCompletionOption/HttpCompletionOptionSnippets.csproj
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
```
MihaZupan, line 5:
```suggestion
    <TargetFramework>net9.0</TargetFramework>
```
Bump sample to .NET 9 (using new `LoadIntoBufferAsync` overload)

PR #10808, opened 2025-01-05
Added OpenFileDialog.SelectReadOnly Property Value

denyshon:
## Summary

Added OpenFileDialog.Select the ReadOnly Property Value section, which includes information about the property's default value.


gewarren (approved):
Thanks @denyshon

PR #10821, opened 2025-01-08
Moving details from blog post

ManickaP:
Added notes to remarks for SP and SPM properties about their behavior on modern .NET.

Ready for review.

**Use "Hide whitespace changes" for comparison.**

### DIFF
--- a/xml/System.Net/ServicePoint.xml
+++ b/xml/System.Net/ServicePoint.xml
@@ -324,7 +324,7 @@ Although a <xref:System.Net.ServicePoint> object can make multiple connections t
 ## Remarks
 
 > [!NOTE]
->  Since .NET 9, this property is not implemented.
+>  This property is only implemented on .NET Framework.
 
 ## Examples
  The following code example displays the value of this property.
@@ -658,7 +658,7 @@ Although a <xref:System.Net.ServicePoint> object can make multiple connections t
  The <xref:System.Net.ServicePoint.CurrentConnections> property contains the number of open Internet connections associated with this <xref:System.Net.ServicePoint> object. The value of <xref:System.Net.ServicePoint.CurrentConnections> cannot exceed that of <xref:System.Net.ServicePoint.ConnectionLimit>.
 
 > [!NOTE]
->  Since .NET 9, this property is not implemented.
+>  This property is only implemented on .NET Framework.
 
 ## Examples
  The following code example uses the <xref:System.Net.ServicePoint.CurrentConnections> property to determine the number of open Internet connections associated with this <xref:System.Net.ServicePoint> object.
@@ -1190,7 +1190,7 @@ The `GetHashCode` method returns a hash code of this instance. This value can be
  Pipelining is described in detail in IETF RFC 2616 section 8.1.2.2.
 
 > [!NOTE]
->  Since .NET 9, this property is not implemented.
+>  This property is only implemented on .NET Framework.
 
 ## Examples
  The following code example displays the value of this property.
--- a/xml/System.Net/ServicePointManager.xml
+++ b/xml/System.Net/ServicePointManager.xml
@@ -423,7 +423,7 @@
 ## Remarks
 
 > [!NOTE]
->  Since .NET 9, this property is not implemented.
+>  This property is only implemented on .NET Framework.
 
 ## Examples
  The following code example sets this property.
@@ -484,7 +484,7 @@
  When more than one IP address is associated with a host name, a DNS resolution normally returns only the first IP address. If you set this property to `true`, then subsequent DNS resolutions will cycle through all available IP addresses for a particular host. This option is useful when a service uses DNS as a load balancing mechanism between servers or server clusters.
 
 > [!NOTE]
->  Since .NET 9, this property is not implemented.
+>  This property is only implemented on .NET Framework.
 
 ## Examples
  The following code example sets this property.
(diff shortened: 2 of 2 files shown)


xml/System.Net/ServicePointManager.xml
```diff
@@ -123,22 +123,25 @@
...
-  
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+ When the <xref:System.Net.ServicePointManager.CertificatePolicy%2A> property is set to an <xref:System.Net.ICertificatePolicy> interface object, the <xref:System.Net.ServicePointManager> object uses the certificate policy defined in that instance instead of the default certificate policy.
+
+ The default certificate policy allows valid certificates and valid certificates that have expired.
+
+> [!NOTE]
+>  On modern .NET, maps to [`CertificateRevocationCheckMode`](https://learn.microsoft.com/dotnet/api/system.net.security.sslclientauthenticationoptions.certificaterevocationcheckmode) on [`SocketsHttpHandler.SslOptions`](https://learn.microsoft.com/dotnet/api/system.net.http.socketshttphandler.ssloptions)
```
antonfirsov, line 134:
```suggestion
>  On modern .NET, maps to <xref:System.Net.Security.SslClientAuthenticationOptions.CertificateRevocationCheckMode%2A?displayProperty=nameWithType> on <xref:System.Net.Http.SocketsHttpHandler.SslOptions%2A?displayProperty=nameWithType>.
```

ManickaP: Fixed.

xml/System.Net/ServicePoint.xml
```diff
@@ -187,14 +187,17 @@
...
+## Remarks
 
 [!INCLUDE [obsolete-notice](~/includes/web-request-obsolete.md)]
 
- Some load balancing techniques require a client to use a specific local IP address and port number, rather than <xref:System.Net.IPAddress.Any?displayProperty=nameWithType> (or <xref:System.Net.IPAddress.IPv6Any?displayProperty=nameWithType> for Internet Protocol Version 6) and an ephemeral port. Your <xref:System.Net.ServicePoint.BindIPEndPointDelegate> can satisfy this requirement.  
-  
+ Some load balancing techniques require a client to use a specific local IP address and port number, rather than <xref:System.Net.IPAddress.Any?displayProperty=nameWithType> (or <xref:System.Net.IPAddress.IPv6Any?displayProperty=nameWithType> for Internet Protocol Version 6) and an ephemeral port. Your <xref:System.Net.ServicePoint.BindIPEndPointDelegate> can satisfy this requirement.
+
+> [!NOTE]
+>  On modern .NET, <xref:System.Net.HttpWebRequest> uses <xref:System.Net.Http.SocketsHttpHandler.ConnectCallback?displayProperty=nameWithType> to bind the underlying socket to the endpoint from this delegate.
```
antonfirsov, line 199: > modern .NET

I don't see us using this term anywhere else in the API docs. Moreover, these properties are unimplemented on .NET 8, which is still LTS. I would prefer us being specific here and elaborate the behavior per-version.

ManickaP: Some of these were implemented in the past, some of them are there from the very beginning of .NET Core, some have been implemented this latest release. For that reason I chose the cop out of "Modern .NET" here. As figuring out what exact version to use for each of these is gonna be quite time consuming.
I can do that if you think it's worth the effort, but I don't really want to :smiling_face_with_tear:

antonfirsov: @gewarren any general guidelines about the terminlogy and the level of detail we should use here?

gewarren: I agree with Anton. "Modern .NET" doesn't provide clarity and will almost certainly lead to negative feedback for confused customers. Also, 10 years from now, the versions you're referring to won't be considered "modern .NET". Do you want me to try and help figure out the versions?

ManickaP: As another cop out, I can just put .NET 9 everywhere. It's not wrong info, just in some cases it might be working for older versions as well.

> Do you want me to try and help figure out the versions?

Feel free to try, but just FYI, it's not about looking at the property itself, but at its usages in different classes (mainly `HttpWebRequest` AFAIK) and seeing if it's used in way that it will affect something.

antonfirsov: I'm fine putting .NET 9+ everywhere.

xml/System.Net/ServicePoint.xml
```diff
@@ -317,15 +320,19 @@ Although a <xref:System.Net.ServicePoint> object can make multiple connections t
...
-          <format type="text/markdown"><![CDATA[  
-  
-## Examples  
- The following code example displays the value of this property.  
-  
+          <format type="text/markdown"><![CDATA[
+## Remarks
+
+> [!NOTE]
+>  Since .NET 9, this property is not implemented.
```
antonfirsov, line 327: This sounds like if there was an implementation in .NET 8, but then we removed it. For unimplemented properties, it would be more accurate to state that

```suggestion
>  This property is only implemented on .NET Framework.
```

ManickaP: :facepalm: I did global replace and didn't read it.

ManickaP: Fixed.

antonfirsov (approved):

PR #10822, opened 2025-01-08
File.OpenWrite doc improvements

### LINKED ISSUE #10813: FileInfo docs poor
### Type of issue

Missing information

### Description

1. having the **Microsoft.NETCore.App** you don't need the
```
using System;
using System.IO;
```
2.  remove all the excess "//" lines in the results
3. shorter 2nd run produces "Second run test of the OpenWrite method" not "Second runtest of the OpenWrite method"

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.io.fileinfo.openwrite?view=net-9.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.IO/FileInfo.xml

a55249a4-99a2-f0c8-6036-80739ea66033

### Article author

@dotnet-bot

adamsitnik:
## Summary

Some minor polishing:

1. remove redundant comments from the snippet
2. update the description to reflect the text that would actually got printed

Fixes #10813

### DIFF
--- a/xml/System.IO/FileInfo.xml
+++ b/xml/System.IO/FileInfo.xml
@@ -2137,7 +2137,7 @@ The following example demonstrates moving a file to a different location and ren
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- The <xref:System.IO.FileInfo.OpenWrite%2A> method opens a file if one already exists for the file path, or creates a new file if one does not exist. For an existing file, it does not append the new text to the existing text. Instead, it overwrites the existing characters with the new characters. If you overwrite a longer string (such as "This is a test of the OpenWrite method") with a shorter string (like "Second run"), the file will contain a mix of the strings ("Second run.test the OpenWrite method.").
+ The <xref:System.IO.FileInfo.OpenWrite%2A> method opens a file if one already exists for the file path, or creates a new file if one does not exist. For an existing file, it does not append the new text to the existing text. Instead, it overwrites the existing characters with the new characters. If you overwrite a longer string (such as "This is a test of the OpenWrite method") with a shorter string (like "Second run"), the file will contain a mix of the strings ("Second run test of the OpenWrite method.").
 
 
 


xml/System.IO/FileInfo.xml
```diff
@@ -2137,7 +2137,7 @@ The following example demonstrates moving a file to a different location and ren
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- The <xref:System.IO.FileInfo.OpenWrite%2A> method opens a file if one already exists for the file path, or creates a new file if one does not exist. For an existing file, it does not append the new text to the existing text. Instead, it overwrites the existing characters with the new characters. If you overwrite a longer string (such as "This is a test of the OpenWrite method") with a shorter string (like "Second run"), the file will contain a mix of the strings ("Second runtest of the OpenWrite method").
+ The <xref:System.IO.FileInfo.OpenWrite%2A> method opens a file if one already exists for the file path, or creates a new file if one does not exist. For an existing file, it does not append the new text to the existing text. Instead, it overwrites the existing characters with the new characters. If you overwrite a longer string (such as "This is a test of the OpenWrite method") with a shorter string (like "Second run"), the file will contain a mix of the strings ("Second run.test the OpenWrite method.").
```
gewarren, line 2140:
```suggestion
 The <xref:System.IO.FileInfo.OpenWrite%2A> method opens a file if one already exists for the file path, or creates a new file if one does not exist. For an existing file, it does not append the new text to the existing text. Instead, it overwrites the existing characters with the new characters. If you overwrite a longer string (such as "This is a test of the OpenWrite method") with a shorter string (like "Second run"), the file will contain a mix of the strings ("Second run test of the OpenWrite method.").
```

gewarren (approved):
Thank you

xml/System.IO/FileInfo.xml
adamsitnik: The suggestion is invalid. To tell the long story short we open a file and write:

"This is to test the OpenWrite method."

and then overwrite it with:

""Second run."

which gives us

"Second run.test the OpenWrite method."

It's not intuitive, but this is the content of the file after described change.

gewarren: Are you referring to an actual example, or just the text from the remarks? Because both of your sentences don't seem to match the remarks sentences (the first one has the word "of" and the second one doesn't have a period at the end).

adamsitnik: @gewarren  You are right! Thank you for your thoroughness and apologies for the confusion.

PR #10830, opened 2025-01-10
Clarify index parameters

### LINKED ISSUE #10829: Unclear verbiage
### Type of issue

Typo

### Description

The verbiage "A 32-bit integer that represents the index in array at which copying begins." is extremely unclear/unhelpful.

It only says it is the initial index value when the copy operation begins. It doesn't say if it is the source's index or destination's index. Wording should be clarified.

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.array.copyto?view=net-8.0#system-array-copyto(system-array-system-int32)

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System/Array.xml

60447b8e-6071-b21e-2dfe-e3013ed84225

### Article author

@dotnet-bot

gewarren:
Fixes #10829

### DIFF
--- a/xml/System/Array.xml
+++ b/xml/System/Array.xml
@@ -1471,9 +1471,9 @@
       </Parameters>
       <Docs>
         <param name="sourceArray">The <see cref="T:System.Array" /> that contains the data to copy.</param>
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

PR #10842, opened 2025-01-15
Update HttpRequestException.xml

a1ndreay:
This class does not provide a statusCode property according to .NET Framework 4.6.1

## Summary

Describe your changes here.


a1ndreay:
Mistake on my side, I was looking at the wrong version of the documentation

PR #10853, opened 2025-01-20
Update Path.GetDirectoryName Snippet3

DickBaker:
## Summary
the current docs are outdated to NRT and confuse quotes
1. variables type should be string**?** not string
2. uses **old-style** format string (**new** would use interpolated syntax)
3. _filepath_ reassignment should use conditional syntax
4. results suggest using single-quote [i.e. char not string], _ugh_
5. the final returned result is **null** but demo output shows **''** which is factually and visually correct but confusingly indistinguishable from "" emptystring

Describe your changes here.

I propose this replacement C# code to improve this Snippet3 accuracy and clarity


BillWagner (approved):
This LGTM @DickBaker 

I'll :shipit: now.

PR #10860, opened 2025-01-22
Fix up see also links

gewarren:
- Mark as "article" instead of "external doc" if it links to learn.microsoft.com docs.
- Replace go.microsoft.com with target URL.
- Remove links whose target no longer matches the expected page title.

(Hide whitespace changes.)


ManickaP (approved):
Networking LGTM, thanks.

cheenamalhotra (approved):

BillWagner (approved):
This LGTM @gewarren 

You can :shipit: when ready.

PR #10893, opened 2025-01-31
Remove System.Speech from source-of-truth assembly list

gewarren:
See https://github.com/dotnet/runtime/issues/87711.


BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

Issue #10925, opened 2025-02-07
System.Net.Sockets.NetworkStream.Read(Byte[], Int32, Int32)


algor1:
### Type of issue

Missing information

### Description

Hi
Please add info in chapter Read(Byte[], Int32, Int32) of System.Net.Sockets.NetworkStream to Returns row:

Returns
The number of bytes read from the [NetworkStream] , **or 0 if the socket is closed**

I managed to find this info only in comments in the code but it is really important to know what it returns when socket is closed.

Thanks in advance.

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.networkstream.read?view=net-9.0#system-net-sockets-networkstream-read(system-byte()-system-int32-system-int32)

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Net.Sockets/NetworkStream.xml

### Document Version Independent Id

b40c2ef4-b8a0-8de3-e71d-baf1aaae541b

### Article author

@karelz

antonfirsov:
@algor1 that code comment is incorrect, see https://github.com/dotnet/dotnet-api-docs/pull/10960#pullrequestreview-2617736535

PR #10960, opened 2025-02-14
Update comment on return value of NetworkStream.Read* variants

rzikm:
Fixes https://github.com/dotnet/dotnet-api-docs/issues/10925.

Mentions that returns 0 when socket was closed.

### DIFF
--- a/xml/System.Net.Sockets/NetworkStream.xml
+++ b/xml/System.Net.Sockets/NetworkStream.xml
@@ -1107,7 +1107,7 @@ The Close method frees both unmanaged and managed resources associated with the 
       <Docs>
         <param name="asyncResult">An <see cref="T:System.IAsyncResult" /> that represents an asynchronous call.</param>
         <summary>Handles the end of an asynchronous read.</summary>
-        <returns>The number of bytes read from the <see cref="T:System.Net.Sockets.NetworkStream" />.</returns>
+        <returns>The number of bytes read from the <see cref="T:System.Net.Sockets.NetworkStream" />, or 0 if the underlying socket was closed.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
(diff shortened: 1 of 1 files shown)


antonfirsov (changes requested):
The comment in the code qouted in #10925 is wrong.

When the receiver socket is closed an exception will be thrown (ODE on abortive close  -  or `SocketException(SocketError.Shutdown)` if graceful shutdown of receives has been initiated.).

`0` bytes will be received when the peer has initiated a graceful termination of writes (either by calling `Shutdown(Send / Both)` or by calling `Close(0)`, which means receiving EOF as documented for base class [`Stream.Read*`](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=net-9.0#system-io-stream-read(system-span((system-byte)))) methods:

> The total number of bytes read into the buffer. This can be less than the size of the buffer if that many bytes are not currently available, or zero (0) if the buffer's length is zero or the end of the stream has been reached.

When the the peer socket is aborted and sent RST, an exception `SocketException(ConnectionReset)` will be thrown.

So IMO the docs should be either copypaste `Stream.Read` content, or slightly augment it, eg.:

> The total number of bytes read into the buffer. This can be less than the size of the buffer if that many bytes are not currently available, or zero (0) if the buffer's length is zero or the end of the stream has been reached, _that is the peer socket has initiated graceful shutdown of writes_.

algor1:
Both suggested answers are correct, but they do not answer the user's questions about the method, but simply refer to look for information in other sources

When we come to look at the instructions, we want to see information about how the methods work. If we simply copy from the Read method, the users will find out that this method works a little differently from other streams only by making a bug in their code.

I went to the documentation to find out exactly the differences in the behavior of this method:
What does the Read method return when the data has not yet arrived?
And when does the Read method return 0. (because I already have a bug in my code)
And I did not find the answers.

> `0` bytes will be received when the peer has initiated a graceful termination of writes (either by calling `Shutdown(Send / Both)` or by calling `Close(0)`
Sounds much better and describes behaviour

And of course, this is just my opinion. Perhaps my point of view does not coincide with the point of view of the majority

antonfirsov:
Closing this in favor of https://github.com/dotnet/dotnet-api-docs/pull/10983.

PR #10983, opened 2025-02-19
Detailed docs on Socket.Read* and NetworkStream.Receive* return values

### LINKED ISSUE #10925: System.Net.Sockets.NetworkStream.Read(Byte[], Int32, Int32)

antonfirsov:
The wording is mostly borrowed from the `Stream.Read*` variants but saying `the peer socket performed a graceful shutdown` instead of `end of the stream has been reached`, which describes the EOF reason more specifically in this case.

Fixes #10925.

### DIFF
--- a/xml/System.Net.Sockets/NetworkStream.xml
+++ b/xml/System.Net.Sockets/NetworkStream.xml
@@ -1107,7 +1107,7 @@ The Close method frees both unmanaged and managed resources associated with the 
       <Docs>
         <param name="asyncResult">An <see cref="T:System.IAsyncResult" /> that represents an asynchronous call.</param>
         <summary>Handles the end of an asynchronous read.</summary>
-        <returns>The number of bytes read from the <see cref="T:System.Net.Sockets.NetworkStream" />, between zero (0) and the number of bytes requested. The method returns zero (0) only if zero bytes were requested or if no more bytes are available because the peer socket performed a graceful shutdown. If zero bytes are requested, read operations may complete immediately or may not complete until at least one byte is available (but without consuming any data).</returns>
+        <returns>The number of bytes read from the <see cref="T:System.Net.Sockets.NetworkStream" />, between zero (0) and the number of bytes requested. The method returns zero (0) only if zero bytes were requested or if no more bytes are available because the peer socket performed a graceful shutdown. If zero bytes are requested, read operations might complete immediately or might not complete until at least one byte is available (but without consuming any data).</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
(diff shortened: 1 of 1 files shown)


xml/System.Net.Sockets/NetworkStream.xml
```diff
@@ -1466,7 +1466,7 @@ The Close method frees both unmanaged and managed resources associated with the
       <Docs>
         <param name="buffer">A region of memory to store data read from the <see cref="T:System.Net.Sockets.NetworkStream" />.</param>
         <summary>Reads data from the <see cref="T:System.Net.Sockets.NetworkStream" /> and stores it to a span of bytes in memory.</summary>
-        <returns>The number of bytes read from the <see cref="T:System.Net.Sockets.NetworkStream" />.</returns>
+        <returns>The total number of bytes read into the buffer between zero (0) and the length of the buffer. The method returns zero (0) only if zero bytes were requested or if no more bytes are available because the peer socket performed a graceful shutdown.</returns>
```
gewarren, line 1469:
```suggestion
        <returns>The total number of bytes read into the buffer, between zero (0) and the length of the buffer. The method returns zero (0) only if zero bytes were requested or if no more bytes are available because the peer socket performed a graceful shutdown.</returns>
```

xml/System.Net.Sockets/NetworkStream.xml
```diff
@@ -1107,7 +1107,7 @@ The Close method frees both unmanaged and managed resources associated with the
       <Docs>
         <param name="asyncResult">An <see cref="T:System.IAsyncResult" /> that represents an asynchronous call.</param>
         <summary>Handles the end of an asynchronous read.</summary>
-        <returns>The number of bytes read from the <see cref="T:System.Net.Sockets.NetworkStream" />.</returns>
+        <returns>The number of bytes read from the <see cref="T:System.Net.Sockets.NetworkStream" />, between zero (0) and the number of bytes requested. The method returns zero (0) only if zero bytes were requested or if no more bytes are available because the peer socket performed a graceful shutdown. If zero bytes are requested, read operations may complete immediately or may not complete until at least one byte is available (but without consuming any data).</returns>
```
gewarren, line 1110:
```suggestion
        <returns>The number of bytes read from the <see cref="T:System.Net.Sockets.NetworkStream" />, between zero (0) and the number of bytes requested. The method returns zero (0) only if zero bytes were requested or if no more bytes are available because the peer socket performed a graceful shutdown. If zero bytes are requested, read operations might complete immediately or might not complete until at least one byte is available (but without consuming any data).</returns>
```

rzikm (approved):
LGTM, modulo existing comments

PR #11028, opened 2025-02-28
AppDomain.FriendlyName doesn't include file extension in .NET (Core)

### LINKED ISSUE #10868: Breaking change in AppDomain.FriendlyName
### Type of issue

Missing information

### Description

In the documentation Remarks it says that the Friendly Name of the default application will also contain the extension but it is not the case any more. In .NET Framework 4.7.2 it was in deed including extension but not in .NET 8 that I am using right now. Maybe it is also not including in other NET 5+ versions. I think it shall be described as a breaking change against old .NET Framework.

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.appdomain.friendlyname?view=net-8.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System/AppDomain.xml

d991790a-6925-9ea6-c850-ce25a3e21b70

### Article author

@dotnet-bot

gewarren:
Fixes #10868.

### DIFF
--- a/xml/System/AppDomain.xml
+++ b/xml/System/AppDomain.xml
@@ -5804,7 +5804,7 @@ This method overload uses the <xref:System.AppDomainSetup> information from the 
 
 ## Remarks
 
-The friendly name of the default application domain is the file name of the process executable. For example, if the executable used to start the process is `"c:\MyAppDirectory\MyAssembly.exe"`, the friendly name of the default application domain is `"MyAssembly.exe"`. (In .NET (Core) versions, the friendly name doesn't include the file extension.)
+The friendly name of the default application domain is the file name of the process executable. For example, if the executable used to start the process is `"c:\MyAppDirectory\MyAssembly.exe"`, the friendly name of the default application domain is `"MyAssembly.exe"`. (In .NET Core and .NET 5+, the friendly name doesn't include the file extension.)
 
 ## Examples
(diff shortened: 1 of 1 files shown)


xml/System/AppDomain.xml
```diff
@@ -5804,9 +5803,8 @@ This method overload uses the <xref:System.AppDomainSetup> information from the
           <format type="text/markdown"><![CDATA[
 
 ## Remarks
- The friendly name of the default application domain is the file name of the process executable. For example, if the executable used to start the process is `"c:\MyAppDirectory\MyAssembly.exe"`, the friendly name of the default application domain is `"MyAssembly.exe"`.
-
 
+The friendly name of the default application domain is the file name of the process executable. For example, if the executable used to start the process is `"c:\MyAppDirectory\MyAssembly.exe"`, the friendly name of the default application domain is `"MyAssembly.exe"`. (In .NET (Core) versions, the friendly name doesn't include the file extension.)
```
BillWagner, line 5807: Should this call out .NET 5 + as well as .NET Core?

BillWagner (approved):
This LGTM @gewarren 

I had one clarifying question for you, and then you can :shipit:

xml/System/AppDomain.xml
gewarren:
```suggestion
The friendly name of the default application domain is the file name of the process executable. For example, if the executable used to start the process is `"c:\MyAppDirectory\MyAssembly.exe"`, the friendly name of the default application domain is `"MyAssembly.exe"`. (In .NET Core and .NET 5+, the friendly name doesn't include the file extension.)
```

PR #11030, opened 2025-02-28
Fix (Http)WebRequest (Begin|End)GetResponse samples

### LINKED ISSUE #8662: HttpWebRequest.BeginGetResponse example doesn't correctly follow the APM pattern
Per @stephentoub's comment in https://github.com/microsoft/referencesource/issues/177#issuecomment-1327368738 regarding [the documentation of `HttpWebRequest.BeginGetResponse`](https://learn.microsoft.com/en-us/dotnet/api/system.net.httpwebrequest.begingetresponse):

> That code sample is buggy and does not correctly follow the APM pattern. BeginXx methods may complete the operation synchronously, which is why the IAsyncResult.CompletedSynchronously property exists. The callback needs to check that property, and if it's true, exit immediately and instead allow the BeginXx call site to perform the continuation. Otherwise, a string of synchronously completing operations may stack dive. The docs should be fixed

Also, I think it might be useful for this page to contain a prominent note pointing people to modern alternatives, namely `async` methods on `HttpClient`.

antonfirsov:
These samples were not handling the `asyncResult.CompletedSynchronously == true` case correctly: changing the sample code to use smaller buffer leads to stack overflow because of recursion on synchronous completions.

Besides the fix there the PR adds some consolidation:
- Cleanup and (somewhat) modernize the sample code
- Remove the Timer-based custom timeout implementation, use `WebRequest.Timeout` instead
- Collapse various samples into one sample
  - Unifiy the `HttpWebRequest` and `WebRequest` samples since they are practically identical
  - Delete CPP and VB samples given we don't have the capacity and the intention to maintain them.

Fixes #8662


gewarren (approved):
Thank you!

ManickaP (approved):

PR #11043, opened 2025-03-04
Add missing exception info for `HttpClient` and `SocketsHttpHandler` properties

### LINKED ISSUE #10962: ArgumentOutOfRangeException is also thrown when setting HttpClient.Timeout to more than 24 days
### Type of issue

Missing information

### Description

> [ArgumentOutOfRangeException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception?view=net-9.0)
The timeout specified is less than or equal to zero and is not [InfiniteTimeSpan](https://learn.microsoft.com/en-us/dotnet/api/system.threading.timeout.infinitetimespan?view=net-9.0#system-threading-timeout-infinitetimespan).

From [the source code](https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Net.Http/src/System/Net/Http/HttpClient.cs#L108C21-L108C89), it actually throws when you put in value greater than about 24 days (`int.MaxValue`ms) too:

```
private static readonly TimeSpan s_maxTimeout = TimeSpan.FromMilliseconds(int.MaxValue);
// ...
        public TimeSpan Timeout
        {
            get => _timeout;
            set
            {
                if (value != s_infiniteTimeout)
                {
                    ArgumentOutOfRangeException.ThrowIfLessThanOrEqual(value, TimeSpan.Zero);
                    ArgumentOutOfRangeException.ThrowIfGreaterThan(value, s_maxTimeout);
                }
                CheckDisposedOrStarted();
                _timeout = value;
            }
        }
```

I encountered this issue when trying to set it to `TimeSpan.MaxValue`:

> System.ArgumentOutOfRangeException: value ('10675199.02:48:05.4775807') must be less than or equal to '24.20:31:23.6470000'. (Parameter 'value')

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.timeout?view=net-9.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Net.Http/HttpClient.xml

199a48e3-16ed-9fe0-8957-f14f37a17acd

### Article author

@karelz

antonfirsov:
I went through all `HttpClient` and `SocketsHttpHandler` properties harmonizing the exception docs with the actual implementation.

I didn't touch `HttpClientHandler` docs, because most property setters are forwareded to the underlying handler, making it hard to provide comprehensive and reliable docs.

Fixes #10962

### DIFF
--- a/xml/System.Net.Http/HttpClient.xml
+++ b/xml/System.Net.Http/HttpClient.xml
@@ -276,7 +276,7 @@ The specified `handler` will be disposed of by calling [HttpClient.Dispose](xref
 
  ]]></format>
         </remarks>
-        <exception cref="T:System.ArgumentException">The <see cref="T:System.Uri" /> specified is not an absolute Uri.</exception>
+        <exception cref="T:System.ArgumentException">The <see cref="T:System.Uri" /> specified is not an absolute URI.</exception>
         <exception cref="T:System.InvalidOperationException">The <see cref="T:System.Net.Http.HttpClient" /> instance has already started one or more requests.</exception>
         <exception cref="T:System.ObjectDisposedException">The <see cref="T:System.Net.Http.HttpClient" /> instance has already been disposed.</exception>
       </Docs>


ManickaP (approved):

xml/System.Net.Http/HttpClient.xml
```diff
@@ -276,6 +276,9 @@ The specified `handler` will be disposed of by calling [HttpClient.Dispose](xref
 
  ]]></format>
         </remarks>
+        <exception cref="T:System.ArgumentException">The <see cref="T:System.Uri" /> specified is not an absolute Uri.</exception>
```
gewarren, line 279:
```suggestion
        <exception cref="T:System.ArgumentException">The <see cref="T:System.Uri" /> specified is not an absolute URI.</exception>
```

gewarren (approved):
One nit.

PR #11056, opened 2025-03-06
`IPInterfaceProperties.GetIPv(4|6)Properties` never returns null

antonfirsov:
The null checking in the [code samples](https://learn.microsoft.com/en-us/dotnet/api/system.net.networkinformation.networkinterface.getipproperties?view=net-9.0#examples) is unnecessary and misleading so I'm proposing to delete it.

Moreover, we should call out in the remarks section the necessity of a capability check so users avoid walking into an exception in case of missing protocol support. (As it occurred in https://github.com/dotnet/runtime/issues/111487)


ManickaP (approved):
Thanks.

antonfirsov:
@gewarren can you please help with the build here?
- The previews did not refresh after [2f1f621](https://github.com/dotnet/dotnet-api-docs/pull/11056/commits/2f1f6212e14332bed1690600e7b281ddb1099a87) and I don't know how to trigger it.
- Snippets build complains that there is no project file for `networkexamples.cpp` and `networkexamples.vb`, but this is an issue preceding the PR. Shall we bypass the check somehow?

Issue #11023, opened 2025-02-27
Inconsistent spelling of HMAC SHA256 in Microsoft documentation


FortFerdinand:
### Type of issue

Typo

### Description

On the page about HMAC SHA256, I see two different spellings of HMAC SHA256:
(https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.hmacsha256?view=net-9.0)

- In **Remarks**, the page mentions "SHA-256 hash function".
- In **Fields**, the page mentions "HMAC SHA256 algorithm".

Based on your cited source on the page, the correct spelling should be SHA256 without a dash:
https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/HMACSHA256.cs

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.hmacsha256?view=net-9.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Security.Cryptography/HMACSHA256.xml

### Document Version Independent Id

f4f90565-a499-8742-10e2-7296ea4a33c6

### Article author

@dotnet-bot

gewarren:
@vcsjones Is SHA256 or SHA-256 (when referring to the algorithm, not the class) more correct? Wikipedia says the latter: https://en.wikipedia.org/wiki/SHA-2.

PR #11062, opened 2025-03-08
SHA256 spelling fix

hawkeye101:
https://github.com/dotnet/dotnet-api-docs/issues/11023
Replaced 'SHA-256' with 'SHA256'.

### DIFF
--- a/xml/System.Security.Cryptography/HMACSHA256.xml
+++ b/xml/System.Security.Cryptography/HMACSHA256.xml
@@ -70,7 +70,7 @@
       <format type="text/markdown"><![CDATA[  
   
 ## Remarks  
- <xref:System.Security.Cryptography.HMACSHA256> is a type of keyed hash algorithm that is constructed from the SHA-256 hash function and used as a Hash-based Message Authentication Code (HMAC). The HMAC process mixes a secret key with the message data, hashes the result with the hash function, mixes that hash value with the secret key again, and then applies the hash function a second time. The output hash is 256 bits in length.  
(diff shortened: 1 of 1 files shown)


gewarren:
Thank you @hawkeye101. I want to figure out what the correct spelling is first (I left a question in the issue). Also, we'll probably need to update the SHA256 class docs as well.

vcsjones:
In general, I would expect SHA-256 when referring to the algorithm, and `SHA256` when referring to the class.

It gets a little messier because SHA-2-256 is also acceptable, but in general I think it should only be used when disambiguating SHA-2-256 and SHA-3-256.

We should also address any inconsistencies with SHA384/SHA-384 and SHA512/SHA-512. I suspect they inherited any documentation inconsistencies.

@bartonjs likely has thoughts, too.

bartonjs:
I, personally, prefer always using "SHA-2-256" over "SHA-256", but I'm a pedant.  Kevin's suggested "SHA256" is the class, "SHA-256" is the algorithm is probably the best starting place.  Anything that already says "SHA-2-256" probably isn't worth changing since they'd have to be read to see if they're excluding (or contradistincting against) "SHA-3-256".

hawkeye101:
@gewarren Here is what I've understood. Just want to confirm before I make the changes.

- In the HMAC class, it says "by using the SHA256 hash function" which contains a hyperlink to the SHA256 class. It seems to be referring to the algorithm here, not the class. So this should be SHA-256 and hyperlink removed.
- In the fields, methods of HMAC class it should be SHA-256 as it is referring to the algorithm.
- In the [SHA256 class docs](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha256?view=net-9.0), "The hash size for the SHA256 algorithm is 256 bits" should be changed to "The hash size for the SHA-256 algorithm is 256 bits".

These are the only changes required as far as I understand. Please let me know if there are more.

gewarren:
@hawkeye101 Those changes look correct to me. There are likely additional changes required, but that would require additional research.

gewarren:
Thank you @hawkeye101. I'm going to close this in favor of the more comprehensive https://github.com/dotnet/dotnet-api-docs/pull/11104.

PR #11094, opened 2025-03-14
Update ThreadStatic example

AaronRobinsonMSFT:
## Summary

The examples provided an initialization of `ThreadStatic` fields, which is a bad example for how they should be used. Removed that and made the example explicitly initialize them for each thread.

### DIFF
--- a/snippets/csharp/System/ThreadStaticAttribute/Overview/threadsafe2a.cs
+++ b/snippets/csharp/System/ThreadStaticAttribute/Overview/threadsafe2a.cs
@@ -9,8 +9,8 @@ class Program
 
     static void Main()
     {
-        Thread thread1 = new Thread(ProcessRequest);
-        Thread thread2 = new Thread(ProcessRequest);
+        Thread thread1 = new(ProcessRequest);
+        Thread thread2 = new(ProcessRequest);
 
         thread1.Start("REQ-001");
         thread2.Start("REQ-002");


jkotas:
I think this is a poor sample for ThreadStatics to start with. It does not demonstrate a situation where ThreadStatic must be used (local variable would work just fine - all uses are in scope of a single method) and it has a bunch of irrelevant clutter.

With [help of ChatGPT](https://chatgpt.com/share/67d4d4c5-1540-8012-acb9-8c9dfa197e14), I would modify the sample to something like this:

```
using System;
using System.Threading;

class Program
{
    [ThreadStatic]
    private static string? _requestId;

    static void Main()
    {
        Thread thread1 = new Thread(ProcessRequest);
        Thread thread2 = new Thread(ProcessRequest);

        thread1.Start("REQ-001");
        thread2.Start("REQ-002");

        thread1.Join();
        thread2.Join();

        Console.WriteLine("Main thread execution completed.");
    }

    static void ProcessRequest(object? requestId)
    {
        // Assign the request ID to the thread-static field
        _requestId = requestId as string;

        // Simulate request processing across multiple method calls
        PerformDatabaseOperation();
        PerformLogging();
    }

    static void PerformDatabaseOperation()
    {
        Console.WriteLine($"Thread {Thread.CurrentThread.ManagedThreadId}: Processing DB operation for request {_requestId}");
    }

    static void PerformLogging()
    {
        Console.WriteLine($"Thread {Thread.CurrentThread.ManagedThreadId}: Logging request {_requestId}");
    }
}
```

AaronRobinsonMSFT:
@jkotas Thanks. Used Copilot to create F# and VB.Net versions.

DickBaker:
I Like Jan Kotas' GPT code, but VS build raises two CA1840 Messages
Use 'Environment.CurrentManagedThreadId' instead of 'Thread.CurrentThread.ManagedThreadId'

so after applying the analyzer fixes and Code Cleanup, imho the C# code should become
```
class Program
{
    [ThreadStatic]
    static string? _requestId;

    static void Main()
    {
        var thread1 = new Thread(ProcessRequest);
        var thread2 = new Thread(ProcessRequest);

        thread1.Start("REQ-001");
        thread2.Start("REQ-002");

        thread1.Join();
        thread2.Join();

        Console.WriteLine("Main thread execution completed.");
    }

    static void ProcessRequest(object? requestId)
    {
        // Assign the request ID to the thread-static field
        _requestId = requestId as string;

        // Simulate request processing across multiple method calls
        PerformDatabaseOperation();
        PerformLogging();
    }

    static void PerformDatabaseOperation() => Console.WriteLine($"Thread {Environment.CurrentManagedThreadId}: Processing DB operation for request {_requestId}");

    static void PerformLogging() => Console.WriteLine($"Thread {Environment.CurrentManagedThreadId}: Logging request {_requestId}");
}
```

DickBaker:
@AaronRobinsonMSFT et al, thanks for adopting the CA1840 suggestions in all lang flavours, but I noticed you declined the "var" changes [the long-form code looked tired to me].
However I have since re-read the [.editorconfig](https://github.com/dotnet/dotnet-api-docs/blob/main/snippets/.editorconfig) where this legacy style seems to persist in doc-land
```
# Types: use keywords instead of BCL types, and permit var only when the type is clear
csharp_style_var_for_built_in_types = false:suggestion
csharp_style_var_when_type_is_apparent = false:none
csharp_style_var_elsewhere = false:suggestion
```

so this note is to ridicule such comment since the doc-style edict is _csharp_style_foo = false_ to disable **var** in ALL cases (i.e. I quarrel with the assertion "**permit var only when the type is clear**" since var seems _verboten_ in every case). If var is so toxic, perhaps you should amend that comment to match? `</rant>`

jkotas (approved):
Thanks

PR #11101, opened 2025-03-17
Update summary values TrustDirection.xml

rarestiaMS:
Updating inbound and outbound summary values

## Summary

Describe your changes here.

Updating values for trust summary

### DIFF
--- a/xml/System.DirectoryServices.ActiveDirectory/TrustDirection.xml
+++ b/xml/System.DirectoryServices.ActiveDirectory/TrustDirection.xml
@@ -74,8 +74,7 @@
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>A trusting domain or forest.</summary>
-        <remarks>An inbound (or incoming) trust is where some other domain or forest trusts this domain or forest. This domain or forest does not have access to resources that belong to the other domain or forest.</remarks>
+        <summary>A trusting domain or forest. An inbound (or incoming) trust is where some other domain or forest trusts this domain or forest. This domain or forest does not have access to resources that belong to the other domain or forest.</summary>
       </Docs>
     </Member>
     <Member MemberName="Outbound">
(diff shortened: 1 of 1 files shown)


xml/System.DirectoryServices.ActiveDirectory/TrustDirection.xml
```diff
@@ -98,7 +98,7 @@
       </ReturnValue>
       <MemberValue>2</MemberValue>
       <Docs>
-        <summary>This is a trusted domain or forest. This domain or forest has access to resources of the other domain or forest. The other domain or forest does not have access to the resources of this domain or forest.</summary>
+        <summary>An outbound (or outgoing) trust is where our local domain/forest trusts some other domain/forest.</summary>
```
gewarren, line 101:
```suggestion
        <summary>A trusted domain or forest.</summary>
        <remarks>An outbound (or outgoing) trust is where this domain or forest trusts some other domain or forest. The other domain or forest does not have access to the resources of this domain or forest.</remarks>
```

xml/System.DirectoryServices.ActiveDirectory/TrustDirection.xml
```diff
@@ -74,7 +74,7 @@
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>This is a trusting domain or forest. The other domain or forest has access to the resources of this domain or forest. This domain or forest does not have access to resources that belong to the other domain or forest.</summary>
+        <summary>An inbound (or incoming) trust is some other domain/forest trusting the local domain/forest.</summary>
```
gewarren, line 77:
```suggestion
        <summary>A trusting domain or forest.</summary>
        <remarks>An inbound (or incoming) trust is where some other domain or forest trusts this domain or forest. This domain or forest does not have access to resources that belong to the other domain or forest.</remarks>
```

gewarren (commented):
Some of this info would be better in remarks.

xml/System.DirectoryServices.ActiveDirectory/TrustDirection.xml
rarestiaMS: This change is acceptable. Please proceed.

cnotin:
I was about to suggest something similar so strongly agree with that.
An inbound trust is safe for the current domain, whereas an outbound trust may be dangerous. The current wording reads like the opposite.

PR #11105, opened 2025-03-18
Explain semantics of multiple calls to `CancellationTokenSource.Cancel`

Smaug123:
Also copy a line about LIFO semantics from one of the overloads, to unify the docs for each function.

## Summary

Earlier today, I needed to perform several experiments to discover the semantics of `CancellationTokenSource.Cancel`, because they are undocumented.

I don't know the semantics of `CancelAsync`, so I haven't documented them.


xml/System.Threading/CancellationTokenSource.xml
```diff
@@ -383,8 +383,10 @@
 ## Remarks  
  The associated <xref:System.Threading.CancellationToken> will be notified of the cancellation and will transition to a state where <xref:System.Threading.CancellationToken.IsCancellationRequested%2A> returns true.  
   
- Any callbacks or cancelable operations registered with the <xref:System.Threading.CancellationToken> will be executed.
-  
+ Any callbacks or cancelable operations registered with the <xref:System.Threading.CancellationToken> will be executed, if they have not already been executed by a previous call to <xref:System.Threading.CancellationTokenSource.Cancel>. Subsequent calls to <xref:System.Threading.CancellationTokenSource.Cancel> will not execute the same callback again unless re-registered. (We recommend avoiding multiple calls to <xref:System.Threading.CancellationTokenSource.Cancel>, because the intent of such code is often unclear.)
```
Smaug123, line 386: Didn't really know the corporate-speak phrasing for this. When writing for a human, one would say "Try not to call `Cancel` multiple times, because it breaks the mental model that the `IsCancellationRequested` state is terminal, and is therefore confusing to read".

xml/System.Threading/CancellationTokenSource.xml
```diff
@@ -383,8 +383,10 @@
 ## Remarks  
  The associated <xref:System.Threading.CancellationToken> will be notified of the cancellation and will transition to a state where <xref:System.Threading.CancellationToken.IsCancellationRequested%2A> returns true.
```
gewarren, line 384:
```suggestion

The associated <xref:System.Threading.CancellationToken> is notified of the cancellation and transitions to a state where <xref:System.Threading.CancellationToken.IsCancellationRequested> returns `true`.
```

xml/System.Threading/CancellationTokenSource.xml
gewarren, line 386:
```suggestion
Any callbacks or cancelable operations registered with the <xref:System.Threading.CancellationToken> are executed, if they haven't already been executed by a previous call to <xref:System.Threading.CancellationTokenSource.Cancel>. Subsequent calls to <xref:System.Threading.CancellationTokenSource.Cancel> won't execute the same callback again unless reregistered. (Avoid multiple calls to <xref:System.Threading.CancellationTokenSource.Cancel>, because the intent of such code is often unclear.)
```

gewarren (commented):
I left some grammatical suggestions. Thanks!

Smaug123:
Some scope creep, but I've gone over every use of the word "will" in the file and replaced it with a present tense unless it's describing asynchrony or future actions.

gewarren (approved):

Issue #10998, opened 2025-02-24
Add descriptions for empty namespaces


gewarren:
### Describe the issue or suggestion

- Add descriptions to empty namespace (ns-*.xml) files
- Add descriptions to member groups

PR #11137, opened 2025-03-28
Delete unnecessary namespace files

gewarren:
These namespaces don't seem to exist in the docs.

Contributes to #10998


BillWagner (approved):
The build is clean, so let's do it.

PR #11142, opened 2025-03-28
(More) fix that/which grammar

gewarren:
https://learn.microsoft.com/style-guide/a-z-word-list-term-collections/t/that-vs-which

(Hide whitespace changes.)


BillWagner (approved):
This LGTM. Let's :shipit:

Issue #10183, opened 2024-07-31
Fix up mismatched paramrefs


gewarren:
### Describe the issue or suggestion

Some paramrefs refer to parameters that don't exist on that API. Fix them up.

PR #11146, opened 2025-03-28
Fix mismatched paramrefs (System.Numerics through System.Runtime*)

gewarren:
Contributes to #10183.


BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

PR #11171, opened 2025-04-08
Clarify CookieContainer thread-safety

### LINKED ISSUE #1535: Missing information on thread safety
It is unclear if this class is thread safe. Looking at the reference sources, it uses lock objects, however this is not stated in the documentation, and so one cannot assume thread safety, which is a problem if the `CookieContainer` is about to be shared between `HttpClientHandler` instances.

---
#### Document Details

* ID: c33c84fa-3316-c743-662a-29739fbcc979
* Version Independent ID: 0ec76522-d0b5-329f-3370-43625633a07d
* Content: [CookieContainer Class (System.Net)](https://docs.microsoft.com/en-us/dotnet/api/system.net.cookiecontainer?view=netstandard-2.0)
* Content Source: [xml/System.Net/CookieContainer.xml](https://github.com/dotnet/dotnet-api-docs/blob/master/xml/System.Net/CookieContainer.xml)
* Product: **dotnet-api**
* GitHub Login: @karelz
* Microsoft Alias: **ncldev**

### LINKED ISSUE #10770: CookieContainer has no information on thread safety
### Type of issue

Missing information

### Description

Most classes have details on thread-safety, but CookieContainer does not. In the HttpClient guidelines at:

https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines

it's stated:

_If your app requires cookies, consider disabling automatic cookie handling or avoiding [IHttpClientFactory](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.ihttpclientfactory). Pooling the [HttpMessageHandler](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpmessagehandler) instances results in sharing of [CookieContainer](https://learn.microsoft.com/en-us/dotnet/api/system.net.cookiecontainer) objects. Unanticipated [CookieContainer](https://learn.microsoft.com/en-us/dotnet/api/system.net.cookiecontainer) object sharing often results in incorrect code._

I can't tell if the "results in incorrect code" is because the CookieContainer is not thread safe or some other reason. If I have two instances of HttpClient using the same CookieContainer (because of IHttpClientFactory) and operating on separate threads, but never accessing the same remote, will this result in a threading issue or "incorrect code"?

"Incorrect code" seems a bit vague as well, how do I determine the source of the "incorrect code"? If a shared HttpClientHandler (because of IHttpClientFactory) can never be used with a CookieContainer in a multiple threaded context, then I think it would be better to point that out directly rather than saying "consider disabling automatic cookie handling".

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.net.cookiecontainer?view=net-9.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Net/CookieContainer.xml

0ec76522-d0b5-329f-3370-43625633a07d

### Article author

@karelz

antonfirsov:
This PR supplements https://github.com/dotnet/docs/pull/45658 by clarifying that `CookieContainer` methods are thread-safe but also noting the dangers of unanticipaded `CookieContainer` sharing.

Fixes #1535
Fixes #10770

### DIFF
--- a/xml/System.Net/CookieContainer.xml
+++ b/xml/System.Net/CookieContainer.xml
@@ -82,7 +82,7 @@
   The methods for adding and retrieving <xref:System.Net.Cookie> instances to and from a <xref:System.Net.CookieContainer> are thread-safe and can be used concurrently from multiple threads.
 
   > [!NOTE]
-  > Regardless of thread-safety, unanticipated sharing of <xref:System.Net.CookieContainer> instances can lead to issues when working with <xref:System.Net.Http.HttpClient> and <xref:System.Net.Http.HttpClientHandler> or <xref:System.Net.Http.SocketsHttpHandler>, since parts of the application that aren't intended to share cookies might inadvertently do so.
+  > Regardless of thread-safety, unanticipated sharing of <xref:System.Net.CookieContainer> instances can lead to issues when working with <xref:System.Net.Http.HttpClient> and <xref:System.Net.Http.HttpClientHandler> or <xref:System.Net.Http.SocketsHttpHandler>, since parts of the application that aren't intended to share cookies with each other might inadvertently do so.
   > It is not recommended to use cookies together with <xref:System.Net.Http.IHttpClientFactory>.
 
  ]]></format>


xml/System.Net/CookieContainer.xml
```diff
@@ -79,6 +79,19 @@
 
  ]]></format>
     </remarks>
+    <threadsafe>
```
gewarren, line 82: Is this a legit tag?

antonfirsov: [Yes](https://github.com/search?q=repo%3Adotnet%2Fdotnet-api-docs+threadsafe+language%3AXML&type=code&l=XML&p=1), but looks like it cannot handle embedded `<format>` :( Will move the content to `<remarks>`.

MihaZupan (approved):

xml/System.Net/CookieContainer.xml
```diff
@@ -77,6 +77,14 @@
...
 
+## Thread Safety
+ 
+  The methods for adding and retrieving <xref:System.Net.Cookie> instances to and from a <xref:System.Net.CookieContainer> are thread-safe and can be used concurrently from multiple threads.
+
+  > [!NOTE]
+  > Regardless of thread-safety, unanticipated sharing of <xref:System.Net.CookieContainer> instances can lead to issues when working with <xref:System.Net.Http.HttpClient> and <xref:System.Net.Http.HttpClientHandler> or <xref:System.Net.Http.SocketsHttpHandler>, since parts of the application that aren't intended to share cookies might inadvertently do so.
```
antonfirsov, line 85:
```suggestion
  > Regardless of thread-safety, unanticipated sharing of <xref:System.Net.CookieContainer> instances can lead to issues when working with <xref:System.Net.Http.HttpClient> and <xref:System.Net.Http.HttpClientHandler> or <xref:System.Net.Http.SocketsHttpHandler>, since parts of the application that aren't intended to share cookies with each other might inadvertently do so.
```

PR #11214, opened 2025-04-22
Turn off comments from OPS

gewarren:
I spend a lot of time hiding outdated OPS build reports in the PR comments. Also, now that status checker is a required status check, authors will be forced to look at the build report if it has warnings, so we don't need the PR comments to make them as "in your face".


BillWagner (approved):
Thank you! Thank you! Thank you!  I detest seeing the out of date OPS build comments.

Let's :shipit: now.

PR #11223, opened 2025-04-22
Add language slugs (Security, Threading, Reflection, Runtime)

gewarren:
Prep for https://github.com/dotnet/docs/issues/45892.

(Hide whitespace changes.)

### DIFF
--- a/xml/System.Reflection/MethodInfo.xml
+++ b/xml/System.Reflection/MethodInfo.xml
@@ -729,7 +729,7 @@ For a list of the invariant conditions for other terms used in generic reflectio
 
  If you call <xref:System.Reflection.MethodInfo.GetGenericMethodDefinition%2A> on a <xref:System.Reflection.MethodInfo> that already represents a generic method definition, it returns the current <xref:System.Reflection.MethodInfo>.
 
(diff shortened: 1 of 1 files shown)


xml/System.Reflection/MethodInfo.xml
```diff
@@ -729,26 +729,21 @@ For a list of the invariant conditions for other terms used in generic reflectio
 
  If you call <xref:System.Reflection.MethodInfo.GetGenericMethodDefinition%2A> on a <xref:System.Reflection.MethodInfo> that already represents a generic method definition, it returns the current <xref:System.Reflection.MethodInfo>.
 
- If a generic method definition includes generic parameters of the declaring type, there will be a generic method definition specific to each constructed type. For example, consider the following C#, Visual Basic, and C++ code:
+ If a generic method definition includes generic parameters of the declaring type, there will be a generic method definition specific to each constructed type. For example, consider the following C# code:
```
gewarren, line 732:
```suggestion
 If a generic method definition includes generic parameters of the declaring type, there will be a generic method definition specific to each constructed type. For example, consider the following code:
```

PR #11239, opened 2025-04-25
[UUF] Fix output

gewarren:


snippets/csharp/System/Math/Overview/mathsample.cs
```diff
@@ -75,7 +75,7 @@ static void Main(string[] args)
 			double dxD = trpz.GetLeftBaseDegreeAngle();
 			Console.WriteLine("Trapezoid left base angle is: " + dxD.ToString() + " Degrees");
 			double dyD = trpz.GetRightBaseDegreeAngle();
-			Console.WriteLine("Trapezoid left base angle is: " + dyD.ToString() + " Degrees");
+			Console.WriteLine("Trapezoid right base angle is: " + dyD.ToString() + " Degrees");
```
BillWagner, line 78: While here, should we use string interpolation:

```suggestion
			Console.WriteLine($"Trapezoid right base angle is: {dyD} Degrees");
```

BillWagner (approved):
This LGTM, with one suggestion. Let's :shipit:

snippets/csharp/System/Math/Overview/mathsample.cs
gewarren: If I change this one, I should change all of them. Is it okay to put this off until we do a bulk effort to update string formats?

PR #11244, opened 2025-04-25
[UUF] Remove invalid exceptions

gewarren:
- I don't think you can pass an "invalid" value since it's not an enum.
- Also formats some tables.


BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

PR #11247, opened 2025-04-25
Update SslProtocols property value

### LINKED ISSUE #11233: Erroneous description of EnabledSslProtocols property value
### Type of issue

Other (describe below)

### Description

The Property Value description says: "One of the values defined in the SslProtocols enumeration." This is probably incorrect, because the name "EnabledSslProtocols" implies that this property should support a bitwise combination of the member values of the SslProtocols enumeration, not just "one" of the values.

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.sslprotocols?view=net-9.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Net.Http/HttpClientHandler.xml

e13f05ac-54c0-4575-a031-187c36aa10ad

### Platform Id

dc78e6d0-f549-f3d1-c07a-eaa538a47ba0

### Article author

@karelz

gewarren:
Fixes #11233.

### DIFF
--- a/xml/System.Net.Http/HttpClientHandler.xml
+++ b/xml/System.Net.Http/HttpClientHandler.xml
@@ -1234,8 +1234,8 @@ handler.ServerCertificateCustomValidationCallback = HttpClientHandler.DangerousA
         <ReturnType>System.Security.Authentication.SslProtocols</ReturnType>
       </ReturnValue>
       <Docs>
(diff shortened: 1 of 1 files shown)


adegeo:
Should we use the term "flags" somewhere?

gewarren:
> Should we use the term "flags" somewhere?

Not according to the guidance for flag enum property values: https://github.com/dotnet/dotnet-api-docs/wiki/Property-Value.

PR #11269, opened 2025-05-05
Correct doc on ReadOnlySequence<T>.IsSingleSegment

Rob-Hague:
The current property value was presumably intended for IsEmpty


gewarren (approved):
Thank you @Rob-Hague

PR #11274, opened 2025-05-06
Don't catch all exceptions in Directory.CreateDirectory example

### LINKED ISSUE #11240: [UUF] Don't catch all exceptions
### Type of issue

Other (describe below)

### Description

Why is the example catching all exceptions? We have clearly documented what the exceptions are and not following our own doc. This is a bad precedent to set for our customers.

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.createdirectory?view=net-9.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.IO/Directory.xml

94e4ec5d-b7bc-0306-7d10-90778a66817a

### Platform Id

23cc4013-8b62-3b4c-df47-f3692abdf742

### Article author

@dotnet-bot

adamsitnik:
## Summary

The `CreateDirectory` can fail with few exceptions:
- `ArgumentException`, `ArgumentNullException`, `PathTooLongException`, `DirectoryNotFoundException` or `NotSupportedException` when the path is invalid
- `UnauthorizedAccessException` when the caller does not have the required permission.

Since this example uses a harcoded, short and valid path `c:\MyDir` there is no need to catch any other exceptions than `UnauthorizedAccessException`

Fixes #11240

### DIFF
--- a/snippets/csharp/System.IO/Directory/CreateDirectory/dir_createdir.cs
+++ b/snippets/csharp/System.IO/Directory/CreateDirectory/dir_createdir.cs
@@ -8,29 +8,30 @@ class Test
     {
         // Specify the directory you want to manipulate.
         string path = @"c:\MyDir";
-
+        
+        // Determine whether the directory exists.
+        if (Directory.Exists(path))
+        {
+            Console.WriteLine("That path exists already.");
+            return;
+        }
+        
+        DirectoryInfo di;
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Thanks @adamsitnik 

PR #11276, opened 2025-05-06
Add System.ArgumentException documentation to Math.Clamp

### LINKED ISSUE #10677: Math.Clamp is missing documentation on its thrown exception for degenerate ranges
### Type of issue

Code doesn't work

### Description

`Math.Clamp` is currently documented as returning, in order of precedence:
1. `value` - where `min` <= `value` <= `max`; 
2. `min` - where `value` < `min`; or
3. `max` - where `max` < `value`.

This means it should support degenerate ranges such as `Math.Clamp(10, 0, -2)` and result in the 2nd case, returning 0.

This is actually _not_ how the method is implemented.
It expressly checks for `min` <= `max` and throws an `ArgumentException` in cases that don't fit this constraint.

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.math.clamp?view=net-8.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System/Math.xml

8e48c3bd-8e68-e4c7-167c-07efd236a06f

### Article author

@dotnet-bot

pangeorg:
## Summary

Added missing System.ArgumentException documentation when max < min in Math.Clamp

Fixes #10677

### DIFF
--- a/xml/System/Math.xml
+++ b/xml/System/Math.xml
@@ -1824,6 +1824,8 @@ Dim i2 As Integer = CInt(Math.Ceiling(d2))        ' Result: 7969
             <paramref name="max" /> if <paramref name="max" /> &lt; <paramref name="value" />.</para>
         </returns>
         <remarks>To be added.</remarks>
+        <exception cref="T:System.ArgumentException">
+          <paramref name="max" /> is less than <paramref name="min"/>. </exception>
       </Docs>
     </Member>
     <Member MemberName="Clamp">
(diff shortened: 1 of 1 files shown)


pangeorg:
@dotnet-policy-service agree

BillWagner:
@tannergooding PTAL

tannergooding (approved):

gewarren (approved):
Thanks @pangeorg

PR #11313, opened 2025-05-18
Fix wording of ldloca description

Smaug123:
## Summary

The current phrasing in the numbered list unambiguously and incorrectly means "dereference the local variable, which is guaranteed to contain an address, and push that address to the stack". The rest of the text correctly implies instead that it means "push the address of the local variable".

ECMA-335 6th edition states:

> The ldloca instruction pushes the address of the local variable number indx onto the stack, where local variables are numbered 0 onwards.


jkotas (approved):
Thanks

jkotas:
Could you please resolve the conflict?

Smaug123:
Thanks; done.

jkotas:
Thank you! Could you please fix up https://github.com/dotnet/dotnet-api-docs/pull/11314 as well?

PR #11314, opened 2025-05-18
Fix description of pointer in ldloca.s

Smaug123:
## Summary

ECMA-335 6th edition states:

> The ldloca instruction pushes the address of the local variable number indx onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like ldind and stind. The result is a managed pointer (type &).

The docs for `ldloca` already state that the result is a managed pointer; this PR brings the docs for `ldloca.s` into alignment with the spec and the docs for `ldloca`. (I hope that this change is in fact correct!)

### DIFF
--- a/xml/System.Reflection.Emit/OpCodes.xml
+++ b/xml/System.Reflection.Emit/OpCodes.xml
@@ -10422,7 +10422,7 @@ The following <xref:System.Reflection.Emit.ILGenerator.Emit%2A> method overloads
 
 1.  The address of the local variable at the specified index is pushed onto the stack.
 
- The `ldloca.s` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a managed pointer (type `&`).
+ The `ldloca.s` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a managed pointer (type `&`). The local variable is stored in unmanaged memory, so the return value can be converted to an unmanaged pointer without pinning.
 
  The `ldloca.s` instruction provides an efficient encoding for use with the local variables 0 through 255.
 


xml/System.Reflection.Emit/OpCodes.xml
```diff
@@ -10423,7 +10423,7 @@ The following <xref:System.Reflection.Emit.ILGenerator.Emit%2A> method overloads
   
 1.  The address stored in the local variable at the specified index is pushed onto the stack.  
   
- The `ldloca.s` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a transient pointer (type `*`).
```
Smaug123, line 10426: I'm concerned that I don't have a plausible story for how this kind of mistake could have been made, so I'm not enormously confident that this *is* a mistake. But on `main` the docs are not internally consistent between `ldloca` and `ldloca.s`.

The spec appears to be ambiguous. It contains two usages of the term "transient pointer":

> Instructions that create pointers which are guaranteed not to point into the memory manager’s heaps (e.g., ldloca, ldarga, and ldsflda) produce transient pointers (type `*`) that can be used wherever a managed pointer (type `&`) or unmanaged pointer (type native unsigned int) is expected.

and

> `*`, a “transient pointer,” which can be used only within the body of a single method, that points to a value known to be in unmanaged memory (see the CIL Instruction Set specification for more details. `*` types are generated internally within the CLI; they are not created by the user).

But then the spec for ldloca is explicit:

> The result is a managed pointer (type &).

So really I don't know what the correct resolution is, or even what the semantics are.

jkotas: > transient pointer

This is clarified in https://github.com/dotnet/runtime/blob/main/docs/design/specs/Ecma-335-Augments.md#transient-pointers

Transient pointers are not a thing.

jkotas (changes requested):
This issue was addressed by https://github.com/dotnet/dotnet-api-docs/pull/9574, but this place was missed. We should apply the same change here for consistency.

xml/System.Reflection.Emit/OpCodes.xml
```diff
@@ -10422,7 +10422,7 @@ The following <xref:System.Reflection.Emit.ILGenerator.Emit%2A> method overloads
 
 1.  The address of the local variable at the specified index is pushed onto the stack.
 
- The `ldloca.s` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a transient pointer (type `*`).
+ The `ldloca.s` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a managed pointer (type `&`).
```
jkotas:
```suggestion
 The `ldloca.s` instruction pushes the address of the local variable number at the passed index onto the stack, where local variables are numbered 0 onwards. The value pushed on the stack is already aligned correctly for use with instructions like <xref:System.Reflection.Emit.OpCodes.Ldind_I> and <xref:System.Reflection.Emit.OpCodes.Stind_I>. The result is a managed pointer (type `&`). The local variable is stored in unmanaged memory, so the return value can be converted to an unmanaged pointer without pinning.
```

jkotas (approved):
Thanks!

PR #11318, opened 2025-05-19
Add documentation for System.Text.Ascii class

ericstj:
Fixes https://github.com/dotnet/runtime/issues/88576

API issue: https://github.com/dotnet/runtime/issues/28230

### DIFF
--- a/xml/System.Text/Ascii.xml
+++ b/xml/System.Text/Ascii.xml
@@ -16,8 +16,8 @@
   </Base>
   <Interfaces />
   <Docs>
-    <summary>Static class containing helper methods for working with ASCII encoded text as bytes or characters.</summary>
-    <remarks>For APIs which take both a source and a destination buffer, the behavior of the method is undefined if the source and destination buffers overlap, unless the API description specifies otherwise. The behavior of all APIs is undefined if another thread mutates the buffers while these APIs are operating on them. All case conversion APIs are culture-unaware.</remarks>
+    <summary>Provides helper methods for working with ASCII-encoded text as bytes or characters.</summary>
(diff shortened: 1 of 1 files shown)


xml/System.Text/Ascii.xml
```diff
@@ -16,8 +16,8 @@
   </Base>
   <Interfaces />
   <Docs>
-    <summary>To be added.</summary>
-    <remarks>To be added.</remarks>
+    <summary>Static class containing helper methods for working with ASCII encoded text as bytes or characters.</summary>
```
gewarren, line 19:
```suggestion
    <summary>Provides helper methods for working with ASCII-encoded text as bytes or characters.</summary>
```

xml/System.Text/Ascii.xml
```diff
@@ -16,8 +16,8 @@
...
+    <remarks>For APIs which take both a source and a destination buffer, the behavior of the method is undefined if the source and destination buffers overlap, unless the API description specifies otherwise. The behavior of all APIs is undefined if another thread mutates the buffers while these APIs are operating on them. All case conversion APIs are culture-unaware.</remarks>
```
gewarren, line 20:
```suggestion
    <remarks>For APIs that take both a source and a destination buffer, the behavior of the method is undefined if the source and destination buffers overlap, unless the description specifies otherwise. The behavior of all APIs is undefined if another thread mutates the buffers while these APIs are operating on them. All case-conversion APIs are culture-unaware.</remarks>
```

gewarren (approved):

tarekgh (approved):
With @gewarren comments, LGTM!

PR #11327, opened 2025-05-21
Update CookieContainer.xml

moojek:
## Summary

Adds some missing whitespace

### DIFF
--- a/xml/System.Net/CookieContainer.xml
+++ b/xml/System.Net/CookieContainer.xml
@@ -476,7 +476,7 @@
 ## Remarks
  If you add a <xref:System.Net.Cookie> instance for just one specific host, do not set the <xref:System.Net.Cookie.Domain%2A> property of the <xref:System.Net.Cookie> instance. This is set automatically, based on the URI.
 
- If your URI corresponds to your local domain and sends to all the hosts on the local domain, set the <xref:System.Net.Cookie> <xref:System.Net.Cookie.Domain%2A> property equal to ".local". Otherwise, make sure it matches the host name used in the URI.
+ If your URI corresponds to your local domain and sends to all the hosts on the local domain, set the <xref:System.Net.Cookie.Domain?displayProperty=nameWithType> property equal to ".local". Otherwise, make sure it matches the host name used in the URI.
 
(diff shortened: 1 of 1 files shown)


moojek:
Sorry for such a shitty PR, but caught my eye when browsing docs.

xml/System.Net/CookieContainer.xml
```diff
@@ -476,7 +476,7 @@
 ## Remarks
  If you add a <xref:System.Net.Cookie> instance for just one specific host, do not set the <xref:System.Net.Cookie.Domain%2A> property of the <xref:System.Net.Cookie> instance. This is set automatically, based on the URI.
 
- If your URI corresponds to your local domain and sends to all the hosts on the local domain, set the <xref:System.Net.Cookie><xref:System.Net.Cookie.Domain%2A> property equal to ".local". Otherwise, make sure it matches the host name used in the URI.
+ If your URI corresponds to your local domain and sends to all the hosts on the local domain, set the <xref:System.Net.Cookie> <xref:System.Net.Cookie.Domain%2A> property equal to ".local". Otherwise, make sure it matches the host name used in the URI.
```
gewarren, line 479:
```suggestion
 If your URI corresponds to your local domain and sends to all the hosts on the local domain, set the <xref:System.Net.Cookie.Domain?displayProperty=nameWithType> property equal to ".local". Otherwise, make sure it matches the host name used in the URI.
```

gewarren (approved):
Thanks, I left a suggestion that will display "Cookie.Domain" as the link text, which I think is what we want.

xml/System.Net/CookieContainer.xml
moojek: I'm sure I'm overcomplicating the issue, since Your suggestion will be good enough, but I think the author's intention might have been that the "Cookie" part is a possessive form. If you take a look at the doc page this noun often appears in this context (_Domain property of the Cookie instance_, _the Path property of the Cookie_). Of course my initial edit is then wrong as well - I now suggest "set the Cookie's Domain property"

gewarren: I think Cookie.Domain is pretty clear that Domain is a property of the Cookie, so I think this is good.

PR #11334, opened 2025-05-22
Correct parameter name in ImageLockMode.UserInputBuffer description

mriehm:
The docs for the `UserInputBuffer` value of the `ImageLockMode` enum in System.Drawing refer to the `flags` parameter as being input or output, which makes no sense. It should refer to the `bitmapData` parameter instead.
This can be confirmed by referencing [the docs for the ImageLockMode enum in GDI+](https://learn.microsoft.com/en-us/windows/win32/api/gdiplusimaging/ne-gdiplusimaging-imagelockmode), from which the System.Drawing version of the documentation appears to have been copied.

### DIFF
--- a/xml/System.Drawing.Imaging/ImageLockMode.xml
+++ b/xml/System.Drawing.Imaging/ImageLockMode.xml
@@ -142,7 +142,7 @@
       </ReturnValue>
       <MemberValue>4</MemberValue>
       <Docs>
-        <summary>Specifies that the buffer used for reading or writing pixel data is allocated by the user. If this flag is set, the <paramref name="bitmapData" /> parameter of the <see cref="Overload:System.Drawing.Bitmap.LockBits" /> method serves as an input parameter (and possibly as an output parameter). If this flag is cleared, then the <paramref name="bitmapData" /> parameter serves only as an output parameter.</summary>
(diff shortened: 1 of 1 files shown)


xml/System.Drawing.Imaging/ImageLockMode.xml
```diff
@@ -142,7 +142,7 @@
       </ReturnValue>
       <MemberValue>4</MemberValue>
       <Docs>
-        <summary>Specifies that the buffer used for reading or writing pixel data is allocated by the user. If this flag is set, the <paramref name="flags" /> parameter of the <see cref="Overload:System.Drawing.Bitmap.LockBits" /> method serves as an input parameter (and possibly as an output parameter). If this flag is cleared, then the <paramref name="flags" /> parameter serves only as an output parameter.</summary>
+        <summary>Specifies that the buffer used for reading or writing pixel data is allocated by the user. If this flag is set, the <paramref name="bitmapData" /> parameter of the <see cref="Overload:System.Drawing.Bitmap.LockBits" /> method serves as an input parameter (and possibly as an output parameter). If this flag is cleared, then the <paramref name="bitmapData" /> parameter serves only as an output parameter.</summary>
```
gewarren, line 145:
```suggestion
        <summary>Specifies that the buffer used for reading or writing pixel data is allocated by the user. If this flag is set, the <paramref name="bitmapData" /> parameter of the <see cref="M:System.Drawing.Bitmap.LockBits(System.Drawing.Rectangle,System.Drawing.Imaging.ImageLockMode,System.Drawing.Imaging.PixelFormat,System.Drawing.Imaging.BitmapData)" /> method serves as an input parameter (and possibly as an output parameter). If this flag is cleared, then the <paramref name="bitmapData" /> parameter serves only as an output parameter.</summary>
```

gewarren (approved):
Thanks, good catch. My suggestion is to also link to the specific overload that has a `bitmapData` parameter.

gewarren (approved):

PR #11335, opened 2025-05-22
Remove needs-area-label when area label added

gewarren:


BillWagner (approved):
I'll approve this.

We should make a feature request for the `label` tag to support regex or glob style patterns. This would be so much simpler if we could have one tag:

```
-labelAdded:
    label: `area-.*`
```

PR #11351, opened 2025-05-25
PipeWriter.UnflushedBytes long property wraps to negative numbers after int.MaxValue

pedrobsaila:
## Summary

The method Advance(int size) throws an exception when its internal unflushed buffer size overflows long.MaxValue instead of silently having UnflushedBytes property with negative value

Contributes to dotnet/runtime#109513


pedrobsaila:
I was being advised not to throw OverflowException exception in this method. Closing the PR

PR #11428, opened 2025-06-10
Remove extra whitespace after hyphens and right angle brackets

gewarren:
These are all [whitespace only changes](https://github.com/dotnet/dotnet-api-docs/pull/11428/files?diff=unified&w=1).

First commit:

Regex find: `^-   (\w)`
Replace: `- $1`

Second commit:

Regex find: `^>  (\w)`
Replace: `> $1`


adegeo (approved):
Checked on a few files, looks good!

PR #11437, opened 2025-06-11
Update mentions of older .NET Framework versions

gewarren:
- ".NET Framework type" --> ".NET type"
- Remove paragraphs that are specific to .NET Fx 1.x/2.x
- Remove mentions of .NET Fx 2.0 where it doesn't make sense to include a version

(Different changes are grouped by commit.)


BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

PR #11446, opened 2025-06-12
Fix UnicodeRanges documentation

### LINKED ISSUE #11434: UnicodeRanges static properties return incorrect range
### Type of issue

Typo

### Description

UnicodeRanges provides several static UnicodeRange properties corresponding to Unicode code blocks. For some of these properties, the returned range is not equivalent to the range stated in the documentation.

For example, UnicodeRanges.BasicLatin is stated to return the range from U+0021 to U+007F, but actually returns the range from U+0000 to U+007F.

The only two ranges I have found to be incorrect are BasicLatin and Latin1Supplement (range starts at U+0080 instead of U+00A1), although there could be other instances.

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.text.unicode.unicoderanges?view=net-9.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Text.Unicode/UnicodeRanges.xml

6368ec59-4d47-ee92-f8e7-3186c02ed15e

### Platform Id

2af53177-3df4-d0c5-3ad9-ed54446fc29d

### Article author

@dotnet-bot

JaidenSlifer:
## Summary

Corrected start point of BasicLatin and Latin1Supplement Unicode Ranges to reflect the official [Unicode Blocks File](https://www.unicode.org/Public/UCD/latest/ucd/Blocks.txt)

Fixes #11434 UnicodeRanges static properties return incorrect range

### DIFF
--- a/xml/System.Text.Unicode/UnicodeRanges.xml
+++ b/xml/System.Text.Unicode/UnicodeRanges.xml
@@ -511,8 +511,8 @@ For the full set of characters in this block, see [The Unicode Standard, Version
         <ReturnType>System.Text.Unicode.UnicodeRange</ReturnType>
       </ReturnValue>
       <Docs>
-        <summary>Gets the Basic Latin Unicode block (U+0021-U+007F).</summary>
-        <value>The Basic Latin Unicode block (U+0021-U+007F).</value>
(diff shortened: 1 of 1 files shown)


tarekgh:
Thanks @JaidenSlifer for providing the fix!

PR #11463, opened 2025-06-17
[HTTP] Example for WinHttpHandler.ServerCertificateValidationCallback

ManickaP:
Add example for `ServerCertificateValidationCallback` in `WinHttpHandler`.

### DIFF
--- a/xml/System.Net.Http/WinHttpHandler.xml
+++ b/xml/System.Net.Http/WinHttpHandler.xml
@@ -740,7 +740,7 @@ When this property is set to `true`, all HTTP redirect responses from the server
       </ReturnValue>
       <Docs>
         <summary>Gets or sets a callback method to validate the server certificate. This callback is part of the SSL handshake.</summary>
-        <value>The callback should return <see langword="true" /> if the server certificate is considered valid and the request should be sent. Otherwise, return <see langword="false" />.</value>
+        <value>The callback should return <see langword="true" /> if the server certificate is considered valid and the request should be sent. Otherwise, returns <see langword="false" />.</value>
         <remarks>
           <format type="text/markdown"><![CDATA[
 
@@ -751,7 +751,7 @@ When this property is set to `true`, all HTTP redirect responses from the server
 
 ## Examples
 
-The following code example implements the callback. If there are validation errors, this method displays them and returns `false`, which prevents communication with the unauthenticated server.
+The following code example implements the callback. If there are validation errors, this method returns `false` preventing communication with the unauthenticated server. Otherwise, it allows for additional validation and return `true` if the certificate is valid.
 
  :::code language="csharp" source="~/snippets/csharp/System.Net.Http/WinHttpHandler/program.cs" id="Snippet1":::
  ]]></format>


snippets/csharp/System.Net.Http/WinHttpHandler/program.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Net;
+using System.Net.Http;
+using System.Net.Sockets;
+using System.Threading.Tasks;
+
+class HttpClientHandler_SecureExample
```
antonfirsov: Nit:

```suggestion
class WinHttpHandler_SecureExample
```

snippets/csharp/System.Net.Http/WinHttpHandler/program.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    static async Task Main()
```
antonfirsov: If this was sync  -> fewer usings -> more compact sample code.

snippets/csharp/System.Net.Http/WinHttpHandler/program.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Net;
+using System.Net.Http;
+using System.Net.Sockets;
+using System.Threading.Tasks;
```
antonfirsov: Are all the using statements required?

snippets/csharp/System.Net.Http/WinHttpHandler/program.cs
antonfirsov: Scrap this I noticed you use `<Snippet1>`.

snippets/csharp/System.Net.Http/WinHttpHandler/program.cs
antonfirsov: Scrap this I noticed you use `<Snippet1>`.

snippets/csharp/System.Net.Http/WinHttpHandler/program.cs
```diff
@@ -0,0 +1,25 @@
...
+    static async Task Main()
+    {
+        // <Snippet1>
+        var handler = new WinHttpHandler();
+        handler.ServerCertificateValidationCallback = (httpRequestMessage, certificate, chain, sslPolicyErrors)
+        {
+             if (sslPolicyErrors == SslPolicyErrors.None)
+                return true;
+
+            Console.WriteLine("Certificate error: {0}", sslPolicyErrors);
```
rzikm: This may simply display: "RemoteCertificateChainErrors", which is not really helpful when diagnosing what exactly is wrong with the cert. I understand that the point of the snippet is to be simple, but consider adding something like the following

```cs
                for (int i = 0; i < chain.ChainElements.Count; i++)
                {
                    var element = chain.ChainElements[i];
                    Console.WriteLine($"Certificate {i}:");
                    Console.WriteLine(element.Certificate.ToString(true));
                    foreach (var status in element.ChainElementStatus)
                    {
                        Console.WriteLine($"  Status: {status.Status}");
                        if (status.StatusInformation.Length > 0)
                        {
                            Console.WriteLine($"  Status Information: {status.StatusInformation}");
                        }
                    }
                    Console.WriteLine();
                }
```

snippets/csharp/System.Net.Http/WinHttpHandler/Project.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Library</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
```
gewarren: Can it target .NET 9 instead?

xml/System.Net.Http/WinHttpHandler.xml
```diff
@@ -747,6 +747,12 @@ When this property is set to `true`, all HTTP redirect responses from the server
 ## Remarks
  The default value is `null`. If this property is `null`, the server certificate is validated using standard well-known certificate authorities.
 
+ The delegate's `sslPolicyErrors` argument contains any certificate errors returned by SSPI while authenticating the server. The <xref:System.Boolean> value returned by this delegate determines whether the authentication is allowed to succeed.
+
+## Examples
+ The following code example implements the callback. If there are validation errors, this method displays them and returns `false`, which prevents communication with the unauthenticated server.
```
gewarren:
```suggestion

The following code example implements the callback. If there are validation errors, this method displays them and returns `false`, which prevents communication with the unauthenticated server.
```

gewarren (approved):

snippets/csharp/System.Net.Http/WinHttpHandler/program.cs
ManickaP: The example is just a copy of this https://learn.microsoft.com/dotnet/api/system.net.security.remotecertificatevalidationcallback.

My goal is to show an example that takes into account `sslPolicyErrors` when returning result, i.e. `return true only if none`. So I'd rather remove the `writeline` and replaced it with something else like `// TODO: custom validation` than drown it in printing details. 

WDYT? Should I keep it as-is, change to a TODO comment, change the original?

snippets/csharp/System.Net.Http/WinHttpHandler/Project.csproj
ManickaP: Of course it can. I just copy pasted from another example, so I thought there's some reason for it being stuck at 6.0.

snippets/csharp/System.Net.Http/WinHttpHandler/program.cs
rzikm: I am okay with either decision, maybe with slight preference for the TODO to keep the code example as simple as possible

PR #11487, opened 2025-06-19
Port some docs from IntelliSense

gewarren:


MihaZupan (approved):
Thanks.
The format provider is ignored for these in practice, as in many other cases, but I see that this is just inheriting the default description from the `IUtf8SpanParsable` interface.

Issue #11499, opened 2025-06-20
Organize VB snippet files


gewarren:
### Describe the issue or suggestion

A while back, I organized the C# snippet files by namespace and type/member name. Now, let's use Copilot to organize the VB snippets in the same way.

PR #11540, opened 2025-07-04
Reorganize VB snippets in Remoting directory

gewarren:


gewarren:
Snippets 5000 errors are for missing projects, which don't need to be addressed for a bulk change like this.

PR #11541, opened 2025-07-04
Reorganize VB snippets in WindowsServer directory

gewarren:


gewarren:
Snippets 5000 errors are all "missing project", which is okay in this case.

PR #11544, opened 2025-07-04
Reorganize VB snippets in CLR directory

gewarren:


gewarren:
I'm not planning to address the remaining Snippets 5000 errors, which are either missing project or some object from a referenced project is undefined.

PR #11588, opened 2025-07-19
Add info about failure modes to ConcurrentQueue Try methods

Smaug123:
## Summary

The docs are currently silent on the failure modes of most of the `System.Collections.Generic.ConcurrentQueue.Try*` methods. This PR clarifies what I believe are the semantics: that the `Try` operations can only fail for the trivial reason that the queue was empty at the time the operation was attempted.

### DIFF
--- a/xml/System.Collections.Concurrent/ConcurrentQueue`1.xml
+++ b/xml/System.Collections.Concurrent/ConcurrentQueue`1.xml
@@ -551,7 +551,7 @@
         <param name="item">The object to add to the <see cref="T:System.Collections.Concurrent.IProducerConsumerCollection`1" />. The value can be a null reference (Nothing in Visual Basic) for reference types.</param>
         <summary>Attempts to add an object to the <see cref="T:System.Collections.Concurrent.IProducerConsumerCollection`1" />.</summary>
         <returns>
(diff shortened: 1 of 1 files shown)


xml/System.Collections.Concurrent/ConcurrentQueue`1.xml
```diff
@@ -551,7 +551,7 @@
         <param name="item">The object to add to the <see cref="T:System.Collections.Concurrent.IProducerConsumerCollection`1" />. The value can be a null reference (Nothing in Visual Basic) for reference types.</param>
         <summary>Attempts to add an object to the <see cref="T:System.Collections.Concurrent.IProducerConsumerCollection`1" />.</summary>
         <returns>
-          <see langword="true" /> if the object was added successfully; otherwise, <see langword="false" />.</returns>
+        <see langword="true" /> if the object was added successfully; otherwise, <see langword="false" />. (This operation never returns <see langword="false" /> for <xref:System.Collections.Concurrent.ConcurrentQueue%601>.)</returns>
```
Smaug123: This *is* mentioned in the remarks section below, but that is not the first place I looked for the answer, suggesting that more docs are required.

PR #11600, opened 2025-07-23
Remove redundant item closing tag from dotpoint list

Issung:
## Summary

Was appearing in the documentation like this
image

I guess it's leftover from a conversion to markdown.


gewarren (approved):
Thanks @Issung

PR #11645, opened 2025-08-04
Add clarity that the Online mode can use cached data

bartonjs:
This also updates these members to not specify CRLs specifically, since revocation can come from other sources (e.g. OCSP).

### DIFF
--- a/xml/System.Security.Cryptography.X509Certificates/X509RevocationMode.xml
+++ b/xml/System.Security.Cryptography.X509Certificates/X509RevocationMode.xml
@@ -113,7 +113,7 @@
       </ReturnValue>
       <MemberValue>0</MemberValue>
       <Docs>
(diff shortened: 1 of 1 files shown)


vcsjones (approved):

xml/System.Security.Cryptography.X509Certificates/X509RevocationMode.xml
```diff
@@ -113,7 +113,7 @@
       </ReturnValue>
       <MemberValue>0</MemberValue>
       <Docs>
-        <summary>No revocation check is performed on the certificate.</summary>
+        <summary>Specifies that revocation checks should not be performed.</summary>
```
gewarren:
```suggestion
        <summary>Revocation checks should not be performed.</summary>
```

PR #11650, opened 2025-08-05
Use `nint` for native-sized integers in `Volatile`

xtqqczze:
Related: https://github.com/dotnet/runtime/pull/118360

cc: @jkotas


jkotas:
Same question as https://github.com/dotnet/dotnet-api-docs/pull/11652#issuecomment-3155545818

The change looks fine otherwise.

xtqqczze:
> The member signatures are autogenerated and will be overwritten

PR #11652, opened 2025-08-05
Use `nint` for native-sized integers in `Marshal`

xtqqczze:
Related: https://github.com/dotnet/runtime/pull/118335

cc: @jkotas

### DIFF
--- a/xml/System.Runtime.InteropServices/Marshal.xml
+++ b/xml/System.Runtime.InteropServices/Marshal.xml
@@ -329,7 +329,7 @@
       </Docs>
     </Member>
     <Member MemberName="AllocHGlobal">
-      <MemberSignature Language="C#" Value="public static IntPtr AllocHGlobal (IntPtr cb);" />
+      <MemberSignature Language="C#" Value="public static IntPtr AllocHGlobal (nint cb);" />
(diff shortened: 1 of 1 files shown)


AaronRobinsonMSFT (approved):

jkotas:
This is the first place that where `MemberSignature Language="C#"` uses nint instead IntPtr in this repo. @gewarren Are there any non-obvious consequences?

xtqqczze:
> This is the first place that where `MemberSignature Language="C#"` uses nint instead IntPtr in this repo. @gewarren Are there any non-obvious consequences?

`nint` is used for `Math.Abs`, but with a `FrameworkAlternate` attribute. Since `nint`/`nuint` become full aliases to `IntPtr`/`UIntPtr` in .NET 7 this suggests there could be consequences.

https://github.com/dotnet/dotnet-api-docs/blob/369ed46b4e28b0d0b7615526800933fe87aa1215/xml/System/Math.xml#L463-L469

gewarren:
> This is the first place that where `MemberSignature Language="C#"` uses nint instead IntPtr in this repo. @gewarren Are there any non-obvious consequences?

@jkotas The member signatures are autogenerated and will be overwritten. I'm not sure if the autogenerated signature will be the same as the change that was merged here, or if [this user story](https://dev.azure.com/ceapex/Engineering/_workitems/edit/722510) (which was bulk removed in June :-( ) will prevent the proper display.

PR #11653, opened 2025-08-05
Fix swapped `FrameworkAlternate` attribute for `nint`/`IntPtr`

xtqqczze:
Noticed in https://github.com/dotnet/dotnet-api-docs/pull/11652#issuecomment-3155665650


xtqqczze:
The existing source is technically correct, closing.

PR #11654, opened 2025-08-05
Add `FrameworkAlternate` for `nint`/`IntPtr` in `Marshal`

xtqqczze:
Use `nint`/`nuint` from .NET 7 to clarify intent; they became full aliases to `IntPtr`/`UIntPtr` from this version.

Follow-up https://github.com/dotnet/dotnet-api-docs/pull/11652


xtqqczze:
> The member signatures are autogenerated and will be overwritten

PR #11661, opened 2025-08-06
Update NativeLibrary.xml

ShalokShalom:
This changes the parameter **libraryPath** into **executableName**

Reasoning: 

The path is automatically handled by the abstraction of the API itself. The user is **not** supposed to provide a path: 

That would simply be not sensical, considering macOS, Linux, etc, all have different paths to begin with. 

The argument could never be multiple paths to all of them, and it's the executable name (that has to be on the PATH, obviously) that is our concern. 

I hope this is fine this way, and I also corrected two small typos.  Thanks

## Summary

Describe your changes here.

### DIFF
--- a/xml/System.Runtime.InteropServices/NativeLibrary.xml
+++ b/xml/System.Runtime.InteropServices/NativeLibrary.xml
@@ -133,12 +133,12 @@ Calling this method with an invalid `handle` parameter other than <xref:System.I
       </Docs>
     </Member>
     <Member MemberName="Load">
-      <MemberSignature Language="C#" Value="public static IntPtr Load (string libraryPath);" />
-      <MemberSignature Language="ILAsm" Value=".method public static hidebysig native int Load(string libraryPath) cil managed" />
+      <MemberSignature Language="C#" Value="public static IntPtr Load (string executableName);" />
+      <MemberSignature Language="ILAsm" Value=".method public static hidebysig native int Load(string executableName) cil managed" />
       <MemberSignature Language="DocId" Value="M:System.Runtime.InteropServices.NativeLibrary.Load(System.String)" />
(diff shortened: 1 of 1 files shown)


jkoritzinsky:
Changing the parameter name is a source-breaking change in C#. As a result, the documentation must match the source code (which uses the existing name). As a result, we cannot accept this sort of change.

AaronRobinsonMSFT:
Argument name changes are source breaking for C#. This would also require updating the source code and going through API review. I also don't agree with the argument being made to change this name. This API is intentially made for shared libraries, industry terminology, not simply "executables".

> The path is automatically handled by the abstraction of the API itself. The user is not supposed to provide a path:

That is an option, but not required. In fact it is very common for relative paths to be used.

> I hope this is fine this way, and I also corrected two small typos. Thanks

I would factor out the typos into a different PR.

ShalokShalom:
How are relative paths then supposed to be applied, so that they are cross platform compatible?

AaronRobinsonMSFT:
> How are relative paths then supposed to be applied, so that they are cross platform compatible?

A common pattern is to do something like below. The `IsOSPlatform()` check is treated as an intrinsic by the JIT so there should be no runtime overhead in a Release build.

```csharp
string path;
if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
{
    path = "rsc\foo.dll";
}
else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
{
    path = "rsc/libfoo.dylib";
}
else
{
    path = "rsc/libfoo.so";
}
NativeLibrary.Load(path);
```

ShalokShalom:
Do you agree, that it might be helpful, to include that information on the documentation page of the API?

PR #11677, opened 2025-08-12
Update SemaphoreSlim TimeSpan milliseconds limit

eduardo-vp:
Update according to https://github.com/dotnet/runtime/pull/117398. 

SemaphoreSlim's Wait/WaitAsync methods that take a TimeSpan now support all the TimeSpan range (except negative values other than -1).

### DIFF
--- a/xml/System.Threading/SemaphoreSlim.xml
+++ b/xml/System.Threading/SemaphoreSlim.xml
@@ -822,7 +822,11 @@ If a thread or task is blocked when calling <xref:System.Threading.SemaphoreSlim
  ]]></format>
         </remarks>
         <exception cref="T:System.ArgumentOutOfRangeException">
-          <paramref name="timeout" /> is a negative number other than -1, which represents an infinite timeout.</exception>
+          <paramref name="timeout" /> is a negative number other than -1, which represents an infinite timeout.
+
+-or-
+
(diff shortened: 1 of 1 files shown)


xml/System.Threading/SemaphoreSlim.xml
```diff
@@ -822,11 +822,7 @@ If a thread or task is blocked when calling <xref:System.Threading.SemaphoreSlim
  ]]></format>
         </remarks>
         <exception cref="T:System.ArgumentOutOfRangeException">
-          <paramref name="timeout" /> is a negative number other than -1, which represents an infinite timeout.
-
--or-
-
-<paramref name="timeout" /> is greater than <see cref="F:System.Int32.MaxValue">Int32.MaxValue</see>.</exception>
+          <paramref name="timeout" /> is a negative number other than -1, which represents an infinite timeout.</exception>
```
jkotas:
```suggestion
          <paramref name="timeout" /> is a negative number other than -1, which represents an infinite timeout.

-or-

In .NET 9 and earlier versions, <paramref name="timeout" /> is greater than <see cref="F:System.Int32.MaxValue">Int32.MaxValue</see>.</exception>
```
We have one set of docs for all .NET versions. Behavior differences between versions needs to called out so that the docs are applicable to all .NET version. Here is an example from existing System.Threading docs for reference: https://github.com/dotnet/dotnet-api-docs/blob/b314392d22b44ed9a5143a6984bc15636b096ea0/xml/System.Threading/Interlocked.xml#L1353

jkotas (approved):

Issue #11468, opened 2025-06-18
Delete code examples for obsolete members


gewarren:
### Describe the issue or suggestion

Similar to #11345, but this issue is to find obsolete *members* and delete any code examples in *those members only*.

Rubric is delete code examples for members if the member is:

- Obsolete since .NET Framework 4 or earlier
- Obsolete in .NET and deprecated in .NET Fx (e.g. CAS)
- Obsolete in .NET and not editor browsable in .NET Fx

See https://github.com/gewarren/tools/tree/main/EcmaXmlHelpers/ObsoleteExampleFinder.

gewarren:
**Candidates**:

- [x] FileSystem.FilePut
- [ ] ClipboardProxy.GetData - newly obsolete
- [x] AppDomain.ClearPrivatePath
- [ ] AppDomain.CreateDomain - obsolete in .NET only
- [x] AppDomain.CreateInstanceAndUnwrap
- [x] AppDomain.DefineDynamicAssembly
- [x] AppDomain.DefineDynamicAssembly
- [x] AppDomain.DefineDynamicAssembly
- [x] AppDomain.DefineDynamicAssembly
- [x] AppDomain.DefineDynamicAssembly
- [x] AppDomain.DefineDynamicAssembly
- [x] AppDomain.DefineDynamicAssembly
- [x] AppDomain.ExecuteAssembly
- [x] AppDomain.ExecuteAssembly
- [x] AppDomain.ExecuteAssembly
- [x] AppDomain.ExecuteAssembly
- [x] AppDomain.Load
- [x] AppDomain.SetAppDomainPolicy
- [x] AppDomain.SetDynamicBase
- [x] AppDomain.SetShadowCopyFiles
- [x] AppDomain.SetShadowCopyPath
- [ ] AppDomain.Unload - only obsolete in .NET
- [ ] Exception..ctor
- [ ] Exception.GetObjectData
- [ ] MarshalByRefObject.InitializeLifetimeService
- [ ] String.Copy
- [ ] Type.IsSerializable
- [ ] TypeLoadException..ctor
- [ ] TypeLoadException.GetObjectData
- [x] Uri..ctor
- [x] Uri..ctor
- [x] Uri.MakeRelative
- [x] Hashtable..ctor
- [x] Hashtable..ctor
- [x] Hashtable..ctor
- [x] Hashtable..ctor
- [x] Hashtable..ctor
- [x] SqlParameterCollection.Add
[...]

PR #11721, opened 2025-08-22
Delete code examples for obsolete API members

gewarren:
Contributes to #11468.

` git log --diff-filter=D --summary 0457aba..c8dd029 | grep 'delete mode' | wc -l`:

51 files deleted


gewarren:
Snippets 5000 failures are for missing projects, which I don't plan to address since no new snippets are being added.

PR #11756, opened 2025-09-02
fix the length parameter description for string constructors with sbyte*

### LINKED ISSUE #11136: Length parameter description confusing for constructors with sbyte*
### Type of issue

Typo

### Description

There are two constructors on this page which take an sbyte* input parameter.  For each, the `length` parameter is described as:

> The number of characters within `value` to use.

Unlike the other constructors on this page, using the term 'characters' for these two is misleading.  The number of characters within an sbyte* is variable, depending on how the decoding works out.  Fortunately, what this parameter really wants is a count of the number of sbytes, not the (more difficult to calculate) number of characters.

Proposed replacement text:

> The number of sbytes within `value` to use.

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.string.-ctor?view=net-9.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System/String.xml

90baab9e-9ef3-9bb5-d433-8d861d7fa876

### Platform Id

ac6a36d8-9684-17ab-5b56-e2696ccb8adf

### Article author

@dotnet-bot

StuartMosquera:
## Summary

- Replace all instances of the word 'characters' with 'sbytes' in the description of the `length` parameter in the `String` constructor that uses `sbyte*`.

Fixes #11136

### DIFF
--- a/xml/System/String.xml
+++ b/xml/System/String.xml
@@ -633,7 +633,7 @@ The sum of <paramref name="startIndex" /> and <paramref name="length" /> is grea
       <Docs>
         <param name="value">A pointer to an array of 8-bit signed integers. The integers are interpreted using the current system code page encoding on Windows (referred to as CP_ACP) and as UTF-8 encoding on non-Windows.</param>
         <param name="startIndex">The starting position within <paramref name="value" />.</param>
-        <param name="length">The number of characters within <paramref name="value" /> to use.</param>
+        <param name="length">The number of sbytes within <paramref name="value" /> to use.</param>
         <summary>Initializes a new instance of the <see cref="T:System.String" /> class to the value indicated by a specified pointer to an array of 8-bit signed integers, a starting position within that array, and a length.</summary>
         <remarks>
           <format type="text/markdown"><![CDATA[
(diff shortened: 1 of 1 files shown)


tannergooding:
This likely needs weigh in and comparison with other docs.

While the term `characters` can be confusing, it is not incorrect and matches the Unicode spec. That is, for `sbyte` the length represents number of `8-bit characters` much as for `System.Char` it is the number of `16-bit characters`.

It is not the number of `code units` or the number of `code points`, which are the likely what @rindlespot meant in the issue.

This particularly comes into play for UTF-16 when it comes to surrogate pairs, which are 2x UTF-16 characters, each representing 1x Unicode Code Unit, which combine to form a single `Code Point`.

For UTF-8, you have multiple characters which combine to form code units, which are then also valid code points since surrogates are illegal in UTF-8.

tannergooding:
It's in general a space where terms get thrown around and interchanged and while improving the wording can be good, we should ensure we do so consistently throughout the docs

StuartMosquera:
Could I clarify the bit width instead?

- For `char*`: `length`: The number of characters (16-bit) within `value` to use.
- For `sbyte*`: `length`: The number of characters (8-bit) within `value` to use.

The constructor with `sbyte*` already specifies its '8-bit' width in the value parameter description, while the one with `char*` does not explicitly mention its bit width.

tannergooding:
Looks like other places, like `System.Text.Encoding.Utf8` and `Utf8Formatter`, etc

We're simply using "the number of characters" when referring to UTF-16 and "the number of bytes" when referring to UTF-8. So "the number of sbytes" here is probably fine.

CC. @gewarren for secondary input and wording suggestions

tannergooding:
Thanks all!

PR #11772, opened 2025-09-05
Recommend LibraryImportAttribute in DllImportAttribute docs

AdmiralSnyder:
Updated documentation to recommend using LibraryImportAttribute.

## Summary

Describe your changes here.


AdmiralSnyder:
sorry, i was confused, because i thought i had just edited the docs for the ctor, so i closed the PR. but then I reconsidered. i think the link to LibraryImport should also be in the constructor's page (probably further up), since that's the one people find when they hit F1 in VS. and then i noticed that this is indeed the docs file for all things.

So the link to LibraryImport should also be put at each ctor, imo.

jkoritzinsky:
Thank you!

PR #11827, opened 2025-09-19
Clarify bufferSize parameter description in XML

ViveliDuCh:
Correct `bufferSize` parameter descriptions in `StreamReader` and `StreamWriter` XML docs

- Clarify `StreamReader.bufferSize` sets the internal byte buffer; `StreamWriter.bufferSize` sets the internal char buffer
- Remove incorrect buffer size units and misleading references
-  Omit secondary buffer details for clarity
- Align docs with implementation and recent runtime changes

Note that both classes allocate both buffer types, but `bufferSize` only directly affects one.
Documentation update in response to issue https://github.com/dotnet/runtime/issues/118748 and code changes in PR https://github.com/dotnet/runtime/pull/118958


xml/System.IO/StreamReader.xml
jozkee, line 1: The exceptions are also not accurate:
```xml
          <paramref name="bufferSize" /> is less than or equal to zero.</exception>
```
it should be IMO `<paramref name="bufferSize" /> is not -1 and is less than or equal to zero.`
cc @gewarren

xml/System.IO/StreamReader.xml
```diff
@@ -1008,7 +1008,7 @@
         <exception cref="T:System.ArgumentNullException">
           <paramref name="stream" /> or <paramref name="encoding" /> is <see langword="null" />.</exception>
         <exception cref="T:System.ArgumentOutOfRangeException">
-          <paramref name="bufferSize" /> is less than or equal to zero.</exception>
+          <paramref name="bufferSize" /> is less than or equal to zero, except for -1, which is allowed in stream-based constructors to indicate the default buffer size.</exception>
```
jozkee: After https://github.com/dotnet/runtime/pull/118958, this no longer applies only to stream-based ctors.
```suggestion
          <paramref name="bufferSize" /> is less than or equal to zero, except for -1, which is allowed to indicate the default buffer size.</exception>
```

PR #11877, opened 2025-10-02
Clarified about copying an immutable dictionary

WhiteBlackGoose:
## Summary

I think saying that a copy is made is misleading, because it makes the person assume a copy of the whole dictionary is made, whereas in reality operations like Add are O(log(N)) complex.

### DIFF
--- a/xml/System.Collections.Immutable/ImmutableDictionary`2.xml
+++ b/xml/System.Collections.Immutable/ImmutableDictionary`2.xml
@@ -97,7 +97,7 @@
       <format type="text/markdown"><![CDATA[
 
 ## Remarks
-  When you manipulate an immutable dictionary, a new one is returned with the appropriate changes, while the old one is not altered.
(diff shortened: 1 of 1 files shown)


xml/System.Collections.Immutable/ImmutableDictionary`2.xml
```diff
@@ -97,7 +97,7 @@
       <format type="text/markdown"><![CDATA[
 
 ## Remarks
- When you manipulate an immutable dictionary a copy of the original dictionary is made, manipulations applied and a new immutable dictionary is returned.
+ When you manipulate an immutable dictionary, a new one is returned with the appropriate changes, while the old one is not altered.
```
WhiteBlackGoose, line 100: Alternatively, "a new instance"? Or even "a new instance if necessary"?

WhiteBlackGoose: Alternatively: using the old wording but with "minimal copy" instead of "copy". Or "minimal (optimized) copy"

PR #11894, opened 2025-10-03
Clarify FileShare enum descriptions to remove ambiguity Fixes #10572

### LINKED ISSUE #10572: FileShare Enum descriptions are incorrect/misleading
### Type of issue

Other (describe below)

### Description

The enum values state/imply that the effect is applied on "subsequent opening of the file". This is not correct because the value can cause the current file open request to fail based on previously opened file handles.

For example, the FileShare.Read value states the following:

> Allows subsequent opening of the file for reading. If this flag is not specified, any request to open the file for reading (by this process or another process) will fail until the file is closed.

In the code below, the second attempt to open the file will fail because it is specifying FileShare.Read and the **previous** file open request has Write access to the file.

```
using var fileStream1 = new FileStream("test.file", FileMode.OpenOrCreate, FileAccess.Write, FileShare.Read);
using var fileStream2 = new FileStream("test.file", FileMode.Open, FileAccess.Read, FileShare.Read);
```

Please update the documentation so that it is clear that the FileShare value applies to existing file handles as well as future attempts to open the file.

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.io.fileshare?view=net-8.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.IO/FileShare.xml

3bd9dbb8-0a53-ad3f-cd5d-c18d47f7f1a1

### Article author

@dotnet-bot

oii-nasif:
## Summary

  - Removed misleading word "subsequent" from all FileShare enum member descriptions
  - Clarified that FileShare flags control access for all file operations, not just future ones
  - Updated descriptions for Delete, Read, ReadWrite, and Write enum members

  ## Details
  The original FileShare enum documentation used the word "subsequent" in descriptions like "Allows subsequent
  opening of the file for reading", which incorrectly implied that these flags only affect future file open attempts
   after the current one.

  In reality, FileShare flags control whether **any** attempt to open the file (including the current one) will
  succeed, based on how existing file handles were opened. This was demonstrated in issue #10572 with this example:

  ```csharp
  using var fileStream1 = new FileStream("test.file", FileMode.OpenOrCreate, FileAccess.Write, FileShare.Read);
  using var fileStream2 = new FileStream("test.file", FileMode.Open, FileAccess.Read, FileShare.Read);
```

  The second open fails because the first handle has Write access, even though both specify FileShare.Read. The
  original documentation's use of "subsequent" made this behavior unclear.

  Changes

  - Delete: "Allows subsequent deleting" → "Allows the file to be deleted by this process or other processes"
  - Read: "Allows subsequent opening for reading" → "Allows the file to be opened for reading by this process or
  other processes"
  - ReadWrite: "Allows subsequent opening for reading or writing" → "Allows the file to be opened for reading or
  writing by this process or other processes"
  - Write: "Allows subsequent opening for writing" → "Allows the file to be opened for writing by this process or
  other processes"

  Fixes #10572

### DIFF
--- a/xml/System.IO/FileShare.xml
+++ b/xml/System.IO/FileShare.xml
@@ -142,7 +142,7 @@
       </ReturnValue>
       <MemberValue>4</MemberValue>
       <Docs>
-        <summary>Allows subsequent deleting of a file.</summary>
+         <summary>Allows the file to be deleted by this process or other processes. The file can be deleted while it remains open.</summary>
       </Docs>
     </Member>
     <Member MemberName="Inheritable">
@@ -277,7 +277,7 @@
       </ReturnValue>
       <MemberValue>1</MemberValue>
       <Docs>
-        <summary>Allows subsequent opening of the file for reading. If this flag is not specified, any request to open the file for reading (by this process or another process) will fail until the file is closed. However, even if this flag is specified, additional permissions might still be needed to access the file.</summary>
+         <summary>Allows the file to be opened for reading by this process or other processes. If this flag is not specified, any request to open the file for reading (by this process or another process) will fail until the file is closed. However, even if this flag is specified, additional permissions might still be needed to access the file.</summary>
       </Docs>
     </Member>
     <Member MemberName="ReadWrite">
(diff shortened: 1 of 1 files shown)


danmoseley:
For some libraries in dotnet/runtime, the content in this repo is source of truth; in others, it's the sources (where we want to be eventually). If it's sources for this one, the fix should go in dotnet/runtime, but I don't know.

gewarren:
> For some libraries in dotnet/runtime, the content in this repo is source of truth; in others, it's the sources (where we want to be eventually). If it's sources for this one, the fix should go in dotnet/runtime, but I don't know.

This repo is the correct place for this library.

PR #11895, opened 2025-10-03
Align MemoryStream Position and Seek docs with SetLength

### LINKED ISSUE #9201: Align MemoryStream Position and Seek to mention origin on ArgumentOutOfRangeException
MemoryStream was recently fixed to avoid falling into a state where it could contain a negative `Position` https://github.com/dotnet/runtime/pull/88572. 

Some API docs don't mention the `origin` when they talk about ArgumentOutOfRangeException and only mention that the value must be between int.MaxValue and zero.
* [Seek()](https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream.seek?view=net-7.0#exceptions)
* [Position](https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream.position?view=net-7.0#exceptions)

[SetLength](https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream.setlength?view=net-7.0) does correctly reflect that the limit accounts for the origin:
> value is negative or is greater than the maximum length of the MemoryStream, where the maximum length is (Int32.MaxValue - origin), and origin is the index into the underlying buffer at which the stream starts.

We need to amend the former cases to align with SetLength's.

oii-nasif:
## Summary
  - Updated `ArgumentOutOfRangeException` documentation for `MemoryStream.Position` property and `MemoryStream.Seek`
   method
  - Aligned exception descriptions with the more detailed explanation used in `SetLength` method
  - Added clarification about how the `origin` parameter affects the maximum valid position/offset

  ## Details
  The `SetLength` method already included a precise explanation of the `ArgumentOutOfRangeException` that mentions
  the `origin` parameter:

  > where the maximum length is (Int32.MaxValue - origin), and origin is the index into the underlying buffer at
  which the stream starts.

  However, the `Position` property and `Seek` method only mentioned `Int32.MaxValue` without explaining how the
  origin affects the valid range.

  ### Changes Made

  **Position property** - Updated from:
  > The position is set to a negative value or a value greater than Int32.MaxValue.

  To:
  > The position is set to a negative value or a value greater than the maximum length of the MemoryStream, where
  the maximum length is (Int32.MaxValue - origin), and origin is the index into the underlying buffer at which the
  stream starts.

  **Seek method** - Updated from:
  > offset is greater than Int32.MaxValue.

  To:
  > offset is greater than the maximum length of the MemoryStream, where the maximum length is (Int32.MaxValue -
  origin), and origin is the index into the underlying buffer at which the stream starts.

  This provides developers with consistent and more accurate information about the actual constraints when working
  with MemoryStream instances that may have been initialized with a non-zero origin offset.

  Fixes #9201

### DIFF
--- a/xml/System.IO/MemoryStream.xml
+++ b/xml/System.IO/MemoryStream.xml
@@ -1800,7 +1800,7 @@ The pending operation does not support writing.</exception>
 
  ]]></format>
         </remarks>
-        <exception cref="T:System.ArgumentOutOfRangeException">The position is set to a negative value or a value greater than <see cref="F:System.Int32.MaxValue">Int32.MaxValue</see>.</exception>
+        <exception cref="T:System.ArgumentOutOfRangeException">The position is set to a negative value or a value greater than the maximum length of the <see cref="T:System.IO.MemoryStream" />, where the maximum length is (<see cref="F:System.Int32.MaxValue">Int32.MaxValue</see> - <code>origin</code>), and <code>origin</code> is the index into the underlying buffer at which the stream starts.</exception>
         <exception cref="T:System.ObjectDisposedException">The stream is closed.</exception>
         <related type="Article" href="/dotnet/standard/io/">File and Stream I/O</related>
         <related type="Article" href="/dotnet/standard/io/how-to-read-text-from-a-file">How to: Read Text from a File</related>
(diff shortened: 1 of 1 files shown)


oii-nasif:
Hello @gewarren , Have you had a chance to review my changes yet?

PR #11900, opened 2025-10-04
Clarify Enumerable.Prepend returns a new sequence

### LINKED ISSUE #11745: Confusing Function Definition for Enumerable.Prepend<TSource>(IEnumerable<TSource>, TSource) Method
### Type of issue

Other (describe below)

### Description

Enumerable.Prepend is defined as:

> Adds a value to the beginning of the sequence.

However, later down the documentation within the Remarks section that mentions that:

> This method does not modify the elements of the collection. Instead, it creates a copy of the collection with the new element.

And in the Examples section it demonstrates that this method does not work when trying to just Prepend the List since it does not modify the original list:

```
// Trying to prepend any value of the same type
numbers.Prepend(0);

// It doesn't work because the original list has not been changed
Console.WriteLine(string.Join(", ", numbers));
```

So why is the definition for this function something along the lines of:

"Returns a list with the value added to the beginning of the sequence"

Instead of:

"Adds a value to the beginning of the sequence."

Since that mis-leads you into thinking that it is modifying the original list.

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.prepend?view=net-9.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Linq/Enumerable.xml

9556f661-a131-5b65-913f-50770cc980cf

### Platform Id

acda5d20-c39a-4d6d-e70e-28d92c16a6b3

### Article author

@dotnet-bot

oii-nasif:
## Summary
  - Updated `Enumerable.Prepend<TSource>` summary to clarify it returns a new sequence
  - Changed from "Adds a value" (implies modification) to "Returns a new sequence" (accurate behavior)

  ## Details
  The `Enumerable.Prepend<TSource>` method had a confusing summary that stated "Adds a value to the beginning of the
   sequence." This wording incorrectly implied that the method modifies the original sequence in place.

  ### The Problem
  The original summary was misleading because:
  - "Adds a value" suggests in-place modification
  - Developers might expect the source sequence to be changed
  - This contradicts LINQ's core principle of deferred execution and immutability

  While the Remarks section included a note clarifying "This method does not modify the elements of the collection.
  Instead, it creates a copy of the collection with the new element," the summary itself was still confusing and
  contradictory.

  ### Changes Made
  Updated the summary from:
  > Adds a value to the beginning of the sequence.

  To:
  > Returns a new sequence that begins with the specified element followed by the elements of the source sequence.

  ### Why This Matters
  This clarification helps developers:
  - Understand that `Prepend` follows LINQ's immutability pattern
  - Avoid expecting the source collection to be modified
  - Write correct code that properly handles the returned sequence
  - Align with how other LINQ methods like `Append`, `Select`, `Where`, etc. are documented

  The new wording is consistent with the Returns section which already correctly stated "A new sequence that begins
  with element."

  Fixes #11745

### DIFF
--- a/xml/System.Linq/Enumerable.xml
+++ b/xml/System.Linq/Enumerable.xml
@@ -11681,7 +11681,7 @@ If comparer is `null`, the default comparer <see cref="P:System.Collections.Gene
         <typeparam name="TSource">The type of the elements of <paramref name="source" />.</typeparam>
         <param name="source">A sequence of values.</param>
         <param name="element">The value to prepend to <paramref name="source" />.</param>
-        <summary>Adds a value to the beginning of the sequence.</summary>
+        <summary>Returns a new sequence that begins with the specified element followed by the elements of the source sequence.</summary>
         <returns>A new sequence that begins with <paramref name="element" />.</returns>
         <remarks>
           <format type="text/markdown"><![CDATA[


xml/System.Linq/Enumerable.xml
rsobodosh, line 1: Do you think the Remark section is no longer necessary?

image

eiriktsarpalis: FWIW this is true in general with LINQ methods, mutations are never being made to the source collection(s). While I don't object to the change in question, users reading this part should have already internalized this invariant of LINQ.

gewarren:
The return value and remarks make it clear that the method returns a new sequence, so I will close this PR.

Issue #11939, opened 2025-10-19
Return type of ValueTuple.Create with 8 arguments is rendered incorrectly


KalleOlaviNiemitalo:
### Type of issue

Other (describe below)

### Description

At <https://learn.microsoft.com/en-us/dotnet/api/system.valuetuple.create?view=net-9.0#system-valuetuple-create-8(-0-1-2-3-4-5-6-7)>, the method signature is rendered as:

```csharp
public static(T1, T2, T3, T4, T5, T6, T7,(T8)) Create<T1,T2,T3,T4,T5,T6,T7,T8>(T1 item1, T2 item2, T3 item3, T4 item4, T5 item5, T6 item6, T7 item7, T8 item8);
```

The return type should not have parentheses around `T8`:

```csharp
public static(T1, T2, T3, T4, T5, T6, T7, T8) Create<T1,T2,T3,T4,T5,T6,T7,T8>(T1 item1, T2 item2, T3 item3, T4 item4, T5 item5, T6 item6, T7 item7, T8 item8);
```

According to [C# 7 §8.3.11](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/types.md#8311-tuple-types), `(T1, T2, T3, T4, T5, T6, T7, T8)` means `ValueTuple<T1, T2, T3, T4, T5, T6, T7, ValueTuple<T8>>`.  This is the actual return type of the eight-parameter ValueTuple.Create method.  In this type, the outer structure is [`ValueTuple<T1, T2, T3, T4, T5, T6, T7, TRest>`](https://learn.microsoft.com/dotnet/api/system.valuetuple-8?view=net-9.0), whose `TRest` type parameter requires a tuple type.

`(T1, T2, T3, T4, T5, T6, T7, (T8))` means `ValueTuple<T1, T2, T3, T4, T5, T6, T7, ValueTuple<ValueTuple<T8>>>`.  That is not what ValueTuple.Create returns.

In the following member signatures, C# is wrong, but the others are correct because they use generic type syntax rather than tuple type syntax: <https://github.com/dotnet/dotnet-api-docs/blob/9ba1f1b43dfb2b9bc49eb84a2023b1b1ca0c82df/xml/System/ValueTuple.xml#L953-L959>

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.valuetuple.create?view=net-9.0#system-valuetuple-create-8(-0-1-2-3-4-5-6-7)

### Content source URL

[...]

PR #11940, opened 2025-10-19
Return type of ValueTuple.Create with 8 arguments is rendered incorrectly (#11939)

kanilsz:
## Summary

Fix return type of ValueTuple.Create with 8 arguments 

Fixes dotnet/dotnet-api-docs/issues/#11939


kanilsz:
> @dotnet-policy-service agree

@dotnet-policy-service agree

kanilsz:
Please add hacktoberfest-accepted label in case of a valid contribution.

gewarren:
The signatures are autogenerated. I appreciate you trying to fix but I will have to close this as it will just be overwritten by CI.

PR #12138, opened 2025-12-12
Updated comment on enum System.Net.HttpStatusCode

abatishchev:
## Summary

Updated the language to reduce the repetition of word "defined" and updated the number of the RFC, from [2616](https://datatracker.ietf.org/doc/html/rfc2616) to [7231](https://datatracker.ietf.org/doc/html/rfc7231)

### DIFF
--- a/xml/System.Net/HttpStatusCode.xml
+++ b/xml/System.Net/HttpStatusCode.xml
@@ -50,7 +50,7 @@
     <BaseTypeName>System.Enum</BaseTypeName>
   </Base>
   <Docs>
-    <summary>Contains the values of HTTP status codes as defined by RFC 7231, section 6.</summary>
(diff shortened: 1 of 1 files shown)


rzikm (approved):

xml/System.Net/HttpStatusCode.xml
```diff
@@ -50,7 +50,7 @@
     <BaseTypeName>System.Enum</BaseTypeName>
   </Base>
   <Docs>
-    <summary>Contains the values of status codes defined for HTTP defined in RFC 2616 for HTTP 1.1.</summary>
+    <summary>Contains the values of HTTP status codes as defined by RFC 7231, section 6.</summary>
```
MihaZupan:
```suggestion
    <summary>Contains the values of HTTP status codes as defined by RFC 9110, section 15.</summary>
```
Nit: Might as well reference the newest one

PR #12178, opened 2025-12-28
xml/System.Threading/Timer.xml: Divides logical blocks to separate paragraphs

tesar-tech:
## Summary

I believe the sentence `If period is zero (0) or` should be in separate paragraph. It is the same kind of information as `If  dueTime is zero (0),`(which has it's own separate paragraph)


gewarren (approved):
Thanks @tesar-tech

PR #12191, opened 2026-01-03
Fix wording in DataColumn.DataType remarks section

ericmutta:
The fragment:

> ...is very important **to** guaranteeing...

Should probably read:

> ...is very important **for** guaranteeing...


roji (approved):
Thanks.

ericmutta:
Thanks @roji for the prompt follow-up and Happy New Year 🎆

PR #12203, opened 2026-01-07
Update remarks for TickCount

gewarren:
Contributes to https://github.com/dotnet/docs/issues/50755.


xml/System/Environment.xml
```diff
@@ -2516,22 +2516,22 @@ The following example creates environment variables for the <xref:System.Environ
...
+> Because the value of the <xref:System.Environment.TickCount> property value is a 32-bit signed integer, if the system runs continuously, <xref:System.Environment.TickCount> increments from zero to <xref:System.Int32.MaxValue?displayProperty=nameWithType> for approximately 24.9 days, then jumps to <xref:System.Int32.MinValue?displayProperty=nameWithType>, which is a negative number, then increments back to zero during the next 24.9 days. You can work around this issue by calling the Windows [GetTickCount](/windows/win32/api/sysinfoapi/nf-sysinfoapi-gettickcount) function, which resets to zero after approximately 49.7 days, or by using the <xref:System.Environment.TickCount64> property, which returns a 64-bit integer.
```
tannergooding: nit: The workaround given here is Windows only and dependent on how the user writes the P/Invoke to `GetTickCount`. It is also no longer compatible with the new behavior of `Environment.TickCount`.

A safer alternative which allows the `49.7 day` wraparound is simply to cast the result of `Environment.TickCount` to `uint`. This works because `int` and `uint` are a two's complement representation and is valid for all platforms/scenarios.

xml/System/Environment.xml
```diff
@@ -2516,22 +2516,22 @@ The following example creates environment variables for the <xref:System.Environ
...
 
- :::code language="csharp" source="~/snippets/csharp/System/Environment/TickCount/tickcount.cs" id="Snippet1":::
- :::code language="fsharp" source="~/snippets/fsharp/System/Environment/TickCount/tickcount.fs" id="Snippet1":::
- :::code language="vb" source="~/snippets/visualbasic/System/Environment/TickCount/tickcount.vb" id="Snippet1":::
+The following example demonstrates how to retrieve the positive range of values returned by the <xref:System.Environment.TickCount> property. The <xref:System.Environment.TickCount> property cycles between <xref:System.Int32.MinValue?displayProperty=nameWithType>, which is a negative number, and <xref:System.Int32.MaxValue?displayProperty=nameWithType> once every 49.8 days. This code sample removes the sign bit to yield a nonnegative number that cycles between zero and <xref:System.Int32.MaxValue> once every 24.9 days.
```
tannergooding: While this example works, I think the following example is also good and shows how to persist the full range:
```csharp
using System;

uint result = (uint)Environment.TickCount;
Console.WriteLine("TickCount: {result}");
```

-- The sample linked to here also has some incorrect indentation and looks fairly "outdated". That is, it isn't using top level statements, isn't using interpolated strings, etc.

xml/System/Environment.xml
```diff
@@ -2571,7 +2571,18 @@ The following example creates environment variables for the <xref:System.Environ
...
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+The elapsed time excludes any non-awake time, such as time spent in sleep mode, hibernation mode, or other low-power states.
+
+This value updates at the same frequency as the underlying interrupt timer for the system.
+
+> [!NOTE]
+> On Windows, in .NET 9 and earlier versions, including .NET Framework, the elapsed time does **not** exclude non-awake time. In addition, the value updates at a fixed cadence of 10-16ms (typically 15.5ms).
```
tannergooding:
```suggestion
> On Windows, in .NET 10 and earlier versions, including .NET Framework, the elapsed time does **not** exclude non-awake time. In addition, the value updates at a fixed cadence of 10-16ms (typically 15.5ms).
```

tannergooding: That is, this behavior also exists on .NET 10 and .NET 11 Preview 1 will be the first version with the new behavior

gewarren: Oops, I'm still stuck on the .NET 10 release LOL.

xml/System/Environment.xml
```diff
@@ -2516,25 +2516,18 @@ The following example creates environment variables for the <xref:System.Environ
...
-
- <xref:System.Environment.TickCount%2A> is different from the <xref:System.DateTime.Ticks%2A?displayProperty=nameWithType> property, which is the number of 100-nanosecond intervals that have elapsed since 1/1/0001, 12:00am.
+> Because the value of the <xref:System.Environment.TickCount> property value is a 32-bit signed integer, if the system runs continuously, <xref:System.Environment.TickCount> increments from zero to <xref:System.Int32.MaxValue?displayProperty=nameWithType> for approximately 24.9 days, then jumps to <xref:System.Int32.MinValue?displayProperty=nameWithType>, which is a negative number, then increments back to zero during the next 24.9 days. You can work around this issue by casting the result of <xref:System.Environment.TickCount> to `uint`. This works because `int` and `uint` are a two's complement representation on all platforms.
```
gewarren:
```suggestion
> Because the value of the <xref:System.Environment.TickCount> property value is a 32-bit signed integer, if the system runs continuously, <xref:System.Environment.TickCount> increments from zero to <xref:System.Int32.MaxValue?displayProperty=nameWithType> for approximately 24.9 days, then jumps to <xref:System.Int32.MinValue?displayProperty=nameWithType>, which is a negative number, then increments back to zero during the next 24.9 days. You can work around this issue by casting the result of <xref:System.Environment.TickCount> to `uint`. This works because `int` and `uint` are a two's complement representation. This workaround is valid on all platforms.
```

tannergooding (approved):

PR #12232, opened 2026-01-14
DO NOT SIMPLY MERGE READ DESCRIPTION - InstallUtil doesn't work with net core executables?

AdmiralSnyder:
We tried installing a service executable converted from .net framework 4.8 to net10 using installUtil, as suggested here, but that errored because of a BadImageFormatException:

> Beim Initialisieren der Installation ist eine Ausnahme aufgetreten:
System.BadImageFormatException: Die Datei oder Assembly "file:///D:\Publish\NexusCbfcLite\Nexus.BackgroundService.exe" oder eine Abhängigkeit davon wurde nicht gefunden. Im Modul wurde ein Assemblymanifest erwartet.

I kinda expected that since ServiceUtil is a .net framework process.

using `sc create` worked, though. documentation shoud reflect this (after this is cleared to be correct, of course)

## Summary

Describe your changes here.


AdmiralSnyder:
oh. i just now realized that creating issues on this repo is possible now - so this should probably have been an issue ... well, since it's containing the change in case the question i asked turns out to be correct, i'll leave it a PR.

PR #12275, opened 2026-01-31
Update returns documentation for StringBuilder comparison

mahfeshar:
This update corrects the documentation for the StringBuilder.Equals(StringBuilder) method to align with the actual behavior in modern .NET versions (.NET Core 3.0 and later).

Identified Issues:

Incorrect Returns Description: The <returns> section currently states that Capacity and MaxCapacity are compared. This is legacy behavior from .NET Framework and no longer applies to modern .NET, where only the string content is compared.

Outdated Example Output: The code example's expected output incorrectly shows False when comparing two StringBuilder objects with different capacities but identical content. In modern .NET, this returns True.

Internal Contradiction: The "Remarks" section correctly mentions the change in .NET Core 3.0, but the "Returns" and "Example" sections were never updated, leading to developer confusion.

Proposed Changes:

Refine <returns>: Updated the text to focus on character equality rather than memory capacity.

Update Snippet Output: Adjusted the sample output (case b4) from False to True to match the results when running on modern .NET runtimes.

### DIFF
--- a/xml/System.Text/StringBuilder.xml
+++ b/xml/System.Text/StringBuilder.xml
@@ -5270,6 +5270,7 @@ The `Equals` method performs an ordinal comparison to determine whether the char
           <para>On .NET (Core), <see langword="true" /> if the characters in this instance and <paramref name="sb" /> are equal; otherwise, <see langword="false" />.</para>
           <para>On .NET Framework, <see langword="true" /> if this instance and <paramref name="sb" /> have equal string, <see cref="P:System.Text.StringBuilder.Capacity" />, and <see cref="P:System.Text.StringBuilder.MaxCapacity" /> values; otherwise, <see langword="false" />.</para>
         </returns>
+        <remarks>
           <format type="text/markdown"><![CDATA[
 
 ## Remarks


xml/System.Text/StringBuilder.xml
```diff
@@ -5267,8 +5267,8 @@ The `Equals` method performs an ordinal comparison to determine whether the char
         <param name="sb">An object to compare with this instance, or <see langword="null" />.</param>
         <summary>Returns a value indicating whether this instance is equal to a specified object.</summary>
         <returns>
-          <see langword="true" /> if this instance and <paramref name="sb" /> have equal string, <see cref="P:System.Text.StringBuilder.Capacity" />, and <see cref="P:System.Text.StringBuilder.MaxCapacity" /> values; otherwise, <see langword="false" />.</returns>
-        <remarks>
+          <see langword="true" /> if the characters in this instance and <paramref name="sb" /> are equal; otherwise, <see langword="false" />.
```
gewarren: I think we need to keep both and differentiate between .NET Framework vs. modern .NET since the docs are shared by both.

mahfeshar: We need to keep both I agree but I think we should make it clear more, cause when I am reading it again and I am using modern one, it was not that clear

gewarren:
```suggestion
          <para>On .NET (Core), <see langword="true" /> if the characters in this instance and <paramref name="sb" /> are equal; otherwise, <see langword="false" />.</para>
          <para>On .NET Framework, <see langword="true" /> if this instance and <paramref name="sb" /> have equal string, <see cref="P:System.Text.StringBuilder.Capacity" />, and <see cref="P:System.Text.StringBuilder.MaxCapacity" /> values; otherwise, <see langword="false" />.</para>
```

xml/System.Text/StringBuilder.xml
```diff
@@ -5267,8 +5267,9 @@ The `Equals` method performs an ordinal comparison to determine whether the char
         <param name="sb">An object to compare with this instance, or <see langword="null" />.</param>
         <summary>Returns a value indicating whether this instance is equal to a specified object.</summary>
         <returns>
-          <see langword="true" /> if this instance and <paramref name="sb" /> have equal string, <see cref="P:System.Text.StringBuilder.Capacity" />, and <see cref="P:System.Text.StringBuilder.MaxCapacity" /> values; otherwise, <see langword="false" />.</returns>
-        <remarks>
+          <para>On .NET (Core), <see langword="true" /> if the characters in this instance and <paramref name="sb" /> are equal; otherwise, <see langword="false" />.</para>
+          <para>On .NET Framework, <see langword="true" /> if this instance and <paramref name="sb" /> have equal string, <see cref="P:System.Text.StringBuilder.Capacity" />, and <see cref="P:System.Text.StringBuilder.MaxCapacity" /> values; otherwise, <see langword="false" />.</para>
+        </returns>
```
gewarren, line 5272:
```suggestion
        </returns>
        <remarks>
```

gewarren (approved):

PR #12280, opened 2026-02-04
Explain what "normal" is in Single.IsNormal

sfiruch:
## Summary

The documentation for Single.IsNormal is wholly unhelpful. It says it checks if a number is "Normal". Only in the remarks, "normal" is explained. This changes it to make it clear what it actually checks.


tannergooding:
This should probably be combined into the same PR with #12281 and with the same feedback given there addressed.

PR #12281, opened 2026-02-04
Explain what "normal" is in Double.IsNormal

sfiruch:
## Summary

The documentation for Double.IsNormal is wholly unhelpful. It says it checks if a number is "Normal". Only in the remarks, "normal" is explained. This changes it to make it clear what it actually checks.

### DIFF
--- a/xml/System/Double.xml
+++ b/xml/System/Double.xml
@@ -2947,7 +2947,7 @@ This method correctly handles floating-point values and so `2.0` and `3.0` will 
       </Parameters>
       <Docs>
         <param name="d">A double-precision floating-point number.</param>
-        <summary>Determines whether the specified value is normal.</summary>
+        <summary>Determines whether the specified value is not NaN, not infinite, not subnormal, and not zero.</summary>
         <returns>
           <see langword="true" /> if the value is normal; <see langword="false" /> otherwise.</returns>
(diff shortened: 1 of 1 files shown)


xml/System/Double.xml
```diff
@@ -2947,7 +2947,7 @@ This method correctly handles floating-point values and so `2.0` and `3.0` will
       </Parameters>
       <Docs>
         <param name="d">A double-precision floating-point number.</param>
-        <summary>Determines whether the specified value is normal.</summary>
+        <summary>Determines whether the specified value is not NaN, not infinite, not subnormal, and not zero.</summary>
```
tannergooding, line 2950: If you're wanting to not be repetitive, then repeating `subnormal` doesn't really help either.. The technical definition is:
```suggestion
        <summary>Determines whether the specified value is a finite non-zero value with magnitude greater than or equal to <c>2<sup>-1022</sup></c>.</summary>
```

tannergooding: I would expect `Single`, `Half`, `NFloat`, and `INumberBase<TSelf>` to all be updated in the same PR.

I would also expect other definitions like `IsInfinite`, `IsSubnormal`, `IsFinite`, etc to be updated if required.

tannergooding: And as per the Copilot guidance below, the summary and remarks are effectively just mirroring eachother now, so the remarks need to either be removed or updated.

PR #12288, opened 2026-02-06
Fix parameter reference in TensorPrimitives.Hypot

vcsjones:
## Summary

The remark for `TensorPrimitives.Hypot` referred to `x` twice instead of `x` and `y`.

Before:

> This method effectively computes <code>destination[i] = T.Hypot(x[i], x[i])</code>.

After

> This method effectively computes <code>destination[i] = T.Hypot(x[i], y[i])</code>.


gewarren:
@vcsjones This should be fixed in the runtime repo instead: https://github.com/dotnet/runtime/blob/591b3de9a58d8669f606533d50469e45a6b64198/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/netcore/TensorPrimitives.Hypot.cs#L21

PR #12354, opened 2026-02-27
Added exceptions that DivRem raises

andanteyk:
## Summary

Added exceptions (especially `OverflowException`) that `Math.DivRem` or `Int***.DivRem` raise.

From https://github.com/dotnet/runtime/discussions/124932

### DIFF
--- a/xml/System/Math.xml
+++ b/xml/System/Math.xml
@@ -2882,6 +2882,8 @@ Dim i2 As Integer = CInt(Math.Ceiling(d2))        ' Result: 7969
         <summary>Produces the quotient and the remainder of two unsigned 8-bit numbers.</summary>
         <returns>The quotient and the remainder of the specified numbers.</returns>
         <remarks>To be added.</remarks>
+        <exception cref="T:System.DivideByZeroException">
+          <paramref name="right" /> is zero.</exception>
       </Docs>
     </Member>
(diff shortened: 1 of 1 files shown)


tannergooding:
Could you update the unsigned integer ones to include the `DivideByZeroException` as well?

Would probably also be good to cover `BigInteger` and `IBinaryInteger`, but otherwise changes LGTM

xml/System.Numerics/BigInteger.xml
```diff
@@ -1716,6 +1716,8 @@ The individual bytes in the `value` array should be in little-endian order, from
         <returns>The quotient and remainder of <paramref name="left" /> divided-by <paramref name="right" />.</returns>
         <remarks>To be added.</remarks>
         <inheritdoc cref="M:System.Numerics.IBinaryInteger`1.DivRem(`0,`0)" />
+        <exception cref="T:System.DivideByZeroException">
+          <paramref name="right" /> is zero.</exception>
```
gewarren, line 1720:
```suggestion
          <paramref name="right" /> is 0.</exception>
```

@tannergooding Wouldn't 0 be better than zero here?

tannergooding: I have no preference. There are times where I think `zero` is more readable and others where I think `0` is instead.

PR #12365, opened 2026-03-03
Add GC.KeepAlive example to Marshal.GetFunctionPointerForDelegate

danmoseley:
The existing remarks for both `GetFunctionPointerForDelegate` overloads say "You must manually keep the delegate from being collected" but don't show how. This adds a concrete code example using `GC.KeepAlive`, and explains that if native code stores the function pointer persistently, the delegate must be rooted in a `static` field.

### DIFF
--- a/xml/System.Runtime.InteropServices/Marshal.xml
+++ b/xml/System.Runtime.InteropServices/Marshal.xml
@@ -4170,7 +4170,7 @@ The code retrieves a reference to an instance of Microsoft Word successfully. Ho
  var callback = new MyNativeCallback(MyManagedMethod);
  IntPtr fnPtr = Marshal.GetFunctionPointerForDelegate(callback);
  NativeMethod(fnPtr);
- GC.KeepAlive(callback); // Prevent collection — fnPtr does not root the delegate
+ GC.KeepAlive(callback); // Prevent collection — fnPtr does not root the delegate.
  ```
 
  If native code stores the function pointer beyond the duration of the call, root the delegate for its entire lifetime — for example, by storing it in a `static` field.


danmoseley:
for @AaronRobinsonMSFT to review before merge

danmoseley:
@gewarren feel free to merge if looks ok

xml/System.Runtime.InteropServices/Marshal.xml
```diff
@@ -4155,7 +4164,16 @@ The code retrieves a reference to an instance of Microsoft Word successfully. Ho
...
+ You must manually keep the delegate from being collected by the garbage collector from managed code. The garbage collector does not track references to unmanaged code. Use <xref:System.GC.KeepAlive%2A> to prevent the delegate from being collected before the native call completes:
+
+ ```csharp
+ var callback = new MyNativeCallback(MyManagedMethod);
+ IntPtr fnPtr = Marshal.GetFunctionPointerForDelegate(callback);
+ NativeMethod(fnPtr);
+ GC.KeepAlive(callback); // Prevent collection — fnPtr does not root the delegate.
+ ```
+
+ If native code stores the function pointer beyond the duration of the call, root the delegate for its entire lifetime — for example, by storing it in a `static` field.
```
AaronRobinsonMSFT, line 4176: A mitigation for this is to define the `Delegate` as an actual argument and the system will pin it for you. Both `DllImport` and `LibraryImport` do this, but it isn't uncommon for people to want blittable P/Invokes and write the argument as `IntPtr` which requires the `GC.KeepAlive()` suggestion above.

danmoseley: this missed the merge -- you could offer a PR for that tweak possibly

AaronRobinsonMSFT: Done - https://github.com/dotnet/dotnet-api-docs/pull/12369

PR #12388, opened 2026-03-11
%2A properties - System.Xml and System.Xaml

gewarren:


gewarren:
@Copilot Can you remove the `%2A` from all property xrefs in these files?

PR #12402, opened 2026-03-16
Augment AllowOutOfOrderMetadataProperties remarks with buffering/backtracking detail

eiriktsarpalis:
Expands the Remarks section for `AllowOutOfOrderMetadataProperties` on both `JsonSerializerOptions` and `JsonSourceGenerationOptionsAttribute` to better document the performance tradeoffs of enabling this property.

### `JsonSerializerOptions.AllowOutOfOrderMetadataProperties`
Replaced the single-sentence remark with a full Remarks section covering:
- **What the property does**: relaxes the requirement that metadata properties appear first in JSON objects.
- **Buffering**: the deserializer must buffer entire JSON objects in memory, since later metadata (e.g. ``) can change how earlier properties are interpreted.
- **Backtracking**: with nested objects, the deserializer may re-process buffered data once per nesting level, leading to worst-case **O(n x d)** deserialization time.
- **Mitigation**: an `[!IMPORTANT]` callout recommending a lower `MaxDepth` when processing untrusted input.

### `JsonSourceGenerationOptionsAttribute.AllowOutOfOrderMetadataProperties`
Added condensed matching remarks with an xref link to the main property for full details.

## Motivation
MSRC feedback indicated that the existing docs didn't sufficiently communicate the backtracking cost of enabling this property, which could be surprising to users who enable it without understanding the implications.

### DIFF
--- a/xml/System.Text.Json/JsonSerializerOptions.xml
+++ b/xml/System.Text.Json/JsonSerializerOptions.xml
@@ -238,13 +238,10 @@ similar to how <xref:System.Text.Json.Serialization.Metadata.JsonTypeInfoResolve
 
 By default, `System.Text.Json` requires that JSON metadata properties such as `$id`, `$type`, and `$ref` appear at the very start of a JSON object. Setting this property to `true` relaxes that requirement, allowing metadata to appear after regular properties. This is useful when consuming JSON from producers that don't guarantee metadata ordering.
 
-### Buffering
+However, there are performance implications when this property is enabled:
 
-When this property is enabled, the deserializer can no longer process object properties in a single forward pass. Because metadata properties appearing later in the object (for example, a `$type` discriminator) can change how all earlier properties should be interpreted, the deserializer must buffer the entire JSON object in memory before it can commit to a deserialization strategy. In streaming deserialization scenarios, this can significantly increase memory consumption, especially for large JSON payloads.
-
-### Backtracking
-
-With nested objects, the cost of out-of-order metadata is multiplicative. The deserializer may need to backtrack—re-read and re-process previously buffered data—once for each level of nesting depth in the incoming object graph. This means the worst-case deserialization time is **O(n × d)**, where *n* is the length of the JSON input and *d* is the depth of the object graph.
+- **Buffering**: When this property is enabled, the deserializer can no longer process object properties in a single forward pass. Because metadata properties appearing later in the object (for example, a `$type` discriminator) can change how all earlier properties should be interpreted, the deserializer must buffer the entire JSON object in memory before it can commit to a deserialization strategy. In streaming deserialization scenarios, this can significantly increase memory consumption, especially for large JSON payloads.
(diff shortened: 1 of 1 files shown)


xml/System.Text.Json.Serialization/JsonSourceGenerationOptionsAttribute.xml
```diff
@@ -125,7 +125,19 @@
...
         <value>To be added.</value>
-        <remarks>To be added.</remarks>
+        <remarks>
+          <format type="text/markdown"><![CDATA[
+
+## Remarks
+
+This property sets the default value for <xref:System.Text.Json.JsonSerializerOptions.AllowOutOfOrderMetadataProperties> on the generated serializer options.
+
+Enabling this property allows JSON metadata properties such as `$id`, `$type`, and `$ref` to appear after regular properties during deserialization. However, this has performance implications: the deserializer must buffer entire JSON objects in memory, and with nested objects it may need to backtrack once per nesting level, resulting in worst-case **O(n × d)** deserialization time (where *n* is the payload length and *d* is the object graph depth).
```
gewarren:
```suggestion
When this property is enabled, JSON metadata properties such as `$id`, `$type`, and `$ref` are allowed to appear after regular properties during deserialization. However, this has performance implications. The deserializer must buffer entire JSON objects in memory, and with nested objects, it might need to backtrack once per nesting level, resulting in worst-case **O(n × d)** deserialization time (where *n* is the payload length and *d* is the object graph depth).
```

xml/System.Text.Json/JsonSerializerOptions.xml
```diff
@@ -231,7 +231,22 @@ similar to how <xref:System.Text.Json.Serialization.Metadata.JsonTypeInfoResolve
...
+
+### Buffering
+
+When this property is enabled, the deserializer can no longer process object properties in a single forward pass. Because metadata properties appearing later in the object (for example, a `$type` discriminator) can change how all earlier properties should be interpreted, the deserializer must buffer the entire JSON object in memory before it can commit to a deserialization strategy. In streaming deserialization scenarios, this can significantly increase memory consumption, especially for large JSON payloads.
+
+### Backtracking
+
+With nested objects, the cost of out-of-order metadata is multiplicative. The deserializer may need to backtrack—re-read and re-process previously buffered data—once for each level of nesting depth in the incoming object graph. This means the worst-case deserialization time is **O(n × d)**, where *n* is the length of the JSON input and *d* is the depth of the object graph.
```
gewarren:
```suggestion
However, there are performance implications when this property is enabled:

- **Buffering**: When this property is enabled, the deserializer can no longer process object properties in a single forward pass. Because metadata properties appearing later in the object (for example, a `$type` discriminator) can change how all earlier properties should be interpreted, the deserializer must buffer the entire JSON object in memory before it can commit to a deserialization strategy. In streaming deserialization scenarios, this can significantly increase memory consumption, especially for large JSON payloads.
- **Backtracking**: With nested objects, the cost of out-of-order metadata is multiplicative. The deserializer might need to backtrack (that is, reread and reprocess previously buffered data) once for each level of nesting depth in the incoming object graph. This means the worst-case deserialization time is **O(n × d)**, where *n* is the length of the JSON input and *d* is the depth of the object graph.
```

PR #12514, opened 2026-04-14
Remove .NET Framework-specific remarks (System namespace)

gewarren:
Contributes to #12513

PR #12563, opened 2026-04-28
Add Framework-only API lookup

gewarren:
Contributes to #12514.

File created by Copilot CLI. Will be used as a faster way to determine which APIs are .NET-Framework only when performing cleanup work in the repo.


BillWagner (approved):
This LGTM.

(I confess to only skimming the diff)

PR #12571, opened 2026-04-28
Remove F# examples where no API signature in F#

### LINKED ISSUE #12542: numeric type mismatch
### Type of issue

Other (describe below)

### Description

C# code example defines
enum SingleHue : **short** 
enum MultipleHue : **short** 

but the loop variable is of type **int**. AFAIK the compiler/JIT will have to enlarge before any bit operation and that is a perf no-no
no advantage reason given why code should be based on **short**, so I suggest you either use matching **int** (or better just omit and let compiler choose)

BTW this code shows **all** of C#, F# and VB language examples, whereas imho devs are likely to focus on just **ONE**.
Elsewhere in the docs the UX offers a combo to display lang of choice (to reduce clutter and brain-fade). TIA !

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.flagsattribute.-ctor?view=net-10.0&devlangs=csharp&f1url=%3FappId%3DDev18IDEF1%26l%3DEN-US%26k%3Dk%28System.FlagsAttribute.%23ctor%29%3Bk%28DevLang-csharp%29%26rd%3Dtrue

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System/FlagsAttribute.xml

4c34e6ce-3163-99f4-3d05-251583b8a874

### Platform Id

83a0548e-26e2-3231-958d-bfbb757475a1

### Article author

@dotnet-bot

---
[Associated WorkItem - 572082](https://dev.azure.com/msft-skilling/Content/_workitems/edit/572082)

gewarren:
Fixes #12542

### DIFF
--- a/xml/System.Collections.Generic/Dictionary`2.xml
+++ b/xml/System.Collections.Generic/Dictionary`2.xml
@@ -263,7 +263,6 @@
  This code example is part of a larger example provided for the <xref:System.Collections.Generic.Dictionary`2> class.
 
  :::code language="csharp" source="~/snippets/csharp/System.Collections.Generic/DictionaryTKey,TValue/Overview/source.cs" interactive="try-dotnet-method" id="Snippet2":::
- :::code language="fsharp" source="~/snippets/fsharp/System.Collections.Generic/DictionaryTKey,TValue/Overview/source.fs" id="Snippet2":::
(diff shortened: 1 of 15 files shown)


davidle-ms:
LGTM

adegeo (approved):

PR #12616, opened 2026-05-12
Document .NET 11 ConstructorInfo.GetGenericArguments behavior change

### LINKED ISSUE #128041: GetGenericArguments throws NotSupportedException on ConstructorInfo
### Description

I would expect MethodBase.GetGenericArguments() to return an empty array when calling on any non-generic method including constructors. A NotSupportedException with the message "Derived classes must provide an implementation" is thrown instead when called on a ConstructorInfo object.

### Reproduction Steps

This unit test fails:

``` 
namespace TestProject1
{
	[TestClass]
	public sealed class Test1
	{
		[TestMethod]
		public void TestMethod1()
		{
			var ctor = typeof(SimpleClass).GetConstructor(new[] { typeof(int) });
			Assert.IsEmpty(ctor.GetGenericArguments());
		}
	}

	public class SimpleClass
	{

		public SimpleClass(int x)
		{ }

	}
}
```

I have created a tiny project that reproduces the problem - at least on my machine:

[Constructor_GetGenericArguments.zip](https://github.com/user-attachments/files/27602012/Constructor_GetGenericArguments.zip)

### Expected behavior

GetGenericArguments() should return an empy array.

### Actual behavior

The call throws an exception, seemingly because GetGenericArguments() is not overridden in RuntimeConstructorInfo.

### Regression?

_No response_

### Known Workarounds

I check MethodBase.IsGenericMethod first, because this is false for constructors; I can then assume that GetGenericArguments() is empty.

### Configuration

.NET version: 10.0.203
OS version: Windows 11 Pro 22H2
CPU architecture: x64

### Other information

_No response_

max-charlamb:
## Summary

Updates the `MethodBase.GetGenericArguments` reference docs to reflect the .NET 11 behavior change in dotnet/runtime#128059 (which fixes dotnet/runtime#128041).

Previously, `MethodBase.GetGenericArguments` docs claimed the method throws `NotSupportedException` when invoked on a `ConstructorInfo`. As of .NET 11, `ConstructorInfo` overrides this method to return an empty array.

The remarks and exception text are updated to describe both behaviors:
- .NET Framework and .NET (Core) 5.0 through 10.0: throws `NotSupportedException` on a `ConstructorInfo`.
- .NET 11+: returns an empty array.

## Notes for reviewers

- No dedicated `<Member MemberName="GetGenericArguments">` entry is added in `ConstructorInfo.xml`. The `mdoc` regeneration that runs against .NET 11 ref assemblies should add the skeleton automatically; a follow-up PR can fill in the prose.
- Related: dotnet/runtime#128041 (bug), dotnet/runtime#128059 (fix).

### DIFF
--- a/xml/System.Reflection/MethodBase.xml
+++ b/xml/System.Reflection/MethodBase.xml
@@ -506,7 +506,7 @@
 
 - If the current method is an open constructed method (that is, the <xref:System.Reflection.MethodBase.ContainsGenericParameters> property returns `true`) in which specific types have been assigned to some type parameters and type parameters of enclosing generic types have been assigned to other type parameters, the array contains both types and type parameters. Use the <xref:System.Type.IsGenericParameter> property to tell them apart. For a demonstration of this scenario, see the code example provided for the <xref:System.Reflection.MethodBase.ContainsGenericParameters> property.
 
-- It the current method is a constructor, the returned array is an empty array in .NET 11 and later. In .NET 10 and earlier versions, the <xref:System.Reflection.MethodBase.GetGenericArguments*> method throws <xref:System.NotSupportedException>.
+- If the current method is a constructor, the returned array is an empty array in .NET 11 and later. In .NET 10 and earlier versions, the <xref:System.Reflection.MethodBase.GetGenericArguments*> method throws <xref:System.NotSupportedException>.
 
  For a list of the invariant conditions for terms specific to generic methods, see the <xref:System.Reflection.MethodBase.IsGenericMethod> property. For a list of the invariant conditions for other terms used in generic reflection, see the <xref:System.Type.IsGenericType?displayProperty=nameWithType> property.
 
--- a/xml/System.Reflection/MethodBase.xml
+++ b/xml/System.Reflection/MethodBase.xml
@@ -506,13 +506,13 @@
 
 - If the current method is an open constructed method (that is, the <xref:System.Reflection.MethodBase.ContainsGenericParameters> property returns `true`) in which specific types have been assigned to some type parameters and type parameters of enclosing generic types have been assigned to other type parameters, the array contains both types and type parameters. Use the <xref:System.Type.IsGenericParameter> property to tell them apart. For a demonstration of this scenario, see the code example provided for the <xref:System.Reflection.MethodBase.ContainsGenericParameters> property.
 
- Generic constructors are not supported in the .NET Framework version 2.0. This property throws <xref:System.NotSupportedException> if not overridden in a derived class, so an exception is thrown if the current instance is of type <xref:System.Reflection.ConstructorInfo>.
+ Generic constructors are not supported. The base <xref:System.Reflection.MethodBase.GetGenericArguments*> implementation throws <xref:System.NotSupportedException> if not overridden in a derived class. In .NET Framework and in .NET (Core) 5.0 through 10.0, <xref:System.Reflection.ConstructorInfo> does not override this method, so calling it on a constructor throws <xref:System.NotSupportedException>. Starting in .NET 11, <xref:System.Reflection.ConstructorInfo.GetGenericArguments*?displayProperty=nameWithType> overrides this method to return an empty array.
 
  For a list of the invariant conditions for terms specific to generic methods, see the <xref:System.Reflection.MethodBase.IsGenericMethod> property. For a list of the invariant conditions for other terms used in generic reflection, see the <xref:System.Type.IsGenericType?displayProperty=nameWithType> property.
 
  ]]></format>
         </remarks>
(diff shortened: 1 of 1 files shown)


xml/System.Reflection/MethodBase.xml
```diff
@@ -506,13 +506,13 @@
...
 
  ]]></format>
         </remarks>
-        <exception cref="T:System.NotSupportedException">The current object is a <see cref="T:System.Reflection.ConstructorInfo" />. Generic constructors are not supported in the .NET Framework version 2.0. This exception is the default behavior if this method is not overridden in a derived class.</exception>
+        <exception cref="T:System.NotSupportedException">This method is not overridden in a derived class. This is the default behavior of the base <see cref="T:System.Reflection.MethodBase" /> implementation. In .NET Framework and in .NET (Core) 5.0 through 10.0, the current object is a <see cref="T:System.Reflection.ConstructorInfo" />, because generic constructors are not supported. Starting in .NET 11, calling this method on a <see cref="T:System.Reflection.ConstructorInfo" /> returns an empty array instead of throwing.</exception>
```
jkotas:
```suggestion
        <exception cref="T:System.NotSupportedException">This method is not overridden in a derived class.</exception>
```
I would keep it simple.


> because generic constructors are not supported. Starting in .NET 11, calling this method on a <see cref="T:System.Reflection.ConstructorInfo" /> returns an empty array instead of throwing.

Generic constructors are still unsupported in .NET 11, but the method does not throw anymore. "throws before .NET 11 because generic constructors are not supported is" is not a logical explanation.

xml/System.Reflection/MethodBase.xml
```diff
@@ -506,13 +506,13 @@
...
 
- Generic constructors are not supported in the .NET Framework version 2.0. This property throws <xref:System.NotSupportedException> if not overridden in a derived class, so an exception is thrown if the current instance is of type <xref:System.Reflection.ConstructorInfo>.
+ Generic constructors are not supported. The base <xref:System.Reflection.MethodBase.GetGenericArguments*> implementation throws <xref:System.NotSupportedException> if not overridden in a derived class. In .NET Framework and in .NET (Core) 5.0 through 10.0, <xref:System.Reflection.ConstructorInfo> does not override this method, so calling it on a constructor throws <xref:System.NotSupportedException>. Starting in .NET 11, <xref:System.Reflection.ConstructorInfo.GetGenericArguments*?displayProperty=nameWithType> overrides this method to return an empty array.
```
jkotas:
```suggestion
- It the current method is a constructor, the returned array is an empty array in .NET 11 and later. In .NET 10 and earlier versions, the <xref:System.Reflection.MethodBase.GetGenericArguments*> method throws <xref:System.NotSupportedException>.
```


> overrides this method to return an empty array

The API docs explain the observable behavior. They typically do not explain how the behavior is implemented (e.g. whether it is implemented by overriding or by some other way)

> In .NET Framework and in .NET (Core) 5.0 through 10.0

There was .NET Core 1.0 - 3.1 as well, there are multiple versions of .NET Standard, ... . We typically just mention the  ,NET version that introduced the new behavior to keep it simple.

max-charlamb: Got it, I was trying to match the existing text. This makes more sense.

xml/System.Reflection/MethodBase.xml
```diff
@@ -506,13 +506,13 @@
 
 - If the current method is an open constructed method (that is, the <xref:System.Reflection.MethodBase.ContainsGenericParameters> property returns `true`) in which specific types have been assigned to some type parameters and type parameters of enclosing generic types have been assigned to other type parameters, the array contains both types and type parameters. Use the <xref:System.Type.IsGenericParameter> property to tell them apart. For a demonstration of this scenario, see the code example provided for the <xref:System.Reflection.MethodBase.ContainsGenericParameters> property.
 
- Generic constructors are not supported in the .NET Framework version 2.0. This property throws <xref:System.NotSupportedException> if not overridden in a derived class, so an exception is thrown if the current instance is of type <xref:System.Reflection.ConstructorInfo>.
+- It the current method is a constructor, the returned array is an empty array in .NET 11 and later. In .NET 10 and earlier versions, the <xref:System.Reflection.MethodBase.GetGenericArguments*> method throws <xref:System.NotSupportedException>.
```
jkotas:
```suggestion
- If the current method is a constructor, the returned array is an empty array in .NET 11 and later. In .NET 10 and earlier versions, the <xref:System.Reflection.MethodBase.GetGenericArguments*> method throws <xref:System.NotSupportedException>.
```

jkotas (approved):

steveisok (approved):

gewarren (approved):

Issue #12513, opened 2026-04-13
Move .NET Framework-specific remarks to dotnet/docs repo


gewarren:
### Describe the issue or suggestion

Related to #10722. This is prep work for backporting docs to /// in dotnet/runtime. This work helps to avoid confusing agents that might be looking at the repo as a source of understanding and might not understand the nuances of multi-targeting.

---
[Associated WorkItem - 569834](https://dev.azure.com/msft-skilling/Content/_workitems/edit/569834)

PR #12635, opened 2026-05-21
Remove .NET Framework remarks (System.Globalization)

gewarren:
.NET Framework API ref has moved to its own repo (https://github.com/dotnet/dotnetfw-api-docs), so we can clean up .NET Framework remarks, exceptions, and code examples out of this repo. Contributes to #12513.

Removes remarks related to:

- .NET Framework versions
- Code-access security
- Configuring apps via app.config file
- App domains

*Hide whitespace changes*


tarekgh (commented):
The GetHashCode(ReadOnlySpan<char>, CompareOptions) overload had its "application domain" reference updated, but the GetHashCode(string, CompareOptions) overload (~line 1410 in CompareInfo.xml) still says:

> hash codes should never be used outside of the **application domain** in which they were created

Should this be updated to match the other overload's wording?

> hash codes should not be treated as unique identifiers, they should not be persisted, and they should not be used outside the current process.

tarekgh (approved):

Issue #12613, opened 2026-05-12
Reinstate "additional remarks"


gewarren:
### Describe the issue or suggestion

Reinstate the "additional remarks" that were moved to the dotnet/docs repo to fix issue https://github.com/dotnet/docs/issues/37859. This is prep work for backporting remarks to /// and not having to have links to the conceptual docs from the ///.

Also remove any .NET Framework-specific remarks and examples as text is brought back to this repo.

- [x] Reinstate remarks to this repo.
  - [x] Sync changes to dotnet-api-docs-temp. 
- [x] Reinstate remarks to the dotnetfw-api-docs repo.
- [x] Merge https://github.com/dotnet/runtime/pull/129002.
- [x] Check for any links from the dotnet/runtime repo, e.g. https://github.com/dotnet/runtime/blob/4f916b0fd33fb81708e8ace5e3348ffccedd336d/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.Match.cs#L446.
- [x] Remove files from the dotnet/docs repo and add redirects.

Contributes to #10722.

---
[Associated WorkItem - 578140](https://dev.azure.com/msft-skilling/Content/_workitems/edit/578140)

PR #12709, opened 2026-06-04
Reinstate supplemental remarks (various)

gewarren:
Contributes to https://github.com/dotnet/dotnet-api-docs/issues/12613.

Also removes .NET Framework remarks.


gewarren:
Snippets 5000 failures are for missing .fsproj files.

snippets/csharp/System.Numerics/BigInteger/Overview/Mutability_Examples.cs
```diff
@@ -0,0 +1,75 @@
...
+        Console.WriteLine(number);
+        // </Snippet1>
+    }
+
+    private static void PerformBigIntegerOperation()
+    {
+        Stopwatch sw = Stopwatch.StartNew();
+
+        // <Snippet12>
+        BigInteger number = Int64.MaxValue ^ 5;
```
wadepickett, line 28: Was this meant to be an exponent?  In VB `^` would raise the number to a power, in this case to a power of 5, which would be big.  Here in C# however it is XOR (exclusive OR).  Also used elsewhere in this file for similar samples.

snippets/fsharp/System.Collections.Generic/ListT/Overview/addremoveinsert.fs
```diff
@@ -0,0 +1,66 @@
+//<snippet1>
```
wadepickett, line 1: No projet for these .fs files.  So they won't pass a compilation check.  Maybe that is OK, but will just point it out here in case that was not the intention.

wadepickett (approved):
Approved, but please take a look at the minor items I found and the question I have about using `^` in the code sample.

snippets/csharp/System.Numerics/BigInteger/Overview/Mutability_Examples.cs
gewarren: I'm not sure, I didn't write the code. Just porting it.

gewarren:
Thank you for the reviews, Wade!!

PR #12724, opened 2026-06-08
Clarify NegotiateAuthentication.IsAuthenticated post-failure behavior

liveans:
The `IsAuthenticated` property returns true after the authentication exchange has finished, including in many post-negotiation failure cases (verified across the Windows, Unix, and `ManagedSpnego` PAL implementations in dotnet/runtime). The previous summary (successfully completed and the session was established) implied that a true value means successful authentication, which is misleading.

Update the summary, value, and remarks to clarify that callers must inspect the `NegotiateAuthenticationStatusCode` returned by `GetOutgoingBlob` to determine actual success.


rzikm (approved):
LGTM modulo existing comment

PR #12729, opened 2026-06-09
Fix incorrect comment in Complex.Zero C# example

### LINKED ISSUE #11229: Error in the comment in the example code of System.Numerics.Complex.Zero
### Type of issue

Typo

### Description

The example code contains a comment that does not match the code example, it states, `// Instantiate a complex number with real part 0 and imaginary part 1.` (_it is the imaginary part being 1 that is erroneous_) and then code is as follows, `Complex value1 = new Complex(0, 0);`

The example code should be modified to :
```
using System;
using System.Numerics;

public class Example
{
   public static void Main()
   {
      Complex value = Complex.Zero;
      Console.WriteLine(value.ToString());

      // Instantiate a complex number with real part 0 and imaginary part 0.
      Complex value1 = new Complex(0, 0);
      Console.WriteLine(value.Equals(value1));
   }
}
// The example displays the following output:
//       (0, 0)
//       True
```

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.numerics.complex.zero?view=net-9.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Numerics/Complex.xml

eeb72f25-c237-e76b-a13e-e9789868de91

### Platform Id

3c62b93e-ec1a-2efb-1fcd-7fcf3b7ac6d5

### Article author

@dotnet-bot

biboete:
Fixes #11229.

In the `Complex.Zero` example snippet, the comment read:

> `// Instantiate a complex number with real part 0 and imaginary part 1.`

but the code is `new Complex(0, 0)` (imaginary part **0**), and the documented output is `(0, 0)` / `True`. Corrected the comment to "imaginary part 0" so it matches the code and output.

_Note:_ the analogous VB snippet (`snippets/visualbasic/System.Numerics/Complex/Zero/zero1.vb`) has a mirror-image version of this typo ("real part 1" while the code is `New Complex(0, 0)`) — happy to fix that in this PR too if you'd prefer a single change covering both.


gewarren:
@biboete Yes, can you please fix them both in this PR?

biboete:
@gewarren Done — pushed a commit fixing the VB snippet as well (`' real part 1` → `' real part 0`, matching its `New Complex(0, 0)`). Both snippets are covered in this PR now.

gewarren (approved):
Thank you @biboete

PR #12751, opened 2026-06-16
Clarify ValueTask consumption guidance

### LINKED ISSUE #12704: ValueTask<TResult> documentation is outdated for .NET 10 - repeated await no longer throws
### Type of issue

Typo

### Description

The documentation for ValueTask<TResult> on [learn.microsoft.com](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1) states:

"A ValueTask<TResult> instance may only be awaited once, and consumers may not read Result until the instance has completed."

In .NET 10.0.300 (release), this is no longer true. I tested the following scenarios and none of them threw an exception:

Awaiting the same ValueTask<int> instance multiple times

Calling .AsTask() multiple times

Accessing .Result before the task completed

Calling .Preserve() after .AsTask()

The documentation should be updated to reflect the actual behavior in .NET 10

Example:

Image

Expected behavior (per docs): The second await should throw InvalidOperationException.

Actual behavior (.NET 10.0.300): Both await calls succeed and print the same result.

Environment:

.NET 10.0.300

Windows x64

Console application targeting net10.0

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask?view=net-10.0

### Content source URL

https://github.com/dotnet/dotnet-api-docs-temp/blob/live/xml/System.Threading.Tasks/ValueTask.xml

8146b64b-4225-5572-23b2-c959597dfb55

### Platform Id

5c4fec22-5232-0ebf-2ed8-a486d802cf53

### Article author

@dotnet-bot

YNan-varamor:
Fixes #12704.

This updates the `ValueTask` and `ValueTask<TResult>` remarks so the guidance is less absolute and reflects the backing-source distinction:

- instances backed by `Task` or a synchronous result may allow repeated awaits
- arbitrary `ValueTask` instances should still be treated as single-consumption because they may be backed by pooled `IValueTaskSource`
- `AsTask` remains the documented path when multiple consumption is needed

Validation:
- `git diff --check`
- confirmed the previous may-only-be-awaited-once wording no longer appears in the ValueTask XML docs

Disclosure: I used OpenAI Codex to implement and verify this documentation update.

### DIFF
--- a/xml/System.Threading.Tasks/ValueTask.xml
+++ b/xml/System.Threading.Tasks/ValueTask.xml
@@ -74,9 +74,9 @@
       <format type="text/markdown"><![CDATA[
 
 ## Remarks
-A `ValueTask` instance may either be awaited or converted to a <xref:System.Threading.Tasks.Task> using <xref:System.Threading.Tasks.ValueTask.AsTask*>. A `ValueTask` instance may only be awaited once, and consumers may not call <xref:System.Threading.Tasks.ValueTask.GetAwaiter> until the instance has completed. If these limitations are unacceptable, convert the `ValueTask` to a <xref:System.Threading.Tasks.Task> by calling <xref:System.Threading.Tasks.ValueTask.AsTask*>.
+A `ValueTask` instance may either be awaited or converted to a <xref:System.Threading.Tasks.Task> using <xref:System.Threading.Tasks.ValueTask.AsTask*>. If the instance wraps a <xref:System.Threading.Tasks.Task> or a synchronous result, multiple awaits might succeed. However, a `ValueTask` instance can also be backed by a pooled <xref:System.Threading.Tasks.Sources.IValueTaskSource>, so consumers should treat an arbitrary `ValueTask` as single-consumption unless the provider explicitly documents support for multiple consumption. Consumers may not call <xref:System.Threading.Tasks.ValueTask.GetAwaiter> until the instance has completed. If these limitations are unacceptable, convert the `ValueTask` to a <xref:System.Threading.Tasks.Task> by calling <xref:System.Threading.Tasks.ValueTask.AsTask*>.
 
-The following operations should never be performed on a `ValueTask` instance:
+The following operations should not be performed on a `ValueTask` instance unless the provider explicitly documents that they are supported:
 
 - Awaiting the instance multiple times.
 - Calling <xref:System.Threading.Tasks.ValueTask.AsTask*> multiple times.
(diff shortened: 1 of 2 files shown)


YNan-varamor:
@dotnet-policy-service agree

xml/System.Threading.Tasks/ValueTask.xml
```diff
@@ -74,9 +74,9 @@
...
+A `ValueTask` instance may either be awaited or converted to a <xref:System.Threading.Tasks.Task> using <xref:System.Threading.Tasks.ValueTask.AsTask*>. If the instance wraps a <xref:System.Threading.Tasks.Task> or a synchronous result, multiple awaits might succeed. However, a `ValueTask` instance can also be backed by a pooled <xref:System.Threading.Tasks.Sources.IValueTaskSource>, so consumers should treat an arbitrary `ValueTask` as single-consumption unless the provider explicitly documents support for multiple consumption. Consumers may not call <xref:System.Threading.Tasks.ValueTask.GetAwaiter> until the instance has completed. If these limitations are unacceptable, convert the `ValueTask` to a <xref:System.Threading.Tasks.Task> by calling <xref:System.Threading.Tasks.ValueTask.AsTask*>.
```
gfoidl, line 77: > multiple awaits **might** succeed

(emphasis by me)

Well does this make the docs better if it just "might" succeed? 
And how should a consumer know by what the VT is backed? 

I'd revert this change, as it's very vague and doesn't provide any betterness -- actually it makes it worse.

xml/System.Threading.Tasks/ValueTask.xml
```diff
@@ -74,9 +74,9 @@
...
 
-The following operations should never be performed on a `ValueTask` instance:
+The following operations should not be performed on a `ValueTask` instance unless the provider explicitly documents that they are supported:
```
gfoidl, line 79: This addition is good. But "unless ... documents it's supported" applies nearly to everything that in general should not be done and I don't think it's meaningful to append such a clause everywhere (to avoid bloat).

BillWagner:
ping @tannergooding @jeffhandley for a technical review on the recommendations.

xml/System.Threading.Tasks/ValueTask.xml
teo-tsirpanis, line 77: The thing is that you can't know whether a `ValueTask` is backed by a `Task` or a synchronous result. This is an implementation detail that is legal to change at any time between versions.

PR #12752, opened 2026-06-16
Improve IDisposable note

### LINKED ISSUE #12628: Improve IDisposable note
### Describe the issue or suggestion

Many types that implement IDisposable include a lengthy note block that describes how to dispose. Instead, this information should live in the IDisposable remarks only and we can simply refer readers there.

Sample of existing text:

> This type implements the <xref:System.IDisposable> interface. When you have finished using the type, you should dispose of it either directly or indirectly. To dispose of the type directly, call its <xref:System.IDisposable.Dispose*> method in a `try`/`catch` block. To dispose of it indirectly, use a language construct such as `using` (in C#) or `Using` (in Visual Basic). For more information, see the "Using an Object that Implements IDisposable" section in the <xref:System.IDisposable> interface topic.

---
[Associated WorkItem - 584116](https://dev.azure.com/msft-skilling/Content/_workitems/edit/584116)

gewarren:
Fixes #12628.


BillWagner (approved):
This LGTM.

Let's :shipit:

PR #12807, opened 2026-06-25
Upgrade projects to .NET 10 (System.Xml.Something)

gewarren:
Contributes to #12513.


BillWagner (approved):
This LGTM. (I like removing the framework multi-targeting, where possible)

PR #12848, opened 2026-07-02
Reword some regex safety warnings

GrabYourPitchforks:
I'm updating regex docs across the various docs repos. This is one part of that update.

Changes:
- Remove misleading guidance that timeout values are appropriate ways to guard against _all_ hostile values.
- Consolidate the warnings to the Regex type rather than just two specific constructors.
- Update external link from CISA to OWASP. The CISA guidance is primarily DDoS-related, which isn't quite relevant to the discussion here. OWASP's discussion is geared specifically toward ReDoS, which is on point.

Related PRs:
- https://github.com/dotnet/docs-desktop/pull/2250
- https://github.com/dotnet/docs/pull/54621
- https://github.com/dotnet/dotnet-api-docs/pull/12848 (this one)
- https://github.com/dotnet/runtime/pull/130271

### DIFF
--- a/xml/System.Text.RegularExpressions/Regex.xml
+++ b/xml/System.Text.RegularExpressions/Regex.xml
@@ -199,9 +199,6 @@
 
 ## Remarks
 
-> [!WARNING]
-> When using <xref:System.Text.RegularExpressions> to process untrusted input, pass a [time-out value](/dotnet/standard/base-types/best-practices-regex#use-time-out-values) to prevent malicious users from causing a [denial-of-service attack](https://www.cisa.gov/news-events/news/understanding-denial-service-attacks). A time-out value specifies how long a pattern-matching method should try to find a match before it times out.
-
  The `pattern` parameter consists of regular expression language elements that symbolically describe the string to match. For more information about regular expressions, see the [.NET Regular Expressions](/dotnet/standard/base-types/regular-expressions) and [Regular Expression Language - Quick Reference](/dotnet/standard/base-types/regular-expression-language-quick-reference) topics.
 
  Calling the <xref:System.Text.RegularExpressions.Regex.%23ctor(System.String)> constructor is equivalent to calling the <xref:System.Text.RegularExpressions.Regex.%23ctor(System.String,System.Text.RegularExpressions.RegexOptions)> constructor with a value of <xref:System.Text.RegularExpressions.RegexOptions.None> for the `options` argument.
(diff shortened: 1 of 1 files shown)


gewarren (commented):
I don't know why the GitOps policy we have to prevent edits to these files didn't fire, but System.Text.RegularExpressions has its source of truth for docs in the dotnet/runtime repo as of a few weeks ago. So please make these changes there instead: https://github.com/dotnet/runtime/blob/0a8142e2de2a05ac170dac8e9edffb41d059463b/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs#L24

GrabYourPitchforks:
Thanks @gewarren! I'll try the changes there.

Just for reference, looks like this is an example of the sanctioned way to include markdown in the `<remarks>` section?

https://github.com/dotnet/runtime/blob/901aeddd1c55f5fc5a1c1450f234dd954c430ea5/src/libraries/System.IO.Pipelines/src/System/IO/Pipelines/PipeReader.cs#L141-L152

gewarren:
> Thanks @gewarren! I'll try the changes there.
> 
> Just for reference, looks like this is an example of the sanctioned way to include markdown in the `<remarks>` section?
> 
> https://github.com/dotnet/runtime/blob/901aeddd1c55f5fc5a1c1450f234dd954c430ea5/src/libraries/System.IO.Pipelines/src/System/IO/Pipelines/PipeReader.cs#L141-L152

Yes, that looks right.

GrabYourPitchforks:
@gewarren Part of this PR involves removing duplicative / misleading text from the Regex ctors, putting the proper text directly at the type level.

However, if you look at the runtime code as it's currently checked in, you'll see that the text I'm deleting doesn't exist there.
https://github.com/dotnet/runtime/blob/640673e2dca2e497c55634c12de9afd7daebe872/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs#L94-L108

Thoughts? Do I still need to make hand edits to this .xml file to remove the target text?

gewarren:
> @gewarren Part of this PR involves removing duplicative / misleading text from the Regex ctors, putting the proper text directly at the type level.
> 
> However, if you look at the runtime code as it's currently checked in, you'll see that the text I'm deleting doesn't exist there. https://github.com/dotnet/runtime/blob/640673e2dca2e497c55634c12de9afd7daebe872/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs#L94-L108
> 
> Thoughts? Do I still need to make hand edits to this .xml file to remove the target text?

Hmm, good catch. There's a lot of remarks and examples ([example](https://github.com/dotnet/dotnet-api-docs/blob/d6bcee227128190f4ae3e77a6f193a198e16c392/xml/System.Text.RegularExpressions/Regex.xml#L189-L216)) that weren't backported. @danmoseley Was that on purpose or did Copilot mess up?

@GrabYourPitchforks Yes, please delete the remarks you don't want from this repo then, if they don't appear in dotnet/runtime.

jeffhandley (approved):

GrabYourPitchforks:
@gewarren I updated the PR to *only* track removal of the duplicate / misleading sections. The PR in the runtime repo tracks changes to the docs that live there.

gewarren (approved):

Issue #12648, opened 2026-05-22
Remove out of support .NET Core versions from docs


gewarren:
### Describe the issue or suggestion

- [x] Remove remarks for versions that have been out of support for at least a year (currently this includes versions up to .NET 7)
- [x] Remove all monikers that have been out of support for at least a year (i.e. .NET Core 1.0 - .NET 7).

---
[Associated WorkItem - 584114](https://dev.azure.com/msft-skilling/Content/_workitems/edit/584114)

PR #12861, opened 2026-07-07
Remove remarks about out of support .NET versions (WinForms)

gewarren:
Contributes to #12648.


gewarren:
WinForms docs won't be backported, so this PR isn't necessary.

PR #12866, opened 2026-07-07
Fix customer feedback issue template

gewarren:
Update to match dotnet/docs and prevent description auto-reset.


BillWagner (approved):
This LGTM.

Issue #12713, opened 2026-06-04
Modernize all C# code snippets


gewarren:
### Describe the issue or suggestion

Prep work for https://github.com/dotnet/dotnet-api-docs/issues/10722.

- [x] Upgrade target framework to .NET 10 (example: https://github.com/dotnet/dotnet-api-docs/pull/12654/changes#r3293816438)
- [ ] Modernize C# code - for example, see remarks at https://github.com/dotnet/dotnet-api-docs/pull/12700#pullrequestreview-4428371172
  - [x] System namespace 
  - [x] System.CodeDom
  - [x] System.CodeDom.Compiler
  - [x] System.Collections
  - [x] System.Collections.Concurrent
  - [x] System.Collections.Generic
  - [x] System.Collections.Immutable
  - [x] System.Collections.ObjectModel
  - [x] System.Collections.Specialized
  - [x] System.ComponentModel
  - [x] System.ComponentModel.Composition
  - [x] System.ComponentModel.Composition.Hosting
  - [x] System.ComponentModel.Design
  - [x] System.ComponentModel.Design.Serialization
  - [x] System.Configuration
  - [x] System.Data
  - [x] System.Diagnostics
  - [x] System.Diagnostics.CodeAnalysis
  - [x] System.Diagnostics.Contracts
  - [x] System.Diagnostics.Eventing.Reader
  - [x] System.Diagnostics.SymbolStore
  - [x] System.Diagnostics.Tracing
  - [x] System.Drawing
  - [ ] System.Drawing.Design
  - [ ] System.Drawing.Drawing2D
  - [ ] System.Drawing.Imaging
  - [ ] System.Drawing.Printing
  - [x] System.Dynamic
  - [ ] System.Globalization
  - [ ] System.IO
  - [ ] System.IO.Compression
  - [x] System.IO.Enumeration
  - [ ] System.IO.IsolatedStorage
  - [ ] System.IO.MemoryMappedFiles
  - [ ] System.IO.Packaging
  - [ ] System.IO.Pipes
  - [ ] System.IO.Ports
  - [ ] System.Linq
  - [ ] System.Linq.Expressions
  - [ ] System.Media
  - [ ] System.Net
  - [ ] System.Net.Http
  - [ ] System.Net.Mail
  - [ ] System.Net.NetworkInformation
  - [ ] System.Net.Security
  - [ ] System.Net.Sockets
  - [ ] System.Numerics
  - [ ] System.Printing
[...]

PR #12920, opened 2026-07-31
Delete .NET Framework VB projects, code files, and refs

gewarren:
Contributes to #12713.

VB code won't be backported to the dotnet/runtime repo, so no sense in spending time to upgrade old VB projects.


BillWagner (approved):
This LGTM

PR #12947, opened 2026-08-06
Modernize C# code snippets - System/C*

gewarren:
Contributes to #12713.


gewarren:
Will break this into two PRs so CCR can review.

PR #12953, opened 2026-08-07
Update SKILL.md with additional code style guidelines

gewarren:
Clarify guidelines on string interpolation and expression-bodied members.

Contributes to #12713.

### DIFF
--- a/.agents/skills/dotnet-snippet-modernization/SKILL.md
+++ b/.agents/skills/dotnet-snippet-modernization/SKILL.md
@@ -38,7 +38,7 @@ Apply these changes when they preserve behavior and sample clarity:
   remain unchanged.
 - Convert eligible value-producing `switch` statements to switch expressions.
 - Use auto-implemented properties instead of defining a separate field.
-- Use expression-bodied members for simple single-expression members. If the line gets too long, for example, a method signature with a type parameter constraint, place the expression body on a new line, for example:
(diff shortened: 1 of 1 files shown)


.agents/skills/dotnet-snippet-modernization/SKILL.md
```diff
@@ -25,15 +25,27 @@ Apply these changes when they preserve behavior and sample clarity:
...
+      timeZoneTime.TimeZone == null ? "<null>" : timeZoneTime.TimeZone.ToString(),
+      timeZoneTime.DateTime);
+  ```
 - Use raw string literals (or interpolated raw string literals) for paragraph-style output.
 - Use object and collection initializers when evaluation order and behavior
   remain unchanged.
 - Convert eligible value-producing `switch` statements to switch expressions.
 - Use auto-implemented properties instead of defining a separate field.
-- Use expression-bodied members for simple single-expression members.
+- Use expression-bodied members for simple single-expression members. If the line gets too long, for example, a method signature with a type parameter constraint, place the expression body on a new line, for example:
```
gewarren:
```suggestion
- Use expression-bodied members for simple single-expression members. If the line gets too long, for example, a method signature with a type parameter constraint, place the expression body on a new line:
```

wadepickett (approved):
@gewarren, approved.  Looks good.

PR #13067, opened 2026-09-15
Update GH action SHAs with dependabot

gewarren:
Contributes to dotnet/docs-tools#743.


wadepickett (approved):
@gewarren:  Approved. Looks great, I can't find any issues.

